PR #4, opened 2021-06-24
Allow PV with permission less than 0777

spilchen:
This is for VER-77261.

This relaxes the requirement that PV must have 0777 permissions.  We can
allow 0775 now, which is the default permission for local data in EKS.

Vertica handles changing the ownership of most of the directories,
except the depot.  The depot directory had to be handled specifically.

I created an e2e test that mimics the reduced permission using a custom
config for Rancher's local-path-provisioner.


ningdeng (approved):

roypaulin (commented):
looks good for me

roypaulin (approved):

PR #7, opened 2021-06-28
Operator to control starting of vertica agent

spilchen:
Prior to this, the vertica agent was being started in the docker entrypoint.  This led to some race conditions -- starting the agent at the same time we are doing create_db or scale-out.  The race conditions often ended up causing the admintools.conf to be corrupted.  The file had all of the host information removed and it was owned by root.

The fix was to give the operator the control over when the agent starts.  This avoids the race conditions because there is only a single operator thread running.  A new `ReconcileActor` called `AgentReconciler` was added.  This is called right after the database is created or revived.

This also has some cleanup for `ReconcilerActor`.  All of the actors had to have common boilerplate functions (`GetClient`, `GetVDB`, `CollectPFacts`).  Only two of the actor's made use of this.   The rest of the actors defined them but had no use.  I cleaned this up by making an interface that the two actors can use.

This fixes VER-77811

### DIFF
--- a/pkg/controllers/agent_reconcile_test.go
+++ b/pkg/controllers/agent_reconcile_test.go
@@ -1,0 +1,75 @@
+/*
+ (c) Copyright [2021] Micro Focus or one of its affiliates.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ You may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/cmds"
+	"github.com/vertica/vertica-kubernetes/pkg/names"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"yunion.io/x/pkg/tristate"
+)
+
+var _ = Describe("agent_reconcile", func() {
+	ctx := context.Background()
+
+	It("should start the agent if it isn't running", func() {
+		vdb := vapi.MakeVDB()
+		vdb.Spec.Subclusters[0].Size = 2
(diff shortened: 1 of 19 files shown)


pkg/controllers/podfacts.go
```diff
@@ -271,6 +278,21 @@ func (p *PodFacts) checkIfNodeIsUp(ctx context.Context, pf *PodFact) error {
...
+// checkIfAgentRunning will check if the Vertica agent is running and set state in pf.agentRunning
+func (p *PodFacts) checkIfAgentRunning(ctx context.Context, pf *PodFact) error {
+	if !pf.isPodRunning {
+		pf.agentRunning = false
+		return nil
+	}
+
+	if _, _, err := p.PRunner.ExecInPod(ctx, pf.name, ServerContainer, "/opt/vertica/sbin/vertica_agent", "status"); err != nil {
+		pf.agentRunning = false
+	} else {
```
roypaulin, line 290: Isn't there a way to avoid this if else? I know it does not change anything.

roypaulin (commented):
looks good for me.

roypaulin (approved):

pkg/controllers/podfacts.go
spilchen: Yeah, makes sense.  I can clean this up.

pkg/controllers/agent_reconcile_test.go
```diff
@@ -0,0 +1,75 @@
...
+		vdb := vapi.MakeVDB()
+		vdb.Spec.Subclusters[0].Size = 2
+		createPods(ctx, vdb, AllPodsRunning)
+		defer deletePods(ctx, vdb)
+
+		cmds := reconcileAndFindVerticaAgentStart(ctx, vdb)
+		Expect(len(cmds)).Should(Equal(2))
+	})
+
+	It("should avoid starting the agent if ipv6", func() {
```
ningdeng, line 43: I like this change for 11, agent may support ipv6 post 11, in that case this may need a minor change.

pkg/controllers/agent_reconciler.go
```diff
@@ -0,0 +1,81 @@
...
+		if pod.agentRunning {
+			continue
+		}
+		// Only start the agent for pods that have been added to a database
+		if !pod.dbExists.IsTrue() {
+			continue
+		}
+		// We don't start the vertica agent if running ipv6 since the agent
+		// currently doesn't work in that mode.  This can be removed once
+		// VER-77406 is addressed.
```
ningdeng, line 58: I'm wondering if we can remove JIRA numbers here, as this is not helpful for users.

ningdeng (approved):

pkg/controllers/agent_reconciler.go
spilchen: Yes, I'll remove it.

PR #8, opened 2021-06-29
Github workflow: trigger unit tests

roypaulin:
Each time there is a push or PR on the main branch, actions are triggered and unit tests are run.


.github/workflows/unittests.yml
```diff
@@ -0,0 +1,28 @@
...
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.16
```
spilchen, line 19: Could we use the same version that's included in DEVELOPER.md -- 1.16.2?

.github/workflows/unittests.yml
```diff
@@ -0,0 +1,28 @@
+name: Operator Unit tests
+
+on:
```
spilchen, line 3: Can this just be `on: [push, pull_request]`?  The unit tests are quick, so it should get run on any push or PR.

spilchen (commented):
Looks good.  Just a question on whether we can do this for each push/PR.

.github/workflows/unittests.yml
roypaulin: sure.

spilchen (approved):

PR #9, opened 2021-06-29
Allow image to be built with docker 18.03

spilchen:
This includes a change to the vertica image so that it can be built with docker 18.03.  The `COPY .. --from` doesn't preserve file ownership on the old docker -- that wasn't fixed until docker 19.03.

Also includes changes to run-k8s-int-tests.sh that are needed by our internal CI.


Makefile
```diff
@@ -65,10 +65,13 @@ endif
...
+export OPERATOR_IMG
 # Image URL to use for building/pushing of the webhook
 WEBHOOK_IMG ?= verticadb-webhook:$(TAG)
+export WEBHOOK_IMG
 # Image URL to use for building/pushing of the vertica server
 VERTICA_IMG ?= vertica-k8s:$(TAG)
+export VERTICA_IMG
 # Set this to YES if you want to create a vertica image of minimal size
 MINIMAL_VERTICA_IMG ?=
 # Produce CRDs that work back to Kubernetes 2.11 (no version conversion)
```
davidmathi91, line 77: * The kuttl run in the script saves the output in the working directory.
* I believe to store the result outputs in specific folders, a variable like `INT_TEST_OUTPUT_DIR` should be defined.

ningdeng:
This looks good to me overall but I'm not a docker pro so I'll leave it to david to approve :)

davidmathi91:
* Have deleted my previous comment. I will add those changes once I am able to run my scripts locally.
* The changes in this branch looks good to me

davidmathi91 (approved):

PR #10, opened 2021-07-01
Better support for update strategy RollingUpdate

spilchen:
An issue was opened when a tester tried to add a license secret after
the cluster was created.  Adding the secret causes the statefulset controller
to apply the update in a rolling update fashion.  This didn't work so
well, the update eventually froze while it was trying to restart
vertica.  The changes here are meant to address that.

- change the readiness probe back to using the 'select 1' canary query.
This was critical in making the rolling update strategy to behave better.
The old readiness probe was causing us to go onto the next pod to soon.
If you only have 3 pods in a k-safety 1 system, this would cause us to
lose two pods, which loses quorum so the 3rd pod goes down to.
- the 'select 1' canary query was used before but was changed when we
added support for the superuser password.  We didn't want to hardcode
the password.  We avoided that here by getting the password from a file
in the container; the file is /etc/podinfo/superuser-passwd.  It is
included in the projected volume that currently contains things from the
downward API.
- use 'OnDelete' for the UpdateStrategy for k-safety 0.  Any pod loss
will take down vertica on the remaining pods.  So supporting a rolling
update strategy seemed pointless.
- add a new option 'restartTimeout' that controls what timeout to use
with admintools when restarting pods
- during restart we will kill any outstanding vertica process ahead of
doing the actual restart.  This seemed to help the restart process, but
is mitigated by the readiness probe change.  I kept it in here because I
think it is still good to have.


pkg/controllers/install_reconcile_test.go
```diff
@@ -101,10 +103,14 @@ var _ = Describe("k8s/install_reconcile_test", func() {
 		// Reset the pod runner output to dump the compat21 node number
 		fpr.Results = cmds.CmdResults{
 			names.GenPodName(vdb, sc, 1): []cmds.CmdResult{
-				{}, // Check for stale admintools.conf
+				{}, // Remove old admintools.conf
```
roypaulin, line 106: could you explain these changes to me?

roypaulin: the next 2 lines included

roypaulin (commented):
looks good!

pkg/controllers/install_reconcile_test.go
spilchen: I added new debug code to capture the state of admintools.conf.  This adds extra calls to the pod runner.  There were two testcases that were sensitive to this and had to be updated.

roypaulin (approved):

README.md
```diff
@@ -285,6 +285,7 @@ The following table describes each configurable parameter in the VerticaDB CRD a
...
 | reviveOrder | This specifies the order of nodes when doing a revive.  Each entry contains an index to a subcluster, which is an index in `subclusters[i]`, and a pod count of the number of pods include from the subcluster.<br><br>For example, suppose the database you want to revive has the following setup:<br>- v_db_node0001: subcluster A<br>- v_db_node0002: subcluster A<br>- v_db_node0003: subcluster B<br>- v_db_node0004: subcluster A<br>- v_db_node0005: subcluster B<br>- v_db_node0006: subcluster B<br><br>And the `subclusters[]` list is defined as {'A', 'B'}.  The revive order would be:<br>- {subclusterIndex:0, podCount:2}  # 2 pods from subcluster A<br>- {subclusterIndex:1, podCount:1}  # 1 pod from subcluster B<br>- {subclusterIndex:0, podCount:1}  # 1 pod from subcluster A<br>- {subclusterIndex:1, podCount:2}  # 2 pods from subcluster B<br><br>If InitPolicy is not Revive, this field can be ignored.|Not set
+| restartTimeout | This specifies the timeout to use calling admintools to restart pods.  If not specified, it defaults 0, which means we will use the admintools default of 20 minutes. | 0 |
```
ningdeng, line 288: Minor: this is readme so it would be nice to mention that it's in seconds, but this is very minor as AT specifies timeout in seconds so users could guess that it's also in seconds.

spilchen: Good point, I will include seconds.

ningdeng:
Looks good, I initially got confused with the rollingUpgrade in the PR title and rollingUpdate in the description, but after looking at the changes it was about rolling update.

ningdeng (approved):

PR #12, opened 2021-07-06
Better error handling for install/uninstall

spilchen:
Better error handling for install/uninstall

When we install we create an install indicator file and when we uninstall we remove it.  The management of this is separate from the actual update_vertica call.  As such, there are a class of errors where the update_vertica gets interupted and the install indicator file gets out of sync.

I was able to reproduce this by killing the controller pod during install/uninstall.  In the e2e tests I have seen this happen when the controller loses leader election and new deployment starts up.  

I have handle this situation in the following manner:
- install: if we get an error that the host is already part of the cluster, we drive uninstall then retry the install again.  
- uninstall: if we get an error that the host is not part of the cluster, we simply remove the install indicator file and retry.  Also, I needed to relax a check in reip if it can't find the host.  This can happen because the host is already removed from the cluster.

This fixes VER-77888


ningdeng (approved):

roypaulin (approved):
looks good!

PR #14, opened 2021-07-08
Save contents of /opt/vertica/config/licensing

spilchen:
When running the vertica-k8s container all of the contents in /opt/vertica/config are lost on initial startup because we use a PV to back that.  There are certain files that we want to keep and must be copied back over when the container starts.  We already stored the contents of /opt/vertica/config/licensing in /home/dbadmin/licensing/ce, we are now going to copy that directory over each time the container starts.

Also included some changes to properly maintain the LABELs in the vertica-k8s container.  Assigning a LABEL an environment variable did not work.  I had to pass the value through in the `docker build` command.


dmankins:
chmod -R 0777 seems rather permissive, I would think, perhaps 0775, or 0755?

docker-vertica/docker-entrypoint.sh
```diff
@@ -6,17 +6,22 @@ start_cron(){
...
+copy_config_files() {
     # We must use sudo in case the PV was created with permissions less than 0777.
     sudo cp -r /home/dbadmin/logrotate/* /opt/vertica/config/
     rm -rf /home/dbadmin/logrotate
+
+    sudo mkdir -p /opt/vertica/config/licensing
+    sudo cp -r /home/dbadmin/licensing/ce/* /opt/vertica/config/licensing
+    sudo chown -R dbadmin:verticadba /opt/vertica/config/licensing
+    sudo chmod -R 0777 /opt/vertica/config/licensing
 }
```
dmankins, line 21: 0777 seems a bit permissive.  Is there anything wrong with 0755 or 0775?

dmankins (approved):

PR #15, opened 2021-07-09
Triggers e2e tests on push, PR comment and manually

roypaulin:
A new CI file has been added. When we push on the main branch or write a specific comment on the PR it triggers e2e tests. You can also triggers them manually by going on the section "Actions" on github. they were some problems with create bucket as aws was trying to retrieve the region and in an isolated env as github it was causing an error so we set AWS_EC2_METADATA_DISABLED to true for aws to ignore that. We remove the agent because it was preventing us to scale out.


roypaulin:
I know you wanted to have a step just to run the script but I added krew install on that same step. Every time I installed it on one step and tried to use it on another it did not work even when it was added to the PATH and exported. Hope it is not a big problem.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,74 @@
...
+  issue_comment:
+    types: [created]
+  workflow_dispatch: {}
+
+jobs:
+
+  build:
+    
+    if: | 
+     (github.event.issue.pull_request && github.event.comment.body == '/run-e2e') ||
```
spilchen, line 15: We should update CONTRIBUTING.md to document how to trigger the e2e tests in your PR.  There is a section for the PR (step 7).  We can probably add a part there for CI, unit tests are run automatically for each commit in the PR and '/run-e2e' to trigger the end-to-end tests.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,74 @@
...
+      env:
+          VERTICA_CE_URL: "https://vertica-community-edition-for-testing.s3.amazonaws.com/XCz9cp7m/vertica-10.1.1-0.x86_64.RHEL6.rpm"
+          STERN_URL: "https://github.com/stern/stern/releases/download/v1.19.0/stern_1.19.0_linux_amd64.tar.gz"
+      run: |
+        curl -fsSLO $STERN_URL 
+        tar zxvf stern_1.19.0_linux_amd64.tar.gz
+        sudo cp stern_1.19.0_linux_amd64/stern /usr/local/bin/
+        curl $VERTICA_CE_URL -o vertica-x86_64.RHEL6.latest.rpm
+      
+    - name: Run e2e tests
```
spilchen, line 54: Should we split this step?  Have a step that installs krew and one that runs the e2e test?

spilchen (commented):
This looks really good.  After you have done the edits I suggested, we should try running the e2e tests in this PR by adding the '/run-e2e' comment.

.github/workflows/e2e.yml
spilchen: Oh, just saw your comment about this.  So you did try to separate them but it didn't work.  You can disregard this for now.

roypaulin: yes Each time I install krew on one step and try to use it on a different one I get a "unknown command error". Will continue to investigate but for now I have them on the same step

.github/workflows/e2e.yml
roypaulin: ok will do that.

spilchen (approved):
As discussed, I will push this and open up a dummy PR to test triggering of the e2e tests.

PR #17, opened 2021-07-13
Add support to run generic sidecars

spilchen:
This adds the ability run a custom sidecar alongside the vertica server.
A new property was added to the CR to allow this:  spec.sidescars[]

When the operator adds the new container, it will automatically include
the volume mounts that were added to the server container.  This way it
automatically has access to the same PV.

A new container was added for testing purposes.  Called the vlogger, it
will send output from vertica.log to stdout in a sidecar.  A few e2e
tests use this to ensure we can run with a sidecar.  This container is
not intended to be externalized in Vertica's dockerhub.

To allow this new image to be used in the e2e tests, the kustomize setup
was changed to allow for this image to be set.  Some of the VerticaDB specs
that we use for tests have a new `kustomize-vlogger-image` placeholder
for this purpose.


DEVELOPER.md
```diff
@@ -264,6 +266,26 @@ Events:
...
+```
+  sidecars:
+    - name: vlogger
+      image: <imageName>
+```
+
+The image <imageName> is a container that you build yourself.  An example container can be found in `docker-vlogger`.  Once it is running you can inspect the logs by using this sidecar.
+
+```
+$ kubectl logs vertica-sc1-0 -c vlogger
```
ningdeng, line 286: +1 this is very helpful!

docker-vlogger/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# A docker container that will tail the vertica.log.  This allows vertica to
+# follow the idiomatic way in Kubernetes of logging to stdout.
+
+FROM alpine:3.13
```
ningdeng, line 17: I recall it's small, which is good :)

ningdeng:
"A new container was added for testing purposes. Called the vlogger, it will send output from vertica.log to stdout in a sidecar. " I understand the container was for testing purposes, I just have a very general question: does this container keep sending output to stdout or does it only send output for one time?

ningdeng (approved):

spilchen:
> "A new container was added for testing purposes. Called the vlogger, it will send output from vertica.log to stdout in a sidecar. " I understand the container was for testing purposes, I just have a very general question: does this container keep sending output to stdout or does it only send output for one time?

Thanks for reviewing @ningdeng .  It will keep sending output to stdout.  Kubernetes will capture that in its own logging file.  But you have the ability to do `kubectl logs` on it etc.  A good depiction of what the sidecar does is [here](https://kubernetes.io/docs/concepts/cluster-administration/logging/#sidecar-container-with-logging-agent)

roypaulin (approved):
The sidecar is a good idea.

tests/e2e-disabled/restart-node-multi-sc/setup-vdb/base/setup-vdb.yaml
```diff
@@ -17,6 +17,9 @@ metadata:
   name: v-mc-restart
 spec:
   image: replace-with-kustomize
+  sidecars:
+    - name: vlogger
+      image: kustomize-vlogger-image
   communal:
```
dmankins, line 23: All these repeated modifications almost makes one think there should be a yaml template somehow used to generate all the yaml file boilerplate.

PR #19, opened 2021-07-15
Support upgrade when cluster is in bad state

roypaulin:
Previously a call to stop_db in upgrade_vertica returned an  error when the previous vertica image was a bad one or when the cluster was down. So now we first check if all pods are running (because if a bad image is running inside them their status will  obviously not be "Running"). if they are not, we skip the call to stop_db and proceed to update the bad image with the new one. if the pods are running, we check if there is at least one node of the cluster UP. if there is not we skip the stop_db call.
Moreover these 2 scenarios have been added to upgrade-vertica test.


roypaulin:
I assumed that if the cluster is down we don't want to call stop_db and will just skip it to continue. Am I right?

scripts/upgrade-vertica.sh
```diff
@@ -123,9 +123,21 @@ kubectl patch vdb $VDB_NAME --type=merge -p '{"spec": {"autoRestartVertica": fal
...
+if [ $RUNNING_PODS -eq $TOTAL_PODS ]
+then
+    UP_NODES=$(kubectl exec $NS_OPT $POD_NAME -- admintools -t list_allnodes | awk '{print $5}' | grep UP | wc -l)
+    if [ $UP_NODES -ne 0 ]
+    then
+        echo "starting stop_db..."
+        kubectl exec $NS_OPT $POD_NAME -- admintools -t stop_db -F -d $VDB_DBNAME $PS_OPT 1> /dev/null
+        echo "cluster down"
+    else
+        echo "Skip SHOTDOWN CLUSTER: no nodes are up -- the cluster is already down"
```
spilchen, line 136: minor: spelling SHOTDOWN => SHUTDOWN

scripts/upgrade-vertica.sh
```diff
@@ -123,9 +123,21 @@ kubectl patch vdb $VDB_NAME --type=merge -p '{"spec": {"autoRestartVertica": fal
...
+    fi
+else
+    echo "Skip SHOTDOWN CLUSTER: no nodes are up -- bad image in pods"
```
spilchen, line 139: minor: spelling SHOTDOWN => SHUTDOWN.  

Also, a bad image is just one reason why a pod may not be running.  Maybe just say "Skip SHUTDOWN CLUSTER: no pods are running"

scripts/upgrade-vertica.sh
```diff
@@ -123,9 +123,21 @@ kubectl patch vdb $VDB_NAME --type=merge -p '{"spec": {"autoRestartVertica": fal
 kubectl wait --for=condition=AutoRestartVertica=False vdb/$VDB_NAME --timeout="${TIMEOUT}"s $NS_OPT 1> /dev/null
 echo "autoRestartVertica successfully disabled"
 
-echo "starting stop_db..."
-kubectl exec $NS_OPT $POD_NAME -- admintools -t stop_db -F -d $VDB_DBNAME $PS_OPT 1> /dev/null
-echo "cluster down"
+RUNNING_PODS=$(${GET_POD} --field-selector=status.phase=Running 2> /dev/null | grep ${VDB_NAME} | wc -l)
```
spilchen, line 126: $POD_NAME is a pod that we'll use to call AT.  But we don't verify that it is running.  Should we put the `--field-selector` argument in $POD_NAME and use that?  We don't need to check that all pods are running.  Some may not be running but the db could still be up.  If we have one pod running, that should be enough to check if the cluster is up with `list_allnodes`

tests/e2e/upgrade-vertica/25-assert.yaml
```diff
@@ -0,0 +1,27 @@
...
+status:
+  replicas: 3
+---
+apiVersion: vertica.com/v1beta1
+kind: VerticaDB
+metadata:
+  name: v-upgrade-vertica
+status:
+  subclusters:
+    - installCount: 2
```
spilchen, line 27: Why are we expecting installCount to be 2?  It's a 3 node cluster.

tests/e2e/upgrade-vertica/35-update-auto-restart.yaml
```diff
@@ -0,0 +1,20 @@
+# (c) Copyright [2021] Micro Focus or one of its affiliates.
```
spilchen, line 1: What is the purpose of step 35 and beyond?  We already tested that we can call upgrade-vertica.sh when all pods aren't running.  We have a cluster that fully up at this point, so we are just going from vertica-k8s:latest to vertica-k8s:10.1.1-0?

spilchen:
> I assumed that if the cluster is down we don't want to call stop_db and will just skip it to continue. Am I right?

Yes that is correct.

tests/e2e/upgrade-vertica/25-assert.yaml
roypaulin: because when you update the crd with a wrong image, the last pod will try to update the image and fail with an ImagePullErr status so the installCount will go down to 2. That is why we delete the statefulset, to update all the other pods.

tests/e2e/upgrade-vertica/35-update-auto-restart.yaml
roypaulin: The pods are running but all the nodes are down. I set autoRestartVertica to false and then deleted the sts. New pods are automatically created but all the vertica nodes will be down. This was to emulate the case we have all pods are running but the cluster is down.

scripts/upgrade-vertica.sh
roypaulin: Ok got it

tests/e2e/upgrade-vertica/25-assert.yaml
spilchen: Okay, makes sense.  Lets add a comment here so its obvious.

tests/e2e/upgrade-vertica/35-update-auto-restart.yaml
spilchen: Okay, lets add a comment at the start of this test to explain the purpose of the next steps.

spilchen (approved):
Looks good.  Let's run the e2e tests and I'll push it if its clean.

PR #20, opened 2021-07-16
Allow e2e parallelism to be overridden

roypaulin:
Now we can pass a parameter to override the parallel field in kuttl. Run the workflow manually with workflow_dispatch event and specify the number of tests you want to run at the same time. if you do not specify any number the e2e tests will use parameters set in kuttl-test.yml.


Makefile
```diff
@@ -167,7 +167,11 @@ endif
 
 .PHONY: run-int-tests
 run-int-tests: install-kuttl-plugin vdb-gen ## Run the integration tests
+ifeq (${PARALLEL}, )
```
spilchen, line 170: Can we name the env. var E2E_PARALLELISM?  Also, instead of `ifeq` here, can we just define E2E_PARALLELISM with the `?=` setting so it is set unless overridden.

```
E2E_PARALLELISM?=2
```

We can include this near the top of the Makefile with the other defines and include a comment to document what behaviour is changes.

roypaulin: Ok. Do you like how we are setting the env var through workflow_dispatch event?

.github/workflows/e2e.yml
```diff
@@ -70,7 +69,8 @@ jobs:
         "$KREW" install krew
         )
         export PATH=$PATH:$HOME/.krew/bin
-        [ -z "$E2E_PARALLELISM" ] && export E2E_PARALLELISM
+        E2E_PARALLELISM=${{ github.event.inputs.parallel }}
+        [ ! -z "$E2E_PARALLELISM" ] && export E2E_PARALLELISM
```
roypaulin, line 73: When running manually with workflow_dispatch you can set the parallelism you want and it will override the default value. if the current event is not workflow_dispatch or did not precise any parallelism this env var will get ignored and  the e2e tests will be run with the default value

Makefile
spilchen: I think its okay.  I'm just not sure how often we would use it.  As far as I understood, the machines we use for GitHub CI are limited in size.  So increasing the parallelism made it more likely that we would see test failures due to timeouts.

.github/workflows/e2e.yml
spilchen: How high can the parallelism go in GitHub before we start seeing test failures due to timeout?  I wonder how much value there is in having this configurable.

spilchen (approved):
Looks good.  We just need a clean e2e test run with the latest commit.

PR #21, opened 2021-07-19
Include minio output in ancillary.log for e2e runs

spilchen:
Adding a label 'stern=include' to minio so that it is included in the ancillary.log when running e2e tests (see logic in scripts/save-ancillary-logs.sh)


roypaulin (approved):
looks good!

davidmathi91:
Looks goood.

davidmathi91 (approved):

PR #22, opened 2021-07-20
Replace issue_comment event with pull_request

roypaulin:
Given that issue_comment did not add much as it always targets the main branch, we replace it with pull_request but only when a new PR is opened. For next runs, you can do it manually with workflow_dispatch.


spilchen (commented):
Looks good.  

Can we update DEVELOPER.md?  It mentions the old '/run-e2e' that has now been removed.  It would be good to explain that (a) we run e2e when the PR is first opened and (b) you can use workflow dispatch if you need to run the tests again.

spilchen (approved):

PR #26, opened 2021-08-05
e2e tests to handle dockerhub rate limiting

spilchen:
A bunch of fixes for the e2e tests so that we can avoid getting failures due to dockerhub rate limiting
- avoid the use of latest as this as a pull policy of Always.  Use a specific tag for tests that use amazon/aws-cli
- remove private-pvc test.  It was using the tag latest somewhere.  Rather than fix it, I opted to remove it because it doesn't work when running in the public cloud.
- keep a list of external images that we should push to the kind nodes to avoid going out to dockerhub
- updates to push-to-kind.sh so that it accepts the external image list


davidmathi91:
It's working as expected.

davidmathi91 (approved):

PR #39, opened 2021-08-23
Remove calls to update vertica

spilchen:
- removed calls to update_vertica in the install and uninstall reconcile.  This significantly speeds up scale-up operations.
- new package (atconf) to handle modifying admintools.conf for install/uninstall
- added a method to copy a file into a pod.  It is implemented as a cat exec call.
- had to explicitly accept the EULA now that we don't call update_vertica
- had to explicitly check the existence and permissions of some directories in /opt/vertica/config (share and logrotate)
- we pass the license in during the create_db command now that we don't call update_vertica anymore
- had to change the obfuscation code to check for --password rather than -w.  I added code that did a call to 'test -w _file_', and the obfuscation code was adding '****' for the file name.
- moved ipv6 check function to its own package so that I can use it in the atconf package
- moved name of vertica server container to names package.  This was necessary so that we can use the const in the new atconf package


tests/e2e/crash-before-createdb/07-assert.yaml
```diff
@@ -1,14 +1,4 @@
 apiVersion: v1
-kind: Event
```
roypaulin, line 2: It is been  a while for me so could you tell me why we do not need this anymore?

docker-vertica/docker-entrypoint.sh
```diff
@@ -10,18 +10,19 @@ start_cron(){
 # because we have a Persistent Volume that backs /opt/vertica/config, so
 # it starts up empty and must be populated
 copy_config_files() {
-    # We must use sudo in case the PV was created with permissions less than 0777.
-    sudo cp -r /home/dbadmin/logrotate/* /opt/vertica/config/
+    cp -r /home/dbadmin/logrotate/* /opt/vertica/config/
+    chown -R dbadmin:verticadba /opt/vertica/config/logrotate
     rm -rf /home/dbadmin/logrotate
 
-    sudo mkdir -p /opt/vertica/config/licensing
```
roypaulin, line 17: Why don't we need sudo anymore?

pkg/atconf/file_writer.go
```diff
@@ -0,0 +1,339 @@
...
+// MakeFileWriter will build and return the FileWriter struct
+func MakeFileWriter(log logr.Logger, vdb *vapi.VerticaDB, prunner cmds.PodRunner) Writer {
+	return &FileWriter{
+		Log:     log,
+		Vdb:     vdb,
+		PRunner: prunner,
+	}
+}
+
+// AddHosts will had ips to an admintools.conf.  New admintools.conf, stored in
```
ningdeng, line 63: minor typo: had --> add. 

I'm trying to get more details: is it adding IPs to the hosts = entry in [Cluster] section?

ningdeng: Also, does it check duplicate IPs?

ningdeng: Edit: nvm, upon looking at the called functions I think my questions are answered.

pkg/atconf/file_writer.go
```diff
@@ -0,0 +1,339 @@
...
+		nodeInfo := fmt.Sprintf("%s,%s,%s", ip, f.Vdb.Spec.Local.DataPath, f.Vdb.Spec.Local.DataPath)
+		err := f.Cfg.Set(NodesSection, nodeName, nodeInfo)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// removeNodes will remove the nodes section for the given set of IPs
```
ningdeng, line 244: I was wondering if this function is expected to remove only compat21 node entries, if yes, should we assert or check somewhere to make sure we do not touch regular database nodes, at least mention in the comment? I understand those regular database nodes will be removed and are supposed to be removed by db_remove_subcluster/db_remove_nodes before the uninstallation.

pkg/controllers/at.go
```diff
@@ -0,0 +1,58 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// distributeAdmintoolsConf will copy an admintools.conf to all of the running pods
+func distributeAdmintoolsConf(ctx context.Context, pf *PodFacts, pr cmds.PodRunner, atConfTempFile string) error {
+	for _, p := range pf.Detail {
+		if !p.isPodRunning {
+			continue
+		}
+		_, _, err := pr.CopyToPod(ctx, p.name, names.ServerContainer, atConfTempFile, paths.AdminToolsConf)
```
ningdeng, line 35: Is it possible that the copying succeed on some Pods but fail on other Pods?

tests/e2e/crash-before-createdb/07-assert.yaml
spilchen: Because the install is so fast, we no longer need the Event for it.  So all of the events for install and uninstall were removed.

docker-vertica/docker-entrypoint.sh
spilchen: We were over zealous with the sudo before and so it wasn't needed.  The long term goal is to have no reliance on sudo as that will play nice when we support OpenShift.  Removing calls to update_vertica was a key first step for that.

pkg/atconf/file_writer.go
spilchen: That's right.  We handle duplicate IPs by treating it as a no-op.  I will update the comment here.

pkg/atconf/file_writer.go
spilchen: That's right.  We only remove compat21 node entries.  The AT -t db_remove_nodes will handle removing of regular database nodes.  I will mention in a comment.

pkg/controllers/at.go
spilchen: Yes it is possible.  We make sure that all pods are running before we begin install/uninstall, but that just shortens the timing.  Do you think it is necessary to have a step in the operator to ensure admintools.conf is the same on all of the pods?  And copy if they are different.  The same pod is always picked first, so it should be a well known location that has the correct version of admintools.conf.

ningdeng: >Do you think it is necessary to have a step in the operator to ensure admintools.conf is the same on all of the pods?
-- Having a step regularly checking admintools.conf on all Pods may be overkilling, as long as the initiator of AT commands has the latest admintools.conf it should be ok. Maybe we can keep as is at this time, if tests show any problem, then we can see what we can do.

ningdeng: I'm not sure about the details of err message, just want to mention that it might be helpful if we know on which nodes the copy succeeded and on which the copy failed, so if there's at least one copy succeeded, then that copy can be the source of latest admintools.conf, such information could be helpful to help debugging and fix admintools.conf

ningdeng:
This looks good to me overall, thanks!

ningdeng (approved):

docker-vertica/docker-entrypoint.sh
roypaulin: Ok now I get it thanks.

roypaulin (approved):

pkg/controllers/at.go
spilchen: @ningdeng based on your comments here, I improved the distributeAdmintoolsConf function.
- the first pod we copy to will always be the same.  It is also the pod that we use for the base for any subsequent admintools.conf changes for install/uninstall
- if the first pod copied was okay, copy to all other pods.  We save the errors and do checking at the end so that we attempt to copy to all pods
- if the copy failed at one of the pods, log an event saying that admintools.conf was partially copied.

You can see these changes in the commit c424386

pkg/controllers/at.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	_, _, err = pr.CopyToPod(ctx, basePod, names.ServerContainer, atConfTempFile, paths.AdminToolsConf)
+	if err != nil {
+		return err
+	}
+
+	// Copy the admintools.conf to the rest of the pods.  We will do error
+	// checking at the end so that we try to copy it to each pod.
+	errs := []error{}
+	for _, p := range pf.Detail {
```
ningdeng, line 52: The latest improvements of copying admintools.conf change looks good to me. Thanks!

Just some minor notes as potential future improvements: I see this is a sequential operation as it uses a for-loop, starting from v11, AT uses a new file copying mechanism, which improves the performance of file distribution to the entire cluster a lot. So, just in case that in the future people suggest/request updating this copying to be a parallel operation on a very large cluster, AT distribute_config_files could be used to distribute the files. At this time, using CopyToPod is better IMO.

spilchen: Okay, thanks.  This is good to know and I will keep it in mind.

Issue #29, opened 2021-08-12
Extra volumes for the sidecars


nyikesda:
**STORY**
In many cases the sidecars needs extra volumes, for example to mount configmaps or / and secrets and check the filesystem to get every changes in the related configmaps and/or secrets.
In our case the logger sidecar needs some secret volumes because in our product we have to communicate with mutual tls towards a syslog server with periodically changed certificates.
On the other hand our sidecar has runtime configurations (for example: loglevel) as well which is configured in configmaps.

**SUPPORT REQUEST**
Please extend the current Vertica CRD with new configuration item, where the extra mounts could be configured. This configuration item should allow the same parameters as the kubernetes spec.volumes

PR #42, opened 2021-08-24
VER-78628 custom volume mounts

roypaulin:
- implemented new CR parameter call volumes.
- added a logic to the webhook to prevent users to put a volume name conflicting with existing mount points that are lodal-data and podinfo.
- **create-and-del-crd** has been modified to check this use case too. I did not think it was necessary to create a new e2e test.


api/v1beta1/verticadb_types.go
```diff
@@ -176,6 +176,10 @@ type VerticaDBSpec struct {
 	// operator adds the same volume mounts that are in the vertica server
 	// container to each sidecar container.
 	Sidecars []corev1.Container `json:"sidecars,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// Custom volumes that can be added to sidecars.
```
spilchen, line 181: Lets expand on this comment a bit since it will appear in our CRD.  How about this:
```
// Custom volumes that are added to sidecars.  Each sidecar must include
// the volume in the volumeMounts for it to appear in the container.  It
// accepts any valid volume type.  A unique name must be given for each
// volume and it cannot conflict with any of the internally generated volumes.
```

api/v1beta1/verticadb_types.go
```diff
@@ -176,6 +176,10 @@ type VerticaDBSpec struct {
 	// operator adds the same volume mounts that are in the vertica server
 	// container to each sidecar container.
```
spilchen, line 177: Maybe mention here that it includes any volumes from Volumes too.

spilchen (commented):
Looks good.  We will update the README with this new parm closer to the end of the 11.0SP1 cycle so as not confuse anyone trying out the officially released operator.

api/v1beta1/verticadb_webhook.go
```diff
@@ -428,6 +431,19 @@ func (v *VerticaDB) hasDuplicateScName(allErrs field.ErrorList) field.ErrorList
 	return allErrs
 }
 
+func (v *VerticaDB) hasValidVolumeName(allErrs field.ErrorList) field.ErrorList {
+	for i := range v.Spec.Volumes {
+		vol := v.Spec.Volumes[i]
+		if vol.Name == LocalDataPVC || vol.Name == PodInfoMountName {
```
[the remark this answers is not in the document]
roypaulin: Yes, I had a doubt about it. I will add it

api/v1beta1/verticadb_types.go
roypaulin: I was lacking inspiration on that part.

spilchen (approved):

spilchen:
Closes #29

PR #46, opened 2021-08-27
Relax cert manager dependency

### LINKED ISSUE #28: Do not require the cert-manager as a hard dependency
**STORY**
Currently the cert-manager is a hard dependency in the generated webhook helm chart, but in many cases the customers have other security solutions to generate certificates and they want to use their solution instead of a required one.

**SUPPORT REQUEST**
Please add an option into the helm chart `values.yaml` to switch off the cert-manager specific kubernetes resources. In case of this helm variable is false, then a secret name is required, which contains the tls.crt and the tls.key. On the other hand  a caBundle certificate parameter is required too, and this parameter should be used in the `MutatingWebhookConfiguration` and in the `ValidatingWebhookConfiguration`.

spilchen:
This adds to the helm chart an option to allow specifying your own certs for use with the webhook.  The default behaviour is to continue to rely on cert-manager.  But if the webhook.tlsSecret is specified then cert-manager isn't used.

To support this, changes had to be made to the generation of the helm-chart.
- we no longer keep all manifests in a single file.  We use a tool called kuberenetes-split-yaml that splits up a giant yaml into individual files -- one for Kind.  Splitting this allowed for easier manipulation of the manifests since we template out the cert-manager ones if webhook.tlsSecrets and add a caBundle to the webhook configs.
- logic for creation of the helm chart was getting too big for the Makefile, so was moved to its own script (create-helm-charts.sh)

A new e2e test was added that deploys the webhook with custom certs.  The certs are still generated by cert-manager, but it is done outside the helm chart install so mimics what a customer would do if they were using their own certs.

Resolves #28


le-an (commented):
This PR looks all good to me. I just have a quick question: 
Do we want to add a test to verify whether custom certs work? or show users an example about how to add custom certs in the documentation?

spilchen:
Thanks @le-an for reviewing.  I did add a new e2e test to verify custom certs.  See `tests/e2e/custom-certs-webhook`.  It uses cert-manager to create the certs, but it doesn't use cert-manager in the helm chart.  So it would be similar to how someone who deploy if they had their own certs.

le-an:
> Thanks @le-an for reviewing. I did add a new e2e test to verify custom certs. See `tests/e2e/custom-certs-webhook`. It uses cert-manager to create the certs, but it doesn't use cert-manager in the helm chart. So it would be similar to how someone who deploy if they had their own certs.

I see that, thanks @spilchen.

le-an (approved):

PR #51, opened 2021-09-03
Handle automatic upgrades

### LINKED ISSUE #52: Do not validate the autoRestartVertica during image upgrade
Based on the actual documentation the image can only changed when autoRestartVertica is false.
As far as I know, the main reason behind this limitation to avoid automatic vertica version increase.
It is not the best, beceause many times a new version from an image only contains other fixes, or features and not the version increase of the vertica rpm.
I think it is enough to highlight in the documentation that vertica db upgrade is not supported, and I would like to recommend to do not block other image version increases (versions which are not contain new vertica db version).

spilchen:
This adds the ability to automate an upgrade.  In the v11.0 release, we published a script to use to upgrade a VerticaDB CR. This change pushes much of the logic of that script into the operator. 

If you want to upgrade vertica, you simply update the `.spec.image` field in the VerticaDB CR.  The operator will detect when this happens and go into a mode that handles the upgrade.

New events are logged to know when the operator is doing this:

```
$ kubectl describe vdb sample
 
...<snip>...
Events:
  Type    Reason                   Age    From                Message
  ----    ------                   ----   ----                -------
  Normal  ImageChangeStart         5m10s  verticadb-operator  Vertica server image change has been initiated to 'vertica-k8s:11.0.1-0'
  Normal  ClusterShutdownStarted   5m12s  verticadb-operator  Calling 'admintools -t stop_db'
  Normal  ClusterShutdownSucceeded 4m08s  verticadb-operator  Successfully called 'admintools -t stop_db' and it took 56.22132s
  Normal  ClusterRestartStarted    4m25s  verticadb-operator  Calling 'admintools -t start_db' to restart the cluster
  Normal  ClusterRestartSucceeded  25s    verticadb-operator  Successfully called 'admintools -t start_db' and it took 240s
  Normal  ImageChangeSucceeded     5s     verticadb-operator  Vertica server image change has completed successfully.
```

A new status condition is added so that you can programmatically wait for the upgrade to complete, or timeout and rollback the upgrade.

To wait for the operator to get into the upgrade mode, you can issue this:
`kubectl wait --for=condition=ImageChangeInProgress=True vdb/v --timeout=x`

To wait for the operator to finish the upgrade, you can issue this:
`kubectl wait --for=condition=ImageChangeInProgress=False vdb/v --timeout=x`

where _v_ is the name of the CR and _x_ is a timeout value.

A new reconciler was added to handle the upgrade (called ImageChangeReconciler).  It uses parts of ObjReconciler and RestartReconciler to avoid code duplication of actions those reconcilers already handle.

A few other minor changes to mention:
- removed the AutoRestartVertica column when we issue 'kubectl get vdb'.  It didn't make sense to report that state now that it isn't as important with auto upgrade
- include vertica version, pulled from the rpm, in the vertica container.  This can be used in the future if we want to add logic about what following the proper upgrade path.

Closes #52


api/v1beta1/verticadb_webhook.go
```diff
@@ -149,15 +149,6 @@ func (v *VerticaDB) validateImmutableFields(old runtime.Object) field.ErrorList
 			"shardCount cannot change after creation.")
 		allErrs = append(allErrs, err)
 	}
-	// image will only be allowed to change if autoRestartVertica is disabled
```
ningdeng, line 152: Just would like to understand why this is removed as it seems that I lost track of webhook design.

pkg/controllers/upgrade_reconcile.go
```diff
@@ -0,0 +1,271 @@
...
+func (u *UpgradeReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctrl.Result, error) {
+	if err := u.PFacts.Collect(ctx, u.Vdb); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if ok, err := u.isUpgradeNeeded(ctx); !ok || err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Functions to perform upgrade processing.  Order matters.
```
ningdeng, line 60: +1

pkg/controllers/upgrade_reconcile.go
```diff
@@ -0,0 +1,271 @@
...
+		return false, fmt.Errorf("verticaDB condition '%s' missing from VerticaDBConditionType", vapi.UpgradeInProgress)
+	}
+	if inx < len(u.Vdb.Status.Conditions) && u.Vdb.Status.Conditions[inx].Status == corev1.ConditionTrue {
+		// Set a flag to indicate that we are continuing an upgrade.  This silences the UpgradeStarted event.
+		u.ContinuingUpgrade = true
+		return true, nil
+	}
+
+	// Next check if an upgrade is needed based on the image being different
+	// between the Vdb and any of the statefulset's.
```
ningdeng, line 126: The image being different meaning image name/tag? If so, I'm wondering if we do anything to prevent "downgrade".

pkg/controllers/upgrade_reconcile.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+	if u.PFacts.getUpNodeCount() == 0 {
+		u.Log.Info("No vertica process running so nothing to shutdown")
+		// No pods have vertica so we can avoid stop_db call
+		return ctrl.Result{}, nil
+	}
+
+	// Check the image of each of the up nodes.  We avoid doing a shutdown if
+	// all of the up nodes are already on the new image.
+	if ok, err := u.anyPodsRunningWithOldImage(ctx); !ok || err != nil {
```
ningdeng, line 166: Trying to understand the process: am I understanding correctly that there could be a period where Pods with old images and Pods with new images co-exist, it's just in such a period we make sure the DB is in shutdown status.

pkg/controllers/upgrade_reconcile.go
```diff
@@ -0,0 +1,271 @@
...
+		u.startUpgrade,
+		// Do a clean shutdown of the cluster
+		u.stopCluster,
+		// Delete the sts objects.  We don't rely on k8s rolling upgrade.
+		// Everything must be destroyed then regenerated.
+		u.deleteStatefulSets,
+		// Create the sts object with the new image name.
+		u.recreateStatefulSets,
+		// Start up vertica in each pod.
+		u.restartCluster,
```
ningdeng, line 72: Do we check if all the images are of the same version in/before this step?

pkg/events/event.go
```diff
@@ -51,4 +51,10 @@ const (
 	UnsupportedVerticaVersion       = "UnsupportedVerticaVersion"
 	ATConfPartiallyCopied           = "ATConfPartiallyCopied"
 	S3AuthParmsCopyFailed           = "S3AuthParmsCopyFailed"
+	UpgradeStart                    = "UpgradeStart"
+	UpgradeSucceeded                = "UpgradeSucceeded"
```
ningdeng, line 55: How do we handle upgrade failure?

api/v1beta1/verticadb_webhook.go
spilchen: We wanted this rule in there to make sure users follow the documented steps to upgrade.  But now that the operator automates those steps it wasn't necessary anymore.

pkg/controllers/upgrade_reconcile.go
spilchen: Yes, that is correct.  We don't make any distinction between upgrade or downgrade.  If the image name changes, then we follow the upgrade process (shutdown cluster, update image, restart cluster).  We need this in case you ever want to rollback an upgrade.  I would consider renaming it if you think it is confusing?  upgrade => imageChange

pkg/controllers/upgrade_reconcile.go
spilchen: This is here to see whether we need to do a shutdown.  The entire reconciler is idempotent, so it is just checking to see if a shutdown is necessary.  There are three cases we are considering here:
1. database is up with the old image
2. database is already shutdown
3. the cluster is already up with the new image

We skip the shutdown for case 2 and 3.

I'll update the comments here.

pkg/controllers/upgrade_reconcile.go
spilchen: No, we don't.  But we are using the same image for each statefulset, so it will be the same version.

pkg/events/event.go
spilchen: Yes, we go into a requeue/retry loop.  It will keep trying to start the database.  It is up to the user to change the image back to something else to "rollback" the upgrade.  We have status conditions that you can use with `kubectl wait`, to know if the upgrade times out.

pkg/controllers/upgrade_reconcile.go
spilchen: Sorry, I don't understand the context of this comment.

ningdeng: Oh this is just a "I like this comment"

pkg/controllers/upgrade_reconcile.go
ningdeng: I see, yes "imageChange" is much more clear to me.

ningdeng (approved):

ningdeng (approved):

pkg/controllers/upgrade_reconcile.go
nprabhu2195: Maybe I am nitpicking here but is there anything we could do to differentiate rollback-based downgrade and direct downgrade? Vertica doesn't support downgrade and that might lead to things breaking.

spilchen: Unfortunately, there is nothing we can do for this in the operator.  There isn't any notion of binary version in the catalog.  This likely needs engine support to have some mechanism to know when we have downgraded.

nprabhu2195: I see. Thanks!

PR #58, opened 2021-09-20
Increase memory limit for operator pod

spilchen:
We have seen some cases where the OOMKiller picks the operator pod.  This bumps the memory limit for the pod from 250MB to 750MB.  This also adds a new configuration parameter to helm to allow resources to be customized.


le-an (commented):
The changes look good. 
One minor question: how do we decide the 750M as the new memory limit?

le-an:
I see, thanks for this explanation

le-an (approved):

PR #59, opened 2021-09-21
Add ScheduleOnly initPolicy

spilchen:
This introduces a new initPolicy called ScheduleOnly.  The bootstrap of the database, either create_db or revive_db, is not handled.  Use this policy when you have a vertica cluster running outside of Kubernetes and you want to provision new nodes to run inside Kubernetes.  

Most of the automation is disabled when running in this mode.  The only automation that is done is attempting to restart any down pods with 'admintools -t restart_node'.  The user is resonsible for adding the pods as nodes to a vertica cluster (update_vertica), adding them to a database (admintools -t db_add_node), and handling any restart of the cluster (admintools -t re_ip/start_db).

Here is a sample CR:
```
kind: VerticaDB
metadata:
  name: sample
spec:
  initPolicy: ScheduleOnly
  subclusters:
    - name: sc1
      size: 3
    - name: sc2
      size: 3
```

Notice that the entire `.spec.communal` section is omitted.

The number of pods that are created is dictated by the size of each subcluster.  However, subclusters aren't added by the operator.  We group by subcluster to control the name of each of the pod.  The actual subcluster the pod is part of does not have to match the name in the CR.

### DIFF
--- a/pkg/controllers/podfacts.go
+++ b/pkg/controllers/podfacts.go
@@ -75,7 +75,7 @@ type PodFact struct {
 	vnodeName string
 
 	// The compat21 node name that Vertica assignes to the pod. This is only set
-	// if installation has occurred.
+	// if installation has occurred and the initPolicy is not ScheduleOnly.
 	compat21NodeName string
 
 	// True if the end user license agreement has been accepted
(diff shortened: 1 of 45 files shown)


pkg/controllers/podfacts.go
```diff
@@ -205,30 +205,48 @@ func (p *PodFacts) collectPodByStsIndex(ctx context.Context, vdb *vapi.VerticaDB
...
-			if _, stderr, err := p.PRunner.ExecInPod(ctx, pf.name, names.ServerContainer, cmd...); err != nil {
-				if !strings.Contains(stderr, "No such file or directory") {
-					return err
-				}
-				pf.hasStaleAdmintoolsConf = false
-			} else {
-				pf.hasStaleAdmintoolsConf = true
+		// We can't reliably set compat21NodeName because the operator didn't
+		// originate the install.  We will intentionally leave that blank.
+		pf.compat21NodeName = ""
```
ningdeng, line 225: No change required, just trying to confirm my understanding of the functionalities: does this mean re_ip will not be automated on the schedule only pods? I recall compat21NodeName is used for re_ip when db is down but would like to double check if my memory is correct.

ningdeng: nvm, I saw the changes in restart_reconcile and I think my question is answered.

ningdeng (approved):

spilchen:
Thanks for taking a look @ningdeng

PR #60, opened 2021-09-21
Added possibility to set S3 region

vyaslav:
Currently, it is not possible to use S3 bucket in region other then us-east-1. 
This PR introduces new field `communal.region` to fix this.
```
kind: VerticaDB
metadata:
  name: verticatest
spec:
  dbName: vert
  initPolicy: Create
  communal:
    path: "s3://vertica-k8s-test"
    endpoint: "https://s3.eu-central-1.amazonaws.com"
    region: "eu-central-1"
    credentialSecret: s3-creds
  subclusters:
    - name: defaultsubcluster
      size: 1
```

### DIFF
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -247,6 +247,10 @@ type CommunalStorage struct {
 	// and cannot change after creation.
 	Endpoint string `json:"endpoint"`
 
+	// +kubebuilder:validation:Optional
+	// The s3 region.
+	Region string `json:"region"`
+
 	// +kubebuilder:validation:required
 	// The name of a secret that contains the credentials to connect to the
 	// communal S3 endpoint. The secret must have the following keys set:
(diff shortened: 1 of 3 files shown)


spilchen:
@vyaslav thanks for your contribution.  I was in the middle of doing the fix when you opened this PR.  There were a few extra things I wanted to include in the fix so I took elements of yours and added it to mine.

vyaslav:
> @vyaslav thanks for your contribution. I was in the middle of doing the fix when you opened this PR. There were a few extra things I wanted to include in the fix so I took elements of yours and added it to mine.

@spilchen sounds good 👍🏼 thanks a lot for fixing it. Could you please also clarify if it is already possible to use azure blob storage (azb://) as communal storage within operator? (AFAIK it is possible in vertical 11 but not sure if operator supports it)

spilchen:
Currently, we don't support azure blob storage in the operator.  I would suggest you open a GitHub issue for this so that it is tracked and others can see the request and like it if it is something they want.

PR #64, opened 2021-09-23
Install krew through make target

spilchen:
Will download and install krew through a make target.  This will also harden the krew version rather than always getting latest.


davidmathi:
* Working as expected.

davidmathi (approved):

PR #92, opened 2021-11-17
Bugfix/no ac in operator readme rjs

rs-vertica:
Updated text to reflect that the admission controller is included in the operator.


spilchen (approved):
Looks good.

PR #93, opened 2021-11-17
Remove the Kerberos initContainer hack

spilchen:
The Vertica engine has very specific requirements for the Kerberos keytab: it must have read/write permissions for dbadmin and no one else.  It isn't possible to mount a file in k8s with specific requirements.  I originally added a hack that created an initContainer and copied the keytab so that the permissions could be set.  I'm removing that here as we now have a config knob in the engine to relax the check.  We just have to set `KerberosEnableKeytabPermissionCheck = 0`, and the engine won't care about keytab permissions.  This also means we don't need to restart the pod whenever the keytab changes as the change will get automatically picked up whenever the Secret changes.  Since this relies on an engine change, I put in a version check that Kerberos cannot be used except on 11.0.2.

### DIFF
--- a/api/v1beta1/verticadb_webhook.go
+++ b/api/v1beta1/verticadb_webhook.go
@@ -33,22 +33,21 @@ import (
 )
 
 const (
-	invalidDBNameChars      = "$=<>`" + `'^\".@*?#&/-:;{}()[] \~!%+|,`
-	dbNameLengthLimit       = 30
-	KSafety0MinHosts        = 1
-	KSafety0MaxHosts        = 3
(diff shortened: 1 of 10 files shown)


roypaulin:
looks good. But won't it be a problem to have less check or this specific is not that important?

spilchen:
We aren't too concerned about relaxing the check.  Removing the check out weighs any cons.  The check is there to make sure you don't expose the keytab to users who shouldn't have access.  But since we are in a container, there is an extra level of security (i.e. exec access through k8s roles) if this is really a concern.

roypaulin (approved):

PR #94, opened 2021-11-22
Add ability to specify custom ssh keys

spilchen:
This adds the ability to specify a custom ssh keys in the container.  The keys, stored in a Secret, will get mounted in /home/dbadmin/.ssh.  If this Secret is omitted, the ssh files from the image are used.  You can this option if you have a cluster that talks to Vertica notes outside of Kubernetes, as it has the public keys to be able to ssh to those nodes.  

The secret can be installed in the CR like this:
```
kind: VerticaDB
metadata:
  name: v
spec:
  communal:
    ...
  subclusters:
    - name: sc1
  sshSecret: ssh-key  
```

It must have the following keys present: id_rsa, id_rsa.pub and authorized_keys.


roypaulin (approved):
looks good

PR #96, opened 2021-11-24
Add security context to enable ssh on Openshift

roypaulin:
Deploy a multi nodes cluster on Openshift fail because nodes cannot ssh each other.
This security context is added to enable that.

### DIFF
--- a/pkg/controllers/builder.go
+++ b/pkg/controllers/builder.go
@@ -340,6 +340,12 @@ func makeServerContainer(vdb *vapi.VerticaDB, sc *vapi.Subcluster) corev1.Contai
 				},
 			},
 		},
+		// Is needed to run sshd on Openshift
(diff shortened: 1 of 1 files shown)


spilchen:
Looks good.  You hit an error in the azb test while it was running the unit test.  Can you rerun that one?

roypaulin:
Yeah I am manually rerunning it on the forked branch.

roypaulin:
I reran it. looks fine to me:
https://github.com/roypaulin/vertica-kubernetes/actions/workflows/e2e-azb.yml?query=branch%3Aadd-security-context++

spilchen:
It should have been rerun by following the Details link.  That way the PR is updated with its run result.  I have rerun it now.  So I'll monitor the run and push when it is clean.

roypaulin:
How do you do that? Can I do that too?

PR #98, opened 2021-11-24
Add serviceaccount to run pods with

roypaulin:
With the security context added to enable ssh on pods on Openshift, it is necessary to have pods associated to privileged scc.
For that, instead of running them with default serviceaccount, we reference the operator serviceaccount which of course is to be associated to a privileged scc in the CSV. This will not break helm install as among the manifests generated in the chart there is already a service account with the same name.


spilchen (approved):
Looks good

PR #99, opened 2021-11-25
Prevent additional restart after upgrade

spilchen:
If a cluster runs with k-safety of 1, we use the UpdateStrategy of RollingUpdate in the StatefulSet.  This should apply to everything but changes to the image because we have our own method of changing the image.  Otherwise, what can happen is that after the image change is successful (i.e. the cluster is brought online), the StatefulSet controller will go through the RollingUpdate process: it will kill the last pod in the state, wait for it become ready (i.e. Vertica running), then kill the next pod and repeat.  This ends up causing another restart to occur after the upgrade.

The change in this PR is to change the UpdateStategy to OnDelete when changing the image.  The UpdateStrategy will go changed back to RollingUpdate, but sometime after the image change.

There are a lot of changes in this PR.  I bulk of those changes was renaming the e2e test upgrade-vertica to upgrade-vertica-ks-0.  A new e2e test upgrade-vertica-ks-1 was added to test upgrade for k-safety 1 clusters.


roypaulin (approved):
looks good

PR #101, opened 2021-11-26
Image change can be marked complete before restart of the pods

spilchen:
This closes a timing window during upgrade where we can indicate the image change has finished before completing the restart of the pods.  There is a check during the restart portion that calls `areAllPodsRunningAndZeroInstalled`.  This returns true if none of the pods exist yet, which is not correct.  State that the function uses was being set after checking if the pod exists.  If the pod doesn't exist, the state was left unset.  The timing hole is closed now that the state is being set correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -303,7 +303,6 @@ docker-build-bundle: bundle ## Build the bundle image
 docker-push-bundle: ## Push the bundle image
 	docker push $(BUNDLE_IMG)
 
-OLM_CATALOG_BASE?=quay.io/operatorhubio/catalog:latest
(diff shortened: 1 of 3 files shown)


spilchen:
fyi @roypaulin, this will fix intermittent failures in the upgrade-vertica-ks-1 e2e testcase

roypaulin:
looks good. but checks were not successful. I guess it has nothing to do with the changes.

spilchen:
There is a problem building the opm package.  It is affecting all builds, so nothing related to this change.  I'm looking into it.

PR #102, opened 2021-11-26
Setup deploy olm

roypaulin:
setup-olm.sh and deploy-olm.sh are used to create the catalog and deploy the operator using olm. They have been modified to handle OpenShift as it requires the catalog to be created in the default namespace openshift-marketplace. Moreover, Openshift comes with olm-operator in openshift-operator-lifecycle-manager namespace so we made sure to not try to install the olm-operator.


scripts/setup-olm.sh
```diff
@@ -76,14 +77,22 @@ then
     $OPERATOR_SDK olm uninstall || true
 fi
 
+COUNT=$(kubectl get namespace openshift-operator-lifecycle-manager -o yaml 2>/dev/null | grep config.openshift.io/v1 | wc -l)
```
spilchen, line 80: We have duplicated this check.  Can we add a script, call it `is-openshift.sh` and just call the script to remove the duplication?

The `if` check then becomes:  `if $SCRIPT_DIR/is-openshift.sh`

scripts/setup-olm.sh
```diff
@@ -76,14 +77,22 @@ then
...
+
+if [ $COUNT -gt 0 ]
+then
+    OLM_NS=openshift-marketplace
+    OLM_OP_NS=openshift-operator-lifecycle-manager
+fi
+
 # Setup olm if not already present
-if ! kubectl get -n olm deployment olm-operator
+if ! kubectl get -n $OLM_OP_NS deployment olm-operator
```
spilchen, line 89: I think it would be more clear if we just skipped this entire if block if it is openshift.  That way we don't need the OLM_OP_NS variable.

```
if ! $SCRIPT_DIR/is-openshift.sh && ! kubectl get -n $OLM_NS deployment olm-operator
then
...
```

spilchen (commented):
Looks good.  You will need the fix that I'm adding in #101 to get a clean run.  You can either cherry pick the last two commits from that PR, or wait for it to be merged in.

scripts/setup-olm.sh
roypaulin: ok

scripts/setup-olm.sh
roypaulin: I will address both in my next commit

scripts/deploy-olm.sh
```diff
@@ -69,8 +70,15 @@ then
   NAMESPACE=default
 fi
 
+# Get the namespace where the catalog is
+if scripts/is-openshift.sh
```
spilchen, line 74: Instead of hardcoding `scripts`, can we define SCRIPT_DIR and use that.  That way you can use the script from any directory.  We set SCRIPT_DIR in other scripts.

```
SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```

spilchen (approved):
Looks good.  Thanks for applying the comments.

PR #103, opened 2021-11-26
Use proper azb endpoint for Azurite

spilchen:
When we run the CI against Azurite, we did not have the proper azb endpoint. 

The endpoint we were using was:
```
azb://<accountName>@<blobEndpoint>/<accountName>/<containerName>
```

It should be:
```
azb://<accountName>@<blobEndpoint>/<containerName>
```

To use this endpoint, we need to ensure we set isMultiAccountEndpoint to true.  But at the same time, when using a real Azure backend, isMultiAccountEndpoint is false.  Vertica already manages the default value for this, so the fix in the operator was not to set that field in the azure config.  That allows the Vertica engine to pick the correct value.

This PR also moves one of the revive tests from e2e-extra to e2e.  The azb CI only runs e2e.  But with this fix, we can now run the revive tests, so moving one of the tests over.


roypaulin (approved):
looks good

PR #106, opened 2021-11-30
Change vertica image version

roypaulin:
upgrade-vertica-ks-0/1 fail because of the use of the version 11.0.0 of the vertica image. We replaced it with version 11.0.1 and updated all related files.


spilchen (approved):
Looks good.

PR #107, opened 2021-12-02
updated broken DEVELOPER link at the top of the repo

rs-vertica:
DEVELOPER link was broken, replaced with correct link.


spilchen (approved):
Good catch.

PR #110, opened 2021-12-02
Reduce space requirements when running e2e in CI

spilchen:
Some of the CI tests are failing because we run out of disk space on the machines that run CI.  This reduces the number of images we need to push into the kind cluster prior to the e2e run.  There were some images that weren't needed for e2e.  Just needed for setup of the kind cluster itself.  So this PR moves some of those images into a new file external-images-build.txt.


roypaulin (approved):
looks good

PR #111, opened 2021-12-02
Add marker to generate missing fields in CSV

roypaulin:
Some scorecard tests were failing because there were some missing fields in the CSV so we add a new marker in  verticadb.types.go file to generate those fields in the CSV.


api/v1beta1/verticadb_types.go
```diff
@@ -39,10 +39,12 @@ const VerticaDBAPIVersion = "vertica.com/v1beta1"
 type VerticaDBSpec struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default:=IfNotPresent
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
```
spilchen, line 42: It is probably out of scope in this PR, but are you going to fill in the x-descriptors?  You can indicate if a field is expecting a Secret or ConfigMap (see here https://github.com/operator-framework/operator-lifecycle-manager/blob/e0eea22/doc/design/building-your-csv.md)

api/v1beta1/verticadb_types.go
```diff
@@ -39,10 +39,12 @@ const VerticaDBAPIVersion = "vertica.com/v1beta1"
...
 	// This dictates the image pull policy to use
```
spilchen, line 43: Some of the data structures that this one refers to don't have the new markers.  Why is that?  I'm thinking ones like LocalStorage, CommunalStorage, etc.

roypaulin: I put this new marker above every field of VerticaDBSpec so that information about the field will be used to fill the specDescriptors field in the CSV as that part was missing leading to scorecard tests failure

api/v1beta1/verticadb_types.go
roypaulin: Yes I was focused on mandatory fields but I guess on the next PR I can add x-descriptors too.

api/v1beta1/verticadb_types.go
spilchen: But why don't we include the fields in LocalStorage and CommunalStorage?  Can you fill those in when you create a CR from the webUI?

roypaulin: Yeah you are right we can also add that to fields in data structures. It is just that to pass tests  adding that marker to just fields inside VerticaDBSpec is enough. And yes we can already fill them all but local.requestSize from the webUI.

roypaulin: I think I will add the marker to all subfields for spec and status.

roypaulin:
I will let x-descriptors for a separate PR as the goal of this one is to fix scorecard tests an x-descriptors are not needed for that.

spilchen (approved):
Looks good.  You should merge with latest main to pick up the azb fix so we can get a clean PR run.

PR #113, opened 2021-12-06
Restart to handle read-only Vertica nodes

spilchen:
This is a change to the restart processing to handle the new read-only state in 11.0SP2.  If you lose cluster quorum, nodes that were up stay up node and are put into read-only state.  The operator needs to kill those processes before it can do a full cluster restart.

This collects the read-only state (11.0SP2 only), then treats them as down nodes when we restart.  The restart reconciler already had code that send a kill signal to running vertica processes.

I also rename the restart testcase to restart-sanity.  This was done so that we can run just this testcase within kuttl.  kuttl --test option will match all test cases that have the argument in its name.  So, trying to run with `kubectl kuttl test --test restart` would match auto-restart-vertica, restart-node-multi-sc and restart.  The new name means we can run just that single test via `kubectl kuttl --test restart-sanity`


pkg/controllers/podfacts.go
```diff
@@ -349,13 +370,56 @@ func (p *PodFacts) checkIfNodeIsUp(ctx context.Context, vdb *vapi.VerticaDB, pf
 			return err
 		}
 		pf.upNode = false
+		pf.readOnly = false
 	} else {
 		pf.upNode = true
+		pf.readOnly = false
```
roypaulin, line 376: What happens here with respect to line 373?

roypaulin:
Other than my comment, it looks good.

roypaulin:
looks good thanks

roypaulin (approved):

PR #115, opened 2021-12-06
Add x-descriptors to CSV - 1st commit

roypaulin:
Add x-descriptors. They are used to improve the form on the WebUI. Each field will get a graphical representation that suits it the best. Example a boolean field will be represented by a boolean switch, field with a limited number of values will be represented by a dropdown.

### DIFF
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// nolint:lll
 package v1beta1
 
 import (
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -44,7 +44,7 @@ type VerticaDBSpec struct {
 	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:io.kubernetes:Secret"}
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// ImagePullSecrets is an optional list of references to secrets in the same
 	// namespace to use for pulling the image. If specified, these secrets will
 	// be passed to individual puller implementations for them to use. For
@@ -88,13 +88,13 @@ type VerticaDBSpec struct {
 
 	// +kubebuilder:default:=12
 	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number"
 	// The number of shards to create in the database. This cannot be updated
 	// once the CRD is created.
 	ShardCount int `json:"shardCount"`
 
 	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:io.kubernetes:Secret"}
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:Secret"
 	// An optional name for a secret that contains the password for the
 	// database's superuser. If this is not set, then we assume no such password
 	// is set for the database. If this is set, it is up the user to create this
@@ -102,7 +102,7 @@ type VerticaDBSpec struct {
 	SuperuserPasswordSecret string `json:"superuserPasswordSecret,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:io.kubernetes:Secret"}
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:Secret"
 	// The name of a secret that contains the contents of license files. The
 	// secret must be in the same namespace as the CRD. Each of the keys in the
 	// secret will be mounted as files in /home/dbadmin/licensing/mnt. If this
@@ -123,7 +123,7 @@ type VerticaDBSpec struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default:=Create
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:Create","urn:alm:descriptor:com.tectonic.ui:select:Revive"}
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:Create","urn:alm:descriptor:com.tectonic.ui:select:Revive","urn:alm:descriptor:com.tectonic.ui:select:ScheduleOnly"}
 	// The initialization policy defines how to setup the database.  Available
 	// options are to create a new database or revive an existing one.
 	InitPolicy CommunalInitPolicy `json:"initPolicy"`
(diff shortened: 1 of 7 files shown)


roypaulin:
This is not complete as there is still a few descriptors to add but  I wanted a review on what I have done until now.

roypaulin:
To understand x-descriptors in detail go to https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md

api/v1beta1/verticadb_types.go
```diff
@@ -39,12 +39,12 @@ const VerticaDBAPIVersion = "vertica.com/v1beta1"
...
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default:=IfNotPresent
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:imagePullPolicy"
 	// This dictates the image pull policy to use
 	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:io.kubernetes:Secret"}
```
spilchen, line 47: This takes a list of secrets.  Is it able to handle that?

api/v1beta1/verticadb_types.go
```diff
@@ -115,21 +115,21 @@ type VerticaDBSpec struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	// Ignore the cluster lease when doing a revive or start_db.  Use this with
 	// caution, as ignoring the cluster lease when another system is using the
 	// same communal storage will cause corruption.
 	IgnoreClusterLease bool `json:"ignoreClusterLease,omitempty"`
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default:=Create
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:Create","urn:alm:descriptor:com.tectonic.ui:select:Revive"}
```
spilchen, line 126: We should support 'ScheduleOnly' too.

api/v1beta1/verticadb_types.go
```diff
@@ -94,15 +94,15 @@ type VerticaDBSpec struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:io.kubernetes:Secret"}
 	// An optional name for a secret that contains the password for the
 	// database's superuser. If this is not set, then we assume no such password
 	// is set for the database. If this is set, it is up the user to create this
 	// secret before deployment. The secret must have a key named password.
 	SuperuserPasswordSecret string `json:"superuserPasswordSecret,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:io.kubernetes:Secret"}
```
spilchen, line 105: Why is this one defined with a brace?  i.e. {"..."} instead of "...".

spilchen (commented):
Are you going to include all of the x-descriptors in 1 PR or over multiple?

api/v1beta1/verticadb_types.go
roypaulin: Yeah, it lets you specify a list of secrets but unlike other secret fields where it allows you to choose from a pool of existing secrets, here you have to manually write each secret name. So I think this x-descriptor does not apply here well because it is an array but I will see if there is a work around or if we have to let the user write them manually

api/v1beta1/verticadb_types.go
roypaulin: ok

api/v1beta1/verticadb_types.go
roypaulin: Yeah, I will change that. Braces are for many not one.

roypaulin:
I will do most of them on this PR and if later there is a need to add others I will just open another PR.

roypaulin:
There are certain changes that can't be done automatically through markers so they are to be added  directly on the CSV.

roypaulin:
Any idea if and how markers can be split through multiple lines? Because each time there will be an error with some line length.

spilchen:
Try adding this comment near the top of the file -- right after the copyright.

```
// nolint:lll
```

This tells the linter to ignore the line limit check.

roypaulin:
the time on my linux was wrong so that's why there was no notification for the latest commit. so this comment is just to notify.

api/v1beta1/verticadb_types.go
```diff
@@ -374,7 +383,7 @@ type CommunalStorage struct {
```
spilchen, line 383: HadoopConfig should have `xDescriptors="urn:alm:descriptor:io.kubernetes:ConfigMap`

api/v1beta1/verticadb_types.go
```diff
@@ -480,7 +489,7 @@ type Subcluster struct {
 	ServiceType corev1.ServiceType `json:"serviceType,omitempty"`
```
spilchen, line 489: Can we add a selector for this one?  We should select between ClusterIP, NodePort, and LoadBalancer

api/v1beta1/verticadb_types.go
```diff
@@ -496,6 +505,21 @@ type Subcluster struct {
 	ExternalIPs []string `json:"externalIPs,omitempty"`
 }
 
+type Affinity struct {
```
spilchen, line 508: Lets add a comment to explain why this exists and why we don't use the Kubernetes one.

api/v1beta1/verticadb_types.go
```diff
@@ -317,7 +326,7 @@ type CommunalStorage struct {
 	Endpoint string `json:"endpoint"`
```
spilchen, line 326: This should have a dependency on initPolicy too.  It isn't needed for ScheduleOnly.

api/v1beta1/verticadb_types.go
```diff
@@ -317,7 +326,7 @@ type CommunalStorage struct {
...
 
 	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:Secret"
```
spilchen, line 329: This isn't needed if ScheduleOnly

spilchen (commented):
This looks good.  I had some suggestions about other x-descriptors we could add.

api/v1beta1/verticadb_types.go
```diff
@@ -302,7 +311,7 @@ type CommunalStorage struct {
```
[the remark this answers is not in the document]
roypaulin: Yeah I wanted to do something like that but I think  fieldDependency does not support  multi-CONTROL_FIELDs yet.

roypaulin:
Right now fieldDependency supports only one dependency so there is not a way to tell that we want a field to be relevant only if initPolicy is either Create or Revive.

api/v1beta1/verticadb_types.go
```diff
@@ -251,6 +251,12 @@ type VerticaDBSpec struct {
 	SSHSecret string `json:"sshSecret,omitempty"`
 }
 
+// Is used instead of corev1.LocalObjectReference and behaves the same.
```
spilchen, line 254: minor: comments like these should start with the name of the type, which is LocalObjectReference.  It is a golang thing.  Same comment for Affinity.

roypaulin:
About x-descriptors I think I will stop there for now and when there will be a need to add more I will open another PR.

spilchen (approved):
Changes look good.  Thanks for doing the revisions.

PR #118, opened 2021-12-08
Updated version link in vlogger README

rs-vertica:
previous link: https://github.com/vertica/vertica-kubernetes/blob/**main**/docker-vlogger/Dockerfile

updated link: https://github.com/vertica/vertica-kubernetes/blob/**v1.0.0**/docker-vlogger/Dockerfile

### DIFF
--- a/docker-vlogger/README.md
+++ b/docker-vlogger/README.md
@@ -5,7 +5,7 @@
 
 # Supported Tags
 
-* [1.0.0, latest](https://github.com/vertica/vertica-kubernetes/blob/v1.0.0/docker-vlogger/Dockerfile)
+* [1.0.0, latest](https://github.com/vertica/vertica-kubernetes/blob/v1.2.0/docker-vlogger/Dockerfile)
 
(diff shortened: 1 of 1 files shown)


qindotguan (approved):
Looks good to me.

docker-vlogger/README.md
```diff
@@ -5,7 +5,7 @@
 
 # Supported Tags
 
-* [1.0.0](https://github.com/vertica/vertica-kubernetes/blob/main/docker-vlogger/Dockerfile)
+* [1.0.0, latest](https://github.com/vertica/vertica-kubernetes/blob/v1.0.0/docker-vlogger/Dockerfile)
```
spilchen, line 8: Unfortunately, the repo tag will be different than the docker tag.  We have a different release cadence for the vlogger and the repo as we don't expect the vlogger to change much.  So it is pretty confusing, but we should link to the v1.2.0 repo -- this tag will be created on December 22.

PR #120, opened 2021-12-08
Add imageChangeStatus message

spilchen:
This is prep for online upgrade. This adds a new status message
that we will use to report the phase that the upgrade is in.
This also has a fix that will retry status changes for any
transient error. You may have seen the error I'm talking about
-- it usually complains about an update failing because we are
not using the latest copy of the object.


roypaulin (approved):
looks good

PR #122, opened 2021-12-09
Refactor of offline image change reconciler

spilchen:
This pulls out some common functions with the new online image change reconciler.  A new Go struct, ImageChangeInitiator, was created that handles the common logic.

This also adds new parameters to the CR to allow for control of what type of upgrade you want.  No reconciler has been added yet for online image change.


roypaulin (approved):
looks good

PR #124, opened 2021-12-10
Add clusterPermissions to CSV to run pods in priviledged mode

roypaulin:
This uses clusterRole and clusterRoleBinding files to add and populate clusterPermissions field in the CSV. In the CSV we will something like this:

```
clusterPermissions:
      - rules:
        - apiGroups:
          - security.openshift.io
          resourceNames:
          - privileged
          resources:
          - securitycontextconstraints
          verbs:
          - use
        serviceAccountName: verticadb-operator-controller-manager
```

### DIFF
--- a/config/clusterpermissions/cluster_role.yaml
+++ b/config/clusterpermissions/cluster_role.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: openshift-cluster-role
+rules:
+- apiGroups:
+  - security.openshift.io
+  resources:
+  - securitycontextconstraints
+  resourceNames: 
+  - privileged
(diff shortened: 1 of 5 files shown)


config/clusterpermissions/cluster_role.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
spilchen, line 1: I don't see how we are adding the ClusterRoleBinding to the CSV.  When I generated the bundle on your branch, I didn't see anything about clusterPermissions in the CSV.  Also, shouldn't there be an update to the CSV with clusterPermissions in it?

spilchen: I understand this now.  Thanks for clarifying.

PR #125, opened 2021-12-13
Flow subcluster info for online-upgrade process

spilchen:
The online-upgrade process needs more information about the subcluster.  This introduces a new data structure SubclusterHandle that has the Subcluster struct that is stored in etcd plus additional runtime info that is needed for the online-upgrade process.  

There is two parts to this change.  First is fetching the additional information through sc_finder.  The second is flowing the new SubclusterHandle through various functions.

In a subsequent PR, I will start to fill out some of the functions in onlineupgrade_reconcile.go using the data collected in SubclusterHandle.


roypaulin (commented):
looks good. Will approve after the tests.

PR #126, opened 2021-12-14
Add scorecard tests to CI

roypaulin:
This simply adds scorecard tests check to the CI. A new target has been added in the Makefile for that.

### DIFF
--- a/.github/workflows/scorecardtests.yml
+++ b/.github/workflows/scorecardtests.yml
@@ -9,19 +9,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
         
-    - name: Set up Go
-      uses: actions/setup-go@v2
-      with:
-        go-version: 1.16.2
-    
-    - name: Install kubectx   
-      shell: bash
-      env:
-          KUBECTX_URL: https://github.com/ahmetb/kubectx/releases/download/v0.9.3/kubectx
-      run: |
-        curl -fsSLO $KUBECTX_URL
-        chmod +x kubectx
-        sudo install -o root -g root -m 0755 kubectx /usr/local/bin/kubectx
+    - name: Set up Go and kubectx
+      uses: ./.github/actions/setup-go-kctx
         
     - name: Create kind cluster
       run: |
@@ -30,4 +19,4 @@ jobs:
         kubectx kind-vertica
 
     - name: Run scorecard tests
-      run: make scorecard-tests
+      run: make run-scorecard-tests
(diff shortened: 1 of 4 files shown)


.github/workflows/scorecardtests.yml
```diff
@@ -0,0 +1,33 @@
...
+on: [push, pull_request]
+
+jobs:
+
+  sct:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+        
+    - name: Set up Go
```
spilchen, line 12: Some of these steps are common with the e2e tests.  Can we setup a new action that we use in this new test and in the e2e tests?  See .github/actions/setup-e2e for an example of the action I created for the e2e tests.

Makefile
```diff
@@ -428,3 +428,6 @@ operator-sdk: $(OPERATOR_SDK)  ## Download operator-sdk locally if necessary
 $(OPERATOR_SDK):
 	curl --silent --show-error --location --fail "https://github.com/operator-framework/operator-sdk/releases/download/v1.10.1/operator-sdk_linux_amd64" --output $(OPERATOR_SDK)
 	chmod +x $(OPERATOR_SDK)
+
+scorecard-tests: bundle ## Run the scorecard tests
```
spilchen, line 432: Can we call the new target `run-scorecard-tests`?  This will keep it common with the other test targets (run-unit-tests, run-int-tests, etc.).

spilchen (commented):
I just had a couple of suggestions.

roypaulin:
Do you think it is fine like this or was it better to not call the new action from setup-e2e action and call both actions for e2e tests?

Makefile
```diff
@@ -428,3 +428,6 @@ operator-sdk: $(OPERATOR_SDK)  ## Download operator-sdk locally if necessary
 $(OPERATOR_SDK):
 	curl --silent --show-error --location --fail "https://github.com/operator-framework/operator-sdk/releases/download/v1.10.1/operator-sdk_linux_amd64" --output $(OPERATOR_SDK)
 	chmod +x $(OPERATOR_SDK)
+
+run-scorecard-tests: bundle ## Run the scorecard tests
+	$(OPERATOR_SDK) scorecard bundle
```
spilchen, line 433: There is a --wait-time duration parameter.  It currently defaults to 30s.  Can we change it to 2m?  I'm concerned about slow environments.

spilchen:
> Do you think it is fine like this or was it better to not call the new action from setup-e2e action and call both actions for e2e tests?

I like what you did as it cuts down on the amount of code duplication.

spilchen (approved):
Looks good.

PR #127, opened 2021-12-16
Online image change: handling of the standby subcluster

spilchen:
This is another PR for online-upgrade.  It will handle creation and removal of the standby subcluster during the online-image change process.

- new state was added to the vapi.Subcluster for this.  Originally, I was planning to keep most of this in the SubclusterHandle struct, but we already pass around vapi.Subcluster so it made it easier to have it their
- new status conditions for offline and online image change.  These are intended to be used by the operator to know what image change to continue with once an image change has started
- filled out more of the logic in onlineimagechange_reconciler.go.  It will scale-out a new standby subcluster for each primary, then scale them down when we are finishing the image change.
- moved more logic into imagechange.go that is common between online and offline image change
- restart logic was changed to allow option to restart read-only nodes.  When restarting for online, we will skip the read-only modes.  Offline restarts everything.

### DIFF
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -798,7 +798,7 @@ func MakeVDB() *VerticaDB {
 			DBName:     "db",
 			ShardCount: 12,
 			Subclusters: []Subcluster{
-				{Name: "defaultsubcluster", Size: 3, ServiceType: corev1.ServiceTypeClusterIP},
+				{Name: "defaultsubcluster", Size: 3, ServiceType: corev1.ServiceTypeClusterIP, IsPrimary: true},
 			},
 		},
 	}
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -512,6 +512,13 @@ type Subcluster struct {
 	// for.  This is state internally managed for online image change.
 	StandbyParent string `json:"standbyParent,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:hidden"
+	// This allows a different image to be used for the subcluster than the one
+	// in VerticaDB.  This is intended to be used internally by the online image
+	// change process.
+	ImageOverride string `json:"imageOverride,omitempty"`
+
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// A map of label keys and values to restrict Vertica node scheduling to workers
 	// with matching labels.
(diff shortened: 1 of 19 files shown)


api/v1beta1/verticadb_types.go
```diff
@@ -499,9 +499,29 @@ type Subcluster struct {
...
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:hidden"
+	// Internal state that indicates whether this is a standby subcluster for a
+	// primary.  Standby are transient subclusters that are created during an
+	// online image change.
+	IsStandby bool `json:"isStandby,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:hidden"
+	// If a standby, this is the name of the primary subcluster it is a standby
```
roypaulin, line 511: Isn't there a missing word on this line?

pkg/controllers/builder.go
```diff
@@ -384,26 +393,26 @@ func getStorageClassName(vdb *vapi.VerticaDB) *string {
...
 			Name:        nm.Name,
 			Namespace:   nm.Namespace,
-			Labels:      makeLabelsForObject(vdb, scHandle),
+			Labels:      makeLabelsForObject(vdb, sc),
 			Annotations: makeAnnotationsForObject(vdb),
 		},
 		Spec: appsv1.StatefulSetSpec{
 			Selector: &metav1.LabelSelector{
-				MatchLabels: makeSvcSelectorLabels(vdb, scHandle),
+				MatchLabels: makeSvcSelectorLabels(vdb, sc),
```
roypaulin, line 406: Is sticking to sc instead of scHandle only due to the fact that vapi.Subcluster is already passed around?

spilchen: Yes, so this undoes some of the work that I did earlier.  Overall, I think it makes it simpler.

api/v1beta1/verticadb_types.go
spilchen: How about this?

```
// If this is a standby, this is the name of the primary subcluster the
// subcluster was created for.  This is state internally managed for
// an online image change. 
```

roypaulin: Yeah I think it is better. Maybe because my english is no so good but I feel like having 2 consecutive occurrences of subscluster is repetitive. What do you thin about:
```

// If this is a standby subcluster, this is the name of the primary subcluster 
// it was created for.  This is state internally managed for
// an online image change. 
```

pkg/controllers/builder.go
roypaulin: Ok got it.

roypaulin:
looks good. let's wait for the checks.

roypaulin:
looks good. Thanks.

roypaulin (approved):

PR #128, opened 2021-12-17
Online image change: manage client traffic over multiple subclusters

spilchen:
This adds manipulating of the service objects during an online image change.  It will route client traffic to the standby when we are upgrading, then reroute the traffic back to the original subclusters when completing the upgrade.

As a side affect this also adds the ability for multiple subclusters to share the same service object.  This has benefits outside of the online image change process.


roypaulin (approved):
looks good

PR #131, opened 2021-12-20
Online image change: route to temporary subcluster to keep online

spilchen:
This is the next set of changes for online image change.  It will route to a temporary subcluster, called a transient, so that client connections connect to an up node.  It will automatically route back to the original subcluster once the subcluster is back up.  It will also process secondary subclusters that they are brought back up.

This includes some rework in the onlineimagechange_reconciler to cut down on the amount of code duplication.

Added the ability to specify a template for the transient cluster.  This will get created when the image change starts and will get cleanup up when the image change is done.

### DIFF
--- a/pkg/controllers/builder.go
+++ b/pkg/controllers/builder.go
@@ -29,8 +29,8 @@ import (
 )
 
 const (
-	SuperuserPasswordPath   = "superuser-passwd"
-	TransientSubclusterName = "transient"
+	SuperuserPasswordPath          = "superuser-passwd"
+	DefaultTransientSubclusterName = "transient"
 )
 
 // buildExtSvc creates desired spec for the external service.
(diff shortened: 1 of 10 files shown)


pkg/controllers/builder.go
```diff
@@ -586,21 +586,35 @@ func getK8sAffinity(a vapi.Affinity) *corev1.Affinity {
...
+		PriorityClassName: vdb.Spec.TransientSubclusterTemplate.PriorityClassName,
+		Tolerations:       vdb.Spec.TransientSubclusterTemplate.Tolerations,
+		Resources:         vdb.Spec.TransientSubclusterTemplate.Resources,
 		ServiceType:       sc.ServiceType,
 		ServiceName:       sc.GetServiceName(),
 		NodePort:          sc.NodePort,
 		ExternalIPs:       sc.ExternalIPs,
 	}
 }
+
```
roypaulin, line 607: Shouldn't we add a comment to describe the function? below too.

roypaulin (approved):

PR #133, opened 2021-12-21
Online image change: allow reuse of existing sc for temporary routing

spilchen:
This is the next set of changes for online image change.  The CR parm TransientSubclusterTemplate was renamed to TemporarySubclusterRouting.  The parm can be used to provide a template of a subcluster to use for temporary routing while subclusters are down.  Or the parm can also be used to specify an existing subcluster.  This later option may be useful to those that want to just reuse existing subclusters.

This PR also cleans up how we route traffic to the subclusters.  We previously had relied on a 'transient' label.  But now we route to service name or subcluster name.

### DIFF
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -197,7 +197,7 @@ type VerticaDBSpec struct {
 	// accept traffic while the other subclusters restart.  The designated
 	// subcluster is specified here.  The name of the subcluster can refer to an
 	// existing one or an entirely new subcluster.  If the subcluster is new, it
-	// will be exist only for the duration of the image change.
+	// will exist only for the duration of the image change.
 	TemporarySubclusterRouting SubclusterSelection `json:"temporarySubclusterRouting,omitempty"`
 
 	// +kubebuilder:default:="1"
@@ -304,7 +304,10 @@ type SubclusterSelection struct {
 
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// +kubebuilder:validation:Optional
-	// A new subcluster will be created using this as a template.
+	// A new subcluster will be created using this as a template.  This
+	// subcluster will only exist for the life of the online image change.  It
+	// will accept client traffic for a subcluster that are in the process of
+	// being restarted.
 	Template Subcluster `json:"template,omitempty"`
 }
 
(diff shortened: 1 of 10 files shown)


api/v1beta1/verticadb_types.go
```diff
@@ -192,13 +192,13 @@ type VerticaDBSpec struct {
...
-	// template to create that subcluster.  This subcluster is created at the
-	// beginning of the online image change and is removed when that process
-	// cleans up.
-	TransientSubclusterTemplate Subcluster `json:"transientSubclusterTemplate,omitempty"`
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// When doing an online image change, we designate a subcluster to
+	// accept traffic while the other subclusters restart.  The designated
+	// subcluster is specified here.  The name of the subcluster can refer to an
+	// existing one or an entirely new subcluster.  If the subcluster is new, it
+	// will be exist only for the duration of the image change.
```
roypaulin, line 200: I think that's a typo. should be "will exist" right?

api/v1beta1/verticadb_types.go
```diff
@@ -291,6 +291,23 @@ type LocalObjectReference struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +kubebuilder:validation:Optional
+	// Names of existing subclusters to use for temporary routing of client
+	// connections.  The operator will use the first subcluster that is online.
+	Names []string `json:"names,omitempty"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +kubebuilder:validation:Optional
+	// A new subcluster will be created using this as a template.
+	Template Subcluster `json:"template,omitempty"`
```
roypaulin, line 308: Will we always use a single subcluster as temporary?

api/v1beta1/verticadb_types.go
```diff
@@ -916,3 +933,16 @@ func (s *Subcluster) GetServiceName() string {
 	}
 	return s.ServiceName
 }
+
+// RequiresTransientSubcluster checks if an online image change requires a
+// transient subcluster.  A transient subcluster exists if the template is
+// filled out or the name of the temporary routing subcluster doesn't exist.
```
roypaulin, line 939: I don't quite get this line.

api/v1beta1/verticadb_types.go
spilchen: Yes, if we use the template a single subcluster is created.  If we reuse an existing subcluster 1 or 2 subclusters can be specified -- the 2nd subcluster is only used when we are upgrading the 1st.

I'll add more comments here.

api/v1beta1/verticadb_types.go
spilchen: If subclusters were specified by name, but none of them actually exist, I was going to have it default to creating a transient subcluster from a template.

roypaulin: Ok thanks

roypaulin (approved):

PR #135, opened 2021-12-23
Online image change: maintain status message

spilchen:
This fills in the Status.ImageChangeStatus message as we progress through the online image change.  It also reworks the status message updates we do for offline image change to share the same infrastructure.

I'm also including a change to keep the ssh key in the Vertica container stable between builds.  The ssh key is just used for communication between the Vertica nodes.  A stable key allows Vertica nodes from a different container image to be able to communicate.  This becomes an issue when doing an online image change because the Vertica pods are either running the old or new image, yet they all have to talk to each other.


roypaulin (approved):
looks good

PR #138, opened 2022-01-07
Online image change: e2e tests

spilchen:
This adds e2e tests for online upgrade.  I added a new directory (e2e-11.1) since we cannot run these yet in our GitHub CI.  This directory will contain all of the e2e tests that must run on a Vertica server 11.1 or higher.  We can fold these into the main e2e after 11.1 is GA'd.

Also, we use another vertica image in e2e tests called BASE_VERTICA_IMG.  The online upgrade tests will change the image from BASE_VERTICA_IMG to VERTICA_IMG.

During the testing, a few issues were found that I have fixes for:
- ObjReconciler, DBAddNodeReconciler and DBAddSubclusterReconciler will work only on the transient subcluster.  We previously added the transient to the VerticaDB, then run these reconciler as-is.  But it can pick up other changes that can interfere with the upgrade -- namely scaling out changes.  So I no longer update the VerticaDB with the transient, and run the reconciler's just with the transient subcluster.
- avoid creating the transient if the cluster is down.  We need the cluster to be up to create the transient, so skipping that entire part if the cluster is down
- Restart reconciler will avoid restarting pods for the transient subcluster.  This subcluster intentionally stays on the old image and there is no way to restart an old image if the primaries are already updated to the new image.


roypaulin (approved):
Looks good. Do you plan to do another demo on the online image change at some point?

spilchen:
Thanks Roy.  Yeah, I was thinking of recording a demo and including it in the spec.

PR #141, opened 2022-01-14
Online image change: new webhook rules

spilchen:
This adds rules to the webhook for online image change:
- prevent changes to upgradePolicy when imageChange is in progress
- transient subcluster template: isPrimary == false, name cannot be an existing subcluster, size > 0 if name present
- transient subcluster cannot be added/removed during an online image change
- if multiple subclusters share a ServiceName, service specific things must be common between them (serviceType, NodePort, externalIPs, etc).

### DIFF
--- a/api/v1beta1/verticadb_webhook.go
+++ b/api/v1beta1/verticadb_webhook.go
@@ -688,7 +688,7 @@ func (v *VerticaDB) matchingServiceNamesAreConsistent(allErrs field.ErrorList) f
 				}
 			}
 		}
-		// Set a flag so that we don't porcess this service name in another subcluster
+		// Set a flag so that we don't process this service name in another subcluster
 		processedServiceName[sc.ServiceName] = true
 	}
 	return allErrs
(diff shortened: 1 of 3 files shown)


api/v1beta1/verticadb_webhook.go
```diff
@@ -610,3 +653,86 @@ func (v *VerticaDB) isSubclusterTypeIsChanging(oldObj *VerticaDB) (ok bool, scIn
...
+				}
+				if sc.ServiceType != osc.ServiceType {
+					err := field.Invalid(fieldPrefix.Child("serviceType").Index(i),
+						sc.ServiceType,
+						"serviceType doesn't match other subcluster(s) sharing the same serviceName")
+					allErrs = append(allErrs, err)
+				}
+			}
+		}
+		// Set a flag so that we don't porcess this service name in another subcluster
```
roypaulin, line 691: I think you meant "process" here

roypaulin (commented):
looks good

PR #142, opened 2022-01-18
Online image change: changes for new admintools options in 11.1.0

spilchen:
- when running AT start_db, we need to run from one of the primary nodes.  It won't work if we try from a read-only node that isn't being restarted.
- when calling re_ip, use the --force option.  This option is new in 11.1.0, so we needed conditional logic to know when we could use this.
- when calling start_db, we use the host list.  This option exists first in 11.0.1, so like reip, we needed conditional logic to know when we can use that option

One thing that isn't related to the title of this PR is some new logic needed in DBAddNodeReconicler.  That reconciler will now requeue if some pods aren't yet ready.  This was needed so that the upgrade properly waits for the transient subcluster to scale out.  Prior to this change, it was possible that the image change went ahead and restarted the primaries before the transient was up.  This should be solved now.

### DIFF
--- a/pkg/controllers/dbaddnode_reconcile.go
+++ b/pkg/controllers/dbaddnode_reconcile.go
@@ -83,7 +83,7 @@ func (d *DBAddNodeReconciler) reconcileSubcluster(ctx context.Context, sc *vapi.
 			}
 		}
 		if unknownState {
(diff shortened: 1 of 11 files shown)


roypaulin (commented):
looks good

roypaulin (approved):

PR #143, opened 2022-01-20
Make logging more consumable

### LINKED ISSUE #56: Configurable log level
Currently the log level could not be configured through helm variables. Many times the customer would like to get the ERROR logs only and not INFO or DEBUG messages.
Please make it configurable instead of the hard-coded version.

### LINKED ISSUE #116: Vertica operator: Write logs to a file
We are trying to integrate the Vertica operator container into our log setup which means we need to send its logs to our main syslog server (with a bunch of additional metadata). Currently it can't connect to a syslog server directly, it can only log to stdout so we'd need to use a node level logging agent to solve this. Due to internal security requirements this isn't an option in our case.

We would like to work around this with a Filebeat based sidecar container which could send logs to the syslog server with all the required fields. This setup would require the operator container to write its logs to a file on a volume shared between the containers.

Please add the ability for the Vertica operator to log to a file:

* The ability to specify a file inside the Vertica operator container as the target.
* and the ability to handle log rotation.
* We are using a modified version of the helm chart from this repository so an official way to add the mentioned sidecar container and volume isn't important in our case.

roypaulin:
Before, logging for the operator wrote everything to stdout.  The logging levels were not tunable, all of the error, warning and info messages are written out. 
Now thanks to new helm parameters, we can log to file too and we built log rotation into the operator. We add 2 extra parameters, stdout and devMode that we use respectively to enforce logging to stdout and to select developement/production mode. 
The scenario are as follow:

-    if no filepath is specified, we log to stdout only and keep as level debug. So the ability to select the level is something supported only when a filepath is given. (That could be easily changed)
-    if filepath is invalid or you do not have access to the file you will have error msgs all over stdout .
-    if a valid filepath is specififed and stdout is false, the logging will be only to file, if stdout is true, the logging will be to both file and stdout

First it was decided to define more helm parameters for sidecar and volume that are used for testing, as the manager container is distroless, but I decided instead to patch to the already installed operator the sidecar container and the volume and then verify that  the log has been done to the file using that sidecar.           

Closes #56 
Closes #116


roypaulin:
I use busybox for the sidecar container but I don't think we want that. Any suggestion for a sidecar container?

Also I ran into a funlen issue as the 50 statements/func limit has been exceeded in the main (now there are 53). So I am wondering whether we can move the flag parameters setting inside init func in order to decrease the number of statements in the main.

spilchen:
> Also I ran into a funlen issue as the 50 statements/func limit has been exceeded in the main (now there are 53). So I am wondering whether we can move the flag parameters setting inside init func in order to decrease the number of statements in the main.

Yes, by all means move to a separate function to keep the code readable.

spilchen (commented):
Some initial comments.

helm-charts/verticadb-operator/values.yaml
```diff
@@ -37,6 +37,27 @@ webhook:
...
+  # maxFileAge, when logging to a file, is the maximum age, in days,
+  # of the logging before log rotation gets rid of it.
+  maxFileAge: 7
+  # maxFileRotation, when logging to a file, is the maximum number of files 
+  # that are kept in rotation before the old ones are removed.
+  maxFileRotation: 3
+  # level is the minimum logging level. Valid values are: debug, info, warn, and error
+  level: info
+  # stdOut is used to enforce logging to stdout.
+  stdOut: false
```
spilchen, line 56: This is a hard parameter to explain.  Instead of a new one, could we just piggyback on dev.  If dev is true, we always log to stdout, even if a file path is choosen.  If dev is false, we only log to the file or stdout never both.

scripts/create-patch.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
spilchen, line 1: Does this need to be a script?  Could it be just a yaml manifest in the custom-cert-webhook directory?

scripts/create-patch.sh
```diff
@@ -0,0 +1,42 @@
...
+  template:
+    spec:
+      containers:
+      - name: manager
+        volumeMounts:
+        - mountPath: /logs
+          name: shared-data
+      - args:
+        - -c
+        - while true; do tail -n 1 /var/logs/try.log; sleep 5;done
```
spilchen, line 29: Can we be consistent with the log directory between the two containers?  Any reason why couldn't make it /logs in both?

tests/e2e/custom-cert-webhook/12-check-log-file.yaml
```diff
@@ -0,0 +1,20 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      PN=$(kubectl get pods -n $NAMESPACE --no-headers -o custom-columns=":metadata.name" | grep verticadb ) \
+      && kubectl wait --for=condition=Ready=true pod/$PN -n $NAMESPACE \
+      && kubectl exec $PN -n $NAMESPACE -c sidecar-container -- tail -n 1  var/logs/try.log
```
spilchen, line 20: Can we add a leading '/' to the path?

cmd/operator/main.go
```diff
@@ -81,11 +102,75 @@ func getIsWebhookEnabled() bool {
...
+		MaxAge:     age, // days
+	}
+	return zapcore.AddSync(lumberJackLogger)
+}
+
+// getZapcoreLevel takes the level as string and returns the corresponding
+// zapcore.Level. If the string level is invalid, it returns the default
+// level
+func getZapcoreLevel(lvl string) zapcore.Level {
+	lvls := map[string]zapcore.Level{
```
spilchen, line 130: I think you can use `zapcore.UnmarshalText` here instead of having the map.

cmd/operator/main.go
```diff
@@ -81,11 +102,75 @@ func getIsWebhookEnabled() bool {
...
+
+// getLogger is a wrapper that calls other functions
+// to build a logger.
+func getLogger(logArgs Logging, devMode bool) *zap.Logger {
+	encoderConfig := getEncoderConfig(devMode)
+	writes := []zapcore.WriteSyncer{}
+	lvl := zap.NewAtomicLevelAt(zap.DebugLevel)
+	if logArgs.FilePath != "" {
+		w := getLogWriter(logArgs.FilePath, logArgs.MaxFileAge, logArgs.MaxFileSize, logArgs.MaxFileRotation)
+		lvl = zap.NewAtomicLevelAt(getZapcoreLevel(logArgs.Level))
```
spilchen, line 152: We will want the level to be configurable, even when not using FilePath.

cmd/operator/main.go
```diff
@@ -81,11 +102,75 @@ func getIsWebhookEnabled() bool {
...
+}
+
+// getLogger is a wrapper that calls other functions
+// to build a logger.
+func getLogger(logArgs Logging, devMode bool) *zap.Logger {
+	encoderConfig := getEncoderConfig(devMode)
+	writes := []zapcore.WriteSyncer{}
+	lvl := zap.NewAtomicLevelAt(zap.DebugLevel)
+	if logArgs.FilePath != "" {
+		w := getLogWriter(logArgs.FilePath, logArgs.MaxFileAge, logArgs.MaxFileSize, logArgs.MaxFileRotation)
```
spilchen, line 151: Can we just pass in the Logging struct instead of each individual field?

spilchen (commented):
I like the approach.  Here are some remaining comments.

scripts/create-patch.sh
roypaulin: No. I thought  it was not a good idea to change the dir structure. I will add a manifest in custom-cert-webhook

helm-charts/verticadb-operator/values.yaml
roypaulin: Yes sounds good

tests/e2e/custom-cert-webhook/12-check-log-file.yaml
roypaulin: Yes I forgot that, will add it.

cmd/operator/main.go
roypaulin: Ok I will take a look, thanks.

cmd/operator/main.go
roypaulin: Yes I am already changing that.

helm-charts/verticadb-operator/values.yaml
```diff
@@ -52,11 +52,15 @@ logging:
...
-  # stdOut is used to enforce logging to stdout.
-  stdOut: false
+  # dev Enables development mode if true and production mode otherwise and also affects
+  # logs format. Afew differences on logging will be: in dev mode the the level is displayed
+  # in capital letters, the time is in a ISO8601-formatted string, in prod, the level is in
+  # lower case, the time is in epoch time format.
+  # example:
+  # dev ->  2022-01-21T12:50:10.078Z        INFO    controller-runtime.certwatcher  Starting certificate watcher
+  # prod -> 1.6427683530752437e+09  info    controller-runtime.certwatcher  Starting certificate watcher
+  dev: false
```
spilchen, line 62: I think dev mode disables sampling of logging messages and produces stack traces more liberally.  The are probably the key things to mention.

scripts/create-helm-charts.sh
```diff
@@ -51,12 +51,11 @@ sed -i 's/resources: template-placeholder/resources:\n          {{- toYaml .Valu
 
 # 6.  Template the logging
 sed -i "s/--filepath=/--filepath={{ .Values.logging.filePath }}/" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
-sed -i "s/--maxfilesize=/--maxfilesize={{ .Values.logging.maxFileSize }}/" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
-sed -i "s/--maxfileage=/--maxfileage={{ .Values.logging.maxFileAge }}/" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
-sed -i "s/--maxfilerotation=/--maxfilerotation={{ .Values.logging.maxFileRotation }}/" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
-sed -i "s/--level=/--level={{ .Values.logging.level }}/" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
-sed -i "s/--dev=/--dev={{ .Values.dev }}/" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
-sed -i "s/--stdout=/--stdout={{ .Values.logging.stdOut }}/" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
+sed -i "s/--maxfilesize=500/--maxfilesize={{ .Values.logging.maxFileSize }}/" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
```
spilchen, line 54: We are duplicating these defaults around a lot.  Can we do pattern matching for the value of each of the parms?

roypaulin: Sorry what do you mean by pattern matching?

helm-charts/verticadb-operator/values.yaml
roypaulin: Yeah. Those are more meaningful.

scripts/create-helm-charts.sh
roypaulin: You mean with reg exp?

spilchen: Yeah, add a reg exp portion so that we don't hard code the default in the pattern.

cmd/operator/main.go
```diff
@@ -191,7 +209,12 @@ func getLogger(logArgs Logging) *zap.Logger {
 		zapcore.NewMultiWriteSyncer(writes...),
 		lvl,
 	)
-	return zap.New(core)
+	opts = append(opts, getStackTrace(logArgs.DevMode))
+	if !logArgs.DevMode {
+		// This enables sampling only in prod
+		core = zapcore.NewSamplerWithOptions(core, time.Second, First, ThereAfter)
```
roypaulin, line 215: I don't know much about sampling so those values were taken from  [NewProductionConfig()](https://github.com/uber-go/zap/blob/feeb9a050b31b40eec6f2470e7599eeeadfe5bdd/config.go#L119). How important is sampling?

spilchen: Why can't we use `NewProductionConfig`?  Sampling will help reduce IO and CPU cost of the logger.  So I think its something that we want.

roypaulin: To create a zap.Logger there are 3 options: 

1. using  `NewProductionConfig/NewDevelopmentConfig` + build() or simply `NewProduction/NewDevelopment` if you don't want to configure manually. This is quite easy as most of the configuration has already been done but it is less flexible.
2. defining a Config struct and then do all the configuration manually. This is used when you really wants to configure at a deep level. functions mentionned above used a Config struct. Then use `Build()` method to create a logger.
3. using `New(core zapcore.Core,options ...Option)` to create a logger from a `zapcore.Core.` This is the most flexible but some configuration are not done automatically (like stack trace) and you need to add them through options.

There is not one thing you can do with one option that you cannot with the others but of Course the level of difficulty is different. In our case we are using lumberjack so the better option is the 3rd as recommended by zap. The use of lumberjack is what makes me choose the 3rd.

I think there is still a way to use NewProductionConfig by using some of these [functions](https://pkg.go.dev/go.uber.org/zap@v1.15.0#Option) to pass the zapcore.Core as option to the `Build()` method but this will require to define some global variables and to change some functions. So basically replacing/altering the current Core. I think it is more complicated and the code is less organized.

spilchen: Thanks for the detailed explanation.  I agree with your approach of using option 3.

Back to your original question, I think taking the sampling settings you copied from `NewProductionConfig` makes the most sense.

spilchen (approved):
Looks good.

PR #144, opened 2022-01-21
Online image change: adding drain logic

spilchen:
This adds drain logic so that we wait for active connections to disappear before taking down a subcluster.  I added finer granular messaging for imageChangeStatus so that we will have a clear idea if it is waiting for the drain of a particular subcluster.

This change involves sorting the output from sc_finder.  This was necessary to match up the status message with order that we will process the subclusters.

Also including a fix that waits for the transient pod to be in a ready state.  There was a small timing window where we started to route client traffic to the transient before it was ready.  The ready probe is run every 10 seconds, so there was a window where vertica was up but k8s didn't yet know about it.

A new e2e test was added to make sure draining works for the primary and secondary subclusters.


roypaulin (approved):
looks good

PR #146, opened 2022-01-27
Allow env vars to be set in the CR

### LINKED ISSUE #137: Make it possible to set env variables for the container env list in the generated statefulset
Unfortunately the admintools process does not get the environment variables what are set by the entrypoint.sh or /etc/environment or ~/.profile or ~/.bash_profile or ~/.bashrc, when it is called by the vertica operator through kubernetes exec request.

This happens because the kubernetes exec request will set environment variables only from the env list from the statefulset and the original env variables what are declared in the dockerfile.

In our organization it is a requirement to set TZ environment variable as a deployment variable during the helm chart deployment, and to do so we have to set it as an env variable in the container env list in the generated statefulset.

It is enough for us, if the annotation list is converted to env list too, and CRD change is not neccessary.

roypaulin:
This PR set the environment variables in the container at runtime. This is done by encoding the annotations in the CR as env vars.
For each annotation, all special characters (not a number a letter or underscore) is turned into underscore and all letters are turned to uppercase. 

For e2e tests update-strategy has been modified and instead of creating new files, step 60 has been modified to take into account annotations too.

Closes #137


pkg/controllers/builder.go
```diff
@@ -374,6 +378,21 @@ func makeContainers(vdb *vapi.VerticaDB, sc *vapi.Subcluster) []corev1.Container
 	return cnts
 }
 
+// getEnvVars returns a list of EnvVars from the annotations
+// in the CR
+func getEnvVars(vdb *vapi.VerticaDB) []corev1.EnvVar {
+	envVars := []corev1.EnvVar{}
+	m := regexp.MustCompile(`[^a-zA-Z0-9]`)
+	for k, v := range vdb.Spec.Annotations {
+		name := strings.ToUpper(m.ReplaceAllString(k, "_"))
```
spilchen, line 387: One other case we need to handle is when the annotation starts with a number.  This is a valid annotation, but is invalid as an environment variable.

pkg/controllers/builder.go
```diff
@@ -374,6 +378,21 @@ func makeContainers(vdb *vapi.VerticaDB, sc *vapi.Subcluster) []corev1.Container
 	return cnts
 }
 
+// getEnvVars returns a list of EnvVars from the annotations
+// in the CR
+func getEnvVars(vdb *vapi.VerticaDB) []corev1.EnvVar {
```
spilchen, line 383: Suggest a name like `translateAnnotationsToEnvVars`.  I think `gentEnvVars` is too generic and doesn't say what is actually happening.

pkg/controllers/builder.go
```diff
@@ -356,6 +358,8 @@ func makeServerContainer(vdb *vapi.VerticaDB, sc *vapi.Subcluster) corev1.Contai
 		},
 		VolumeMounts: buildVolumeMounts(vdb),
 	}
+	cnt.Env = append(cnt.Env, getEnvVars(vdb)...)
```
spilchen, line 361: What happens if we have two environment variables with the same name?  We have some ones that already have (e.g. POD_IP).  If we had an annotation called `pod-ip`, what happens?

spilchen, line 361: I think we should add the environment variables to any sidecars too.

tests/e2e/update-strategy/86-verify-env-vars-exist.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Verify the license exists in each of the pods
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: kubectl exec -t -n $NAMESPACE v-ks-1-update-strategy-sc1-0 -- sh -c 'test $TZ = CET'
+  - script: kubectl exec -t -n $NAMESPACE v-ks-1-update-strategy-sc1-0 -- sh -c 'test $VERTICA_COM_GIT_REF = 1fab313k'
+  - script: kubectl exec -t -n $NAMESPACE v-ks-1-update-strategy-sc1-1 -- sh -c 'test $TZ = CET'
+  - script: kubectl exec -t -n $NAMESPACE v-ks-1-update-strategy-sc1-1 -- sh -c 'test $VERTICA_COM_GIT_REF = 1fab313k'
+  - script: kubectl exec -t -n $NAMESPACE v-ks-1-update-strategy-sc1-2 -- sh -c 'test $TZ = CET'
```
spilchen, line 23: I noticed this step failed in the azb run.  It is probably timing related.  I wonder if the statefulset controller is going through another rolling update for some reason?  We should look into that some more.

spilchen (commented):
Looks good.  I like the regex usage.  Just had a few comments.

pkg/controllers/builder.go
roypaulin: Yes it sounds better.

pkg/controllers/builder.go
roypaulin: Good remark. It will overwrite in this case the previous POD_IP value and that is not something we want. So we probably need to disapprove any annotation that translates to one of the internally defined env vars (POD_IP, DATA_PATH, DEPOT_PATH).

spilchen: Could we reorder things so that we add POD_IP etc. _after_ adding the annotations?

roypaulin: Yes that way if any of these env vars is set in the annotations, its value will be overwritten by the correct one.

pkg/controllers/builder.go
roypaulin: Ok do we want to discard completely the annotation or just reformat it into a valid env var?

roypaulin: I would like to discard annotations that do not start with a letter. Do you want to reformat them  instead?

spilchen: Discarding them is fine.

spilchen (commented):
The changes look good.  Lets see if we can figure out why the e2e test failed.

roypaulin:
Looks like all tests passed this time.

roypaulin:
> Can you look at the failure a bit? I don't like to see tests that fail every so often. Maybe run it manually and observe what happens to the pods when we get to the step you added. Does the statefulset controller go through another rolling upgrade cycle? What if we put a sleep at the beginning of the new step, does that increase the likelyhood of hitting the issue?

Took a look on the error. When I run it locally it never fails. When I opened a PR in the forked repo, it failed even before the step I added (at step 75). When I added a sleep before step 86 and opened a PR, e2e (s3) tests failed instead and the others passed. Another thing I did is instead of merging both annotations and license at step 60, I added extra steps to repeat steps 60 to 85 but for annotations but I could never reach a state where all pods are up, they kept been deleted and recreated. However I modified update-strategy by only patching the annotations ( not the license), I ran tests 3 times and I have not hit the issue yet. So I am thinking of adding the test for that to a separated e2e test instead of patching both annotations and license together in update-strategy.

spilchen:
> Another thing I did is instead of merging both annotations and license at step 60, I added extra steps to repeat steps 60 to 85 but for annotations but I could never reach a state where all pods are up, they kept been deleted and recreated.

This is an interesting finding.  It sounds like this was the issue.  Can you open an internal jira for this to remember to investigate this?

> So I am thinking of adding the test for that to a separated e2e test instead of patching both annotations and license together in update-strategy.

This sounds good.

roypaulin:
> This is an interesting finding. It sounds like this was the issue. Can you open an internal jira for this to remember to investigate this?

I have never opened something on jira. Any idea what info I should put? like the issue type etc...

roypaulin:
I really expected to hit no issues this time. I will investigate more on that.

roypaulin:
I will resolve the conflicts and update the issue page after everything goes well.

roypaulin:
The checks could not start because of the conflicts in builder.go, so I had to resolve them first.

PR #158, opened 2022-02-18
Webhook to account for empty serviceName in checks

spilchen:
The webhook makes sure that two subclusters that share the same service
name have similar service specific settings (serviceType, nodePort,
etc.).  It didn't account for subclusters that don't fill in the
serviceName.  A blank service name will just use the subcluster name.
So two subclusters that have a blank service name will have different
service objects.  Using the GetServiceName() fixes this bug.


roypaulin (approved):
looks good

PR #161, opened 2022-02-24
Fix bug to support Endpoints with trailing slashes

nprabhu2195:
Any endpoints ending with "/" will not affect DB creation. Not sure if we need a changie entry for this.


spilchen (commented):
Looks good.  Can you add a changie entry for this?  Something along the lines of 'Fixed: tolerate slashes being at the end of the communal endpoint url'.

nprabhu2195:
Sure, will add it. I think for first time you may have to approve the workflows.

nprabhu2195:
Looks like the lint fails because of the long test line. I have fixed that in my next commit.

PR #164, opened 2022-03-02
Order the env vars converted from annotations

spilchen:
This fixes a bug when spec.annotations has multiple annotations.  Each annotation is converted to an environment variable in the server container.  Prior to this fix, if you had multiple annotations defined the StatefulSet we generate each reconcile cycle could change.  This causes the statefulset controller to go into a rolling upgrade -- update pod, wait for it to be ready, update next pod, etc.

The StatefulSet wasn't really changing though.  It was just the order that we declared the environment variables.  The fix is to sort the environment variables so that the order is consistent.  That way the StatefulSet we generate doesn't change
if the annotations didn't change.

I also update the operator version as I would like to release a new version of the operator with this fix.


roypaulin (commented):
Looks good. Now I understand why each time I experienced the rolling upgrade, I had set multiple env vars through the annotations. I had forgot about the lack of order on maps. Thanks for fixing that.
I will approve once all the checks are successful.

roypaulin (approved):

PR #166, opened 2022-03-08
Batch 'admintools -t db_add_node' for faster scale up

spilchen:
Prior to this PR we did 'admintools -t db_add_node' one at a time.  So if you were adding 3 nodes to a subcluster we would call admintools 3 times.  We now batch that into a single call.  This greatly speeds up the time to do scale up.  I have seen it take about 52 seconds to add 3 nodes, where previously we did 3 calls to admintools that took about 40 seconds each.

Shard rebalance is always done after add node.  I moved this to a separate reconciler now.  This allows for better error handling and better separation of concerns.

### DIFF
--- a/pkg/controllers/dbaddnode_reconcile.go
+++ b/pkg/controllers/dbaddnode_reconcile.go
@@ -75,30 +75,27 @@ func (d *DBAddNodeReconciler) Reconcile(ctx context.Context, req *ctrl.Request) 
 // reconcileSubcluster will reconcile a single subcluster.  Add node will be
 // triggered if we determine that it hasn't been run.
 func (d *DBAddNodeReconciler) reconcileSubcluster(ctx context.Context, sc *vapi.Subcluster) (ctrl.Result, error) {
-	if missingDB, unknownState := d.PFacts.anyPodsMissingDB(sc.Name); missingDB || unknownState {
-		if missingDB {
-			res, err := d.runAddNode(ctx, sc)
-			if err != nil || res.Requeue {
-				return res, err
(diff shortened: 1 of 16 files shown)


pkg/controllers/dbaddnode_reconcile.go
```diff
@@ -116,36 +110,37 @@ func (d *DBAddNodeReconciler) runAddNode(ctx context.Context, sc *vapi.Subcluste
...
-		debugDumpAdmintoolsConf(ctx, d.PRunner, atPod.name)
+	debugDumpAdmintoolsConf(ctx, d.PRunner, atPod.name)
 
-		// Invalidate the cached pod facts now that some pods have a DB now.
-		d.PFacts.Invalidate()
-	}
-	err := d.rebalanceShards(ctx, atPod, sc.Name)
-	return ctrl.Result{}, err
+	// Invalidate the cached pod facts now that some pods have a DB now.
+	d.PFacts.Invalidate()
```
roypaulin, line 131: Do you mind telling me what this part means?

roypaulin (commented):
Looks good. Just some minor remarks.

pkg/controllers/dbaddnode_reconcile.go
spilchen: The Invalidate() call will mark the podfacts as being stale.  The next reconciler (RebalanceShards) needs update pod facts to properly work since there are no new nodes that were not yet accounted for in the pod facts.

roypaulin (approved):

PR #171, opened 2022-03-10
Allow existing serviceaccount to be used

roypaulin:
Until now, we could only use the serviceaccount (role, rolebinding) provided by the helm charts. This PR offers the possibility to use a custom serviceaccount to deploy the operator through helm.
-  A new helm parameter (serviceAccountNameOverride) has been added and when it is not empty, no new serviceaccount, roles, rolebindings will be created but the serviceaccount referenced will be used instead.
- A new helm parameter (webhook.enable)has been added which can disable/enable the access to the webhook. This can be useful to allow users with limited permissions to install the operator.
- Previously, the serviceaccount (the same as the operator) passed to the vertica pods was constant. Now we added a serviceAccount field to VerticaDBReconciler that will get the value from a flag and and will use that value in builder.go to set the correct serviceaccount name.
- A new e2e test has been added to check this use case as well as a few samples to create a custom serviceaccount, role rolebinding.


config/samples/custom-rb.yaml
```diff
@@ -0,0 +1,26 @@
+# (c) Copyright [2021-2022] Micro Focus or one of its affiliates.
```
spilchen, line 1: General comment about these new files in config/samples.  Can we have them generated?  It'll be error prone to ensure they are kept in sync with the roles/rolebindings in the helm chart.  We can give them the default name (i.e. verticadb-operator-manager-role).  When using them in the new e2e script, we can use kustomize to change the name of the roles.

config/samples/custom-role.yaml
```diff
@@ -0,0 +1,117 @@
+# (c) Copyright [2021-2022] Micro Focus or one of its affiliates.
```
spilchen, line 1: Do we need to provide 2 roles?  One for the manager and one for the leader election?  Or is the leader election role a subset of the manager one?

spilchen (commented):
I like these changes.  I had some minor suggestions around naming and a few questions.

config/samples/custom-role.yaml
roypaulin: Yes this file combines both manager and leader election rules. Do you think it is better to keep them separate?

spilchen: I don't mind having one.

spilchen:
Forgot to mention that we should add a changie entry for this PR.

roypaulin:
Yes sure. As usual I will add that at the end.

roypaulin:
My wsl time is not sync with my windows machine so that is why this new commit seems old. Anyway I had the default rbac manifests been generated. I could not keep the roles as one with kustomize so we could add a section in the docs on what each role is used for.

.gitignore
```diff
@@ -58,6 +58,9 @@ helm-charts/verticadb-webhook/crds/*
 # Omit overlays files
 config/overlays/
 
+# Omit gernerated default rbac manifests
```
spilchen, line 61: spelling: gernerated => generated

scripts/gen-rbac.sh
```diff
@@ -0,0 +1,56 @@
...
+- verticadb-operator-controller-manager-sa.yaml
+- verticadb-operator-manager-role-role.yaml
+- verticadb-operator-manager-rolebinding-rb.yaml
+- verticadb-operator-leader-election-role-role.yaml
+- verticadb-operator-leader-election-rolebinding-rb.yaml
+EOF
+
+sed -i '$d' $RBAC_DIR/verticadb-operator-controller-manager-sa.yaml
+sed -i '$d' $RBAC_DIR/verticadb-operator-manager-rolebinding-rb.yaml
+sed -i '$d' $RBAC_DIR/verticadb-operator-leader-election-rolebinding-rb.yaml
```
spilchen, line 56: I don't quite understand this part.  I don't see '$d' be defined, so I am wondering if this even does anything?

scripts/gen-rbac.sh
```diff
@@ -0,0 +1,56 @@
...
+namePrefix: verticadb-operator-
+
+bases:
+- ../../rbac
+EOF
+
+mkdir -p $RBAC_DIR
+
+cd $REPO_DIR
+$KUSTOMIZE build config/overlays/rbac | $KUBERNETES_SPLIT_YAML --outdir $RBAC_DIR -
```
spilchen, line 41: Do we have to run this through kubernetes split yaml?  I think it would be cleaner if we just produced a single file that had the roles, rolebindings and service account in it.  It will simplify the documentation and the GitHub release.

scripts/gen-rbac.sh
```diff
@@ -0,0 +1,56 @@
...
+
+cd $RBAC_DIR
+
+cat <<- EOF > kustomization.yaml
```
spilchen, line 45: I believe this kustomization.yaml is created for our e2e tests so that we can adjust the names of everything.  Can we put a comment to indicate why we are creating this file?

spilchen:
> My wsl time is not sync with my windows machine so that is why this new commit seems old. Anyway I have now the default rbac manifests been generated. I could not keep the roles as one with kustomize so we could add a section in the docs on what each role is used for.

What issue did you hit?  I think it's cleaner if we have a single file with everything in it (role, rolebindings, sa)

scripts/gen-rbac.sh
roypaulin: this deletes the last line of each file which is the namespace line as I wanted the manifests to not have any namespace specified.

roypaulin:
Ok I will keep all the rbac manifests in one file.

scripts/gen-rbac.sh
roypaulin: It deletes the line: "namespace: system"

spilchen: Okay, this wasn't obvious to me.  Lets have a comment.  But, I guess this is complicated if we combine all of the manifests in a single file.  Is there another approach to take to remove the namespace in that case?

roypaulin: Yes I think there is a possibility to delete a line matching a pattern with sed so I will try that.

spilchen (approved):
Thanks for doing these changes

PR #172, opened 2022-03-10
Utilize the new errors package to encapsulate reconcile requeue/error handling

nprabhu2195:
As part of #168 we added a new error handling package. The function IsReconcileAborted verifies if the reconcile needs a re-queue, needs to be re-queued after a delay or returned an error. This function can be utilized at other points of reconcilers where similar check happens.


nprabhu2195:
I have left the err!=nil checks as they are. Not sure if we need to complicate that check with a function call. I can do that as well if there is any value to it.

nprabhu2195:
Not sure why the test failed here. It passed on my local

spilchen:
> Not sure why the test failed here. It passed on my local

It seemed like it had trouble rescheduling pods after it deleted them.  Step 10 in crash-before-createdb will delete all 3 pods, then wait for them to get rescheduled by checking the install count in the VerticaDB.  Something is happening that prevents this.  This likely isn't related to your change.  Do you mind adding this to 10-assert.yaml in the crash-before-createdb testcase?
```
apiVersion: v1
kind: Pod
metadata:
  name: v-crash-before-createdb-defsc-0
status:
  phase: Running
---
apiVersion: v1
kind: Pod
metadata:
  name: v-crash-before-createdb-defsc-1
status:
  phase: Running
---
apiVersion: v1
kind: Pod
metadata:
  name: v-crash-before-createdb-defsc-2
status:
  phase: Running
```

If we hit this again we will have pod information dumped.

Also, can you remove these lines from kuttl-test.yaml?  

```
 22 suppress:
 23   - events
```
Dumping out the events might be helpful here.

pkg/controllers/offlineupgrade_reconcile.go
```diff
@@ -223,7 +223,7 @@ func (o *OfflineUpgradeReconciler) checkForNewPods(ctx context.Context) (ctrl.Re
 	}
 	if !foundPodWithNewImage {
 		o.Log.Info("Requeue to wait until at least one pod exists with the new image")
-		return ctrl.Result{Requeue: true}, nil
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Duration(o.Vdb.Spec.UpgradeRequeueTime)}, nil
```
spilchen, line 226: We shouldn't need to change this line.  The caller (`Reconcile`) will deal with changing the Result.

pkg/controllers/verticadb_controller.go
```diff
@@ -162,7 +162,7 @@ func (r *VerticaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			// Functions such as Upgrade may already set RequeueAfter and Requeue to false
 			if (res.Requeue || res.RequeueAfter > 0) && vdb.Spec.RequeueTime > 0 {
 				res.Requeue = false
-				res.RequeueAfter = time.Second * time.Duration(vdb.Spec.RequeueTime)
+				res.RequeueAfter = time.Duration(vdb.Spec.RequeueTime)
```
spilchen, line 165: Can we keep the `time.Second` in here?

spilchen (commented):
This looks good.  I had a suggestion for a few places we missed.  Also, it seems like there was some extra changes from the previous PR?

nprabhu2195:
> Also, can you remove these lines from kuttl-test.yaml?

Sure. I can add this.

pkg/controllers/offlineupgrade_reconcile.go
nprabhu2195: I apologise, I think this creeped back in when I merged the main after a previous PR

pkg/controllers/onlineupgrade_reconcile_test.go
```diff
@@ -38,6 +38,7 @@ var _ = Describe("onlineupgrade_reconcile", func() {
 	ctx := context.Background()
 	const OldImage = "old-image"
 	const NewImageName = "different-image"
+	const UpgradeRequeueTime = 100
```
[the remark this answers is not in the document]
nprabhu2195: We dont. Something must have gone wrong when I merged the fork main and these creeped back in. Will fix these

tests/e2e/crash-before-createdb/10-assert.yaml
```diff
@@ -11,12 +11,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: vertica.com/v1beta1
```
spilchen, line 14: Sorry if I caused confusion here, but can we keep the old check too?  So it would check each of the 3 pods, plus the VerticaDB.

nprabhu2195: Oh ok. Will do.

spilchen (approved):

PR #174, opened 2022-03-11
Use a helper function to set UpgradeRequeueTime in the test

nprabhu2195:
Instead of defining the UpgraderequeueTime for each test, use a helper function GetURTime()

### DIFF
--- a/pkg/test/helpers.go
+++ b/pkg/test/helpers.go
@@ -18,6 +18,7 @@ package test
 import (
 	"context"
 	"fmt"
+	"time"
 
 	. "github.com/onsi/gomega" // nolint:revive,stylecheck
 	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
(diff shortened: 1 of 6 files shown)


nprabhu2195:
Something I should have done in the previous PR. Let me know if this make sense.

pkg/test/helpers.go
```diff
@@ -175,3 +176,8 @@ func DeleteSvcs(ctx context.Context, c client.Client, vdb *vapi.VerticaDB) {
 		ExpectWithOffset(1, c.Delete(ctx, svc)).Should(Succeed())
 	}
 }
+
+// GetURTime is a helper function that returns the upgrade requeue time for the test to assert.
+func GetURTime(vdb *vapi.VerticaDB) time.Duration {
```
spilchen, line 181: I like the approach you took here, but lets take it one step further.  I think we can move this logic right into GetUpgradeRequeueTime().  So, change that function to return a `time.Duration` instead of an int.

PR #176, opened 2022-03-15
Skip service packs in upgrade path detection

spilchen:
The upgrade path detection that was added in v1.3.0 of the operator was too strict when it came to checking for service packs.  As per the Vertica docs, those can be skipped when doing upgrades.  This PR fixes that bug.


roypaulin (commented):
looks good. Do you mind telling me more about what service packs are?

spilchen:
Service packs are typically small bug fixes for the Vertica server.  The next server release that we are planning for next month is a service pack update (11.1.1).  More details on supported upgrade paths in Vertica: 
https://www.vertica.com/docs/11.1.x/HTML/Content/Authoring/InstallationGuide/Upgrade/UpgradePaths.htm

roypaulin (approved):

PR #177, opened 2022-03-15
docker log <container> was showing errors

NerdLogic:
I'm pretty sure they were coming from these copy commands.  It's OK that
these logs don't exist, so I piped stderr to /dev/null.
And, "mv" can be more efficient than "cp;rm".
And, mkdir -p could create /opt/vertica/config, so move it before the mv
And, chmod 755 is wrong for non-executables.  Instead add 644 to perms

### DIFF
--- a/docker-vertica/docker-entrypoint.sh
+++ b/docker-vertica/docker-entrypoint.sh
@@ -10,12 +10,13 @@ start_cron(){
 # because we have a Persistent Volume that backs /opt/vertica/config, so
 # it starts up empty and must be populated
 copy_config_files() {
-    cp -r /home/dbadmin/logrotate/* /opt/vertica/config/
+    mkdir -p /opt/vertica/config/licensing
+
+    mv /home/dbadmin/logrotate/* /opt/vertica/config/ 2>/dev/null
     rm -rf /home/dbadmin/logrotate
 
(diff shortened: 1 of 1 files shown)


docker-vertica/docker-entrypoint.sh
```diff
@@ -10,12 +10,13 @@ start_cron(){
 # because we have a Persistent Volume that backs /opt/vertica/config, so
 # it starts up empty and must be populated
 copy_config_files() {
-    cp -r /home/dbadmin/logrotate/* /opt/vertica/config/
+    mkdir -p /opt/vertica/config/licensing
+
+    mv /home/dbadmin/logrotate/* /opt/vertica/config/ 2>/dev/null || true
     rm -rf /home/dbadmin/logrotate
```
NerdLogic, line 16: Matt, why is this directory getting removed?  It looks like the dockerfile is copying a couple files (directory?) and then this entrypoint is copying them back.  I think everything I changed is safe, but I wanted to remove this rm command and wasn't sure about it.

spilchen: It is fine to skip this.  I think we only removed it because the files were copied to /opt.  It might be confusing to have the files in two places.

PR #178, opened 2022-03-16
Prevent cluster restart after offline upgrade

spilchen:
This fixes a bug that caused a second cluster restart to occur after a successful offline upgrade.  The cached pod facts weren't getting invalidated.  So after the upgrade, it will proceed to the restart reconciler and using the stale facts think that a cluster restart is needed.  Adding an invalidate to the restart reconciler fixed this problem.

Also fixing a minor issue with online upgrade.  We could try to restart the primaries with 'AT -t restart_node'.  This will fail because the cluster is read-only.  Added a wait mechanism so that we only proceed with the primary restart once the server moves the secondaries into read-only mode.  This last issue was minor, so I didn't bother adding a changie entry for it.

### DIFF
--- a/changes/unreleased/Fixed-20220316-092618.yaml
+++ b/changes/unreleased/Fixed-20220316-092618.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: Avoid a second cluster restart after offline upgrade has completed
+  successfully.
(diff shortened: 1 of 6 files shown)


roypaulin (approved):
looks good.

PR #179, opened 2022-03-16
Clean up unnecessary files

roypaulin:
This is the 1st of 2 PR to decrease the size of the vertica-k8s image.
It adds more files to clean up, removes the backup files and some libraries.

### DIFF
--- a/docker-vertica/packages/cleanup.sh
+++ b/docker-vertica/packages/cleanup.sh
@@ -34,11 +34,17 @@ rm -r -f \
    /opt/vertica/examples \
    /opt/vertica/sdk \
    /opt/vertica/packages/*/examples \
-   /opt/vertica/oss/python3/lib/python3.7/test \
-   /opt/vertica/oss/python3/lib/python3.7/pip \
-   /opt/vertica/oss/python3/lib/python3.7/site-packages/pip \
-   /opt/vertica/oss/python3/lib/python3.7/config-3.7*
+   /opt/vertica/packages/kafka \
+   /opt/vertica/oss/python3/lib/python3.9/test \
+   /opt/vertica/oss/python3/lib/python3.9/unittest/test \
+   /opt/vertica/oss/python3/lib/python3.9/pip \
+   /opt/vertica/oss/python3/lib/python3.9/site-packages/pip \
+   /opt/vertica/oss/python3/lib/python3.9/config-3.9* \
+   /opt/vertica/oss/python3/lib/python3.9/tkinter \
+   /opt/vertica/oss/python3/lib/python3.9/idlelib 
 
+# cleanup many of the __pycache__ directories 
+find /opt/vertica/oss/python3/lib/python3.9/  -type d -name "__pycache__" -exec rm -rf {} +
    
 # many of these directories contain things that aren't binaries
 # thus divert error output to /dev/null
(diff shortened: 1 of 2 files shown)


docker-vertica/packages/cleanup.sh
```diff
@@ -34,11 +34,17 @@ rm -r -f \
...
+   /opt/vertica/oss/python3/lib/python3.9/test \
+   /opt/vertica/oss/python3/lib/python3.9/unittest/test \
+   /opt/vertica/oss/python3/lib/python3.9/pip \
+   /opt/vertica/oss/python3/lib/python3.9/site-packages/pip \
+   /opt/vertica/oss/python3/lib/python3.9/config-3.9* \
+   /opt/vertica/oss/python3/lib/python3.9/tkinter \
+   /opt/vertica/oss/python3/lib/python3.9/idlelib 
 
+# cleanup many of the __pycache__ directories 
+find /opt/vertica/oss/python3/lib/python3.9/  -type d -name "__pycache__" -exec rm -rf {} +
```
spilchen, line 47: There are __pycache__ directories off of /opt/vertica/oss/python3/bin.  Can we use /opt/vertica/oss as the directory?  That way we don't have to update this if the python version changes again.

docker-vertica/packages/cleanup.sh
```diff
@@ -34,11 +34,17 @@ rm -r -f \
    /opt/vertica/examples \
    /opt/vertica/sdk \
    /opt/vertica/packages/*/examples \
-   /opt/vertica/oss/python3/lib/python3.7/test \
-   /opt/vertica/oss/python3/lib/python3.7/pip \
-   /opt/vertica/oss/python3/lib/python3.7/site-packages/pip \
-   /opt/vertica/oss/python3/lib/python3.7/config-3.7*
+   /opt/vertica/packages/kafka \
```
spilchen, line 37: Lets leave a comment about why we are removing this.  Maybe something like "# Removing because we don't have a java runtime installed".

docker-vertica/packages/cleanup.sh
```diff
@@ -34,11 +34,17 @@ rm -r -f \
...
+   /opt/vertica/oss/python3/lib/python3.9/test \
```
spilchen, line 38: General: can we use wildcards for the python versions?  Will this work?
/opt/vertica/oss/python*/lib/python*/test

spilchen (commented):
Just a few suggestions.  How much does this reduce the size of the image?

spilchen:
The `latest` is the minimal image.  So you need to set the `MINIMAL_VERTICA_IMG` env variable to `yes`.  Also, make sure you have the official RPM for the 11.1.0 release.  I think you have it based on the name, but I just wanted to double check.

roypaulin:
the current size is 888MB.

roypaulin:
I am a little confused by the python2.7 part. What is it use for? Is that something that does not change even when the version is upgraded?

spilchen:
Where do you see python2.7?

roypaulin:
In cleanup.sh we remove /usr/lib/python2.7 /usr/bin/python2.7 etc. So why are there folders/files relative to python2.7 given that we are running python3.9?

spilchen:
Not sure.  It's possible they aren't even in the image, or maybe 2.7 is part of standard CentOS.  But good point, we look into this further to see if they are still necessary and why.

roypaulin:
the python2.7 that should be deleted in cleanup.sh are still present in the container. It comes from CentOS. As cleanup.sh is called in the builder, I think all those lines on python2.7 are useless.

spilchen:
Okay, if they are from the base image, then we should just remove them from the cleanup script.  There is nothing we can do to free up that space.

spilchen (approved):

PR #180, opened 2022-03-17
Add online upgrade tests to GitHub CI

nprabhu2195:
In the last phase when we added online upgrade, we created a test suite specifically for those. We couldn't run it in the GitHub CI because it required that we had a 11.1 server. Now that 11.1 is released, we can move those tests to the CI.

- New workload was added to workflows (e2e-online-upgrade). 
- The e2e-11.1 tests were renamed to e2e-online-upgrade.
- Updated the RPM used in CI to 11.1.0
- Updated the upgrade tests in e2e to use vertica-k8s:11.1.0-0
- Updated the Image used in downgrade block test to 11.0.1


.github/workflows/e2e-online-upgrade.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+
+    - uses: actions/checkout@v2
+
+    - name: Set up e2e environment
+      uses: ./.github/actions/setup-e2e
+      
+    - name: Run e2e tests
+      run: |
+        # Parallelism must be 1.  The testcases will uninstall and install CRDs,
```
spilchen, line 27: We don't need to set the parallelism to 1.  That was true only for the operator upgrade tests.  We should set it based on the github input (if present).

Makefile
```diff
@@ -256,7 +256,7 @@ endif
 ifeq ($(BASE_VERTICA_IMG), <not-set>)
```
spilchen, line 256: Can we change the name of the Makefile target too?

spilchen (commented):
This looks really good.  I just had some minor suggestions.

.github/workflows/e2e-online-upgrade.yml
nprabhu2195: Makes sense

Makefile
nprabhu2195: Must have missed this one, will change

spilchen (approved):

PR #181, opened 2022-03-18
Control pod routing in service objects with label

spilchen:
This adds a new label to pods to control when they are visible for Service objects.  This is separate from the ready state.  A label gives us the ability to control routing for scaling events.  For instance, it will allow us to drain out nodes ahead of them being scaled down.  Or, we can add them after we do an add node and have verified that Vertica has finished rebalancing the nodes.

The label isn't part of the statefulset spec.  Rather we add (or remove) it using a new reconcile actor.

This should close timing scenarios during scaling seen occasionally in the e2e tests (e.g. step 40 in k-safety-0-scaling test). 

The online upgrade reconciler got a bit of an overhaul.  When adding the pending delete state in pod facts, it broke handling of transient subclusters.  So I am now adding transient subclusters to the VerticaDB during online upgrade.  I think this is a cleaner solution anyway.  It allowed me to cleanup transient handling in sc_finder.go for instance.


roypaulin (approved):
looks good.

PR #183, opened 2022-03-22
Drain node during scale down

spilchen:
This will add drain logic when we do a scale down so that we don't kill nodes that have active connections.  This builds on the work done in #181.  Namely the pending delete state, to know if a pod is going to be scaled down.  And the client routing change so that the Service object does not route new connections to the pod we are draining.


roypaulin (approved):
looks good.

PR #189, opened 2022-03-25
subcluster: Add loadBalancerIP and serviceAnnotations

harisokanovic:
Additional subcluster options to better customize network load
balancers, i.e. when subcluster.serviceType is "LoadBalancer".

loadBalancerIP: Specifies front-end IP address of services. This is
useful in baremetal K8s (e.g. using MetalLB) to assign static IPs [1].

serviceAnnotations: Additional annotations for implementation-specific
customization of services. Managed K8s (e.g. Amazon EKS) heavily
use annotations to configure NLBs [2]. This is useful for choosing a
VPC/subnet, enabling logging, and other configurations.

[1] https://metallb.universe.tf/usage/#requesting-specific-ips
[2] https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/annotations/

Example configuration:
```
kind: VerticaDB
metadata:
  name: example
  namespace: example
spec:
  subclusters:
    - name: example
      serviceType: LoadBalancer
      loadBalancerIP: 192.168.1.2
      serviceAnnotations:
        service.beta.kubernetes.io/aws-load-balancer-foo: bar
```

### DIFF
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -619,6 +619,18 @@ type Subcluster struct {
 	// specify. If not set, the external IP list is left empty in the service object.
 	// More info: https://kubernetes.io/docs/concepts/services-networking/service/#external-ips
 	ExternalIPs []string `json:"externalIPs,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// Specify IP address of LoadBalancer service for this subcluster.
+	// This field is ignored when serviceType != "LoadBalancer".
+	// More info: https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer
+	LoadBalancerIP string `json:"loadBalancerIP,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// A map of key/value pairs appended to service metadata.annotations.
+	ServiceAnnotations map[string]string `json:"serviceAnnotations,omitempty"`
 }
 
 // Affinity is used instead of corev1.Affinity and behaves the same.
--- a/pkg/controllers/obj_reconcile.go
+++ b/pkg/controllers/obj_reconcile.go
@@ -239,8 +239,28 @@ func (o ObjReconciler) reconcileExtSvc(ctx context.Context, expSvc *corev1.Servi
 	if o.Mode == ObjReconcileModeIfNotFound {
 		return nil
 	}
+
+	newSvc := o.reconcileExtSvcFields(curSvc, expSvc, sc)
+
+	if newSvc != nil {
+		o.Log.Info("updating svc", "Name", svcName)
+		return o.VRec.Client.Update(ctx, newSvc)
+	}
+	return nil
+}
+
+// reconcileExtSvcFields merges relevant expSvc fields into curSvc, and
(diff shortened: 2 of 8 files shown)


api/v1beta1/verticadb_types.go
```diff
@@ -619,6 +619,18 @@ type Subcluster struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// Specify IP address of LoadBalancer service for this subcluster.
+	// This field is ignored when serviceType != "LoadBalancer".
+	// More info: https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer
+	LoadBalancerIP string `json:"loadBalancerIP,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// A map of key/value pairs appended to service metadata.annotations.
+	ServiceAnnotations map[string]string `json:"serviceAnnotations,omitempty"`
```
spilchen, line 633: You can set annotations at `.spec.annotations` and this will be set for all of the objects created, including service objects.  I guess one advantage to define this here is the annotations will only be set for a particular service object.  If you had multiple subclusters each with their own service objects, are there use cases where you would want different annotations ?  I'm guessing there is, but I wanted to confirm.

api/v1beta1/verticadb_types.go
```diff
@@ -619,6 +619,18 @@ type Subcluster struct {
 	// specify. If not set, the external IP list is left empty in the service object.
 	// More info: https://kubernetes.io/docs/concepts/services-networking/service/#external-ips
 	ExternalIPs []string `json:"externalIPs,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// Specify IP address of LoadBalancer service for this subcluster.
+	// This field is ignored when serviceType != "LoadBalancer".
+	// More info: https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer
+	LoadBalancerIP string `json:"loadBalancerIP,omitempty"`
```
spilchen, line 628: Awesome, this was on our list of things we wanted to do!

pkg/controllers/obj_reconcile.go
```diff
@@ -227,6 +227,7 @@ func (o *ObjReconciler) checkForDeletedSubcluster(ctx context.Context) (ctrl.Res
 }
 
 // reconcileExtSvc verifies the external service objects exists and creates it if necessary.
+//gocyclo:ignore
```
spilchen, line 230: Not a big fan of ignoring this.  The function is getting quite big now.  Any suggestions on how we could reorganize it so we don't need this?  Maybe move the code that merges expSvc into curSvc into a separate function.

BTW, I updated this function the other day, so best to merge up before doing any big changes to it.

spilchen (commented):
This is awesome, thanks for this contribution.  Just some minor comment/question.  I'll add a changie entry for this so that we update the CHANGELOG at the next release.

harisokanovic:
Hi @spilchen, thanks for looking it over.

> are there use cases where you would want different annotations?

For example, specifying EIP allocations on Amazon is done by annotation instead of loadBalancerIP. These should be unique to each service/NLB instance.

> Any suggestions on how we could reorganize it so we don't need this? Maybe move the code that merges expSvc into curSvc into a separate function.

I'll try to clean it up in the next push.

spilchen (approved):

PR #190, opened 2022-03-25
GitHub CI workflow to produce release artifacts

nprabhu2195:
The current release procedure is a manual process. This introduces a workflow (Build Release Artifacts) to run in GitHub CI that generates the artifacts that are used in the GitHub release.

Specifically it will create the following files and include them as artifacts:

    helm-charts/verticadb-operator/crds/verticadbs.vertica.com-crd.yaml
    config/samples/custom-scc.yaml
    config/samples/rbac/default-rbac.yaml
    helm chart package (*.tgz)
    vdb-gen
    bundle/*

All the files are uploaded to a single artifact (release-artifacts) which can be then downloaded as a zip.

This CI will only be run on-demand using work_dispatch.


.github/workflows/build-release-artifacts.yml
```diff
@@ -0,0 +1,105 @@
...
+
+    - name: Build RBAC yaml
+      run: |
+        mkdir -p config/samples/rbac
+        make create-default-rbac
+        ls -lhrt config/samples/rbac/
+
+    - name: Build Bundle
+      run: | 
+        make bundle
```
spilchen, line 42: I think the image in the bundle for verticadb-operator won't be correct.  It will likely be `verticadb-operator:latest`.  We need to use the official image in dockerhub (i.e. `vertica/verticadb-operator:1.3.1`).  The tag comes from the operator version.

Currently, the image is defined in the Makefile like this:
```
OPERATOR_IMG ?= verticadb-operator:$(TAG)
```

Where `$(TAG)` is either `kind` or `latest`.

I think we should fix this by updating the default image/tag in the Makefile.  We don't need to pick the tag between kind/latest, lets just use the version.  We can add a new `$(REPO)` variable that allows you to set the repository.  This can apply to all of the `*_IMG` variables.

Then we use the correct image with this command:
```
REPO=vertica/ make bundle
```

.github/workflows/build-release-artifacts.yml
```diff
@@ -0,0 +1,105 @@
...
+        name: release-artifacts
+        path: /home/runner/work/vertica-kubernetes/vertica-kubernetes/bin/vdb-gen
+
+    - name: Upload Bundle
+      uses: actions/upload-artifact@v2
+      if: always()
+      continue-on-error: false
+      with:
+        name: release-artifacts
+        path: /home/runner/work/vertica-kubernetes/vertica-kubernetes/bundle
```
spilchen, line 98: I noticed in your runs of this workflow that it doesn't include the bundle directory.  Just the subdirectories in it.  Since the bundle is not used when creating the GitHub release, only when updating the operator in operatorhub, maybe add this to a different artifact called olm-bundle?  Maybe call the other artifact `github-release-assets` (to align with the name they use when creating the GitHub release)?

spilchen (commented):
Thanks for doing this.  This will save lots of time each time we cut a release.

.github/workflows/build-release-artifacts.yml
nprabhu2195: Sounds good.

.github/workflows/build-release-artifacts.yml
spilchen: Lets default to an empty string.

.github/workflows/build-release-artifacts.yml
```diff
@@ -0,0 +1,105 @@
...
+        cd helm-charts
+        helm package verticadb-operator
+        ls -lhrt
+    
+    - name: Upload CRD
+      uses: actions/upload-artifact@v2
+      if: always()
+      continue-on-error: false
+      with:
+        name: release-artifacts
```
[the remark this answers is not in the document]
spilchen: Okay, maybe go back to upload in multiple steps.  I like that they are were all in the root directory.  It is more align with how those files will be uploaded.

Makefile
```diff
@@ -71,11 +71,13 @@ else
   TAG?=latest 
 endif
 
+# Image Repo to use when pushing/pulling any image
+IMG_REPO?=
 # Image URL to use for building/pushing of the operator
-OPERATOR_IMG ?= verticadb-operator:$(TAG)
+OPERATOR_IMG ?= $(IMG_REPO)verticadb-operator:$(VERSION)
```
spilchen, line 77: Can we keep using `$(TAG)`, but lets set it to equal `$(VERSION)`?  These is some conditional logic to set `$(TAG)` based on whether we are running kind.  That can be removed entirely.

helm-charts/verticadb-operator/README.md
```diff
@@ -4,6 +4,7 @@ This helm chart will install the operator and an admission controller webhook.
 |----------------|-------------|---------------|
 | image.name | The name of image that runs the operator. | vertica/verticadb-operator:1.0.0 |
 | image.repo | Repo server hosting image.name | null (DockerHub) |
+| image.pullPolicy | When to pull the image from DockerHub  | IfNotPresent |
```
spilchen, line 7: We might not be pulling from DockerHub.  I suggest this for the description:
`The pull policy for the image that runs the operator`

scripts/create-helm-charts.sh
```diff
@@ -36,7 +36,9 @@ sed -i 's/verticadb-operator-.*-webhook-configuration/{{ .Release.Namespace }}-&
 # 2. Template image names
 sed -i "s|image: controller|image: '{{ with .Values.image }}{{ join \"/\" (list .repo .name) }}{{ end }}'|" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
 sed -i "s|image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0|image: '{{ with .Values.rbac_proxy_image }}{{ join \"/\" (list .repo .name) }}{{ end }}'|" $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
-# 3. Append imagePullSecrets
+# 3. Template imagePullPolicy
+sed -i 's/imagePullPolicy: policy/imagePullPolicy: {{ default "IfNotPresent" .Values.image.pullPolicy }}/' $TEMPLATE_DIR/verticadb-operator-controller-manager-deployment.yaml
+# 4. Append imagePullSecrets
```
spilchen, line 41: We'll need to renumber the rest of the comments too.

helm-charts/verticadb-operator/values.yaml
```diff
@@ -15,11 +15,10 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-
-
 image:
   repo: null # null = dockerhub
   name: vertica/verticadb-operator:1.3.1
+  pullPolicy: IfNotPresent # For kind environments this needs to be IfNotPresent
```
spilchen, line 21: The comment is misleading.  I suggest we just remove it.  It is best to have that comment in the Makefile, which we already do.

spilchen (approved):
The changes look great.  Thanks for the revisions.

PR #191, opened 2022-03-29
Allow uppercase dbName

harisokanovic:
podfacts.go and suite_test.go format incorrect node names when dbName
contains uppercase chars: e.g. "v_MYDB_node1234" instead of the
lowercase'd "v_mydb_node1234". As a result, operator continuously
re-creates database in this scenario because podfacts cannot detect
successful creation due to a malformed filepath.


spilchen (approved):
Thanks for finding and fixing this! I added a commit to increase our e2e test coverage of this scenario and to add a changie entry for the CHANGELOG.

harisokanovic:
Awesome. Thanks for looking it over. I'll be sure to include a changie entry next time.

PR #198, opened 2022-04-19
Expose prometheus service for operator

spilchen:
This exposes a /metrics endpoint for prometheus.  Prometheus metrics are already built into the operator-sdk framework that we use.  It provides some Go specific metrics (# of go routines, memory usage, etc.) and some controller specific metrics (how long the reconciliation iterations are, latency of webhook, etc.).  The work here is exposing the metrics for prometheus.

There are new helm parameters to control the deployment.  By default, we expose the endpoint with an auth proxy sidecar (prometheus.expose = EnableWithAuthProxy).  This sidecar will authenticate each request and only return the metrics if the caller is using a service account that is linked with the correct rbac rules.  

You can also expose the endpoint without any auth checking (prometheus.expose = EnableWithoutAuth).  This is suitable if you want to make the endpoints available outside of kubernetes.

There is an option to create a ServiceMonitor CR (prometheus.createServiceMonitor).  This is a CRD that is provided by the prometheus operator.  We don't make this the default because it would add a dependency on the prometheus operator.

There are a bunch of rbac rules that you need to apply if you want to access the protected endpoint (prometheus.createProxyRBAC).  These have to be cluster scoped rules, so by default the helm install won't include them.  Otherwise, you cannot install the operator more than once on the same system.

This is a merge into a prometheus branch.  It will get merged into main after we cut the next release.  I will create a changie entry when that happens.


scripts/authorize-metrics.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# A script that will authorize the given service account in the given namespace
+# the ability to access a protected metrics endpoint.
+
+set -o errexit
+set -o pipefail
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+REPO_DIR=$(dirname $SCRIPT_DIR)
+OP_SA=verticadb-operator-controller-manager
```
roypaulin, line 24: In case of custom sa what do we do?

scripts/config-transformer.sh
```diff
@@ -0,0 +1,46 @@
...
+rm $TEMPLATE_DIR/*yaml 2>/dev/null || true
+$KUSTOMIZE build $REPO_DIR/config/default | $KUBERNETES_SPLIT_YAML --outdir $TEMPLATE_DIR -
+mv $TEMPLATE_DIR/verticadbs.vertica.com-crd.yaml $CRD_DIR
+mv $TEMPLATE_DIR/verticaautoscalers.vertica.com-crd.yaml $CRD_DIR
+
+# Delete openshift clusterRole and clusterRoleBinding files
+rm $TEMPLATE_DIR/verticadb-operator-openshift-cluster-role-cr.yaml 
+rm $TEMPLATE_DIR/verticadb-operator-openshift-cluster-rolebinding-crb.yaml
+
+# Generate release artifacts from the split yaml's just generated.  This is
```
roypaulin, line 39: What were we doing before?

scripts/authorize-metrics.sh
spilchen: I use this in a single test currently, which does not provide its own ServiceAccount.  So, the default is fine.  I will update the script to add an option to override this in case we ever use this in other places.

scripts/config-transformer.sh
spilchen: Before it was just generating of the rbac manifest in the gen-rbac.sh script that you wrote.  Its more general now to handle the additional rbac and ServiceMonitor manifest.

PR #199, opened 2022-04-22
New prometheus metrics for the operator

spilchen:
This exports metrics specific to the verticadb operator to the /metrics endpoint.

- number of nodes restarted (Counter)
- number of subclusters (Gauge)
- number of pods (Gauge) using subcluster as label
- number of ready pods (Gauge) using subcluster as label
- number of running pods (Gauge) using subcluster as label
- number of cluster restarts (Counter)
- number of upgrades (Counter)
    
A new reconciler was added to maintain the gauges.  We increment the counters when we do various operations, like cluster restart and upgrade.

### DIFF
--- a/.github/workflows/build-release-assets.yml
+++ b/.github/workflows/build-release-assets.yml
@@ -30,20 +30,18 @@ jobs:
         make vdb-gen
         ls -lhrt bin/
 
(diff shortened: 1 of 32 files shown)


roypaulin (approved):
Looks good.

PR #204, opened 2022-05-02
Prevent negative values for requeue times

spilchen:
There are two parms in the CR that allow you to control the requeue time of a controller reconciler iteration.  Both of these parameters could be assigned negative values.  This PR adds new rules to the webhook to prevent that.

### DIFF
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -1029,3 +1029,14 @@ func (v *VerticaDB) BuildTransientSubcluster(imageOverride string) *Subcluster {
 		// service objects.
 	}
 }
+
+// FindSubclusterStatus will find a SubclusterStatus entry for the given
+// subcluster name.  Returns false if none can be found.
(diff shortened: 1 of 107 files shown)


api/v1beta1/verticadb_types.go
```diff
@@ -890,6 +890,7 @@ func MakeVDB() *VerticaDB {
 				DataPath:  "/data",
 				DepotPath: "/depot",
 			},
+			KSafety:    KSafety1,
```
roypaulin, line 893: Why adding this now?

roypaulin (commented):
Looks good. I just have one question.

api/v1beta1/verticadb_types.go
spilchen: I used the `MakeVDB()` function in the webhook UT.  When I tried to test the successful case, where the requeue times were valid, it still complained that the kSafety was not valid.  We just never set it properly, so any webhook test will complain about it.

roypaulin (approved):

PR #206, opened 2022-05-06
Push down more state into /etc/podinfo

spilchen:
This adds more information to /etc/podinfo.  The motivation is that we want the vertica server to include this stuff in a new k8s specific DC table.  

A summary of the new information in /etc/podinfo is as follows:
- k8s-build-date: Build date of the k8s server
- k8s-git-commit: The git commit of the k8s server
- k8s-version: The version of the k8s server
- annotations: A list of all of the annotations set for the pod
- namespace: The namespace the pod is created in
- operator-deployment-method: What method did we use to deploy the operator
- operator-version: What version of the operator are we using

As well, we are exposing the k8s node name and IP as environment variables:
- HOST_IP
- HOST_NODENAME

Two of the states (operator-deployment-method and operator-version) are fixed for the life of the operator.  For this reason they come from a configMap that is populated either in the helm chart or the OLM bundle.

Also, changed PodFacts so that we have access to a Log.


pkg/builder/labels_annotations.go
```diff
@@ -54,6 +54,13 @@ const (
 	OperatorVersion131 = "1.3.1"
 	OperatorVersion140 = "1.4.0"
 	OperatorVersion150 = CurOperatorVersion
+
+	// Annotations that we set in each of the pod.  These are set by the
+	// PodAnnotationReconciler.  They are available in the pod with the
+	// downwardAPI so they can be picked up the Vertica data collector (DC).
```
roypaulin, line 60: I think there is a "by" missing here.

pkg/controllers/vdb/podannotation_reconcile.go
```diff
@@ -0,0 +1,131 @@
...
+				return ctrl.Result{}, err
+			}
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+// generateAnnotations will generate static annotations that will be applied to each running pod
+func (s *PodAnnotationReconciler) generateAnnotations() (map[string]string, error) {
+	// We get the k8s server information from the client.  It would better to
```
roypaulin, line 72: Should be "It would be better" right?

roypaulin (commented):
Looks good. Just some typo in some comments.

pkg/builder/labels_annotations.go
```diff
@@ -57,7 +57,7 @@ const (
 
 	// Annotations that we set in each of the pod.  These are set by the
 	// PodAnnotationReconciler.  They are available in the pod with the
-	// downwardAPI so they can be picked up the Vertica data collector (DC).
+	// downwardAPI so they can be picked by up the Vertica data collector (DC).
```
roypaulin, line 60: picked up by

roypaulin (approved):

PR #224, opened 2022-06-20
Allow spread encryption to be set in the VerticaDB CR

### LINKED ISSUE #222: Setting for enabling TLS for internode Spread communication
Currently there is no option in the operator to turn on TLS for internode Spread communication.
As a workaround we set it inside the container's entrypoint, but it needs a complete database restart after the modification what can cause a lot of unpredictable misbehavior.

We would like to ask for a setting where we can enable this, so the database creation can happen with this option enabled.

spilchen:
The following parm was added to the VerticaDB CR:

```
spec.enableSpreadEncryption
```

When this is set to *vertica*, then the `EncryptSpreadComm` parameter will be set for the database right after it is created.  Setting encryption in spread requires a server restart, which the operator handles automatically.  The default value for this parameter is an empty string, which means no encryption is set.

The events for the VerticaDB will look like the following when enabling spread encryption.  We will show a successful create database, then a complete server restart.

```
Events:
  Type    Reason                   Age   From                Message
  ----    ------                   ----  ----                -------
  Normal  CreateDBStart            78s   verticadb-operator  Calling 'admintools -t create_db'
  Normal  CreateDBSucceeded        48s   verticadb-operator  Successfully created database with subcluster 'my-sc'. It t
ook 30.065841637s
  Normal  StopDBStart              47s   verticadb-operator  Calling 'admintools -t stop_db'
  Normal  StopDBSucceeded          43s   verticadb-operator  Successfully stopped the database.  It took 4s
  Normal  ClusterRestartStarted    42s   verticadb-operator  Calling 'admintools -t start_db' to restart the cluster
  Normal  ClusterRestartSucceeded  33s   verticadb-operator  Successfully called 'admintools -t start_db' and it took 9s
```

This can only be set for new instances of VerticaDB.  It is ignored if the initPolicy is something other than *Create*.  It only accepts an empty string (to clear encryption) or *vertica*.  The webhook enforces these rules.

The restart was generalized with a new condition variable: *VerticaRestartNeeded*.  In the future, if we have other cases where Vertica needs to be restarted we can use this condition variable.

Closes #222

### DIFF
--- a/pkg/controllers/vdb/stopdb_reconcile.go
+++ b/pkg/controllers/vdb/stopdb_reconcile.go
@@ -1,0 +1,126 @@
+/*
+ (c) Copyright [2021-2022] Micro Focus or one of its affiliates.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ You may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package vdb
+
+import (
+	"context"
+	"time"
+
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/cmds"
+	"github.com/vertica/vertica-kubernetes/pkg/controllers"
+	"github.com/vertica/vertica-kubernetes/pkg/events"
+	"github.com/vertica/vertica-kubernetes/pkg/names"
+	"github.com/vertica/vertica-kubernetes/pkg/vdbstatus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// StopDBReconciler will stop the cluster and clear the restart needed status condition
+type StopDBReconciler struct {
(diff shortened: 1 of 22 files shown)


spilchen:
@nyikesda the stop DB doesn't wait for active connection.  If there are any, those connections will be terminated.  This happens immediately after create db though.  Do you expect to have a lot of connections come in between?  You can use various status conditions to know when the database is fully setup.

```
kubectl wait --for=condition=DBInitialized=True vdb/v   # Wait for DB to be created
kubectl wait --for=condition=VerticaRestartNeeded=False vdb/v  # Wait for stop db
kubectl wait --for=condition=Ready=True pod/v-my-sc-0 # Wait for start db by checking ready state of one pod
```

nyikesda:
I do not see --force argument in your code after the admintools -t stop_db, so I thought that it waites for the the active connections. If it the --force argument is not neccessary, then sorry for my previous comment.

spilchen:
Okay, I understand what you mean.  You are right I am missing a force.  I'll add that in.

nyikesda (approved):
Thanks

pkg/controllers/vdb/stopdb_reconcile.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		s.VRec.Event(s.Vdb, corev1.EventTypeWarning, events.StopDBFailed, "Failed to stop the database")
+		return err
+	}
+	s.VRec.Eventf(s.Vdb, corev1.EventTypeNormal, events.StopDBSucceeded,
+		"Successfully stopped the database.  It took %ds", int(time.Since(start).Seconds()))
+	return nil
+}
+
+// genCmd will return the command to run in the pod to create the database
```
roypaulin, line 119: I think what you meant is "stop" instead of "create".

roypaulin (commented):
Looks good. I don't have too much to add.

PR #234, opened 2022-07-19
Better support for upgrade of the operator

spilchen:
A bunch of changes and fixes related to upgrade of the operator:
- This increases the operator to 1.7.0.
- Added a new testcase to verify the upgrading of the operator from 1.6.0.
- Fixed an issue where the operator wouldn't update the operator version in k8s objects (app.kubernetes.io/version).  It stays on the version that originally created it. 
- Fixes an issue that every time we updated the operator version, it would force all of the pods to be rescheduled.  This happened because we changed a label for the pod in the sts template portion.  I now add a label for the version through a reconciler rather than through the statefulset template.  Unfortunately, we can't do anything for the next release (1.7.0) -- since we have to update the sts template to remove the label.  But the next operator version after 1.7.0 should not require a pod restart.
- readOnly status fields were made optional.  These fields were added in 1.2.0 of the operator as required field.  This prevented upgrade of the operator from versons <= 1.1.0 to the current version.  So, this change unblocks that.

### DIFF
--- a/pkg/controllers/vdb/podannotatelabel_reconcile.go
+++ b/pkg/controllers/vdb/podannotatelabel_reconcile.go
@@ -29,23 +29,23 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-// PodAnnotationReconciler will maintain annotations in the pod about the running system
-type PodAnnotationReconciler struct {
+// PodAnnotateLabelReconciler will maintain annotations and labels in pods about the running system
+type PodAnnotateLabelReconciler struct {
 	VRec   *VerticaDBReconciler
 	Vdb    *vapi.VerticaDB
 	PFacts *PodFacts
 }
 
-// MakePodAnnotationReconciler will build a PodAnnotationReconciler object
-func MakePodAnnotationReconciler(vdbrecon *VerticaDBReconciler,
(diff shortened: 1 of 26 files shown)


pkg/controllers/vdb/podannotatelabel_reconcile.go
```diff
@@ -29,23 +29,23 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-// PodAnnotationReconciler will maintain annotations in the pod about the running system
-type PodAnnotationReconciler struct {
+// PodAnnotateLabelReconciler will maintain annotations and labels in pods about the running system
+type PodAnnotateLabelReconciler struct {
```
roypaulin, line 33: Minor question here Why did you change the name of this struct? And why is it PodAnnotateLabel and not something like PodAnnotationLabel or PodAnnotationAndLabel?

spilchen: This reconciler does more stuff now.  In addition to adding annotations to pods, it can add labels too.  I was trying to keep the name small.  I agree that the current name isn't ideal.  I'll change it to AnnotateAndLabelPodReconciler

PR #249, opened 2022-08-19
e2e udx tests for python and C++

jizhuoyu:
The python and cpp UDx tests are ready. The cpp UDx test for SourceFunctions.sql is not included for now because errors occur while vertica is loading those corresponding .so libraries. This might be an issue resulted from the fact that the lib binaries now are not built within the context of containers in k8s but rather built alsewhere and uploaded to the pods.


tests/e2e-udx/udx-cpp/setup-vdb/base/setup-vdb.yaml
```diff
@@ -0,0 +1,31 @@
...
+  local:
+    requestSize: 100Mi
+  dbName: vertdb
+  encryptSpreadComm: vertica
+  subclusters:
+    - name: sc1
+      size: 1
+  kSafety: "0"
+  certSecrets: []
+  imagePullSecrets: []
```
spilchen, line 31: We need to set the volumes and volumeMounts to empty lists.  This is so that kustomize can add volumes.
```suggestion
  imagePullSecrets: []
  volumes: []
  volumeMounts: []
```

tests/e2e-udx/udx-python/setup-vdb/base/setup-vdb.yaml
```diff
@@ -0,0 +1,31 @@
...
+  local:
+    requestSize: 100Mi
+  dbName: vertdb
+  encryptSpreadComm: vertica
+  subclusters:
+    - name: sc1
+      size: 1
+  kSafety: "0"
+  certSecrets: []
+  imagePullSecrets: []
```
spilchen, line 31: We need to set the volumes and volumeMounts to empty lists.  This is so that kustomize can add volumes.
```suggestion
  imagePullSecrets: []
  volumes: []
  volumeMounts: []
```

spilchen (commented):
Looks good.  Just minor changes.  I will hook these into the GitHub CI as a follow-on.

PR #262, opened 2022-10-05
Fix vdbgen

roypaulin:
This fixes some vdbgen issues by:
- fetching the ksafety from the server
- fetching the depot and data sizes from the server, and using them to calculate the local requestSize.
- fetching the server version and using it to build the image name.

### DIFF
--- a/pkg/vdbgen/vdb.go
+++ b/pkg/vdbgen/vdb.go
@@ -52,10 +52,13 @@ type DBGenerator struct {
 type QueryType string
 
 const (
-	ShardCountKey      QueryType = "shardCount"
-	DBCfgKey           QueryType = "dbCfg"
-	StorageLocationKey QueryType = "storageLocation"
-	SubclusterQueryKey QueryType = "subcluster"
+	ShardCountKey               QueryType = "shardCount"
+	DBCfgKey                    QueryType = "dbCfg"
+	StorageLocationKey          QueryType = "storageLocation"
+	SubclusterQueryKey          QueryType = "subcluster"
+	KSafetyQueryKey             QueryType = "ksafety"
+	StorageLocationSizeQueryKey QueryType = "storageLocationSize"
+	VersionQueryKey             QueryType = "version"
 
 	SecretAPIVersion = "v1"
 	SecretKindName   = "Secret"
@@ -65,10 +68,13 @@ const (
 )
 
 var Queries = map[QueryType]string{
-	ShardCountKey:      "SELECT COUNT(*) FROM SHARDS WHERE SHARD_TYPE != 'Replica'",
-	DBCfgKey:           "SHOW DATABASE DEFAULT ALL",
-	StorageLocationKey: "SELECT NODE_NAME, LOCATION_PATH FROM STORAGE_LOCATIONS WHERE LOCATION_USAGE = ?",
-	SubclusterQueryKey: "SELECT SUBCLUSTER_NAME, IS_PRIMARY FROM SUBCLUSTERS ORDER BY NODE_NAME",
+	ShardCountKey:               "SELECT COUNT(*) FROM SHARDS WHERE SHARD_TYPE != 'Replica'",
+	DBCfgKey:                    "SHOW DATABASE DEFAULT ALL",
+	StorageLocationKey:          "SELECT NODE_NAME, LOCATION_PATH FROM STORAGE_LOCATIONS WHERE LOCATION_USAGE = ?",
+	SubclusterQueryKey:          "SELECT SUBCLUSTER_NAME, IS_PRIMARY FROM SUBCLUSTERS ORDER BY NODE_NAME",
+	KSafetyQueryKey:             "SELECT GET_DESIGN_KSAFE()",
+	StorageLocationSizeQueryKey: "SELECT MAX(MAX_SIZE) FROM STORAGE_LOCATIONS WHERE LOCATION_USAGE = ?",
+	VersionQueryKey:             "SELECT VERSION()",
 }
 
 // Create will generate a VerticaDB based the specifics gathered from a live database
@@ -80,12 +86,15 @@ func (d *DBGenerator) Create() (*KObjs, error) {
 		d.readLicense,
 		d.connect,
 		d.setShardCount,
+		d.setKSafety,
+		d.setImage,
 		d.setCommunalPath,
 		d.fetchDatabaseConfig,
 		d.setCommunalEndpointAWS,
 		d.setCommunalEndpointGCloud,
 		d.setCommunalEndpointAzure,
 		d.setLocalPaths,
+		d.setRequestSize,
 		d.setSubclusterDetail,
 		d.setLicense,
 		d.setPasswordSecret,
@@ -167,6 +176,30 @@ func (d *DBGenerator) setShardCount(ctx context.Context) error {
 	if err := rows.Scan(&d.Objs.Vdb.Spec.ShardCount); err != nil {
 		return fmt.Errorf("failed running '%s': %w", q, err)
 	}
+
+	return nil
+}
+
+// setKsafety will fetch the ksafety from the database and set it inside v.vdb
+func (d *DBGenerator) setKSafety(ctx context.Context) error {
+	q := Queries[KSafetyQueryKey]
+	rows, err := d.Conn.QueryContext(ctx, q)
+	if err != nil {
+		return fmt.Errorf("failed running '%s': %w", q, err)
+	}
+	defer rows.Close()
+
+	if rows.Err() != nil {
+		return fmt.Errorf("failed running '%s': %w", q, rows.Err())
+	}
+	if !rows.Next() {
+		return errors.New("could not get ksafety from meta-function GET_DESIGN_KSAFE()")
+	}
+	var ksafety string
+	if err := rows.Scan(&ksafety); err != nil {
+		return fmt.Errorf("failed running '%s': %w", q, err)
+	}
+	d.Objs.Vdb.Spec.KSafety = vapi.KSafetyType(ksafety)
 
 	return nil
 }
@@ -338,6 +371,23 @@ func (d *DBGenerator) setLocalPaths(ctx context.Context) error {
 	return nil
 }
 
+// setRequestSize will fetch the local data size and set it in v.vdb.
+func (d *DBGenerator) setRequestSize(ctx context.Context) error {
+	depotMaxSize, err := d.queryLocalDataSize(ctx, "DEPOT")
+	if err != nil {
+		return err
+	}
+
+	dataMaxSize, err := d.queryLocalDataSize(ctx, "DATA,TEMP")
+	if err != nil {
+		return err
+	}
+	requestSize := depotMaxSize + dataMaxSize
+	d.Objs.Vdb.Spec.Local.RequestSize = *resource.NewQuantity(requestSize, resource.BinarySI)
+
+	return nil
+}
+
 // setCommunalPath will query the catalog and set the communal path in the vdb.
 func (d *DBGenerator) setCommunalPath(ctx context.Context) error {
 	var communalPath string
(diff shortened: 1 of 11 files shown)


roypaulin:
Should the user still have the possibility to set the image on command line? Should we keep image flag?

spilchen:
> Should the user still have the possibility to set the image on command line? Should we keep image flag?

I think if the image was specified on the command line, we should honour it.  But if nothing was selected, then we should pick an image from the Vertica docker repository.

pkg/vdbgen/vdb.go
```diff
@@ -475,6 +551,37 @@ func (d *DBGenerator) setSubclusterDetail(ctx context.Context) error {
 	return nil
 }
 
+// setImage will fetch the server version and use it to build
+// and set the image inside v.vdb.
```
spilchen, line 555: Suggest the following.
```suggestion
// setImage will fetch the server version and use it to pick
// an image that is hosted on our docker repository.
```

pkg/vdbgen/vdb.go
```diff
@@ -475,6 +551,37 @@ func (d *DBGenerator) setSubclusterDetail(ctx context.Context) error {
...
+	}
+	if !rows.Next() {
+		return errors.New("could not get Vertica version from meta-function")
+	}
+	var image string
+	if err := rows.Scan(&image); err != nil {
+		return fmt.Errorf("failed running '%s': %w", q, err)
+	}
+	// regex to match Vertica version
+	exp := regexp.MustCompile(`\d+(?:\.\d+){2}`)
```
spilchen, line 575: We should have a unit test for this.  Some sample inputs:
- Vertica Analytic Database 12.0.2-20221006
- Vertica Analytic Database 11.0.1-0

pkg/vdbgen/vdb.go
```diff
@@ -475,6 +551,37 @@ func (d *DBGenerator) setSubclusterDetail(ctx context.Context) error {
...
+	version := exp.FindString(image)
+	if version == "" {
+		return errors.New("could not find Vertica version")
+	}
+	d.Objs.Vdb.Spec.Image = fmt.Sprintf("vertica/vertica-k8s:%s-0", version)
```
spilchen, line 580: I suggest we add some explanation about how we come up with the image name.
```suggestion
        // Pick an image we hosted in Docker Hub.  We always publish an image for
        // hotfix 0. Rarely do we publish one for subsequent hotfixes, so always use
        // hotfix 0 regardless of what hotfix was currently in use.
	d.Objs.Vdb.Spec.Image = fmt.Sprintf("vertica/vertica-k8s:%s-0", version)
```

pkg/vdbgen/vdb.go
```diff
@@ -475,6 +551,37 @@ func (d *DBGenerator) setSubclusterDetail(ctx context.Context) error {
 	return nil
 }
 
+// setImage will fetch the server version and use it to build
+// and set the image inside v.vdb.
+func (d *DBGenerator) setImage(ctx context.Context) error {
+	q := Queries[VersionQueryKey]
```
spilchen, line 557: As suggested earlier, we should just exit out if an image was specified on the command line.

pkg/vdbgen/vdb.go
```diff
@@ -475,6 +551,37 @@ func (d *DBGenerator) setSubclusterDetail(ctx context.Context) error {
...
+	}
+	defer rows.Close()
+
+	if rows.Err() != nil {
+		return fmt.Errorf("failed running '%s': %w", q, rows.Err())
+	}
+	if !rows.Next() {
+		return errors.New("could not get Vertica version from meta-function")
+	}
+	var image string
```
spilchen, line 570: Suggest renaming this variable to `fullVersion`.  It's not the image name yet.  Its the output of the `version()` meta-function.

pkg/vdbgen/vdb.go
```diff
@@ -338,6 +371,23 @@ func (d *DBGenerator) setLocalPaths(ctx context.Context) error {
 	return nil
 }
 
+// setRequestSize will fetch the local data size and set it in v.vdb.
+func (d *DBGenerator) setRequestSize(ctx context.Context) error {
+	depotMaxSize, err := d.queryLocalDataSize(ctx, "DEPOT")
```
spilchen, line 376: I don't think we are handling the catalog in this function.  I wonder if we should pick the size of the disk where the depot, catalog, and data is hosted.  

You can do this via:
select max(disk_space_used_mb+disk_space_free_mb) from disk_storage where storage_usage = 'DEPOT';
select max(disk_space_used_mb+disk_space_free_mb) from disk_storage where storage_usage in ('CATALOG','DATA,TEMP');

pkg/vdbgen/vdb.go
```diff
@@ -171,6 +180,30 @@ func (d *DBGenerator) setShardCount(ctx context.Context) error {
 	return nil
 }
 
+// setKsafety will fetch the ksafety from the database and set it inside v.vdb
+func (d *DBGenerator) setKSafety(ctx context.Context) error {
```
spilchen, line 184: We should update the e2e test for this case.  It currently is k-safety 0, but ends up having 3 nodes.  We likely recreated the vdb with k-safety of 1 and didn't even realize.  I suggest we had a few more steps to this test to scale down to 1 or 2 nodes and run vdb-gen again.

spilchen (commented):
A few comments:
- we should have a changie entry so that this change is included in the CHANGELOG when we create the new release
- we should have unit tests for most of the new code added

pkg/vdbgen/vdb.go
roypaulin: Yes i don't know why I named it `image`, `fullVersion` is better.

pkg/vdbgen/vdb.go
roypaulin: I was looking for a more accurate query. Thanks.

pkg/vdbgen/vdb.go
roypaulin: I will add unit tests for the new  code.

roypaulin:
I will add the changie entry at the end.

pkg/vdbgen/vdb.go
```diff
@@ -171,6 +188,65 @@ func (d *DBGenerator) setShardCount(ctx context.Context) error {
...
+		return errors.New("could not get ksafety from meta-function GET_DESIGN_KSAFE()")
+	}
+	var designKSafe string
+	ksafety := vapi.KSafety1
+	if err := rows.Scan(&designKSafe); err != nil {
+		return fmt.Errorf("failed running '%s': %w", q, err)
+	}
+	if designKSafe == "0" {
+		if nodeCount, err := d.countNodes(ctx); err == nil {
+			// vdbgen will fail if kasety is 0 and there are more than 3 nodes
```
spilchen, line 213:
```suggestion
			// vdbgen will fail if ksafety is 0 and there are more than max nodes
```

pkg/vdbgen/vdb.go
```diff
@@ -171,6 +188,65 @@ func (d *DBGenerator) setShardCount(ctx context.Context) error {
...
+			if nodeCount > MaxNodeCountForKSafety0 {
+				return fmt.Errorf("ksafety 0 is not recommended for a %d nodes cluster", nodeCount)
```
spilchen, line 215:
```suggestion
				return fmt.Errorf("VerticaDB does not support ksafety 0 of more than %d nodes", nodeCount)
```

pkg/vdbgen/vdb.go
```diff
@@ -171,6 +188,65 @@ func (d *DBGenerator) setShardCount(ctx context.Context) error {
...
+	}
+	if !rows.Next() {
+		return errors.New("could not get ksafety from meta-function GET_DESIGN_KSAFE()")
+	}
+	var designKSafe string
+	ksafety := vapi.KSafety1
+	if err := rows.Scan(&designKSafe); err != nil {
+		return fmt.Errorf("failed running '%s': %w", q, err)
+	}
+	if designKSafe == "0" {
```
spilchen, line 211: Can we use the constant for "0"?

```suggestion
	if designKSafe == string(vapi.KSafety0) {
```

pkg/vdbgen/vdb.go
```diff
@@ -171,6 +188,65 @@ func (d *DBGenerator) setShardCount(ctx context.Context) error {
...
+		if nodeCount, err := d.countNodes(ctx); err == nil {
+			// vdbgen will fail if kasety is 0 and there are more than 3 nodes
+			if nodeCount > MaxNodeCountForKSafety0 {
```
spilchen, line 214: There is a constant already for this.  Suggest this:
```suggestion
			if nodeCount > vapi.KSafety0MaxHosts {
```

pkg/vdbgen/vdb.go
```diff
@@ -52,23 +52,37 @@ type DBGenerator struct {
...
 	ConfigKindName   = "ConfigMap"
+
+	MaxNodeCountForKSafety0 = 3
 )
 
 var Queries = map[QueryType]string{
 	ShardCountKey:      "SELECT COUNT(*) FROM SHARDS WHERE SHARD_TYPE != 'Replica'",
 	DBCfgKey:           "SHOW DATABASE DEFAULT ALL",
 	StorageLocationKey: "SELECT NODE_NAME, LOCATION_PATH FROM STORAGE_LOCATIONS WHERE LOCATION_USAGE = ?",
+	NodeCountQueryKey:  "SELECT COUNT(NODE_NAME) FROM SUBCLUSTERS",
```
spilchen, line 79: Suggest querying from nodes. It will be the same result, just more intuitive.
```suggestion
	NodeCountQueryKey:  "SELECT COUNT(*) FROM NODES",
```

pkg/vdbgen/vdb.go
```diff
@@ -419,6 +512,37 @@ func (d *DBGenerator) queryLocalPath(ctx context.Context, usage string) (string,
 	return commonPrefix, nil
 }
 
+// queryLocalDataSize will find data/depot size. It will pick the max among all nodes
+func (d *DBGenerator) queryLocalDataSize(ctx context.Context, usage string) (int64, error) {
+	var qtype QueryType
+	if usage == "DEPOT" {
+		qtype = DepotSizeQueryKey
+	} else if usage == "DATA,TEMP" {
+		qtype = CatalogSizeQueryKey
```
spilchen, line 521: Rather than have a mapping from `string` to key, I suggest have the caller pass in the key to use, and use that directly.

tests/e2e-extra/vdb-gen/70-run-vdbgen.yaml
```diff
@@ -0,0 +1,28 @@
...
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # Copy vdb-gen into the container.  We run it in the container so that it has
+  # access to k8s network.  Running it outside k8s, it would only have access to
+  # what is exposed.
+  - command: kubectl -n $NAMESPACE cp ../../../bin/vdb-gen v-vdb-gen-sc1-0:/tmp/vdb-gen
+  - command: kubectl -n $NAMESPACE exec v-vdb-gen-sc1-0 -- chmod +x /tmp/vdb-gen
+  - command: kubectl -n $NAMESPACE cp run-vdb-gen.sh v-vdb-gen-sc1-0:/tmp/run-vdb-gen.sh
+  - command: kubectl -n $NAMESPACE exec v-vdb-gen-sc1-0 -- chmod +x /tmp/run-vdb-gen.sh
```
spilchen, line 23: Haven't we already copied vdb-gen into the container at step 45?

spilchen (commented):
Thanks for applying the comments. I had some more that I thought that should be added.

tests/e2e-extra/vdb-gen/70-run-vdbgen.yaml
roypaulin: Yes, I was not careful enough. Too many copy paste without reading. Thanks

pkg/vdbgen/vdb.go
```diff
@@ -233,18 +231,18 @@ func (d *DBGenerator) countNodes(ctx context.Context) (int, error) {
 	}
 	defer rows.Close()
 
-	if rows.Err() != nil {
-		return 0, fmt.Errorf("failed running '%s': %w", q, rows.Err())
-	}
-	if !rows.Next() {
-		return 0, errors.New("could not find any nodes in the cluster")
+	if rows.Next() {
```
roypaulin, line 234: There was code duplication with another method, so I changed this one. I relaxed the checks here as `setSubclusterDetail` is the part handling the node count of each subcluster. Here we only care about Ksafety.

spilchen (approved):
Looks good. I'll merge once the changie entry has been added.

PR #270, opened 2022-10-19
Fix EULA acceptance issues

roypaulin:
Sometimes `admintools` is run before the run accept the EULA which cause failures. So we accept the EULA ahead of the restart.
operator-upgrade test has been update to test this particular scenario. Running the test with the official image will fail because of EULA acceptance.


roypaulin:
I do not like `acceptEulaIfMissing` duplication, but I wanted to check something with you. I wanted to move the EULA code to restart_reconcile. Correct me if I am wrong but by accepting the EULA in restart_reconcile we do not need to do it again in install_reconcile, right?

tests/e2e-operator-upgrade-template/template/47-restart-pod.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
```
spilchen, line 16: We may need to do the delete and `rm` in separate steps. If they are in the same steps, kuttl will do them more or less at the same time. And probably with that logic you would want to do the `rm` first, then delete the pods.

pkg/controllers/vdb/restart_reconcile.go
```diff
@@ -646,3 +655,16 @@ func (r *RestartReconciler) shouldRequeueIfPodsNotRunning() bool {
...
+	for _, p := range r.PFacts.Detail {
+		if !p.eulaAccepted.IsFalse() || !p.isPodRunning {
+			continue
+		}
+		if err := acceptEulaInPod(ctx, p, r.PRunner); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
spilchen, line 670: I think this is largely duplicated from install_reconciler.go. I am guessing that you couldn't use the one in that file because we couldn't change the function parameters. I suggest moving this logic to a new file (eula.go), and have this file call it and install_reconciler.go. The caller from install_reconciler.go might need a wrapper to call it. Something like this:
```
func (d *InstallReconciler) acceptEulaIfMissing(ctx context.Context) error {
        return acceptEulaIfMissing(ctx, d.PFacts)
}
```

pkg/controllers/vdb/install_reconcile.go
```diff
@@ -329,7 +329,7 @@ func (d *InstallReconciler) genCmdRemoveOldConfig() []string {
 }
 
 // acceptEulaInPod will run a script that will accept the eula in the given pod
-func (d *InstallReconciler) acceptEulaInPod(ctx context.Context, pf *PodFact) error {
```
spilchen, line 332: I suggest moving this to a new file eula.pod, as its not install specific anymore.

spilchen:
> I do not like acceptEulaIfMissing duplication, but I wanted to check something with you. I wanted to move the EULA code to restart_reconcile. Correct me if I am wrong but by accepting the EULA in restart_reconcile we do not need to do it again in install_reconcile, right?

I think you are on the correct path. I put a suggestion of how to get rid of the duplication in my review comments.

pkg/controllers/vdb/restart_reconcile.go
roypaulin: Thanks, I like this suggestion.

spilchen (approved):
Looks good. I added a changie entry. I'll merge once the CI passes.

PR #281, opened 2022-10-28
Add pod security policy to helm charts

roypaulin:
On  a cluster with Pod Security Policies (PSP) enabled, vertica pods need to be validated by at least 1 psp.
This PR adds to the helm charts:
- psp that has the minimum features to validate our pods without giving them more permissions than they need.
- a role and rolebinding to allow pods in the same namespace as the operator to get validated.

As psp is a cluster scoped resource, it is disabled by default.
As the operator is namespace scoped I prefered role/rolebinding to clusterrole/clusterrolebinding. So there will be a psp object and a pair of role/rolebinding for each helm instance of the operator.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -24,6 +24,7 @@ resources:
 - ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
 - ../prometheus
+- ../psp
 
 patchesStrategicMerge:
 # Protect the /metrics endpoint by putting it behind auth.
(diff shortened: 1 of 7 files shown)


roypaulin:
I could not test it because I could not enable psp on my kind cluster. So I worked on a minikube cluster as it was easy to enable psp there.

roypaulin:
I also tested the case without SYS_CHROOT and AUDIT_WRITE on both my kind and minikube cluster and yes we do not need to specify them on plain k8s. If we could find a way to know we are on OpenShift we could be remove them for non-OpenShift environment.
It is also worth trying out OpenShift to see if we still need to specify these 2 capabilities.

spilchen:
Closing for now. We can pick these changes up again in the future.

PR #283, opened 2022-11-07
Improve the stability of the operator in big clusters

spilchen:
This change improves the stability of the operator for big clusters. When used to create big clusters (> 40 nodes), the number of things it had to do per pod starts adding up. Calls to 'kubectl exec' to get various state (cmds package) can take a while. There is also a lot of memory pressure when 'kubectl exec' is called frequently -- we send the command to the REST endpoint that uses gzip to compress/uncompress the input/output. The compression library, which we don't have any control over, uses a lot of memory. Go is a garbage collection language, so the memory is eventually freed, but the GC can take some time to kick in. 

The approach taken in this change is to batch as many 'kubectl exec' commands as we can. This change impacts the podfacts collector, installer reconciler and create db reconciler. The end result is that there is less memory pressure and it is quicker for a reconcile iteration to get through its various state checks.

Refactor of the main.go module. Most of the command line argument handling was moved to a separate package called opcfg. This was changed because I wanted to flow down the command line arguments into the controllers so that we can log things only when in dev mode. So, we need to put the arguments in an exported struct from a new package that the controllers would have access to.

Included in this is a change to make it easier to run the operator locally. A new script was added that handles proper setup and teardown.

### DIFF
--- a/pkg/controllers/vdb/eula.go
+++ b/pkg/controllers/vdb/eula.go
@@ -29,7 +29,7 @@ import (
 // acceptEulaIfMissing will accept the end user license agreement if any pods have not yet signed it
 func acceptEulaIfMissing(ctx context.Context, pFacts *PodFacts, pRunner cmds.PodRunner) error {
 	for _, p := range pFacts.Detail {
-		if !p.eulaAccepted.IsFalse() || !p.isPodRunning {
+		if p.eulaAccepted || !p.isPodRunning {
 			continue
 		}
 		if err := acceptEulaInPod(ctx, p, pRunner); err != nil {
--- a/pkg/controllers/vdb/eula.go
+++ b/pkg/controllers/vdb/eula.go
@@ -61,14 +61,8 @@ func acceptEulaInPod(ctx context.Context, pf *PodFact, pRunner cmds.PodRunner) e
 	}
 	tmp.Close()
 
-	_, _, err = pRunner.CopyToPod(ctx, pf.name, names.ServerContainer, tmp.Name(), paths.EulaAcceptanceScript)
-	if err != nil {
-		return err
-	}
-
-	_, _, err = pRunner.ExecInPod(ctx, pf.name, names.ServerContainer, "/opt/vertica/oss/python3/bin/python3", paths.EulaAcceptanceScript)
-	if err != nil {
-		return err
-	}
-	return nil
+	// Copy and execute the script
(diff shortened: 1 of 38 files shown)


pkg/controllers/vdb/eula.go
```diff
@@ -61,14 +61,8 @@ func acceptEulaInPod(ctx context.Context, pf *PodFact, pRunner cmds.PodRunner) e
...
-		return err
-	}
-
-	_, _, err = pRunner.ExecInPod(ctx, pf.name, names.ServerContainer, "/opt/vertica/oss/python3/bin/python3", paths.EulaAcceptanceScript)
-	if err != nil {
-		return err
-	}
-	return nil
+	// Copy and execute the script
+	_, _, err = pRunner.CopyToPod(ctx, pf.name, names.ServerContainer, tmp.Name(), paths.EulaAcceptanceScript,
```
roypaulin, line 65: So now you copy and execute the script in a single function. Is that as you mentioned to reduce the number of "exec"?

scripts/run-operator.sh
```diff
@@ -0,0 +1,113 @@
...
+function logError {
+    logGeneric ${RED} "!" $@
+}
+
+cd $REPO_DIR
+
+mkdir -p $INT_TEST_OUTPUT_DIR
+OP="${INT_TEST_OUTPUT_DIR}/local-verticadb-operator.log"
+
+if helm list -o json | jq --exit-status '.[] | select(.chart|test("^verticadb-operator")) | .name'
```
roypaulin, line 89: Why do we check this if we run the operator in our local env?

roypaulin (commented):
Looks good. Just have a few comments, more like questions.

scripts/run-operator.sh
spilchen: This is just to make sure the operator isn't already deployed through a helm chart. I wasted a bunch of time because I was running the operator locally and had it deployed in helm too. This is just a check to make sure two operators aren't running at the same time.

pkg/controllers/vdb/eula.go
spilchen: Yes. Before we would do a 'kubectl exec' to copy the file in. Then another 'kubectl exec' to run the command we copied in. This just combines the two.

roypaulin (approved):

PR #287, opened 2022-11-17
Allow authorization to /metrics endpoint with TLS certificates

spilchen:
The RBAC auth proxy is a sidecar that runs alongside the operator manager to control access to the operators /metrics endpoint. Previous to this change, if you used the auth proxy, the only authentication method was with bearer tokens that you can get from the service account. This change allows custom client certificates to be used for authentication also.

To get this behaviour, the certificates need to be added to a secret, then referenced in the new helm chart parameter: `prometheus.tlsSecret`. The secret must have the following keys set: tls.key, tls.crt, and ca.crt.

My using custom certificates, authentication to the rbac proxy can be done with a curl command like this:
```
curl --key tls.key --cert tls.crt --cacert ca.crt https://verticadb-operator-metrics-service.<namespace>.svc:8443/metrics
```
If the `prometheus.tlsSecret` is omitted, which is the default behaviour, you can only use the bearer token to authenticate with the RBAC auth proxy.

I also include a change in here to make it easier to setup the RBAC rules for the metrics endpoint. The `prometheus.createRBACProxy` parameter created the RBAC rules. We had a rule that you could not use this parameter when deploying several instances of the operator. This has been relaxed now. As a result, the default value for `prometheus.createRBACProxy` is now true.

This also adds a deprecation warning for the `prometheus.createServiceMonitor` parameter. This will be removed in a future release.


helm-charts/verticadb-operator/README.md
```diff
@@ -23,5 +23,6 @@ This helm chart will install the operator and an admission controller webhook.
...
-| prometheus.createServiceMonitor | Set this to true if you want to create a ServiceMonitor.  This object is a CR provided by the prometheus operator to allow for easy service discovery.  If set to true, the prometheus operator must be installed before installing this chart.<br> See: https://github.com/prometheus-operator/prometheus-operator | false |
-| prometheus.createProxyRBAC | Set this to true if you want the helm install to create the rbac rules for accessing the metrics endpoint when it is protected by the rbac auth proxy.  If setting this to true, installation of this chart will fail if installing multiple times on the same system.  This happens because of the cluster scoped nature of the rbac rules. | false |
+| prometheus.createServiceMonitor | Set this to true if you want to create a ServiceMonitor.  This object is a CR provided by the prometheus operator to allow for easy service discovery.  If set to true, the prometheus operator must be installed before installing this chart.<br> See: https://github.com/prometheus-operator/prometheus-operator<br><br>*This parameter is deprecated and will be removed in a future release.* | false |
```
roypaulin, line 26: Why are we deprecating this?

roypaulin (commented):
Looks good, just have a question.

helm-charts/verticadb-operator/README.md
spilchen: It really isn't usable as-is. I tried to use it on our k8s-load-testing repo, but it needed to be modified to find the operator as a prometheus target. So, it really depends on how the prometheus operator is setup. It might be useful as a sample, but it originally was documented as a way to automatically include the operator in the prometheus stack.

roypaulin (approved):

PR #290, opened 2022-11-24
Expose the http port in the service object

spilchen:
**The http server is experimental at the moment and is hidden behind `spec.httpServerMode: Enabled`**

This exposes a new port (8443) in the service port for the http server that runs inside vertica server. This also removes the agent port in the service object, which was never used because we never ended up running the agent within the container.

I am also renaming the `spec.httpServerSecret` field in the VerticaDB to `spec.httpServerTLSSecret`. Since this feature is experimental, there should not be anyone using it, so no dependency on the field that I am renaming.


tests/e2e-http-server/http-custom-certs/50-verify-version.yaml
```diff
@@ -27,7 +27,7 @@ data:
     set -o errexit
 
     # Version can change only validate a portion of it
-    curl --insecure --fail --verbose --key /certs/tls.key --cert /certs/tls.crt https://v-http-custom-certs-main-0.v-http-custom-certs:8443/version | tee /tmp/curl.out
+    curl --insecure --fail --verbose --key /certs/tls.key --cert /certs/tls.crt https://v-http-custom-certs-main-0.v-http-custom-certs:8443/info/version | tee /tmp/curl.out
```
roypaulin, line 30: Why did you change the url.

roypaulin (approved):
Looks good !

tests/e2e-http-server/http-custom-certs/50-verify-version.yaml
spilchen: The server team changed around the API name. I was just updating the test to reflect this. I will eventually get this testsuite (e2e-http-server) into our CI.

PR #320, opened 2023-01-17
Add an init program to vertica-k8s image

spilchen:
This adds a proper init program to the vertica-k8s container so that it can reap zombie process and automatically restart cron/sshd if they fail. We use [s6-overlay](https://github.com/just-containers/s6-overlay) as our init program. It adds about 7MB (uncompressed) to our images.

There were a few additional items included in this due to that change:
- Removed the /config directory setup that was in the container entry script. This is better suited for the operator when it runs the install reconciler. We already did some setup in the operator, this just moves all of the pieces that were previously in the container.
- Related to the above, we set the securityPolicy.fsGroup so that the PV that is mounted in the container uses the gid of dbadmin. This gives it write access to the PV and no chmod was needed for the PVs in the container.
- Cleanup of the pod_facts.go. We kept track of certain files and directories with individual bools. This is managed in a much better way through a map.
- Removed some hacks that were added for 11.0.0 related to directory permissions of /opt/vertica/config/logrotate. The new minimum version the operator supports is now 11.0.1.


pkg/controllers/vdb/podfacts.go
```diff
@@ -379,12 +380,16 @@ func (p *PodFacts) genGatherScript(vdb *vapi.VerticaDB) string {
 		df --block-size=1 --output=size %s | tail -1
  	`,
 		vdb.GenInstallerIndicatorFileName(),
-		paths.AdminToolsConf,
 		paths.EulaAcceptanceFile,
-		paths.ConfigLogrotatePath,
-		paths.ConfigLogrotatePath,
-		paths.ConfigSharePath,
-		paths.HTTPTLSConfDir, paths.HTTPTLSConfFile,
+		paths.ConfigLogrotatePath, paths.ConfigLogrotatePath,
```
roypaulin, line 384: Why is each argument passed twice?

roypaulin (commented):
Looks Good ! Just have a comment.

pkg/controllers/vdb/podfacts.go
spilchen: We test each file like this:
```
echo -n '  %s: '
test -f %s && echo true || echo false
```

We want the file name to be displayed, then the actual test for the file after it. So, we need to reference the same variable twice.

roypaulin: Ah Ok, did not see it was `fmt.Sprintf`

roypaulin (approved):

PR #325, opened 2023-01-27
Add startupProbe and livenessProbe for vertica server

spilchen:
This adds new health probes to better handle when vertica crashes or becomes unresponsive. This was handled before internally by the operator. It would kill vertica process that were no longer responsive. This approach wasn't configurable and didn't handle some edge cases well. Using health probes is better because it's a more idiomatic way of doing things in k8s. They can also be tuned with new parameters in the VerticaDB.

The following probes are added:
- startupProbe: We wait for up to 20-minutes for the vertica process to start. The probe will monitor the startup.log. This can be tuned using the `spec.startupProbeOverride` field in the VerticaDB.
- livenessProbe: This probe will wait for about 1.5-minutes before failing. The probe checks if the TCP client port is open. This differs from the readinessProbe, as that checks the 'select 1' canary query. We chose a different path to minimize variability as this probe isn't forgiving if it fails (we reschedule the pod). This can be tuned using the `spec.livenessProbeOverride` field in the VerticaDB

The operator will only ever restart vertica when the startupProbe is active. The livenessProbe is too short. If we started vertica when that probe is active, we risk having the pod get rescheduled during the restart.

When moving up to an operator with this fix, the operator will add the new probes. This requires restart of the pods. The operator will handle this automatically, but you will see a rolling update as the change is applied.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -31,11 +31,13 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 const (
 	SuperuserPasswordPath = "superuser-passwd"
 	TestStorageClassName  = "test-storage-class"
+	VerticaClientPort     = 5433
 )
 
 // BuildExtSvc creates desired spec for the external service.
@@ -52,7 +54,7 @@ func BuildExtSvc(nm types.NamespacedName, vdb *vapi.VerticaDB, sc *vapi.Subclust
 			Selector: selectorLabelCreator(vdb, sc),
 			Type:     sc.ServiceType,
 			Ports: []corev1.ServicePort{
-				{Port: 5433, Name: "vertica", NodePort: sc.NodePort},
+				{Port: VerticaClientPort, Name: "vertica", NodePort: sc.NodePort},
 				{Port: 8443, Name: "vertica-http", NodePort: sc.VerticaHTTPNodePort},
 			},
 			ExternalIPs:    sc.ExternalIPs,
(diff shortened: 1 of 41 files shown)


pkg/builder/builder.go
```diff
@@ -476,13 +480,62 @@ func makeReadinessProbe(vdb *vapi.VerticaDB) *corev1.Probe {
...
+		// to minimize variability. If the livenessProbe fails, the pod is
+		// rescheduled. So, it isn't as forgiving as the readinessProbe.
+		ProbeHandler: corev1.ProbeHandler{
+			TCPSocket: &corev1.TCPSocketAction{
+				Port: intstr.FromInt(VerticaClientPort),
+			},
+		},
+		// These values were picked so that the vertica process needs to be
+		// unresponsive for about 1.5-minutes before it gets killed. Formula is:
+		// InitialDelaySeconds + PeriodSeconds * FailureThreshold.
```
roypaulin, line 525: Based on this formula, shouldn't it be 2.5 mins? 60 + 30 * 3 = 150/60 = 2.5

pkg/controllers/vdb/restart_reconcile.go
```diff
@@ -45,6 +47,8 @@ const (
 	AdminToolsMapFile = "/opt/vertica/config/ipMap.txt"
 	// Constant for an up node, this is taken from the STATE colume in NODES table
 	StateUp = "UP"
+	// Percent of livenessProbe time to wait when requeuing due to waiting on livenessProbe
+	PctOfLivenessProbeWait = 0.25
```
roypaulin, line 51: Is this a random value?

roypaulin (commented):
Just a few comments

roypaulin:
Not part of this PR but you could also set a constant for other hard-coded ports we use.

pkg/builder/builder.go
spilchen: Right. I need to update the comment. It's 2.5 minutes at the start of the container, then 1.5 subsequently.

pkg/controllers/vdb/restart_reconcile.go
spilchen: Yeah, pretty much. It's just a rough heuristic of when to retry the reconcile iteration. I'll update the comment.

roypaulin (approved):

PR #330, opened 2023-01-31
Make disk full errors more prominent

spilchen:
This change makes disk full errors more prominent in the operator. The focus is on checking the disk space for the local PV. We handle this with a couple of strategies:
- parsing the output from admintools. Sometimes, disk full errors will be seen in the output.
- add a reconcile actor that will check the disk space available in the local PV and log a k8s event if its below a threshold (currently 10mb)

I did the second approach because there are some times when vertica may have trouble starting and the error doesn't surface in the admintools output.

### DIFF
--- a/pkg/controllers/vdb/localdatacheck_reconcile.go
+++ b/pkg/controllers/vdb/localdatacheck_reconcile.go
@@ -17,6 +17,7 @@ package vdb
 
 import (
 	"context"
+	"fmt"
 
 	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
 	"github.com/vertica/vertica-kubernetes/pkg/controllers"
@@ -51,8 +52,9 @@ func (l *LocalDataCheckReconciler) Reconcile(ctx context.Context, req *ctrl.Requ
 	pods := l.PFacts.findPodsLowOnDiskSpace(FreeSpaceThreshold)
 	l.NumEvents = 0
 	for i := range pods {
-		l.VRec.Eventf(l.Vdb, corev1.EventTypeWarning, events.LowLocalDataAvailSpace,
-			"Low disk space in persistent volume attached to %s", pods[i].name.Name)
+		msg := fmt.Sprintf("Low disk space in persistent volume attached to %s", pods[i].name.Name)
+		l.VRec.Log.Info(msg)
+		l.VRec.Eventf(l.Vdb, corev1.EventTypeWarning, events.LowLocalDataAvailSpace, msg)
 		l.NumEvents++
 	}
 
--- a/pkg/controllers/vdb/localdatacheck_reconcile.go
+++ b/pkg/controllers/vdb/localdatacheck_reconcile.go
@@ -17,7 +17,6 @@ package vdb
 
 import (
 	"context"
-	"fmt"
 
 	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
 	"github.com/vertica/vertica-kubernetes/pkg/controllers"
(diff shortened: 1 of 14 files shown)


pkg/controllers/vdb/restart_reconcile.go
```diff
@@ -390,16 +391,33 @@ func (r *RestartReconciler) execRestartPods(ctx context.Context, downPods []*Pod
...
+		r.logATFailureEvent("restart_node", stdout)
 		return stdout, err
 	}
 	r.VRec.Eventf(r.Vdb, corev1.EventTypeNormal, events.NodeRestartSucceeded,
 		"Successfully called 'admintools -t restart_node' and it took %ds", int(elapsedTimeInSeconds))
 	return stdout, nil
 }
 
+// logATFailureEvent will log k8s events for the admintools restart error.
+func (r *RestartReconciler) logATFailureEvent(atCmd, op string) {
```
roypaulin, line 404: Are you going to use this for other admintools commands? Like if there is a need to have the same kind of function for another reconciler.

pkg/controllers/vdb/localdatacheck_reconcile.go
```diff
@@ -0,0 +1,62 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// LocalDataCheckReconciler will check the free space available in the PV and
+// log events if they it is too low.
+type LocalDataCheckReconciler struct {
+	VRec      *VerticaDBReconciler
+	Vdb       *vapi.VerticaDB
+	PFacts    *PodFacts
+	NumEvents int // Number of events written
```
roypaulin, line 35: What do you use this for?

roypaulin (commented):
Looks good!

pkg/controllers/vdb/restart_reconcile.go
spilchen: There is a bunch of parsing we do for create_db. I debated about throwing everything into a new package that they both use. Do you think I should do that cleanup? It will probably make things easier when we move away from admintools.

pkg/controllers/vdb/localdatacheck_reconcile.go
spilchen: This is for the unit test to prove that we recorded the correct number of events.

pkg/controllers/vdb/restart_reconcile.go
roypaulin: Yes I do like that idea.

roypaulin: Are you going to do that on this PR?

spilchen: Yep, I'll include it in this PR. I'm working on this change now.

pkg/mgmterrors/aterrors.go
```diff
@@ -0,0 +1,183 @@
...
+	"time"
+
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/cloud"
+	"github.com/vertica/vertica-kubernetes/pkg/events"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type ATErrors struct {
```
roypaulin, line 31: Could we add a comment for this struct? Same for some functions below.

pkg/mgmterrors/aterrors.go
```diff
@@ -0,0 +1,183 @@
...
+		GenericFailureReason: genericFailureReason,
+	}
+}
+
+const (
+	// Amount of time to wait after a restart failed because nodes weren't down yet
+	RestartNodesNotDownRequeueWaitTimeInSeconds = 10
+)
+
+// LogFailureError is called when admintools had attempted an option but
```
roypaulin, line 50: function's name is LogFailure.

roypaulin (commented):
Looks good!

pkg/mgmterrors/aterrors.go
spilchen: Good catch. This was an early name of this function. I forget to change this.

roypaulin (approved):

PR #340, opened 2023-02-28
fix case sensitivity of repositories

NerdLogic:


NerdLogic:
I removed the makefile change because that wasn't sufficient to fix the problem and thus it only added complexity.  Hopefully this passes the tests....

spilchen (approved):
Looks good. Thanks for finding and fixing.

PR #341, opened 2023-02-28
use perl instead of sed

NerdLogic:
because the sed on OSX treats the -i flag differently


spilchen (approved):
This is fine. You'll come across other sed usages in scripts as you do more things in OSX (e.g. scripts/template-helm-chart.sh).

PR #342, opened 2023-03-01
Default value of httpServerMode will setup http server

spilchen:
This changes the default value for `spec.httpServerMode`. If it's an empty string, or Auto, we will setup the http server if running with a vertica version of 12.0.3 or higher.

This required some refactoring of the version package. To allow the VerticaDB to check if version is compatible, some of the logic of the version package was moved into the VerticaDB API.


roypaulin (approved):
Looks good

PR #343, opened 2023-03-01
Create custom image that does not have ssh keys stored

roypaulin:
All keys that were stored in the image have been removed.
For dbadmin ssh keys in particular, a new variable has been added (NO_SSH_KEYS) that will allow to create an image that do not have those static keys stored inside. 
e2e-leg-1 as well as github actions have been modified to use that custom image.


roypaulin:
I am curious about why on `build-images->build-server-minimal->Login to Docker Hub`  there is `inputs.full_vertica_image` instead of `inputs.minimal_vertica_image`?

docker-vertica/Dockerfile
```diff
@@ -68,15 +70,15 @@ RUN set -x \
 # versions at once.
```
spilchen, line 70: We should update this comment to explain what we are doing with NO_SSH_KEYS. 
```suggestion
# versions at once. This step is required if not including SSH keys because
# conditional copy isn't a thing in Docker. If no SSH keys are to be included, 
# then the next RUN will remove the key we just added.
```

docker-vertica/Dockerfile
```diff
@@ -160,7 +160,9 @@ RUN set -x \
   && chmod u+s /usr/sbin/cron \
   # Untar the init program that was downloaded earlier
   && tar -C / -Jxpf /tmp/s6-overlay-x86_64.tar.xz \
-  && tar -C / -Jxpf /tmp/s6-overlay-noarch.tar.xz
+  && tar -C / -Jxpf /tmp/s6-overlay-noarch.tar.xz \
+  # delete old host keys
+  && rm -rf /etc/ssh/ssh_host*
```
spilchen, line 165: Where are these host keys being added? I would have thought that removal of `ssh-keygen` would have been enough.

docker-vertica/packages/cleanup.sh
```diff
@@ -23,6 +23,7 @@
 #
 # wander around in the image looking for things you can remove
 rm -r -f \
+   /opt/vertica/config/* \
```
spilchen, line 26: This might be problematic for vertica-demo. They use the vertica-k8s image, and probably rely on the config directory to be set up. Can we change this to remove specific things in `/config`? The agent keys, the keys for the https_certs, etc.

spilchen (commented):
This is good. I had a few comments/suggestions.

spilchen:
> I am curious about why on `build-images->build-server-minimal->Login to Docker Hub` there is `inputs.full_vertica_image` instead of `inputs.minimal_vertica_image`?

I think that is a bug. Can you fix it?

docker-vertica/Dockerfile
roypaulin: I observed that some keys' creation time corresponded to the build time and others to pod creation time, so that is why I removed all the keys before `ssh-keygen`.

.github/workflows/build-images.yml
```diff
@@ -172,6 +179,83 @@ jobs:
...
+      uses: docker/login-action@v2
+      if: ${{ github.event_name != 'pull_request' && inputs.nosshkeys_vertica_image == '' || startsWith(inputs.nosshkeys_vertica_image, 'ghcr.io') }}
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Login to Docker Hub
+      uses: docker/login-action@v2
+      if: ${{ inputs.full_vertica_image != '' && startsWith(inputs.full_vertica_image, 'docker.io') }}
```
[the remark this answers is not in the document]
roypaulin: Yes I will. This is related to my question above.

docker-vertica/packages/cleanup.sh
roypaulin: Ok I will delete specific keys files.

roypaulin:
> > I am curious about why on `build-images->build-server-minimal->Login to Docker Hub` there is `inputs.full_vertica_image` instead of `inputs.minimal_vertica_image`?
> 
> I think that is a bug. Can you fix it?

I will.

docker-vertica/packages/cleanup.sh
```diff
@@ -31,7 +32,9 @@ rm -r -f \
    /opt/vertica/oss/python*/lib/python*/site-packages/pip \
    /opt/vertica/oss/python*/lib/python*/config-[0-9]* \
    /opt/vertica/oss/python*/lib/python*/tkinter \
-   /opt/vertica/oss/python*/lib/python*/idlelib
+   /opt/vertica/oss/python*/lib/python*/idlelib \
+   /opt/vertica/oss/python*/lib/python*/site-packages/Cryptodome/SelfTest/PublicKey/test_vectors/ECC \
```
spilchen, line 36: As a general fix, could we remove all test directories for packages under site-package? We can match `test` and `*Test`.

docker-vertica/Dockerfile
```diff
@@ -65,18 +67,20 @@ RUN set -x \
...
-  && mkdir -p /root/.ssh \
-  && cp -r /home/dbadmin/.ssh /root \
-  && chmod 700 /root/.ssh \
-  && chmod 600 /root/.ssh/* \
   && chmod 700 /home/dbadmin/.ssh \
   && chmod 600 /home/dbadmin/.ssh/* \
   && chown -R dbadmin:verticadba /home/dbadmin/ \
-  && chmod go-w /etc/ssh/sshd_config.d/* /etc/ssh/ssh_config.d/*
+  && chmod go-w /etc/ssh/sshd_config.d/* /etc/ssh/ssh_config.d/* \
+  && if [[ ($NO_SSH_KEYS == "YES" || $NO_SSH_KEYS == "yes") ]] ; then \
```
spilchen, line 81: Can we simplify this to do a case-insensitive compare?
```suggestion
  && if [[ ${NO_SSH_KEYS^^} == "YES" ]] ; then \
```

roypaulin: Ok. I can also do the same with `MINIMAL`.

spilchen: Yes, that would be good thanks.

spilchen (approved):

PR #346, opened 2023-03-06
added supported platforms section

rs-vertica:
New section with link to supported platforms docs.


spilchen (approved):
Thanks for making this change.

PR #349, opened 2023-03-10
Add backdoor to run the agent

roypaulin:
This adds the agent for VA. 
- A new reconciler has been added.
- We start the agent only when this specific annotation is set in the CR: `vertica.com/run-agent: "yes"`
- We have added the agent key back to some of the images (not the nosshkeys version).
- The port 5444 will be added to the extsvc and  every vertica container
- An e2e test has been modified to check that agent is running when enabled.

### DIFF
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -20,6 +20,7 @@ package v1beta1
 import (
 	"fmt"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/vertica/vertica-kubernetes/pkg/paths"
@@ -964,6 +965,10 @@ const (
 	DefaultS3Region       = "us-east-1"
 	DefaultGCloudRegion   = "US-EAST1"
 	DefaultGCloudEndpoint = "https://storage.googleapis.com"
+
+	// VerticaDB.Spec.Annotation to enable the agent
+	StartAgent   = "start-agent"
+	AgentEnabled = "yes"
 )
 
 // ExtractNamespacedName gets the name and returns it as a NamespacedName
@@ -1252,4 +1257,10 @@ func (v *VerticaDB) IsHTTPServerEnabled() bool {
 // overridden to take affect.
 func (v *VerticaDB) IsEON() bool {
 	return v.Spec.ShardCount > 0
+}
+
+// IsAgentEnabled returns true if the annotation to enable the argent
+// has been set to the correct value
+func (v *VerticaDB) IsAgentEnabled() bool {
+	return strings.EqualFold(v.Spec.Annotations[StartAgent], AgentEnabled)
 }
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -961,14 +961,13 @@ const (
 	BuildRefAnnotation  = "vertica.com/buildRef"
 	// Annotation for the database's revive_instance_id
 	ReviveInstanceIDAnnotation = "vertica.com/revive-instance-id"
+	// Annotation to enable the agent
+	RunAgentAnnotation             = "vertica.com/run-agent"
+	RunAgentAnnotationEnabledValue = "yes"
 
 	DefaultS3Region       = "us-east-1"
 	DefaultGCloudRegion   = "US-EAST1"
 	DefaultGCloudEndpoint = "https://storage.googleapis.com"
-
-	// VerticaDB.Spec.Annotation to enable the agent
-	StartAgent   = "start-agent"
-	AgentEnabled = "yes"
 )
 
 // ExtractNamespacedName gets the name and returns it as a NamespacedName
@@ -1259,8 +1258,8 @@ func (v *VerticaDB) IsEON() bool {
 	return v.Spec.ShardCount > 0
 }
 
-// IsAgentEnabled returns true if the annotation to enable the argent
+// IsAgentEnabled returns true if the annotation to enable the agent
 // has been set to the correct value
 func (v *VerticaDB) IsAgentEnabled() bool {
-	return strings.EqualFold(v.Spec.Annotations[StartAgent], AgentEnabled)
+	return strings.EqualFold(v.ObjectMeta.Annotations[RunAgentAnnotation], RunAgentAnnotationEnabledValue)
 }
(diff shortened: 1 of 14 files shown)


pkg/controllers/vdb/agent_reconciler.go
```diff
@@ -0,0 +1,88 @@
...
+		"sudo", "/opt/vertica/sbin/vertica_agent", "start",
+	}
+	a.VRec.Event(a.Vdb, corev1.EventTypeNormal, events.StartAgentStart,
+		"Calling '/opt/vertica/sbin/vertica_agent start'")
+	if _, _, err := a.PRunner.ExecInPod(ctx, pod.name, ServerContainer, cmd...); err != nil {
+		a.VRec.Event(a.Vdb, corev1.EventTypeWarning, events.StartAgentFailed, "Failed to start the agent")
+		return err
+	}
+	a.VRec.Event(a.Vdb, corev1.EventTypeNormal, events.StartAgentSucceeded,
+		"The agent is now running")
```
spilchen, line 86: Suggest including the pod name in the event messages:
```suggestion
	a.VRec.Event(a.Vdb, corev1.EventTypeNormal, events.StartAgentStart,
		fmt.Sprintf("Calling '/opt/vertica/sbin/vertica_agent start' in pod %s", pod.name))
	if _, _, err := a.PRunner.ExecInPod(ctx, pod.name, ServerContainer, cmd...); err != nil {
		a.VRec.Event(a.Vdb, corev1.EventTypeWarning, events.StartAgentFailed,
			fmt.Sprintf("Failed to start the agent in pod %s", pod.name))
		return err
	}
	a.VRec.Event(a.Vdb, corev1.EventTypeNormal, events.StartAgentSucceeded,
		fmt.Sprintf("The agent is now running in pod %s", pod.name))
```

pkg/controllers/vdb/agent_reconciler.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/vertica/vertica-kubernetes/pkg/controllers"
+	"github.com/vertica/vertica-kubernetes/pkg/events"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// AgentReconciler will ensure the agent is running
+type AgentReconciler struct {
+	VRec    *VerticaDBReconciler
+	Log     logr.Logger
```
spilchen, line 33: I don't think we use `Log`. Plus, we can get it from the VRec, so I suggest we remove it from this struct.

pkg/controllers/vdb/podfacts.go
```diff
@@ -601,6 +608,16 @@ func (p *PodFacts) checkIsDBCreated(ctx context.Context, vdb *vapi.VerticaDB, pf
 	return nil
 }
 
+// checkIfAgentRunning will check if the Vertica agent is running and set state in pf.agentRunning
+func (p *PodFacts) checkIfAgentRunning(ctx context.Context, vdb *vapi.VerticaDB, pf *PodFact, gs *GatherState) error {
```
spilchen, line 612: I don't think we need this function. There is a mapping function from the `GatherState` to `PodFact` (see `checkForSimpleGatherStateMapping`). We can just update that.

api/v1beta1/verticadb_types.go
```diff
@@ -1253,3 +1258,9 @@ func (v *VerticaDB) IsHTTPServerEnabled() bool {
 func (v *VerticaDB) IsEON() bool {
 	return v.Spec.ShardCount > 0
 }
+
+// IsAgentEnabled returns true if the annotation to enable the argent
```
spilchen, line 1262: Spelling
```suggestion
// IsAgentEnabled returns true if the annotation to enable the agent
```

pkg/events/event.go
```diff
@@ -77,6 +77,9 @@ const (
 	MgmtFailed                      = "MgmtFailed"
 	MgmtFailedDiskFull              = "MgmtFailedDiskfull"
 	LowLocalDataAvailSpace          = "LowLocalDataAvailSpace"
+	StartAgentStart                 = "StartAgentStart"
+	StartAgentSucceeded             = "StartAgentSucceeded"
+	StartAgentFailed                = "StartAgentFailed"
```
spilchen, line 82: `StartAgentStart` seems like a weird name. Suggested values:
```suggestion
	RunAgentStart                 = "RunAgentStart"
	RunAgentSucceeded             = "RunAgentSucceeded"
	RunAgentFailed                = "RunAgentFailed"
```

api/v1beta1/verticadb_types.go
```diff
@@ -964,6 +965,10 @@ const (
 	DefaultS3Region       = "us-east-1"
 	DefaultGCloudRegion   = "US-EAST1"
 	DefaultGCloudEndpoint = "https://storage.googleapis.com"
+
+	// VerticaDB.Spec.Annotation to enable the agent
+	StartAgent   = "start-agent"
+	AgentEnabled = "yes"
```
spilchen, line 971: Suggest a rename so that we know this corresponds to the annotation:
```suggestion
	RunAgentAnnotationEnabledValue = "yes"
```

api/v1beta1/verticadb_types.go
```diff
@@ -1253,3 +1258,9 @@ func (v *VerticaDB) IsHTTPServerEnabled() bool {
 func (v *VerticaDB) IsEON() bool {
 	return v.Spec.ShardCount > 0
 }
+
+// IsAgentEnabled returns true if the annotation to enable the argent
+// has been set to the correct value
+func (v *VerticaDB) IsAgentEnabled() bool {
+	return strings.EqualFold(v.Spec.Annotations[StartAgent], AgentEnabled)
```
spilchen, line 1265: Lets check the annotations that are present in the vdb's metadata. The annotations that are set in the `spec` apply to objects that are created by the operator.
```suggestion
	return strings.EqualFold(v.ObjectMeta.Annotations[StartAgent], AgentEnabled)
```

spilchen (commented):
This looks great. I only had some minor suggestions. Mainly around names and such. I'll add a changie entry and update the PR description so that we are clear on what annotation needs to be present.

spilchen (approved):
Thanks for applying the review comments

PR #352, opened 2023-03-15
Update metrics endpoint

roypaulin:
Replace /metrics with /v1/metrics to by in sync with the server metrics endpoint change.

### DIFF
--- a/tests/e2e-http-server/http-custom-certs/50-verify-metrics.yaml
+++ b/tests/e2e-http-server/http-custom-certs/50-verify-metrics.yaml
@@ -27,7 +27,7 @@ data:
     set -o errexit
 
     # metrics can change only validate a portion of it
-    curl --insecure --fail --verbose --key /certs/tls.key --cert /certs/tls.crt https://v-http-custom-certs-main-0.v-http-custom-certs:8443/metrics | tee /tmp/curl.out
+    curl --insecure --fail --verbose --key /certs/tls.key --cert /certs/tls.crt https://v-http-custom-certs-main-0.v-http-custom-certs:8443/v1/metrics | tee /tmp/curl.out
     grep -cq 'vertica_build_info' /tmp/curl.out
 ---
 apiVersion: v1


tests/e2e-http-server/http-custom-certs/50-verify-metrics.yaml
```diff
@@ -27,7 +27,7 @@ data:
     set -o errexit
 
     # metrics can change only validate a portion of it
-    curl --insecure --fail --verbose --key /certs/tls.key --cert /certs/tls.crt https://v-http-custom-certs-main-0.v-http-custom-certs:8443/metrics | tee /tmp/curl.out
+    curl --insecure --fail --verbose --key /certs/tls.key --cert /certs/tls.crt https://v-http-custom-certs-main-0.v-http-custom-certs:8443/v1/metrics | tee /tmp/curl.out
```
spilchen, line 30: We still need this test to be compatible when run against 12.0.3 image. We can't really rely on the vertica version since 12.0.4-* will have both old and new API. Can we try both endpoints and make sure that one works?

tests/e2e-http-server/http-custom-certs/50-verify-metrics.yaml
```diff
@@ -27,8 +27,11 @@ data:
     set -o errexit
 
     # metrics can change only validate a portion of it
-    curl --insecure --fail --verbose --key /certs/tls.key --cert /certs/tls.crt https://v-http-custom-certs-main-0.v-http-custom-certs:8443/v1/metrics | tee /tmp/curl.out
-    grep -cq 'vertica_build_info' /tmp/curl.out
+    for endpoint in metrics v1/metrics
+    do
+      curl --insecure --fail --verbose --key /certs/tls.key --cert /certs/tls.crt https://v-http-custom-certs-main-0.v-http-custom-certs:8443/$endpoint | tee /tmp/curl.out
+      grep -cq 'vertica_build_info' /tmp/curl.out && break
```
spilchen, line 33: We have `set -o errexit` set, so when the endpoint is v1/metrics, won't this fail during the first iteration?

roypaulin: No I think the exception is when the command is part of an AND/OR-list. I will reverse the endpoints order just to be sure.

spilchen: Tricky. It's because we don't have `set -o pipefail`. The `curl` command that pipes to `tee` will actually succeed because the `tee` command is a success. Maybe add a comment here.

spilchen:
The HDFS tests are failing because an image was removed. I'm in the process of changing the CI so we don't rely on that image anymore. (See #354)

tests/e2e-http-server/http-custom-certs/50-verify-metrics.yaml
roypaulin: I will!

PR #353, opened 2023-03-15
Setup keys for client side agent access

roypaulin:
In order to use the agent within the operator, we require certain keys to be setup in well known locations under /opt/vetrica. This will setup those keys to allow for client side agent access.

A change to the vertica image is necessary as we had to save off the keys that were generated in install_vertica. This applies to all but the nosshkeys image. This follows the same pattern we do for licensing and logrotate state.


docker-vertica/Dockerfile
```diff
@@ -58,6 +58,8 @@ RUN set -x \
   && mkdir -p /home/dbadmin/logrotate \
   && cp -r /opt/vertica/config/logrotate /home/dbadmin/logrotate/  \
   && cp /opt/vertica/config/logrotate_base.conf /home/dbadmin/logrotate/ \
+  && mkdir -p /home/dbadmin/share \
+  && cp -r /opt/vertica/config/share/agent* /home/dbadmin/share/ \
```
spilchen, line 62: Can we add the NO_SSH_KEYS check here so we don't have to delete the file later?

pkg/controllers/vdb/install_reconcile.go
```diff
@@ -320,6 +320,12 @@ func (d *InstallReconciler) genCreateConfigDirsScript(p *PodFact) string {
 		numCmds++
 	}
 
+	if !p.fileExists[paths.AgentKeyFile] || !p.fileExists[paths.AgentCertFile] {
```
spilchen, line 323: We need to skip this if the image doesn't have the agent keys (i.e. it's a NO SSH KEY build or an older image). We probably need to collect that in podfacts in the gather script. Something like 'does the image have the agent keys?'.

docker-vertica/Dockerfile
spilchen, line 1: Are the keys static across multiple builds of the image? Basically, does the installer regenerate them new each time? We may want them static. Otherwise, they will change during an upgrade, which will make management of them a nightmare.

An alternative approach is to mount the keys through a secret. But, then we'd have to tell the agent where to find the mounted keys.

roypaulin: No keys are not static across multiple builds.

roypaulin: I did not consider secrets first because I thought we wanted something simple. So could we mount the keys and then copy them into the correct folder?

roypaulin: This is just temporary right, for development, so why don't we just keep static keys like the ones for dbadmin?

spilchen: Actually, lets just defer this part for now. I think there might be some broader security concerns here. If we do nothing about this we will have static keys for the same image. That's probably fine for our internal development. No one running the operator in production should set the knob to enable the agent.

spilchen:
Is it possible to add a simple e2e test to verify we can read/write to the agent? We should be able to use one of the agent e2e tests that we use to have before.

pkg/controllers/vdb/podfacts.go
```diff
@@ -399,6 +410,8 @@ func (p *PodFacts) genGatherScript(vdb *vapi.VerticaDB, pf *PodFact) string {
 		df --block-size=1 --output=avail %s | tail -1
 		echo -n 'agentRunning: '
 		/opt/vertica/sbin/vertica_agent status | grep --quiet "running" && echo true || echo false
+		echo -n 'hasAgentKeys: '
+		test -f %s/agent.key && test -f %s/agent.cert && test -f %s/apikeys.dat && echo true || echo false
```
spilchen, line 414: I wonder if we should check if the directory is empty. Is there ever a case where only some keys are present? There is an example of checking for an empty dir above in the `dbExists` check.

pkg/controllers/vdb/podfacts.go
```diff
@@ -399,6 +410,8 @@ func (p *PodFacts) genGatherScript(vdb *vapi.VerticaDB, pf *PodFact) string {
 		df --block-size=1 --output=avail %s | tail -1
 		echo -n 'agentRunning: '
 		/opt/vertica/sbin/vertica_agent status | grep --quiet "running" && echo true || echo false
+		echo -n 'hasAgentKeys: '
```
spilchen, line 413: Suggest a different name:
```suggestion
		echo -n 'imageHasAgentKeys: '
```

pkg/controllers/vdb/podfacts.go
```diff
@@ -157,6 +157,10 @@ type PodFact struct {
 
 	// Is the agent running in this pod?
 	agentRunning bool
+
+	// Check if the pod has the agent keys
+	// and an API Key, in apikeys.dat, used for agent calls
+	hasAgentKeys bool
```
spilchen, line 163: Doesn't this mean if the image has the agent keys?
```suggestion
	// Check if the image has agent keys saved in the dbadmin directory.
	imageHasAgentKeys bool
```

pkg/controllers/vdb/podfacts.go
```diff
@@ -1037,3 +1055,12 @@ func getHostList(podList []*PodFact) []string {
 	}
 	return hostList
 }
+
+// needAgentKeysCopy returns true if all agent keys are present in /home/dbadmin/share
+// and have not yet been copied to /opt/vertica/config/
+func needAgentKeysCopy(pod *PodFact) bool {
```
spilchen, line 1061: Lets make this function part of `PodFact`:
```suggestion
func (p *PodFact) needAgentKeysCopy() bool {
```

spilchen (commented):
I just had a few comments around names and such

pkg/controllers/vdb/podfacts.go
roypaulin: We want all keys to the here so what will be the benefit of checking if the dir is empty?

roypaulin: We want all keys to the here so what will be the benefit of checking if the dir is empty?

spilchen: That way you don't have to spell out all of the keys you want. It makes it easier in case there are other things added in here in the future. You are basically just checking to see if the image that we are using preserved agent artifacts that need to be copied over to /opt/vertica/confg. So, why do we care that specific files are there.

pkg/controllers/vdb/podfacts.go
roypaulin: Yes I really need to improve my naming skills!

pkg/controllers/vdb/podfacts.go
roypaulin: Yes there are no cases where only some of them are here.

docker-vertica/Dockerfile
```diff
@@ -58,6 +59,11 @@ RUN set -x \
   && mkdir -p /home/dbadmin/logrotate \
   && cp -r /opt/vertica/config/logrotate /home/dbadmin/logrotate/  \
   && cp /opt/vertica/config/logrotate_base.conf /home/dbadmin/logrotate/ \
+  && if [[ ${NO_SSH_KEYS^^} != "YES" ]] ; then \
```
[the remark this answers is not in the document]
roypaulin: Ok, We need some doc changes.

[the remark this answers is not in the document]
spilchen: Understood, I opened VER-86201 for this.

spilchen (commented):
I think we're almost there. Just had one minor comment.

spilchen (approved):

PR #357, opened 2023-03-22
rename nosshkeys to nokeys, part 1

NerdLogic:
This part is about adding the new build image "nokeys".  After we change the jenkins process to  build "nokeys", we will remove the build image "nosshkeys".

### DIFF
--- a/.github/actions/setup-e2e/action.yaml
+++ b/.github/actions/setup-e2e/action.yaml
@@ -26,7 +26,7 @@ runs:
       run: |
         find . -name \*.tar
         docker load --input full-vertica-image/full-vertica-image.tar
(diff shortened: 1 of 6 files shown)


spilchen:
I'm not sure why the tests didn't run automatically. I'm going to close then reopen the PR to see if this gets things going.

spilchen:
This might be a clue. In the action on your fork (https://github.com/NerdLogic/vertica-kubernetes/actions/runs/4494004180). I see this error:
> Invalid workflow file: .github/workflows/e2e.yml#L1
> you may only define up to 10 `inputs` for a `workflow_dispatch` event

We may need to temporarily steal a parameter. You can safely remove `security_scan_exit_code` for now, then add it back when the `nosshkeys_vertica_image` parm is removed.

PR #362, opened 2023-04-03
Support subcluster names with underscore

roypaulin:
Before this, a subcluster's name couldn't contain the underscore character because it is part of sts and pod names and according to FQDN rules, an underscore is not a valid character. 

This adds better support by simply mapping '_' to a '-' before building pod/service/sts name. This will allow us to support the default subcluster name in vertica – default_subcluster. 

The webhook now allows subcluster name with '_'. However, two subclusters could have have identical sts names: e.g. default_subcluster and default-subcluster. We reject that case now in the webhook.

### DIFF
--- a/api/v1beta1/verticadb_types.go
+++ b/api/v1beta1/verticadb_types.go
@@ -1131,11 +1131,18 @@ func (s *Subcluster) GetType() string {
 	return SecondarySubclusterType
 }
 
+// GenNameWithoutUnderscore returns the subcluster name after
+// replacing all `_` occurrences with `-`
+func (s *Subcluster) GenNameWithoutUnderscore() string {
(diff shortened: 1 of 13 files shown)


spilchen (commented):
Just a few minor suggestions

spilchen:
> I had a problem with problem with golangci-lint. The current version consumes way to much memory, I had to upgrade to v1.51.1 to run it locally. We may want to change version. Also there are some linter that are deprecated since v1.49.0.

I was aware that some of the linters were deprecated. There is no replacements for them, so I am inclined to keep using them for now. If you want to upgrade the linter, feel free to provide that in a separate PR.

spilchen (approved):

PR #369, opened 2023-04-06
Fix helm install without cluster admin priv

spilchen:
Normally, when installing the operator via a helm chart you need full cluster admin privileges. There is a path that you can take install the operator if you don't have cluster admin priv. That was broken a few months ago. This PR is fixing that.

The root cause was that cluster roles/rolebindings were added to the operator so that it could patch the webhook config objects. This is only needed when the operator generates its own cert for the webhook. We are adding conditional logic in those manifests to only apply them if the operator is generating the certs.

New e2e testcase was added to verify install without cluster admin priv.


helm-charts/verticadb-operator/values.yaml
```diff
@@ -119,14 +119,19 @@ resources:
...
+# Set this to true to force the helm chart to skip creation of any Roles,
+# RoleBindings and ServiceAccount. This assumes that the roles that the
+# operator relies on have been setup outside of this helm chart.
 #
 # Use this option if you are installing the helm chart with k8s privileges that
 # prevent you from creating Roles/RoleBindings. We provide the Roles and
-# RoleBindings that the operator needs as a artifact of the GitHub release (see
-# https://github.com/vertica/vertica-kubernetes/releases).
+# RoleBindings that the operator needs as a artifact of the GitHub release. For
+# instance, have your cluster admin run this against the namespace you are
```
roypaulin, line 129:
```suggestion
# instance, have your cluster admin run this against the namespace you
```

tests/e2e-leg-3/non-cluster-admin-deployment/15-run-cluster-admin-tasks.yaml
```diff
@@ -0,0 +1,48 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: fred-cluster-access
+subjects:
+- kind: User
+  name: fred
+  apiGroup: rbac.authorization.k8s.io
+roleRef:
+  kind: ClusterRole
```
roypaulin, line 32: Couldn't this be a `Role`?

tests/e2e-leg-3/non-cluster-admin-deployment/15-run-cluster-admin-tasks.yaml
```diff
@@ -0,0 +1,48 @@
...
+  # Setup the namespace for the operator
+  - script: kubectl apply -n $NAMESPACE -f ../../../config/release-manifests/operator-rbac.yaml
+  # Install the CRDs
+  - command: sh -c "cd ../../.. && make install"
+---
+# Give user 'fred' general access to edit things inside its namespace
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: fred-cluster-access
```
roypaulin, line 26: Wasn't the goal to give limited access to the cluster.

roypaulin (commented):
I have a few questions for now.

tests/e2e-leg-3/non-cluster-admin-deployment/15-run-cluster-admin-tasks.yaml
spilchen: User Fred needs some access to do things like create pods, etc. This is a RoleBinding so it only gives Fred access to to do things inside the namespace.

tests/e2e-leg-3/non-cluster-admin-deployment/15-run-cluster-admin-tasks.yaml
spilchen: It doesn't matter. The point is that we are binding it with a RoleBinding, so all of the things that it grants are limited to the current namespace. Choosing CusterRole allows me to reuse one of the pre-canned ones rather than creating my own Role.

0001-Fixes.patch
```diff
@@ -0,0 +1,150 @@
+From 4d0a9a1a82d93b241810446bf0b19eb9d9793f2b Mon Sep 17 00:00:00 2001
```
roypaulin, line 1: What is this file for?

spilchen: This was included by mistake. I'll remove it.

roypaulin (approved):

PR #377, opened 2023-04-17
Event Trigger

albertogg:
We need the ability to run arbitrary scripts immediately following a new database creation.

### DIFF
--- a/pkg/controllers/et/createet_reconciler.go
+++ b/pkg/controllers/et/createet_reconciler.go
@@ -1,0 +1,91 @@
+package et
+
+import (
+	"context"
+	"fmt"
+
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/controllers"
+	"github.com/vertica/vertica-kubernetes/pkg/etstatus"
+	batchv1 "k8s.io/api/batch/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type CreateETReconciler struct {
+	VRec *EventTriggerReconciler
+	Et   *vapi.EventTrigger
+}
+
+func MakeCreateETReconciler(r *EventTriggerReconciler, et *vapi.EventTrigger) controllers.ReconcileActor {
+	return &CreateETReconciler{VRec: r, Et: et}
+}
+
+func (r *CreateETReconciler) Reconcile(ctx context.Context, req *reconcile.Request) (reconcile.Result, error) {
+	for _, reference := range r.Et.Spec.References {
+		if reference.Object.Kind != "VerticaDB" {
+			err := fmt.Errorf("unexpected type: %s", reference.Object.Kind)
+			r.VRec.Log.Error(err, "checking for reference")
+			return ctrl.Result{}, err
+		}
+
+		vdb := &vapi.VerticaDB{}
+
+		nm := types.NamespacedName{
+			Namespace: r.Et.Namespace,
+			Name:      reference.Object.Name,
+		}
+
+		if err := r.VRec.Client.Get(ctx, nm, vdb); err != nil {
+			if errors.IsNotFound(err) {
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, err
+		}
+
+		if len(vdb.Status.Conditions) <= vapi.DBInitializedIndex {
+			return ctrl.Result{}, nil
+		}
+
+		for _, match := range r.Et.Spec.Matches {
+			if vdb.Status.Conditions[vapi.DBInitializedIndex].Status == match.Condition.Status {
+				// Check if job already created.
+				if r.Et.Status.References != nil {
+					return ctrl.Result{}, nil
+				}
+
+				job := batchv1.Job{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: r.Et.Spec.Template.Metadata.Name,
+					},
+					Spec: r.Et.Spec.Template.Spec,
+				}
+
+				if err := r.VRec.Client.Create(ctx, &job); err != nil {
+					return ctrl.Result{}, err
+				}
+
+				// Update status to fill in JobName,JobNamespace
+				refStatus := vapi.ETRefObjectStatus{
+					Kind:         reference.Object.Kind,
+					JobNamespace: r.Et.Namespace,
+					JobName:      job.Name,
+				}
+				if err := etstatus.Apply(ctx, r.VRec.Client, r.Et, &refStatus); err != nil {
+					return ctrl.Result{}, err
+				}
+			} else {
+				r.VRec.Log.Info(
+					"status was not met",
+					"expected", match.Condition.Status,
+					"found", vdb.Status.Conditions[vapi.DBInitializedIndex].Status,
+				)
+			}
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -128,6 +128,7 @@ func addReconcilersToManager(mgr manager.Manager, restCfg *rest.Config, oc *opcf
 	if err := (&et.EventTriggerReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
+		Log:    ctrl.Log.WithName("controllers").WithName("EventTrigger"),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "EventTrigger")
 		os.Exit(1)
--- a/pkg/controllers/et/eventtrigger_controller.go
+++ b/pkg/controllers/et/eventtrigger_controller.go
@@ -29,7 +29,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
@@ -44,6 +43,7 @@ import (
 type EventTriggerReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
+	Log    logr.Logger
 }
 
 const (
@@ -61,35 +61,35 @@ const (
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
 func (r *EventTriggerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	lg := log.FromContext(ctx).WithValues("et", req.NamespacedName)
-	lg.Info("starting reconcile of EventTrigger")
+	log := r.Log.WithValues("et", req.NamespacedName)
+	log.Info("starting reconcile of EventTrigger")
 
 	et := &vapi.EventTrigger{}
 	err := r.Get(ctx, req.NamespacedName, et)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			// Request object not found, cound have been deleted after reconcile request.
-			lg.Info("EventTrigger resource not found.  Ignoring since object must be deleted")
+			log.Info("EventTrigger resource not found.  Ignoring since object must be deleted")
 			return ctrl.Result{}, nil
 		}
-		lg.Error(err, "failed to get EventTrigger")
+		log.Error(err, "failed to get EventTrigger")
 		return ctrl.Result{}, err
 	}
 
 	// Iterate over each actor
-	actors := r.constructActors(lg, et)
+	actors := r.constructActors(et)
 	var res ctrl.Result
 	for _, act := range actors {
-		lg.Info("starting actor", "name", fmt.Sprintf("%T", act))
+		log.Info("starting actor", "name", fmt.Sprintf("%T", act))
 		res, err = act.Reconcile(ctx, &req)
 		// Error or a request to requeue will stop the reconciliation.
 		if verrors.IsReconcileAborted(res, err) {
-			lg.Info("aborting reconcile of VerticaDB", "result", res, "err", err)
+			log.Info("aborting reconcile of VerticaDB", "result", res, "err", err)
 			return res, err
 		}
 	}
 
-	lg.Info("ending reconcile of EventTrigger", "result", res, "err", err)
+	log.Info("ending reconcile of EventTrigger", "result", res, "err", err)
 	return res, err
 }
 
@@ -156,7 +156,9 @@ func (r *EventTriggerReconciler) findObjectsForVerticaDB(vdb client.Object) []re
 // constructActors will a list of actors that should be run for the reconcile.
 // Order matters in that some actors depend on the successeful execution of
 // earlier ones.
-func (r *EventTriggerReconciler) constructActors(lg logr.Logger, et *vapi.EventTrigger) []controllers.ReconcileActor {
+func (r *EventTriggerReconciler) constructActors(et *vapi.EventTrigger) []controllers.ReconcileActor {
 	// The actors that will be applied, in sequence, to reconcile an et.
-	return []controllers.ReconcileActor{}
+	return []controllers.ReconcileActor{
+		MakeCreateETReconciler(r, et),
+	}
 }


pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
+/*
+Copyright [2021-2023] Micro Focus or one of its affiliates.
```
spilchen, line 2: We recently updated the copyright text. 
```suggestion
Copyright [2021-2023] Open Text.
```

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+		nm := types.NamespacedName{
+			Namespace: r.Et.Namespace,
+			Name:      reference.Object.Name,
+		}
+
+		if err := r.VRec.Client.Get(ctx, nm, vdb); err != nil {
+			if errors.IsNotFound(err) {
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, err
```
spilchen, line 62: We should go onto the next reference if we couldn't find the object. To be clear, we could have multiple matches, but they apply to individual objects. Only some objects have a matches.

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/vertica/vertica-kubernetes/pkg/etstatus"
+	batchv1 "k8s.io/api/batch/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type CreateETReconciler struct {
```
spilchen, line 34: Suggest a different name for this struct.
```suggestion
type VerticaDBRefReconciler struct {
```

If we support other types of references, we may end up have a reconciler for each.

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+	return &CreateETReconciler{VRec: r, Et: et}
+}
+
+func (r *CreateETReconciler) Reconcile(ctx context.Context, req *reconcile.Request) (reconcile.Result, error) {
+	for _, reference := range r.Et.Spec.References {
+		if reference.Object.Kind != "VerticaDB" {
+			err := fmt.Errorf("unexpected type: %s", reference.Object.Kind)
+			r.VRec.Log.Error(err, "checking for reference")
+			return ctrl.Result{}, err
+		}
```
spilchen, line 49: Lets add this logic into the webhook api/v1beta1/eventtrigger_webhook.go. For this reconciler, if the GVK doesn't match we treat it like a no-op.

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+	Et   *vapi.EventTrigger
+}
+
+func MakeCreateETReconciler(r *EventTriggerReconciler, et *vapi.EventTrigger) controllers.ReconcileActor {
+	return &CreateETReconciler{VRec: r, Et: et}
+}
+
+func (r *CreateETReconciler) Reconcile(ctx context.Context, req *reconcile.Request) (reconcile.Result, error) {
+	for _, reference := range r.Et.Spec.References {
+		if reference.Object.Kind != "VerticaDB" {
```
spilchen, line 45: We should check the apiVersion too.

spilchen, line 45: There is a constant for this, see `VerticaDBKind` in the vapi package.

spilchen (commented):
Some early comments. I'll take another pass tomorrow.

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		if err := r.VRec.Client.Get(ctx, nm, vdb); err != nil {
+			if errors.IsNotFound(err) {
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, err
+		}
+
+		if len(vdb.Status.Conditions) <= vapi.DBInitializedIndex {
```
spilchen, line 65: This is specific to the DB initialized condition. We need this to be more generic. Depending on the status condition in the spec, we need to search for it. It may make sense to just iterate through the conditions and see if the type matches the one in the et spec.

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+			return ctrl.Result{}, nil
```
spilchen, line 66: We will need to continue to next reference if we get here rather than exiting.

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		for _, match := range r.Et.Spec.Matches {
+			if vdb.Status.Conditions[vapi.DBInitializedIndex].Status == match.Condition.Status {
+				// Check if job already created.
+				if r.Et.Status.References != nil {
+					return ctrl.Result{}, nil
```
spilchen, line 73: We will need to go to the next reference object. It might be better if we pushed all of the logic for a specific reference object into its own separate function. Then you can call return if we are to skip the object for any reason.

pkg/controllers/et/createet_reconciler.go
spilchen, line 45: We should have a check to see if the reference is a no-op to begin with. If the job has already been created, we should just skip the reference object.

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+				}
+
+				job := batchv1.Job{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: r.Et.Spec.Template.Metadata.Name,
```
spilchen, line 78: There are other things we need to fill in here. Basically everything in `r.Et.Spec.Template.Metadata` should find a home in the new job.

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+				}
+
+				if err := r.VRec.Client.Create(ctx, &job); err != nil {
+					return ctrl.Result{}, err
+				}
+
+				// Update status to fill in JobName,JobNamespace
+				refStatus := vapi.ETRefObjectStatus{
+					Kind:         reference.Object.Kind,
+					JobNamespace: r.Et.Namespace,
```
spilchen, line 90: Does the job that we construct have the namespace filled in? We should use that if it was set.

pkg/controllers/et/createet_reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		for _, match := range r.Et.Spec.Matches {
+			if vdb.Status.Conditions[vapi.DBInitializedIndex].Status == match.Condition.Status {
+				// Check if job already created.
+				if r.Et.Status.References != nil {
+					return ctrl.Result{}, nil
+				}
+
+				job := batchv1.Job{
```
spilchen, line 76: When we support other types other than VerticaDB, Job creation will likely be the same. So, we may want to future-proof it by putting it in a common function.

pkg/controllers/et/createet_reconciler_test.go
spilchen, line 1: I like to pair each .go file with a `_test.go` file. Lets keep `eventtrigger_controller_test.go` and add a new one for your new reconciler file.

pkg/controllers/et/verticadbref_reconciler.go
```diff
@@ -0,0 +1,121 @@
...
+		vdb := &vapi.VerticaDB{}
+
+		nm := types.NamespacedName{
+			Namespace: r.Et.Namespace,
+			Name:      ref.Object.Name,
+		}
+
+		if err := r.VRec.Client.Get(ctx, nm, vdb); err != nil {
+			if errors.IsNotFound(err) {
+				continue
```
spilchen, line 57: We still need to update the status before we get here. Only part of the status needs to be updated. We won't have resourceVersion, UID, etc. But I want to see a status update as soon the reconciler runs. Even if there is no reference object created yet.

pkg/controllers/et/verticadbref_reconciler.go
```diff
@@ -0,0 +1,121 @@
...
+
+			return ctrl.Result{}, err
+		}
+
+		for _, match := range r.Et.Spec.Matches {
+			// Grab the condition based on what was given.
+			conditionType := vapi.VerticaDBConditionType(match.Condition.Type)
+			conditionTypeIndex, ok := vapi.VerticaDBConditionIndexMap[conditionType]
+			if !ok {
+				return ctrl.Result{}, fmt.Errorf("vertica DB condition '%s' missing from VerticaDBConditionType", conditionType)
```
spilchen, line 68: An error might be a bit hard here. Keep in mind the condition is something the user provides. So, if they misspell the condition, they will hit this error each time. It will make for a lot of noise in the operator log. I would just write a log.info message and continue to the next reference object.

pkg/controllers/et/verticadbref_reconciler.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+
+			if len(vdb.Status.Conditions) <= conditionTypeIndex {
+				continue
```
spilchen, line 72: We need all of the matches to pass before we consider creating the job. We can only have one so far, but the design was if there were multiple matches they are all treated as if they were ANDed together.

pkg/controllers/et/verticadbref_reconciler.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+
+			if vdb.Status.Conditions[conditionTypeIndex].Status == match.Condition.Status {
+				// Check if job already created.
+				if r.Et.Status.References != nil {
+					continue
```
spilchen, line 78: Similar comment for here. This continue should really continue to the next reference object, not the next match.

pkg/controllers/et/verticadbref_reconciler.go
```diff
@@ -0,0 +1,121 @@
...
+				}
+
+				// Update status to fill in JobName,JobNamespace
+				refStatus := vapi.ETRefObjectStatus{
+					APIVersion:      ref.Object.APIVersion,
+					Namespace:       ref.Object.Namespace,
+					Name:            ref.Object.Name,
+					Kind:            ref.Object.Kind,
+					ResourceVersion: r.Et.ResourceVersion,
+					UID:             r.Et.GetUID(),
```
spilchen, line 103: Both of these needs to be from the reference object (i.e. from vdb)

pkg/controllers/et/verticadbref_reconciler.go
```diff
@@ -0,0 +1,121 @@
...
+					JobNamespace:    r.Et.Namespace,
```
spilchen, line 104: Can we get the name space from job instead?
```suggestion
					JobNamespace:    job.Namespace,
```

pkg/controllers/et/verticadbref_reconciler_test.go
```diff
@@ -0,0 +1,95 @@
...
+	ctx := context.Background()
+
+	It("should succeed with no-op on objects types other than VerticaDB", func() {
+		et := vapi.MakeET()
+		et.Spec.References[0].Object.Kind = "Pod"
+
+		Expect(k8sClient.Create(ctx, et)).Should(Succeed())
+		defer func() { Expect(k8sClient.Delete(ctx, et)).Should(Succeed()) }()
+
+		Expect(etRec.Reconcile(ctx, ctrl.Request{NamespacedName: et.ExtractNamespacedName()})).Should(Equal(ctrl.Result{}))
```
spilchen, line 39: We should check the status of the eventTrigger. The reference object should be set. We may need to create the EventTrigger object for this to work.

pkg/controllers/et/verticadbref_reconciler_test.go
```diff
@@ -0,0 +1,95 @@
...
+		et.Spec.Matches[0].Condition.Type = "NotWorking"
+
+		Expect(k8sClient.Create(ctx, et)).Should(Succeed())
+		defer func() { Expect(k8sClient.Delete(ctx, et)).Should(Succeed()) }()
+
+		_, err := etRec.Reconcile(ctx, ctrl.Request{NamespacedName: et.ExtractNamespacedName()})
+		Expect(err).ShouldNot(Succeed())
+	})
+
+	It("should succeed with no-op when object already exists", func() {
```
spilchen, line 76: We should have a test that does create the job when the condition variables pass.

api/v1beta1/eventtrigger_webhook.go
spilchen, line 1: We should add some checks to make sure that we have exactly 1 reference object and 1 match.

pkg/controllers/et/verticadbref_reconciler.go
albertogg: I'm not sure I understand here

spilchen: I would like the status to be updated anytime the controller runs. It will be an indication that the controller ran and saw the ET. So, we need status updates for cases where we haven't yet trigger the job.

pkg/controllers/et/verticadbref_reconciler_test.go
albertogg: I've done this in other tests, but should we also update the status of the eventTrigger when the Object Kind is not VerticaDB?

spilchen: I think that's fine to leave out. We have a webhook that will prevent it, so it's not a code path that will get run.

pkg/controllers/et/verticadbref_reconciler.go
```diff
@@ -0,0 +1,131 @@
...
+		shouldCreateJob := true
+		for _, match := range r.Et.Spec.Matches {
+			if !r.matchStatus(vdb, ref, match) {
+				shouldCreateJob = false
+				break
+			}
+		}
+
+		// Check if job already created.
+		if len(r.Et.Status.References) > 0 && r.Et.Status.References[refIdx].JobName != "" {
```
spilchen, line 82: Let's not assume the reference object in the spec shares the same index in the status. Can we iterate over all of the status references checking each one?

pkg/controllers/et/verticadbref_reconciler.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// matchStatus will check if the matching condition given from the manifest
+// matches with the reference object and return false when it doesn't match.
+func (r *VerticaDBRefReconciler) matchStatus(vdb *vapi.VerticaDB, ref vapi.ETReference, match vapi.ETMatch) bool {
+	// Grab the condition based on what was given.
+	conditionType := vapi.VerticaDBConditionType(match.Condition.Type)
+	conditionTypeIndex, ok := vapi.VerticaDBConditionIndexMap[conditionType]
+	if !ok {
+		r.VRec.Log.Info(fmt.Sprintf("vertica DB condition %s missing from VerticaDBConditionType", match.Condition.Type))
```
spilchen, line 112: minor: we don't need to use sprintf here. Info() takes multiple arguments that allow you to pass in key/value pairs.
```suggestion
		r.VRec.Log.Info("vertica DB condition missing from VerticaDBConditionType", "condition", match.Condition.Type)
```

pkg/controllers/et/verticadbref_reconciler_test.go
spilchen, line 1: A few other suggestions for test:
- case where we actually the job object. We should be able to fetch the job from the apiserver and verify its contents
- where it's a no-op because the status condition doens't match

api/v1beta1/eventtrigger_webhook.go
```diff
@@ -47,13 +57,25 @@ var _ webhook.Validator = &EventTrigger{}
...
 func (e *EventTrigger) ValidateUpdate(old runtime.Object) error {
 	eventtriggerlog.Info("validate update", "name", e.Name)
-	return nil
+
+	allErrs := e.validateVerticaDBRefSpec()
+	if allErrs == nil {
+		return nil
+	}
+
+	return apierrors.NewInvalid(schema.GroupKind{Group: Group, Kind: EventTriggerKind}, e.Name, allErrs)
```
spilchen, line 78: We can replace the first function with `GkET`. Similar comment in other spots in this file.

api/v1beta1/eventtrigger_webhook_test.go
spilchen, line 1: Can we have a test that will validate update? I think all of the test cases are only validating create.

spilchen (commented):
It's looking in good shape. Just a few more comments. But, I think we are really close here.

pkg/controllers/et/verticadbref_reconciler.go
albertogg: I changed this but I'm not sure if what I did works or I need to match additional things

api/v1beta1/eventtrigger_webhook.go
```diff
@@ -126,7 +125,7 @@ func (e *EventTrigger) validateVerticaDBReferencesSize(allErrs field.ErrorList)
 		err := field.Invalid(
 			field.NewPath("spec").Child("reference"),
 			ref,
-			fmt.Sprintf("only %d referece object is allowed, number received: %d", allowedNumberReferences, len(ref)),
+			fmt.Sprintf("only %d referece object allowed, number received: %d", allowedNumberReferences, len(ref)),
```
spilchen, line 128: spelling
```suggestion
			fmt.Sprintf("only %d reference object allowed, number received: %d", allowedNumberReferences, len(ref)),
```

pkg/controllers/et/verticadbref_reconciler.go
```diff
@@ -102,14 +101,25 @@ func (r *VerticaDBRefReconciler) Reconcile(ctx context.Context, req *ctrl.Reques
 	return ctrl.Result{}, nil
 }
 
+func (r *VerticaDBRefReconciler) isJobCreated(ref vapi.ETReference) bool {
+	for refStatusIdx := range r.Et.Status.References {
+		refStatus := r.Et.Status.References[refStatusIdx]
+		if refStatus.Name == ref.Object.Name && refStatus.JobName != "" {
```
spilchen, line 107: We should check the kind and apiVersion as well.

pkg/controllers/et/verticadbref_reconciler_test.go
```diff
@@ -0,0 +1,156 @@
...
+func getEventTriggerStatus(ctx context.Context, nm types.NamespacedName) vapi.EventTrigger {
+	etrigger := vapi.EventTrigger{}
+	etStatus := k8sClient.Get(ctx, nm, &etrigger)
+	Expect(etStatus).Should(Succeed())
+
+	return etrigger
+}
+
+func setVerticaStatus(ctx context.Context, clnt client.Client, vdb *vapi.VerticaDB, conditions []vapi.VerticaDBCondition) error {
+	return vdbstatus.Update(ctx, clnt, vdb, func(vdb *vapi.VerticaDB) error {
```
spilchen, line 152: can you use `vdbstatus.UpdateCondition` for this? I know I said otherwise when you asked. But now I get the full context. The test that calls into here fails for me, and I think using `UpdateCondition` will resolve that.

spilchen (approved):

PR #378, opened 2023-04-18
Duane/add dblink

NerdLogic:
This installs dblink on the image, but doesn't load it into vertica.

### DIFF
--- a/.github/actions/download-rpm/action.yaml
+++ b/.github/actions/download-rpm/action.yaml
@@ -7,5 +7,7 @@ runs:
       shell: bash
       env:
           VERTICA_CE_URL: "https://vertica-community-edition-for-testing.s3.amazonaws.com/XCz9cp7m/vertica-12.0.3-0.x86_64.RHEL6.rpm"
+          DBLINK_URL: "https://github.com/vertica/dblink/releases/download/v0.2.1/ldblink.so.ubuntu-v12.0.4"
       run: |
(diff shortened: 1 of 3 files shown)


spilchen (commented):
This will break the nightly build, and I don't see an easy way to fix this. The nightly build runs against the latest server version, which is not yet released. However, this adds a new dependency that we pull a released dblink from the dblink repo. But, we can never fulfill this dependency because we are using an unreleased version of the server.

NerdLogic:
> This will break the nightly build, and I don't see an easy way to fix this. The nightly build runs against the latest server version, which is not yet released. However, this adds a new dependency that we pull a released dblink from the dblink repo. But, we can never fulfill this dependency because we are using an unreleased version of the server.

Yes.  Don't merge this until we have the UDX container build and the DBLink build automated and a mechanism to upload ldblink.so in the same way that the rpm is uploaded.  There are JIRAs for that effort already.  I didn't want to lose track of these changes, so I created a PR.

Optionally, we could just add the unixodbc package and that already saves the customer from needing to create their own image.  They would just need to manually download the correct ldblink.so and install it on a running k8s container just like they would for any other udx.  Well, they would also need to configure their odbc settings on each k8s container which is something we can't really package for them.

spilchen:
I kind of like your idea of just adding the unixodbc package. Otherwise, this PR may sit around a while waiting for dblink automation to happen.

spilchen:
Closing for now. We can pick these changes up again in the future.

PR #385, opened 2023-04-24
Support emptyDir as the depot volume

roypaulin:
A new option is provided to allow the depot to be created as an emptyDir.
For that a new CR field `depotVolume` has been added and can take 2 possible values: `PersistentVolume` and `EmptyDir`.
- `PersistentVolume` is the default value and represents the way it means all data are to be stored in a PV. This is the way the operator has been working so far.
- `EmptyDir`: only the depot will be stored in an emptyDir. Here is an example:
  ```
  kind: VerticaDB
  metadata:
    name: sample
  spec:
    ......
    local:
      depotVolume: EmptyDir
      requestSize: 500Gi  
    subclusters:
      - name: sc1
  ```

### DIFF
--- a/api/v1beta1/verticadb_webhook.go
+++ b/api/v1beta1/verticadb_webhook.go
@@ -891,7 +891,7 @@ func (v *VerticaDB) validateDepotVolume(allErrs field.ErrorList) field.ErrorList
 	if !v.IsKnownDepotVolumeType() {
 		err := field.Invalid(field.NewPath("spec").Child("local").Child("depotVolume"),
 			v.Spec.Local.DepotVolume,
-			fmt.Sprintf("valid values are %s and %s", EmptyDir, PersistentVolume))
+			fmt.Sprintf("valid values are %s, %s or an empty string", EmptyDir, PersistentVolume))
 		allErrs = append(allErrs, err)
 	}
 	return allErrs
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -113,10 +113,9 @@ func buildVolumeMounts(vdb *vapi.VerticaDB) []corev1.VolumeMount {
 	if vdb.Spec.Local.DataPath != vdb.Spec.Local.DepotPath {
 		if vdb.IsDepotVolumeEmptyDir() {
 			// If depotVolume is EmptyDir, the depot is stored in its own 'emptyDir' volume
-			vmts := []corev1.VolumeMount{
-				{Name: vapi.DepotMountName, MountPath: vdb.Spec.Local.DepotPath},
-			}
-			volMnts = append(volMnts, vmts...)
+			volMnts = append(volMnts, corev1.VolumeMount{
+				Name: vapi.DepotMountName, MountPath: vdb.Spec.Local.DepotPath,
+			})
 		} else {
 			volMnts = append(volMnts, corev1.VolumeMount{
 				Name: vapi.LocalDataPVC, SubPath: vdb.GetPVSubPath("depot"), MountPath: vdb.Spec.Local.DepotPath,
@@ -433,8 +432,8 @@ func buildHTTPServerSecretVolume(vdb *vapi.VerticaDB) corev1.Volume {
 	}
 }
 
-// buildEmtyDirVolume returns a generic 'emptyDir' volume
-func buildEmtyDirVolume(volName string) corev1.Volume {
+// buildEmptyDirVolume returns a generic 'emptyDir' volume
+func buildEmptyDirVolume(volName string) corev1.Volume {
 	return corev1.Volume{
 		Name: volName,
 		VolumeSource: corev1.VolumeSource{
@@ -445,7 +444,7 @@ func buildEmtyDirVolume(volName string) corev1.Volume {
 
 // buildDepotVolume returns an 'emptyDir' volume for the depot
 func buildDepotVolume() corev1.Volume {
-	return buildEmtyDirVolume(vapi.DepotMountName)
+	return buildEmptyDirVolume(vapi.DepotMountName)
 }
 
 // buildPodSpec creates a PodSpec for the statefulset
--- a/pkg/reviveplanner/analyze.go
+++ b/pkg/reviveplanner/analyze.go
@@ -106,13 +106,14 @@ func (a *ATPlanner) ApplyChanges(vdb *vapi.VerticaDB) (updated bool, err error) 
 	if depotPath != vdb.Spec.Local.DepotPath {
 		a.logPathChange("depot", vdb.Spec.Local.DepotPath, depotPath)
 		vdb.Spec.Local.DepotPath = depotPath
-		if vdb.IsDepotVolumeEmptyDir() && !vdb.Spec.Local.IsDepotPathUnique() {
-			a.Log.Info("depot path not unique, depotVolume has to change to PersistentVolume")
-			// Because when depotVolume is EmptyDir, we cannot have depot path
-			// equal to catalog or data path. We will instead have PersistentVolume
-			// as depot volume.
-			vdb.Spec.Local.DepotVolume = vapi.PersistentVolume
-		}
+		updated = true
+	}
+	if vdb.IsDepotVolumeEmptyDir() && !vdb.Spec.Local.IsDepotPathUnique() {
+		a.Log.Info("depot path not unique, depotVolume has to change to PersistentVolume")
+		// Because when depotVolume is EmptyDir, we cannot have depot path
+		// equal to catalog or data path. We will instead have PersistentVolume
+		// as depot volume.
+		vdb.Spec.Local.DepotVolume = vapi.PersistentVolume
 		updated = true
 	}
 
--- a/pkg/reviveplanner/analyze_test.go
+++ b/pkg/reviveplanner/analyze_test.go
@@ -158,7 +158,6 @@ var _ = Describe("analyze", func() {
 		origVdb := vdb.DeepCopy()
 
 		// Change some things in vdb that the planner will change back
-		vdb.Spec.Local.DepotPath = "/depot-dir"
 		vdb.Spec.Local.DataPath = "/data-dir"
 		vdb.Spec.Local.DepotVolume = vapi.EmptyDir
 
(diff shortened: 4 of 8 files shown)


api/v1beta1/verticadb_webhook.go
```diff
@@ -869,6 +876,26 @@ func (v *VerticaDB) validateLocalPaths(allErrs field.ErrorList) field.ErrorList
...
+	}
+	return allErrs
+}
+
+func (v *VerticaDB) validateDepotVolume(allErrs field.ErrorList) field.ErrorList {
+	if !v.IsKnownDepotVolumeType() {
+		err := field.Invalid(field.NewPath("spec").Child("local").Child("depotVolume"),
+			v.Spec.Local.DepotVolume,
+			fmt.Sprintf("local.depotVolume, if specified, can only be %s or %s",
+				EmptyDir, PersistentVolume))
```
spilchen, line 896: Looking at the output from this on the CLI, the `local.depotVolume` is kind of repetitive.

```
 » k apply -f ~/m/vdb-1node-12.0.3-unstripped-hostpath.yaml
The VerticaDB "v" is invalid: spec.local.depotVolume: Invalid value: "EmptyDirs": local.depotVolume, if specified, can only be EmptyDir or PersistentVolume
```

```suggestion
			fmt.Sprintf("valid values are %s and %s", EmptyDir, PersistentVolume))
```

api/v1beta1/verticadb_webhook.go
```diff
@@ -1011,6 +1038,18 @@ func (v *VerticaDB) checkImmutableS3ServerSideEncryption(oldObj *VerticaDB, allE
 	return allErrs
 }
 
+// checkImmutableDepotVolume will make sure local.depotVolume
+// does not change after creation.
+func (v *VerticaDB) checkImmutableDepotVolume(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {
+	if v.Spec.Local.DepotVolume != oldObj.Spec.Local.DepotVolume {
+		err := field.Invalid(field.NewPath("spec").Child("local").Child("depotVolume"),
+			v.Spec.Local.DepotVolume,
+			"local.depotVolume cannot change after creation")
```
spilchen, line 1047: Similar comment here about repeating the field name.

Here is the output if you try to violate this rule:
```
# verticadbs.vertica.com "v" was not valid:
# * spec.local.depotVolume: Invalid value: "PersistentVolume": local.depotVolume cannot change after creation
```

Suggest the following:
```suggestion
			"cannot change after creation")
```

pkg/builder/builder.go
```diff
@@ -111,9 +111,18 @@ func buildVolumeMounts(vdb *vapi.VerticaDB) []corev1.VolumeMount {
 	// container. Otherwise, you will get multiple mount points shared the same
 	// path, which will prevent any pods from starting.
 	if vdb.Spec.Local.DataPath != vdb.Spec.Local.DepotPath {
-		volMnts = append(volMnts, corev1.VolumeMount{
-			Name: vapi.LocalDataPVC, SubPath: vdb.GetPVSubPath("depot"), MountPath: vdb.Spec.Local.DepotPath,
-		})
+		if vdb.IsDepotVolumeEmptyDir() {
+			// If depotVolume is EmptyDir, the depot is stored in its own 'emptyDir' volume
+			vmts := []corev1.VolumeMount{
+				{Name: vapi.DepotMountName, MountPath: fmt.Sprintf("%s/%s", paths.LocalDataPath, vdb.GetPVSubPath("depot"))},
```
spilchen, line 117: I don't think we need this mount path. I think the /home/dbadmin/local-data should just point to things that are stored in the PV. The mount path the next line should be all that we need to setup the depot dir.

tests/e2e-leg-2/restart-node-multi-sc/setup-vdb/base/setup-vdb.yaml
spilchen, line 1: Can we convert a few more e2e tests to use emptyDir? They don't have to be restart ones.

pkg/vdbgen/options.go
```diff
@@ -33,4 +33,5 @@ type Options struct {
 	CACertName         string
 	Krb5Conf           string
 	Krb5Keytab         string
+	DepotVolume        string
```
spilchen, line 36: Can we update the vdb-gen e2e test by using this new parameter?

api/v1beta1/verticadb_types.go
```diff
@@ -1312,6 +1329,28 @@ func (v *VerticaDB) IsAgentEnabled() bool {
...
+// IsDepotVolumeEmptyDir returns true if the depot volume's type
+// is emptyDir.
+func (v *VerticaDB) IsDepotVolumeEmptyDir() bool {
+	return v.Spec.Local.DepotVolume == EmptyDir
+}
+
+// IsDepotVolumePersistentVolume returns true if the depot volume's type
+// is persistentVolume.
+func (v *VerticaDB) IsDepotVolumePersistentVolume() bool {
+	return v.Spec.Local.DepotVolume == PersistentVolume
```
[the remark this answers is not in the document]
roypaulin: I followed the way `InitPolicy` has been defined. The default value is `PersistentVolume` so if `depotVolume` is not set in the CR, this value will be assumed; if it is set to empty string, it will be rejected.

api/v1beta1/verticadb_types.go
```diff
@@ -629,6 +638,13 @@ type LocalStorage struct {
 	// depot path used when the database was first created.
 	DepotPath string `json:"depotPath"`
 
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=PersistentVolume
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:PersistentVolume","urn:alm:descriptor:com.tectonic.ui:select:EmptyDir"}
+	// The type of volume to use for the depot. Allowable values will be: EmptyDir and PersistentVolume.
+	// If omitted, the default value will be PersistentVolume.
+	DepotVolume DepotVolumeType `json:"depotVolume"`
```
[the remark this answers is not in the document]
roypaulin: Empty string is not a valid value. This is handled like `InitPolicy`.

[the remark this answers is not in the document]
roypaulin: Do we need `omitempty` because of VerticaDB's that were created with an older operator?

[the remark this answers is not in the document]
spilchen: yeah, empty string needs to be a valid value.

api/v1beta1/verticadb_webhook.go
```diff
@@ -869,6 +876,26 @@ func (v *VerticaDB) validateLocalPaths(allErrs field.ErrorList) field.ErrorList
 			fieldRef, fmt.Sprintf("%s cannot be set to %s. This is a restricted path.", fieldPathName, invalidPath))
 		allErrs = append(allErrs, err)
 	}
+	// When depotVolume is EmptyDir, depotPath must be different from data and catalog paths.
+	if v.IsDepotVolumeEmptyDir() {
+		if v.Spec.Local.DepotPath == v.Spec.Local.DataPath ||
```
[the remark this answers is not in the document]
roypaulin: Now I see what you mean and this webhook rule will reject the update on this specific scenario. A way to fix this would be to also update depotVolume to `PersistentVolume` and log a msg like `cannot use emptyDir as depot volume to revive db with depotPath=dataPath, a persistent volume will be used.`

pkg/builder/builder.go
```diff
@@ -111,9 +111,17 @@ func buildVolumeMounts(vdb *vapi.VerticaDB) []corev1.VolumeMount {
...
 	if vdb.Spec.Local.DataPath != vdb.Spec.Local.DepotPath {
-		volMnts = append(volMnts, corev1.VolumeMount{
-			Name: vapi.LocalDataPVC, SubPath: vdb.GetPVSubPath("depot"), MountPath: vdb.Spec.Local.DepotPath,
-		})
+		if vdb.IsDepotVolumeEmptyDir() {
+			// If depotVolume is EmptyDir, the depot is stored in its own 'emptyDir' volume
+			vmts := []corev1.VolumeMount{
+				{Name: vapi.DepotMountName, MountPath: vdb.Spec.Local.DepotPath},
+			}
+			volMnts = append(volMnts, vmts...)
```
spilchen, line 119: Could we combine these statements?
```suggestion
			volMnts = append(volMnts, corev1.VolumeMount{
				Name: vapi.DepotMountName, MountPath: vdb.Spec.Local.DepotPath,
			})
```

pkg/vdbgen/options.go
spilchen: I don't see this change. There should be an update to `tests/e2e-leg-4/vdb-gen/run-vdb-gen.sh`

pkg/builder/builder.go
```diff
@@ -422,6 +433,21 @@ func buildHTTPServerSecretVolume(vdb *vapi.VerticaDB) corev1.Volume {
 	}
 }
 
+// buildEmtyDirVolume returns a generic 'emptyDir' volume
+func buildEmtyDirVolume(volName string) corev1.Volume {
```
spilchen, line 437: spelling: Emty => Empty
```suggestion
// buildEmptyDirVolume returns a generic 'emptyDir' volume
func buildEmptyDirVolume(volName string) corev1.Volume {
```

tests/e2e-leg-4/revivedb-3-node/vdb-to-revive/base/setup-vdb.yaml
```diff
@@ -23,7 +23,8 @@ spec:
   initPolicy: Revive
   local:
     dataPath: /my-data
-    depotPath: /my-data
```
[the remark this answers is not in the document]
roypaulin: There is a webhook rule preventing these paths to be the same. I guest I will just get rid of it. Now when depotVolume is `EmptyDir` and datapath == depotpath/catalogpath, we will update depotVolume to `PersistentVolume`

spilchen (approved):
Looks good. Thanks for applying all of the revisions.

roypaulin:
Thanks for your comments!!

PR #390, opened 2023-05-03
Start http server if httpServerMode changes in the VerticaDB

roypaulin:
The operator now starts the http server in the vertica pods. When a CR is created with httpServerMode Disabled, the http server will not be running. If the CR is changed so that httpServerMode is Enabled, the operator will start the http service. The only allowed transitions are from Disabled to Enabled/Auto and Auto to Enabled. Any other will be rejected.

A new e2e test has been added and tests a transition from `Disabled` to `Enabled`.


pkg/controllers/vdb/httpserverctrl_reconciler.go
```diff
@@ -0,0 +1,106 @@
...
+		return ctrl.Result{}, err
+	}
+
+	for _, pod := range h.PFacts.Detail {
+		if pod.isHTTPServerRunning {
+			continue
+		}
+		// Only start the http server for pods that have been added to a database
+		if !pod.upNode {
+			h.VRec.Log.Info("Not all pods are up.  Need to requeue httpserverctrl reconciler.", "pod", pod.name)
```
spilchen, line 66: We aren't doing a requeue. Should we just log the pod?
```suggestion
			h.VRec.Log.Info("Skipping http server start on pod because vertica is not running", "pod", pod.name)
```

pkg/controllers/vdb/httpserverctrl_reconciler.go
```diff
@@ -0,0 +1,106 @@
...
+		// We need this config file for the http server.
+		if !pod.fileExists[paths.HTTPTLSConfFile] {
+			h.VRec.Log.Info("Skipping http server start because https config is still missing in pod", "pod", pod.name)
+			continue
+		}
+		if err := h.startHTTPServerInPod(ctx, pod); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		// Invalidate the pod facts cache since its out of date due to the agent starting
```
spilchen, line 78:
```suggestion
		// Invalidate the pod facts cache because we started the http server
```

pkg/controllers/vdb/httpserverctrl_reconciler.go
```diff
@@ -0,0 +1,106 @@
...
+		"-tAc", genHTTPServerCtrlQuery("start"),
+	}
+	h.VRec.Event(h.Vdb, corev1.EventTypeNormal, events.HTTPServerStartStarted,
+		fmt.Sprintf("Calling select http_server_ctrl('start', '') in pod %s", pod.name))
+	if _, _, err := h.PRunner.ExecVSQL(ctx, pod.name, names.ServerContainer, cmd...); err != nil {
+		h.VRec.Event(h.Vdb, corev1.EventTypeWarning, events.HTTPServerStartFailed,
+			fmt.Sprintf("Failed to start the http server in pod %s", pod.name))
+		return err
+	}
+	h.VRec.Event(h.Vdb, corev1.EventTypeNormal, events.HTTPServerStartSucceeded,
```
spilchen, line 97: The start of the server is pretty quick. So, I don't think we need this event. Lets just log starting the http server and the error case.

pkg/controllers/vdb/httpserverctrl_reconciler.go
```diff
@@ -0,0 +1,106 @@
...
+	return ctrl.Result{}, nil
+}
+
+// startHTTPServerInPod will start the http server in the given pod
+func (h *HTTPServerCtrlReconciler) startHTTPServerInPod(ctx context.Context, pod *PodFact) error {
+	cmd := []string{
+		"-tAc", genHTTPServerCtrlQuery("start"),
+	}
+	h.VRec.Event(h.Vdb, corev1.EventTypeNormal, events.HTTPServerStartStarted,
+		fmt.Sprintf("Calling select http_server_ctrl('start', '') in pod %s", pod.name))
```
spilchen, line 91: I suggest we leave out the meta-function name. I'm not sure it will even be documented. So, it may be confusing for some.
```suggestion
		fmt.Sprintf("Starting http server in pod %s", pod.name))
```

pkg/controllers/vdb/http_server.go
```diff
@@ -0,0 +1,49 @@
...
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/events"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// hasCompatibleVersionForHTTPServer checks, in case http server is enabled, if
+// the server has a required version for http server.
+func hasCompatibleVersionForHTTPServer(vrec *VerticaDBReconciler, vdb *vapi.VerticaDB, logEvent bool, action string) bool {
+	eventMsg := "Skipping %s because the Vertica version doesn't have " +
+		"support for it. A Vertica version of '%s' or newer is needed"
```
spilchen, line 30: Can we move this down near where we log the event? It seems out of place here.

spilchen (commented):
Looks good. I just had minor comments.

spilchen (approved):

PR #392, opened 2023-05-09
Additional ports added to headless service

spilchen:
This adds new ports to the headless service. This was done for compatibility with istio. When istio is enabled with TPROXY redirection and strict mTLS, the ports are needed to ensure vertica pods can continue to communicate with each other. See this for more info: https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services

The headless service previously only had the ssh port. That was renamed from ssh to tcp-ssh to follow the istio port naming guidelines (see https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/)


roypaulin (approved):
Looks good!
Why do we need istio?

spilchen:
> Looks good! Why do we need istio?

A team within OpenText is using it for their vertica k8s deployment.

PR #396, opened 2023-05-11
Avoid mounting dbadmin password if not needed

spilchen:
Currently, the dbadmin password is mounted in the container at /etc/podinfo/superuser-passwd. This is needed so that readiness/startup probes can log into vertica and run the 'select 1' canary query. This change is to satisfy a request to avoid mounting the password if the probes are change to not refer to the password.

There are two ways you can change the probes:

1. Check if the client port is open.

Instead of checking the canary query. Just check if the client port is opened. Include overrides for the readiness and startup probes in the vdb like this:

```
kind: VerticaDB
metadata:
  name: v
spec:
  ...
  readinessProbeOverride:
    tcpSocket:
      port: 5433
  startupProbeOverride:
    tcpSocket:
      port: 5433
```

2. Setup local trust for the dbadmin

Keep using the canary query but setup local trust for dbadmin so a password isn't needed. After the database is up, setup a local trust using SQL like the following:

```
CREATE AUTHENTICATION v_dbadmin_trust METHOD 'trust' LOCAL;
GRANT AUTHENTICATION v_dbadmin_trust TO dbadmin;
ALTER AUTHENTICATION v_dbadmin_trust PRIORITY 10000;
SELECT sync_catalog();
```

This can be included in an EventTrigger so that it's done automatically after a DB was created. Then create the vdb with an override for the command.

```
kind: VerticaDB
metadata:
  name: v
spec:
  readinessProbeOverride:
    exec:
      command: ["vsql", "-c", "select 1"]
  startupProbeOverride:
    exec:
      command: ["vsql", "-c", "select 1"]
```

In both cases, the operator was changed to not include the mount if we no longer reference dbadmin password in any of the probes.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -361,9 +361,43 @@ func buildOperatorConfigMapProjection(deployNames *DeploymentNames) *corev1.Conf
 	}
 }
 
+// probeContainsSuperuserPassword will check if the probe uses the superuser
+// password.
+func probeContainsSuperuserPassword(probe *corev1.Probe) bool {
+	if probe.Exec == nil {
+		return false
+	}
+	for _, v := range probe.Exec.Command {
+		if strings.Contains(v, SuperuserPasswordPath) {
+			return true
+		}
+	}
+	return false
+}
+
+// requiresSuperuserPasswordSecretMount returns true if the superuser password
+// needs to be mounted in the pod.
+func requiresSuperuserPasswordSecretMount(vdb *vapi.VerticaDB) bool {
+	if vdb.Spec.SuperuserPasswordSecret == "" {
+		return false
+	}
+
+	// Construct each probe. If don't use the superuser password in them, then
+	// it is safe to not mount this in the downward API projection.
+	funcs := []func(*vapi.VerticaDB) *corev1.Probe{
+		makeReadinessProbe, makeStartupProbe, makeLivenessProbe,
+	}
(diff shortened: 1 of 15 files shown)


pkg/builder/builder.go
```diff
@@ -590,8 +624,33 @@ func overrideProbe(probe, ov *corev1.Probe) {
 		return
 	}
 	// Merge in parts of the override into the default probe
+	//
+	// You can only set one handler (exec, tcpSocket, httpGet or grpc). If the
+	// override has any one of those set, we always clear the other ones. A
+	// webhook exists that prevents setting more than one in the CR.
 	if ov.Exec != nil {
 		probe.Exec = ov.Exec
+		probe.TCPSocket = nil
```
roypaulin, line 633: I get why you added this but why do you need to if we know that the default handler is `exec` and there is a webhook rule preventing more than one to be set in the CR. When a new probe is created aren't all handlers `nil` before setting?

roypaulin (commented):
Looks good! Just have one question.

pkg/builder/builder.go
spilchen: The main reason I did this was for the case where the default probe uses the `exec` handler, but we overwrite it in the CR with a `tcpSocket` handler. I ended up generalizing this for all handler as I didn't want to make assumptions about the default probe.

roypaulin (approved):

PR #398, opened 2023-05-12
Reuse node names when installing new k8s pods

spilchen:
This corrects node name reuse when the operator installs new hosts in the admintools.conf. The old reuse logic was a bit too simple and could end up skip reusing a name under certain circumstances. 

Note, there are two kinds of node names in vertica. The node name before a node is added to the cluster -- in the format of node####. And the vnode name, which is the name after its added to the cluster -- in the format of v_<dbname>_node####. This impacts only the first format.

### DIFF
--- a/pkg/atconf/file_writer.go
+++ b/pkg/atconf/file_writer.go
@@ -230,14 +230,15 @@ func (f *FileWriter) removeFromClusterHosts(ips []string) error {
 // addNodes will add the given set of installIPs as new nodes in the Nodes
 // section.  The updates are done in-place in the ConfigParser.
 func (f *FileWriter) addNodes(oldHosts map[string]bool, installIPs []string) error {
-	nextNodeNumber := f.getNextNodeNumber()
+	nodes := f.buildNodeSlice()
+	var nextNodeNumber int
 	for _, ip := range installIPs {
 		// If host already exists, we treat as a no-op and skip the host
 		if _, ok := oldHosts[ip]; ok {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


pkg/atconf/file_writer.go
```diff
@@ -293,13 +293,28 @@ func (f *FileWriter) getNextNodeNumber() int {
...
 	}
-	return nextNodeNumber
+	return nodes
+}
+
+// getNextNodeNumber returns the number to use for the next vertica node.  It
+// determines this by finding a free spot in the nodes slice.
+func (f *FileWriter) getNextNodeNumber(nodes []bool) (nextNodeNumber int, newNodes []bool) {
+	for i := 0; i < len(nodes); i++ {
+		if !nodes[i] {
```
roypaulin, line 310: I don't understand what's going on here. Same for `buildNodeSlice()`. I get what you are trying to do, but not the implementation.

spilchen: We are picking the next available node number based on nodes[]. We update this slice as we go, which may involve extending it, so we always return a new copy of nodes[].

Any suggestions on how I can make it easier to understand?

roypaulin: I think some "why" comments in `buildNodeSlice()` and `getNextNodeNumber()` would help. Especially where you extend the slice. And maybe adding, in `buildNodeSlice() description`, what we are going to do with the slice.

roypaulin (approved):
Looks good!

PR #404, opened 2023-05-23
Annotation to suspend reconcile on a CR

spilchen:
This adds the ability to suspend all reconcile activity against a custom resource. This aids in debugging because the operator uses an infinity retry model (i.e. keep retrying the reconcile and use an exponential back-off for repeated failures). This works great unless you want to dig into a specific behaviour that you are seeing.

If any of the CRs (VerticaDB, VerticaAutoscaler, EventTrigger) have the `vertica.com/pause` annotation set to `true`, then no further reconcile will get scheduled for the CR.

This also swaps the create DB parameter --force-cleanup-on-failure with --force-removal-at-creation. It has the same affect, but if you have the operator paused you can see the left over log files from a failed create.

### DIFF
--- a/tests/e2e-leg-3/labels-annotations/50-assert.yaml
+++ b/tests/e2e-leg-3/labels-annotations/50-assert.yaml
@@ -1,0 +1,21 @@
+# (c) Copyright [2021-2023] Open Text.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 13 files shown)


tests/e2e-leg-3/labels-annotations/50-assert.yaml
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: Pod
+metadata:
+  name: vdb-label-ant-cluster1-0
```
roypaulin, line 17: I thought the `pause`  would prevent the operator to start the pod.

spilchen: The controller for the statefulset isn't paused. So, the pod will still come up. We will never restart vertica in the pod though because the operator is paused.

roypaulin: Got it.

roypaulin (approved):

PR #407, opened 2023-05-25
Auto delete PVC on VerticaDB deletion

spilchen:
When the VerticaDB is deleted, we will also delete all of the PVCs too. We get this behaviour automatically by filling in the ownerReferences in the PVC template.

Only new StatefulSet's created by the operator will get this behaviour. If upgrading the operator to this version, and the StatefulSet isn't recreated, then auto delete won't happen. We did this because you cannot change the PVC template in the StatefulSet after it is created.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -778,6 +778,7 @@ func getStorageClassName(vdb *vapi.VerticaDB) *string {
 
 // BuildStsSpec builds manifest for a subclusters statefulset
 func BuildStsSpec(nm types.NamespacedName, vdb *vapi.VerticaDB, sc *vapi.Subcluster, deployNames *DeploymentNames) *appsv1.StatefulSet {
+	isControllerRef := true
 	return &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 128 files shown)


pkg/builder/builder.go
```diff
@@ -804,6 +805,16 @@ func BuildStsSpec(nm types.NamespacedName, vdb *vapi.VerticaDB, sc *vapi.Subclus
...
 					ObjectMeta: metav1.ObjectMeta{
 						Name: vapi.LocalDataPVC,
+						// Set the ownerReference so that we get auto-deletion
+						OwnerReferences: []metav1.OwnerReference{
+							{
+								APIVersion: vapi.GroupVersion.String(),
+								Kind:       vapi.VerticaDBKind,
+								Name:       vdb.Name,
+								UID:        vdb.UID,
+								Controller: &isControllerRef,
```
roypaulin, line 815: What do they mean by `Controller`? `verticadb controller`?

spilchen: Yeah, the operator is a controller so we set that to true. We do the same thing for other objects created by the operator.

roypaulin (approved):

PR #413, opened 2023-06-06
Base infrastructure for vclusterOps

spilchen:
vclusterOps is the integration with a Go library that will handle admin tasks via HTTP REST APIs rather than admintools and ssh. This is the first code drop to begin developing this integration.

- fork the vertica-k8s container. We are going to be making some wholesale changes to the container. We don't want to touch the one we use for admintools style management. It's a copy of the current vertica-k8s, save for a few minor changes, but will have some bigger changes coming up soon. This will be built with the `make docker-build-vertica-v2` target.
- add a feature flag so that we can hide stuff in the operator. The feature flag will be the 'vertica.com/vcluster-ops' annotation. If this is set in the VerticaDB, the operator will start to take a new path
- start the node management agent (nma) in the container. This requires setting of some environment variables in the container so that the nma knows where to find the certs to use


roypaulin:
Looks good! Is vclusterOps a standalone library that comes with vertica?

roypaulin (approved):

spilchen:
Yes, vclusterOps is the new go library that we are developing that will handle admin tasks via node management agent and https service.

PR #414, opened 2023-06-08
SSH secret mounting improvements

spilchen:
When using your own ssh keys through a ssh secret, we now mount the secret for the root user too. This comes in handy if you are running a hybrid k8s server and need to run update_vertica from within the pod. Prior to this change you couldn't do that because the ssh key for the root user was unrecognized.

### DIFF
--- a/pkg/builder/builder_test.go
+++ b/pkg/builder/builder_test.go
@@ -21,6 +21,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/paths"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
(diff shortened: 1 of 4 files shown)


pkg/builder/builder_test.go
```diff
@@ -230,8 +231,29 @@ var _ = Describe("builder", func() {
...
 	})
+
+	It("should mount ssh secret for dbadmin and root", func() {
+		vdb := vapi.MakeVDB()
+		vdb.Spec.SSHSecret = "my-secret"
+		c := buildPodSpec(vdb, &vdb.Spec.Subclusters[0], &DeploymentNames{})
+		cnt := &c.Containers[0]
+		i, ok := getFirstSSHSecretVolumeMountIndex(cnt)
+		Expect(ok).Should(BeTrue())
+		Expect(cnt.VolumeMounts[i].MountPath).Should(Equal(paths.DBAdminSSHPath))
```
roypaulin, line 242: I think this will always fail

pkg/paths/paths.go
```diff
@@ -57,15 +57,16 @@ const (
...
+	RootSSHPath               = "/root/.ssh"
 	HTTPServerCACrtName       = "ca.crt"
 )
 
 // MountPaths lists all of the paths for internally generated mounts.
 var MountPaths = []string{LocalDataPath, CELicensePath, MountedLicensePath,
 	HadoopConfPath, ConfigPath, ConfigSharePath, ConfigLogrotatePath,
 	LogPath, PodInfoPath, AdminToolsConf, AuthParmsFile, EulaAcceptanceFile,
-	EulaAcceptanceScript, CertsRoot, Krb5Conf, Krb5Keytab, SSHPath}
+	EulaAcceptanceScript, CertsRoot, Krb5Conf, Krb5Keytab, DBAdminSSHPath}
```
roypaulin, line 69: Shouldn't you also add `RootSSHPath`?

pkg/builder/builder_test.go
spilchen: I made my last change too fast.

roypaulin: Sorry if the comment was confusing but I saw that you fixed the tests.

roypaulin (approved):

PR #417, opened 2023-06-15
Operator installer changes for vclusterOps

roypaulin:
Get past the install phase when running the vclusterOps feature flag. 
-  change the operator so that instead of mounting /opt/vertica/config from the PV, we mount /opt/vertica/config/https_certs.
- continue to generate httpstls.json file and skip all other aspects of the installer.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -103,11 +103,27 @@ func BuildHlSvc(nm types.NamespacedName, vdb *vapi.VerticaDB) *corev1.Service {
 	}
 }
 
+// buildConfigVolumeMount returns the volume mount for config.
+// If vclusterops flag is enabled we mount only /opt/vertica/config/https_certs
+func buildConfigVolumeMount(vdb *vapi.VerticaDB) corev1.VolumeMount {
+	subPath := "config"
+	mountPath := paths.ConfigPath
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		subPath = "config/https_certs"
+		mountPath = paths.HTTPTLSConfDir
+	}
+	return corev1.VolumeMount{
+		Name:      vapi.LocalDataPVC,
+		SubPath:   vdb.GetPVSubPath(subPath),
+		MountPath: mountPath,
+	}
+}
+
 // buildVolumeMounts returns the volume mounts to include in the sts pod spec
 func buildVolumeMounts(vdb *vapi.VerticaDB) []corev1.VolumeMount {
 	volMnts := []corev1.VolumeMount{
 		{Name: vapi.LocalDataPVC, MountPath: paths.LocalDataPath},
-		{Name: vapi.LocalDataPVC, SubPath: vdb.GetPVSubPath("config"), MountPath: paths.ConfigPath},
+		buildConfigVolumeMount(vdb),
 		{Name: vapi.LocalDataPVC, SubPath: vdb.GetPVSubPath("log"), MountPath: paths.LogPath},
 		{Name: vapi.LocalDataPVC, SubPath: vdb.GetPVSubPath("data"), MountPath: vdb.Spec.Local.DataPath},
 		{Name: vapi.PodInfoMountName, MountPath: paths.PodInfoPath},
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -106,16 +106,20 @@ func BuildHlSvc(nm types.NamespacedName, vdb *vapi.VerticaDB) *corev1.Service {
 // buildConfigVolumeMount returns the volume mount for config.
 // If vclusterops flag is enabled we mount only /opt/vertica/config/https_certs
 func buildConfigVolumeMount(vdb *vapi.VerticaDB) corev1.VolumeMount {
-	subPath := "config"
-	mountPath := paths.ConfigPath
+	// When using vclusterOps, we don't have any state in /opt/vertica/config to preserve. There is no
+	// equivalent file like admintools.conf. The only state we need to preserve is the the contents of
+	// the https_certs directory.
 	if vmeta.UseVClusterOps(vdb.Annotations) {
-		subPath = "config/https_certs"
-		mountPath = paths.HTTPTLSConfDir
+		return corev1.VolumeMount{
+			Name:      vapi.LocalDataPVC,
+			SubPath:   vdb.GetPVSubPath("config/https_certs"),
+			MountPath: paths.HTTPTLSConfDir,
+		}
 	}
 	return corev1.VolumeMount{
 		Name:      vapi.LocalDataPVC,
-		SubPath:   vdb.GetPVSubPath(subPath),
-		MountPath: mountPath,
+		SubPath:   vdb.GetPVSubPath("config"),
+		MountPath: paths.ConfigPath,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


pkg/controllers/vdb/podfacts.go
```diff
@@ -1027,6 +1027,17 @@ func (p *PodFacts) anyInstalledPodsNotRunning() (bool, types.NamespacedName) {
 	return false, types.NamespacedName{}
 }
 
+// anyIPodsNotRunning returns true if any pod isn't running.  It will
```
spilchen, line 1030: typo
```suggestion
// anyPodsNotRunning returns true if any pod isn't running.  It will
```

pkg/builder/builder.go
```diff
@@ -103,11 +103,27 @@ func BuildHlSvc(nm types.NamespacedName, vdb *vapi.VerticaDB) *corev1.Service {
...
+	mountPath := paths.ConfigPath
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		subPath = "config/https_certs"
+		mountPath = paths.HTTPTLSConfDir
+	}
+	return corev1.VolumeMount{
+		Name:      vapi.LocalDataPVC,
+		SubPath:   vdb.GetPVSubPath(subPath),
+		MountPath: mountPath,
+	}
```
spilchen, line 119: I think it is clearer if you don't overwrite the variables. How about something like this?
```suggestion
	if vmeta.UseVClusterOps(vdb.Annotations) {
		return corev1.VolumeMount{
			Name:      vapi.LocalDataPVC,
			SubPath:   vdb.GetPVSubPath("config/https_certs"),
			MountPath: paths.HTTPTLSConfDir,
		}
	}
	return corev1.VolumeMount{
		Name:      vapi.LocalDataPVC,
		SubPath:   vdb.GetPVSubPath("config"),
		MountPath: paths.ConfigPath,
	}}
```

pkg/builder/builder.go
```diff
@@ -103,11 +103,27 @@ func BuildHlSvc(nm types.NamespacedName, vdb *vapi.VerticaDB) *corev1.Service {
 	}
 }
 
+// buildConfigVolumeMount returns the volume mount for config.
+// If vclusterops flag is enabled we mount only /opt/vertica/config/https_certs
+func buildConfigVolumeMount(vdb *vapi.VerticaDB) corev1.VolumeMount {
+	subPath := "config"
```
spilchen, line 109: I didn't think I provided enough rationale for this change. Can you include a comment? Something like this:
```
// When using vclusterOps, we don't have any state in /opt/vertica/config to preserve. There is no
// equivalent file like admintools.conf. The only state we need to preserve is the the contents of
// the https_certs directory.
```

pkg/controllers/vdb/http_server.go
```diff
@@ -20,12 +20,16 @@ import (
...
 	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
 	"github.com/vertica/vertica-kubernetes/pkg/events"
+	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
 	corev1 "k8s.io/api/core/v1"
 )
 
 // hasCompatibleVersionForHTTPServer checks, in case http server is enabled, if
 // the server has a required version for http server.
 func hasCompatibleVersionForHTTPServer(vrec *VerticaDBReconciler, vdb *vapi.VerticaDB, logEvent bool, action string) bool {
+	if vmeta.UseVClusterOps(vdb.Annotations) {
```
spilchen, line 30: I assume this is needed because we need the http service to communicate with the library. Did the vertica version check not work?

pkg/controllers/vdb/install_reconciler.go
```diff
@@ -69,6 +70,16 @@ func (d *InstallReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (c
...
 		return ctrl.Result{}, err
 	}
+	// We generate https conf file and skip the install phase when running
+	// the vclusterOps feature flag
+	if vmeta.UseVClusterOps(d.Vdb.Annotations) {
+		if ok, podNotRunning := d.PFacts.anyPodsNotRunning(); ok {
+			d.Log.Info("At least one pod isn't running.  Aborting the install.", "pod", podNotRunning)
+			return ctrl.Result{Requeue: true}, nil
+		}
+		return ctrl.Result{}, d.generateHTTPCerts(ctx)
```
spilchen, line 80: To keep the same level of abstraction in this function, can we move these contents to its own function? It might make sense to create the new function as `installForVClusterOps` and rename `analyzeFacts` to `installForAdmintools`.

pkg/controllers/vdb/install_reconciler.go
```diff
@@ -69,6 +70,16 @@ func (d *InstallReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (c
 	if err := d.PFacts.Collect(ctx, d.Vdb); err != nil {
 		return ctrl.Result{}, err
 	}
+	// We generate https conf file and skip the install phase when running
+	// the vclusterOps feature flag
+	if vmeta.UseVClusterOps(d.Vdb.Annotations) {
+		if ok, podNotRunning := d.PFacts.anyPodsNotRunning(); ok {
```
spilchen, line 76: Do we need to wait for all pods to be running? The old installer did that because it needed to sync the admintools.conf to all pods. But that is no longer necessary. We may need to prevent going to the next actor until all pods have done the install. But that shouldn't prevent generating the certs for the pods that are running.

pkg/controllers/vdb/http_server.go
roypaulin: I might have overthought this one. I thought that given that we use the feature only for development we did not care about all the checks, we would assume http server is running. I also did not want it to depend on `httpServerMode`.We don't need this. I suppose by the time vclusterOps is made public, http server will be mandatory and `httpServerMode` may be deprecated.

spilchen: Yeah, we'll probably need to have smarter checks in the operator that if you are using the version of vertica without admintools that you also need the http service started. Lets leave this off for now.

pkg/controllers/vdb/install_reconciler.go
```diff
@@ -69,11 +70,33 @@ func (d *InstallReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (c
...
+// installForVClusterOps will go through the install phase for vclusterOps.
+// It only generates the http certs and requeue if at least one pod has not done the install
+func (d *InstallReconciler) installForVClusterOps(ctx context.Context) (ctrl.Result, error) {
+	hasUninstalledPods, err := d.generateHTTPCerts(ctx)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if hasUninstalledPods {
+		// We do not proceed to the next actor until
+		// all pods have done the install
```
spilchen, line 91: Lets log a message. I like to have a clear reason each time we requeue a reconcile iteration.

pkg/controllers/vdb/install_reconciler_test.go
```diff
@@ -32,6 +33,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+const VClusterOpsAnnotationTrue = "true"
```
spilchen, line 36: Lets move this to meta/annotations.go

spilchen (approved):

PR #419, opened 2023-06-16
Add dispatch abstraction for admintools -t list_allnodes

spilchen:
This is a continuation of the change to hide all admintools commands behind an abstraction so that we can replace them with the vclusterOps library. This adds an abstraction for the call to admintools -t list_allnodes. I called it FetchNodeState.

### DIFF
--- a/pkg/vadmin/fetch_node_state_at.go
+++ b/pkg/vadmin/fetch_node_state_at.go
@@ -26,7 +26,7 @@ import (
 )
 
 // FetchNodeState will determine if the given set of nodes are considered UP
-// or DOWN in our consensous state. It returns a map of vnode to its node state.
+// or DOWN in our consensus state. It returns a map of vnode and its node state.
 func (a Admintools) FetchNodeState(ctx context.Context, opts ...fetchnodestate.Option) (map[string]string, ctrl.Result, error) {
 	s := fetchnodestate.Parms{}
 	if err := s.Make(opts...); err != nil {
--- a/pkg/vadmin/fetch_node_state_at.go
+++ b/pkg/vadmin/fetch_node_state_at.go
@@ -1,0 +1,84 @@
+/*
+ (c) Copyright [2021-2023] Open Text.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ You may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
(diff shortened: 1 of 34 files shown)


pkg/vadmin/fetch_node_state_at.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, res, err2
+	}
+	return a.parseClusterNodeStatus(stdout, a.buildHostsNeededSet(s.Hosts)), ctrl.Result{}, nil
+}
+
+// parseClusterNodeStatus will parse the output from a AT -t list_allnodes call
+func (a Admintools) parseClusterNodeStatus(stdout string, hostsNeeded map[string]bool) map[string]string {
+	stateMap := map[string]string{}
+	lines := strings.Split(stdout, "\n")
+	const ColHeaderCount = 2
```
roypaulin, line 48: If I understand this variable represent the first 2 lines (the header). I think the name is a bit confusing. Would something like `headerRowCount` or `headerRowSize` be better?

pkg/vadmin/fetch_node_state_at.go
```diff
@@ -0,0 +1,84 @@
...
+	"strings"
+
+	"github.com/vertica/vertica-kubernetes/pkg/events"
+	"github.com/vertica/vertica-kubernetes/pkg/names"
+	"github.com/vertica/vertica-kubernetes/pkg/vadmin/opts/fetchnodestate"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// FetchNodeState will determine if the given set of nodes are considered UP
+// or DOWN in our consensous state. It returns a map of vnode to its node state.
```
roypaulin, line 29:
```suggestion
// or DOWN in our consensus state. It returns a map of vnode and its node state.
```
Do you mean consensus?

pkg/vadmin/fetch_node_state_at_test.go
```diff
@@ -0,0 +1,85 @@
...
+				" v_d_node0002 | 10.244.1.7 | DOWN  | vertica-11.0.0.20210309 | d \n" +
+				" v_d_node0003 | 10.244.1.8 | UP    | vertica-11.0.0.20210309 | d \n" +
+				"\n",
+			},
+		}
+		evWriter := mgmterrors.TestEVWriter{}
+		d := MakeAdmintools(logger, vdb, fpr, &evWriter)
+		state, res, err := d.FetchNodeState(ctx,
+			fetchnodestate.WithInitiator(atPod, "10.244.1.6"),
+			fetchnodestate.WithHost("v_d_node0001", "10.244.1.6"),
```
roypaulin, line 51: What happens when you pass the same host twice?

roypaulin (commented):
Looks good!

pkg/vadmin/fetch_node_state_at.go
spilchen: I'll change it to `headerRowCount`

roypaulin (approved):

PR #423, opened 2023-06-21
Create a test suite for vclusterOps integration

chinhtranvan:
This adds a new e2e test suite that we will use for vclusterOps integration. It doesn't do much to start, but we will add to it as we get further along with the vclusterOps integration.

You need to run with the next generation vertica-k8s image (docker-vertica-v2).

To run just this new test suite:
```
E2E_TEST_DIRS=tests/e2e-vcluster make run-int-tests
```


tests/e2e-vcluster/vcluster-ops/setup-vdb/base/setup-vdb.yaml
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: vertica.com/v1beta1
+kind: VerticaDB
+metadata:
+  name: verticadb-sample
```
spilchen, line 17: This test needs to set the feature flag that enables vcluster ops. Lets add the following annotation.
```suggestion
  name: verticadb-sample
  annotations:
    vertica.com/vcluster-ops: "true"
```

spilchen, line 17: I like to use the name of the testcase as the vdb name. It helps if you have many testcases running in parallel. Can you rename it to: `v-vcluster-ops`?

tests/e2e-vcluster/vcluster-ops/setup-vdb/base/setup-vdb.yaml
```diff
@@ -0,0 +1,41 @@
...
+spec:
+  image: kustomize-vertica-image
+  sshSecret: ssh-keys
+  communal:
+    includeUIDInPath: false
```
spilchen, line 22: Let's set this to true. Otherwise the test is more likely to hit communal storage is not empty when trying to create a db -- usually after a failed attempt.

tests/e2e-vcluster/vcluster-ops/setup-vdb/base/setup-vdb.yaml
```diff
@@ -0,0 +1,41 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: vertica.com/v1beta1
+kind: VerticaDB
+metadata:
+  name: verticadb-sample
+spec:
+  image: kustomize-vertica-image
+  sshSecret: ssh-keys
```
spilchen, line 20: Seeing as vcluster-ops won't be using ssh. Lets not use this in this test. I don't think its valuable to test that. Please remove this line.

tests/e2e-vcluster/vcluster-ops/10-deploy-operator.yaml
spilchen, line 1: Lets go one more step and actually try and create the database. It will fail, but lets make sure that pods are created.

spilchen:
I think it will be good to have two test cases. One that creates a single node database with ksafety 0. And another one that create a 3 node database with ksafety 1. The current test is 3-node/ksafety-1, can you rename it? Maybe something like `vcluster-ks-1`

chinhtranvan:
Ready for second round review!

spilchen:
You need to update setup-kustomize.sh so that it creates the overlay files for kustomize. 
```
diff --git a/scripts/setup-kustomize.sh b/scripts/setup-kustomize.sh
index e65f043..4f5f936 100755
--- a/scripts/setup-kustomize.sh
+++ b/scripts/setup-kustomize.sh
@@ -725,11 +725,11 @@ setup_creds_for_private_repo

 # Descend into each test and create the overlay kustomization.
 # The overlay is created in a directory like: overlay/<tc-name>
-for tdir in e2e-leg-*/*/*/base e2e-server-upgrade/*/*/base e2e-operator-upgrade-overlays/*/*/base e2e-udx/*/*/base e2e-http-server/*/*/base e2e-enterprise/*/*/base
+for tdir in e2e-leg-*/*/*/base e2e-server-upgrade/*/*/base e2e-operator-upgrade-overlays/*/*/base e2e-udx/*/*/base e2e-http-server/*/*/base e2e-enterprise/*/*/base e2e-vcluster/*/*/base
 do
     create_vdb_pod_kustomization $(dirname $tdir) $(basename $(realpath $tdir/../..))
 done
-for tdir in e2e-leg-*/* e2e-server-upgrade/* e2e-operator-upgrade-overlays/* e2e-udx/* e2e-http-server/*
+for tdir in e2e-leg-*/* e2e-server-upgrade/* e2e-operator-upgrade-overlays/* e2e-udx/* e2e-http-server/* e2e-vcluster/*
 do
     clean_communal_kustomization $tdir
 done
```

spilchen:
This probably worked on your system because you likely copied a testcase overway, which had the overlays. The overlays are in the .gitignore file so never made it into github.

tests/e2e-vcluster/vcluster-ks-0/setup-vdb/base/setup-vdb.yaml
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: vertica.com/v1beta1
+kind: VerticaDB
+metadata:
+  name: v-vcluster-ks-0
```
spilchen, line 17: We are missing the annotation to enable vcluster ops

tests/e2e-vcluster/vcluster-ks-1/setup-vdb/base/setup-vdb.yaml
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: vertica.com/v1beta1
+kind: VerticaDB
+metadata:
+  name: v-scale-up-and-down
```
spilchen, line 17: We need the annotation to enable the vclusterOps integration

tests/e2e-vcluster/vcluster-ops/05-create-creds.yaml
spilchen, line 1: We should remove this testcase. We should only have two.

tests/e2e-vcluster/vcluster-ks-0/30-scale-up-to-2.yaml
spilchen, line 1: There are a lot of steps in these two new testcases that we won't be able to test until we build out the rest of the vcluster integration. The goal here is to have a testcase that runs clean and tests only what we have integrated with vcluster ops. At this point we should be able to get to the create db. The create db won't succeed, which is fine. As long as our assertions aren't looking for a successful create then we can still have the testcase succeed. 

So, I think the testcase should:
- deploy the operator
- create the vdb with the annotation that enables vcluster ops integration
- ensure that pods are running and look for an event for create db 
- do the standard cleanup (usually the 9x-* files)

Tip on looking for a create db event. Have a look at 65-assert.yaml in this testcase as an example. I think something like the following will work:
```
apiVersion: v1
kind: Event
reason: CreateDBStart
source:
  component: verticadb-operator
involvedObject:
  kind: VerticaDB
  name: v-vcluster-ks-0
```

spilchen:
Lets include the GitHub CI changes in this PR too so that this it ends up running the new testsuite

chinhtranvan:
All comments are addressed! GitHub CI is added to this PR, and pass the e2e tests..

.github/workflows/build-images.yml
```diff
@@ -269,6 +266,89 @@ jobs:
...
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+    - uses: actions/checkout@v3
+      if: ${{ inputs.v2_vertica_image == '' }}
+
+    - name: Download the RPM
+      uses: ./.github/actions/download-rpm
+      if: ${{ inputs.v2_vertica_image == '' }}
+
+    - name: Build and optionally push nokeys server image
```
spilchen, line 312: minor: correct the name of the step
```suggestion
    - name: Build and optionally push v2 server image
```

.github/workflows/build-images.yml
```diff
@@ -269,6 +266,89 @@ jobs:
...
+      if: ${{ inputs.v2_vertica_image == '' }}
+      run: |
+        export VERTICA_IMG=${{ steps.v2_vertica_image.outputs.value }}
+        export NO_KEYS=yes
```
spilchen, line 316: Lets remove this env var. The NO_KEYS image only applies to ones that depend on SSH.
```suggestion
```

.github/actions/download-rpm/action.yaml
```diff
@@ -3,9 +3,16 @@ description: 'Will download the RPM as prep for container building'
...
   steps:
-    - name: Download vertica RPM package
+    - name: Download vertica RPM package to docker-vertica
       shell: bash
       env:
           VERTICA_CE_URL: "https://vertica-community-edition-for-testing.s3.amazonaws.com/XCz9cp7m/vertica-12.0.3-0.x86_64.RHEL6.rpm"
       run: |
         curl $VERTICA_CE_URL -o docker-vertica/packages/vertica-x86_64.RHEL6.latest.rpm
+
+    - name: Download vertica RPM package to docker-vertica-2
```
spilchen, line 13: This does two downloads each time it is called. Let's just have the caller specify the target directory.

.github/workflows/build-images.yml
```diff
@@ -269,6 +266,89 @@ jobs:
...
+        echo "OWNER_LC=${OWNER,,}" >>${GITHUB_ENV}
+
+    - name: Pick the name of the image
+      uses: spilchen/switch-case-action@v2
+      id: v2_vertica_image
+      with:
+        default: ghcr.io/${{ env.OWNER_LC }}/vertica-k8s:${{ github.sha }}-nokeys
+        conditionals-with-values: |
+          ${{ inputs.v2_vertica_image != '' }} => ${{ inputs.v2_vertica_image }}
+          ${{ github.event_name == 'pull_request' }} => vertica-k8s:kind-nokeys
```
spilchen, line 288: Lets not use -nokeys as the default image name. Instead, lets use -v2.
```suggestion
        default: ghcr.io/${{ env.OWNER_LC }}/vertica-k8s:${{ github.sha }}-v2
        conditionals-with-values: |
          ${{ inputs.v2_vertica_image != '' }} => ${{ inputs.v2_vertica_image }}
          ${{ github.event_name == 'pull_request' }} => vertica-k8s:kind-v2
```

.github/workflows/e2e.yml
```diff
@@ -26,6 +26,10 @@ on:
         description: 'Name of an existing vlogger image. Leave blank to build oe with the default name'
         type: string
         required: false
+      v2_vertica_image:
+        description: 'Name of a v2 Vertica server image. If blank we will build one using the default name'
+        type: string
+        required: false
```
spilchen, line 32: I suggest we move this up ahead of the vlogger image. That way we keep all of the parms related to the vertica-k8s image variants together.

.github/workflows/e2e.yml
```diff
@@ -147,6 +147,18 @@ jobs:
       DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
       DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
 
+  e2e-vcluster:
+    if: ${{ inputs.e2e_test_suites == 'all' || inputs.e2e_test_suites == 'hostpath' || inputs.e2e_test_suites == '' }}
```
spilchen, line 151: We want to pick this test suite if vcluster was chosen from the drop down
```suggestion
    if: ${{ inputs.e2e_test_suites == 'all' || inputs.e2e_test_suites == 'vcluster' || inputs.e2e_test_suites == '' }}
```

.github/workflows/e2e.yml
```diff
@@ -147,6 +147,18 @@ jobs:
...
+    needs: [build] 
+    uses: ./.github/workflows/e2e-vcluster.yml
+    with:
+      vlogger-image: ${{ needs.build.outputs.vlogger-image }}
+      operator-image: ${{ needs.build.outputs.operator-image }}
+      vertica-image: ${{ needs.build.outputs.full-vertica-image }}
```
spilchen, line 157: We need to pass in the v2 image here.
```suggestion
      vertica-image: ${{ needs.build.outputs.v2-vertica-image }}
```

tests/e2e-vcluster/vcluster-ops/10-assert.yaml
spilchen, line 1: Lets remove the vcluster-ops testcase. We only need two testcases. One for k-safety 0 and one for k-safety 1.

tests/e2e-vcluster/vcluster-ks-0/65-assert.yaml
spilchen, line 1: I suggest we rename step 65 to be step 15. We are going to add more steps after, so that gives us some room to add them.

chinhtranvan:
All comments are addressed! Could you help me take another look?

.github/actions/download-rpm/action.yaml
```diff
@@ -9,3 +9,4 @@ runs:
           VERTICA_CE_URL: "https://vertica-community-edition-for-testing.s3.amazonaws.com/XCz9cp7m/vertica-12.0.3-0.x86_64.RHEL6.rpm"
       run: |
         curl $VERTICA_CE_URL -o docker-vertica/packages/vertica-x86_64.RHEL6.latest.rpm
+        curl $VERTICA_CE_URL -o docker-vertica-v2/packages/vertica-x86_64.RHEL6.latest.rpm
```
spilchen, line 12: I was hoping we could have a parameter that picks the location. But an alternative is to just do a copy. Then we save having to download the file twice.
```suggestion
        cp docker-vertica/packages/*rpm docker-vertica-v2/packages
```

.github/workflows/build-images.yml
```diff
@@ -53,6 +52,9 @@ on:
       minimal-vertica-image:
         description: "The image name of the vertica server, but with optional software removed"
         value: ${{ jobs.build-server-minimal.outputs.image }}
+      v2_vertica_image:
```
spilchen, line 55: I think this needs to be dashes not underscores. It has to match the variable name that you reference in e2e.yml
```suggestion
      v2-vertica-image:
```

chinhtranvan:
All e2e tests are passed!

spilchen (approved):
Looks great. Thanks for applying all of the revisions.

PR #424, opened 2023-06-22
Dispatch abstractions for 5 more APIs

spilchen:
More work to move admintools commands behind the dispatch abstraction. This is work needed for our vclusterOps integration. This PR adds the following interfaces:
- stop DB
- db add node.
- db remove node
- remove subcluster
- add subcluster

### DIFF
--- a/pkg/controllers/vdb/dbremovenode_reconciler.go
+++ b/pkg/controllers/vdb/dbremovenode_reconciler.go
@@ -114,14 +114,14 @@ func (d *DBRemoveNodeReconciler) removeNodesInSubcluster(ctx context.Context, sc
 	startPodIndex, endPodIndex int32) (ctrl.Result, error) {
 	podsToRemove, requeueNeeded := d.findPodsSuitableForScaleDown(sc, startPodIndex, endPodIndex)
 	if len(podsToRemove) > 0 {
-		atPod, ok := d.PFacts.findPodToRunAdmintoolsAny()
+		initiatorPod, ok := d.PFacts.findPodToRunAdmintoolsAny()
 		if !ok {
 			// Requeue since we couldn't find a running pod
 			d.Log.Info("Requeue since we could not find a pod to run admintools")
 			return ctrl.Result{Requeue: true}, nil
 		}
 
-		if err := d.execAdminCmd(ctx, atPod, podsToRemove); err != nil {
+		if err := d.runRemoveNode(ctx, initiatorPod, podsToRemove); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to call remove node: %w", err)
 		}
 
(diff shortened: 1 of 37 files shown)


pkg/vadmin/add_sc_at.go
```diff
@@ -0,0 +1,52 @@
...
+		"--database", a.VDB.Spec.DBName,
+		"--subcluster", s.Subcluster,
+	}
+
+	// In v11, when adding a subcluster it defaults to a secondary.  Prior
+	// versions default to a primary.  Use the correct switch, depending on what
+	// version we are using.
+	vinf, ok := a.VDB.MakeVersionInfo()
+	const DefaultSecondarySubclusterCreationVersion = "v11.0.0"
+	if ok && vinf.IsEqualOrNewer(DefaultSecondarySubclusterCreationVersion) {
```
roypaulin, line 40: Can't we just check `s.IsPrimary` and append the correct flag?
```
if s.IsPrimary {
     cmd = append(cmd, "--is-primary")
} else {
    cmd = append(cmd, "--is-secondary")
}
```

pkg/controllers/vdb/dbremovenode_reconciler.go
```diff
@@ -135,13 +136,20 @@ func (d *DBRemoveNodeReconciler) removeNodesInSubcluster(ctx context.Context, sc
 	return ctrl.Result{Requeue: requeueNeeded}, nil
 }
 
-// execATCmd will run the admintools command to remove the node
+// execAdminCmd will run the admin command to remove the node
 // This handles recording of the events.
-func (d *DBRemoveNodeReconciler) execATCmd(ctx context.Context, atPod types.NamespacedName, podNames string, cmd []string) error {
+func (d *DBRemoveNodeReconciler) execAdminCmd(ctx context.Context, atPod *PodFact, pods []*PodFact) error {
```
roypaulin, line 141: Why not just `runRemoveNode` or `removeNode`?

roypaulin (commented):
Looks good! Just have a few comments.

pkg/vadmin/add_node_at.go
```diff
@@ -0,0 +1,77 @@
...
+				Msg: stdout,
+			}
+		default:
+			a.EVWriter.Eventf(a.VDB, corev1.EventTypeWarning, events.AddNodeFailed,
+				"Failed when calling 'admintools -t db_add_node' for pod(s) '%s'", strings.Join(s.Hosts, ","))
+		}
+	}
+
+	if a.DevMode {
+		a.debugDumpAdmintoolsConf(ctx, s.InitiatorName)
```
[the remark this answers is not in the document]
spilchen: It gives you a nice before/after snapshot so you can see exactly what changed in the add node call.

pkg/vadmin/add_sc_at.go
spilchen: Let me clean it up. We don't need the version check anymore because the minimum version that we support for the operator is v11.0.1.

pkg/controllers/vdb/dbremovenode_reconciler.go
spilchen: I'll use runRemoveNode

roypaulin (approved):

PR #425, opened 2023-06-22
Automatically start the NMA in v2 vertica-k8s image

roypaulin:
In the v2 vertica-k8s image, the NMA wasn't automatically starting up. The issue was that it wasn't finding the certs because the environment variables it depended on were no loaded. This PR fixes that so it can now find the certs through the environment variables and start.


spilchen (approved):
Nice find. Thanks for fixing. fyi @cchen-vertica

cchen-vertica:
Thank you, Roy.

PR #428, opened 2023-06-26
Move restart_node and start_db behind dispatcher abstraction

spilchen:
This is a continuation of building out the dispatcher interface. The dispatcher interface will be used to switch between admintools and vcluster library. This adds the final two calls: RestartNode and StartDB.

### DIFF
--- a/pkg/vadmin/start_db_vc.go
+++ b/pkg/vadmin/start_db_vc.go
@@ -17,15 +17,16 @@ package vadmin
 
 import (
 	"context"
-	"fmt"
 
-	"github.com/vertica/vertica-kubernetes/pkg/vadmin/opts/removenode"
+	"github.com/vertica/vertica-kubernetes/pkg/vadmin/opts/startdb"
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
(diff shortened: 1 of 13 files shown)


pkg/vadmin/start_db_vc.go
```diff
@@ -0,0 +1,32 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// StartDB will start a subset of nodes. Use this when vertica has lost
+// cluster quorum. The IP given for each vnode *must* match the current IP
+// in the vertica catalog. If they aren't a call to ReIP is necessary.
+func (v VClusterOps) StartDB(ctx context.Context, opts ...startdb.Option) (ctrl.Result, error) {
+	s := startdb.Parms{}
+	s.Make(opts...)
+	return ctrl.Result{}, nil
```
roypaulin, line 31: Just curious, why didn't you return `fmt.Errorf("not implemented")` ?

spilchen: I forgot to. Let me add that in.

roypaulin (approved):

PR #432, opened 2023-06-27
Build images with docker buildx build

roypaulin:
As of Docker Engine 23.0, Buildx is the default build client. So we are now using it to build our images instead of `docker build`.
All `docker build` calls have been replaced by `docker buildx build`.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -21,7 +21,7 @@ This guide explains how to set up an environment to develop and test the Vertica
 # Software Setup
 Prior to developing, the following software needs to be installed manually.  There is other software that needed, but it is downloaded through make targets in the repo's bin directory.
 
-- [docker](https://docs.docker.com/get-docker/) (version 19.03 or newer)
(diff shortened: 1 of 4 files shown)


DEVELOPER.md
```diff
@@ -21,7 +21,7 @@ This guide explains how to set up an environment to develop and test the Vertica
 # Software Setup
 Prior to developing, the following software needs to be installed manually.  There is other software that needed, but it is downloaded through make targets in the repo's bin directory.
 
-- [docker](https://docs.docker.com/get-docker/)
+- [docker](https://docs.docker.com/get-docker/) (version 19.03 or newer)
```
spilchen, line 24: Based on the description of the PR, should this be version 23.0?

spilchen (commented):
Thanks for fixing this.

DEVELOPER.md
roypaulin: Yes It should. I put 19.03 because it is the 1st version supporting buildx but it is very old. Let's put 23.0.

spilchen (approved):

PR #434, opened 2023-06-28
Remove ssh packages and setup from vertica-k8s v2 image for vclusterOps 

roypaulin:
With VClusterOps, we will not be relying on ssh and admintools. This removes ssh and its dependencies:
- removes ssh packages and ssh setup from Dockerfile
- removes static ssh keys


spilchen (commented):
We should remove docker-vertica-v2/s6-rc.d/user/contents.d/sshd too. Other than that, it looks good.

spilchen (approved):

PR #435, opened 2023-06-29
implemented VClusterOps.StopDB()

cchen-vertica:
This is the integration work that calls vclusterops library to stop a database.

The e2e test will cover create_db case as well. In my code, I am using a test version of vcluster library. I will change the go.mod and go.sum after my PR of vcluster library is approved and merged.

### DIFF
--- a/pkg/vadmin/stop_db_vc.go
+++ b/pkg/vadmin/stop_db_vc.go
@@ -19,6 +19,7 @@ import (
 	"context"
 
 	vops "github.com/vertica/vcluster/vclusterops"
+	"github.com/vertica/vcluster/vclusterops/vstruct"
 	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
 	"github.com/vertica/vertica-kubernetes/pkg/net"
 	"github.com/vertica/vertica-kubernetes/pkg/vadmin/opts/stopdb"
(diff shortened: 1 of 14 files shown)


spilchen (commented):
I'm good with these changes. I'll approve once the upstream changes are made and applied here.

spilchen:
BTW, your PR is failing in the vcluster test suite. It fails to create the database. This is because the CI uses an old version of vertica 12.0.3. So, the NMA isn't being started and the HTTPS endpoints are old. @chinhtranvan is working on ensuring all PRs run with an image built from the latest master in the server repo. So, we are blocked until that is resolved.

cchen-vertica:
> BTW, your PR is failing in the vcluster test suite. It fails to create the database. This is because the CI uses an old version of vertica 12.0.3. So, the NMA isn't being started and the HTTPS endpoints are old. @chinhtranvan is working on ensuring all PRs run with an image built from the latest master in the server repo. So, we are blocked until that is resolved.

Thanks for the information. I have added Chinn's PR to the dependent list of stop_db PR(VER-87510).

cchen-vertica:
I've updated the PR to use the latest vclusterops library. It's ready for second-round review.

pkg/vadmin/stop_db_vc.go
```diff
@@ -17,15 +17,54 @@ package vadmin
...
+	return nil
+}
+
+func (v *VClusterOps) genStopDBOptions(s *stopdb.Parms) vops.VStopDatabaseOptions {
+	opts := vops.VStopDatabaseOptionsFactory()
+
+	opts.RawHosts = append(opts.RawHosts, s.InitiatorIP)
+	v.Log.Info("Setup stop db options", "hosts", opts.RawHosts[0])
+	if net.IsIPv6(s.InitiatorIP) {
+		opts.Ipv6 = vstruct.True
```
spilchen, line 54: I wonder if we should have a `vstruct.MakeNullableBool` function to use for this pattern? Something to convert a bool to a NullableBool in a single line. I see this pattern repeated in a few spots.

spilchen (commented):
The new workflow is that you must do vclusterOps integration from a branch off the vertica-kubernetes repo. Not your fork. The fork doesn't have access to the dockerhub credentials. So it cannot run the e2e-vcluster test suite.

pkg/vadmin/stop_db_vc.go
spilchen: Maybe don't need to include this in the PR since it's a change in a different library. Can we add this function to the vcluster library so that we can refactor later?

cchen-vertica: Sure, I will add the function and do the refactoring in VER-86455.

PR #436, opened 2023-06-29
Add additional configuration parameters for VClusterOps create db

roypaulin:
When calling create_db for admintools we would create a file that has a bunch of config options in it. n the vclusterOps, these are all implemented using a parameter called ConfigurationParameters. So we collect all the parameters and pass them as a map to createdb.

### DIFF
--- a/pkg/controllers/vdb/init_db.go
+++ b/pkg/controllers/vdb/init_db.go
@@ -22,13 +22,11 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
-	"github.com/lithammer/dedent"
 	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
 	"github.com/vertica/vertica-kubernetes/pkg/cloud"
 	"github.com/vertica/vertica-kubernetes/pkg/cmds"
 	verrors "github.com/vertica/vertica-kubernetes/pkg/errors"
 	"github.com/vertica/vertica-kubernetes/pkg/events"
-	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
 	"github.com/vertica/vertica-kubernetes/pkg/names"
 	"github.com/vertica/vertica-kubernetes/pkg/paths"
 	"github.com/vertica/vertica-kubernetes/pkg/vdbstatus"
@@ -51,19 +49,19 @@ const (
 type DatabaseInitializer interface {
 	getPodList() ([]*PodFact, bool)
 	findPodToRunInit() (*PodFact, bool)
-	execCmd(ctx context.Context, initiatorPod types.NamespacedName,
-		hostList []string, confParms map[string]string) (ctrl.Result, error)
+	execCmd(ctx context.Context, initiatorPod types.NamespacedName, hostList []string) (ctrl.Result, error)
 	preCmdSetup(ctx context.Context, initiatorPod types.NamespacedName, podList []*PodFact) (ctrl.Result, error)
 	postCmdCleanup(ctx context.Context) (ctrl.Result, error)
 }
 
 type GenericDatabaseInitializer struct {
-	initializer DatabaseInitializer
-	VRec        *VerticaDBReconciler
-	Log         logr.Logger
-	Vdb         *vapi.VerticaDB
-	PRunner     cmds.PodRunner
-	PFacts      *PodFacts
+	initializer         DatabaseInitializer
+	VRec                *VerticaDBReconciler
+	Log                 logr.Logger
+	Vdb                 *vapi.VerticaDB
+	PRunner             cmds.PodRunner
+	PFacts              *PodFacts
+	ConfigurationParams map[string]string
 }
 
 // checkAndRunInit will check if the database needs to be initialized and run init if applicable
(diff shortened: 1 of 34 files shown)


roypaulin:
I know the ticket's description specifies that we move the file generation to the dispatcher abstraction, but I chose this approach as it uses what is already there. If we want to refactor all that part of the code then I will probably close this PR and open a new one.

spilchen:
I would prefer to see the additional refactor. We still create a file on disk for the vclusterOps path. I want to avoid the file creation entirely because we may have some plain text secrets in there -- even if the file is short lived. Instead of creating a file in init_db.go, it should populate a map[string]string and pass that down to the dispatcher abstraction.

roypaulin:
Ok I will refactor the code. But what is wrong with what I did? I made so that we do not create a file anymore.

spilchen:
Okay, I see what you did there. Sorry, I missed that the first time I went through this. I still would like to see the refactor though. I think it will be cleaner to push down the changes. One of the goals with the dispatcher is to localize all of the admintools related functions, so grouping the auth file logic there will further that goal.

roypaulin:
Ok I understand. Thanks!

pkg/vadmin/add_node_at.go
```diff
@@ -57,7 +57,7 @@ func isLicenseLimitError(stdout string) bool {
 }
 
 // genAddNodeCommand returns the command to run to add nodes to the cluster.
-func (a Admintools) genAddNodeCommand(s *addnode.Parms) []string {
+func (a *Admintools) genAddNodeCommand(s *addnode.Parms) []string {
```
roypaulin, line 60: I used a pointer receiver function. I am curious, any reason why you did not?

pkg/vadmin/describe_db_at.go
```diff
@@ -42,7 +42,7 @@ func (a Admintools) DescribeDB(ctx context.Context, opts ...describedb.Option) (
 
 // genDescribeCmd will generate the command line options for calling
 // admintools -t revive_db --display-only.
-func (a Admintools) genDescribeCmd(s *describedb.Parms) []string {
+func genDescribeCmd(s *describedb.Parms) []string {
```
roypaulin, line 45: I turned the method into a simple function because it did not have any dependencies on `Admintools`. I saw that in many cases you used receiver functions even though we are not using any fields from `Admintools`. Any reasons why?

pkg/vadmin/add_node_at.go
spilchen: We weren't changing anything in `Admintools` and it was a small struct. So a copy shouldn't be to bad. I'm okay with changing it to a pointer receiver. Especially if you are adding new fields to Admintools.

pkg/vadmin/describe_db_at.go
spilchen: I like to keep the admintools commands together under the same struct. It makes it obvious that we are generating a command for admintools. I think functionally it would be fine to drop the receiver. But I think that makes sense only if you find a case where you can reuse the function outside of an `Admintools` function.

pkg/controllers/vdb/init_db.go
```diff
@@ -114,7 +117,7 @@ func (g *GenericDatabaseInitializer) runInit(ctx context.Context) (ctrl.Result,
 	}
 
 	if res, err := g.initializer.execCmd(ctx, initiatorPod,
-		getHostList(podList)); verrors.IsReconcileAborted(res, err) {
+		getHostList(podList), genCommunalParmsMap(content)); verrors.IsReconcileAborted(res, err) {
```
spilchen, line 120: I wonder if it would be better to just use a string map from the start? Rather than build a string of key/value pairs, then parse the string into a map. It seems like a lot of processing we can save if we just started with a map at the beginning and build it up that way.

pkg/vadmin/at.go
```diff
@@ -48,3 +53,39 @@ func (a Admintools) execAdmintools(ctx context.Context, initiatorPod types.Names
...
+	// characters then we won't even be able to copy the file.
+	if err != nil {
+		a.EVWriter.Eventf(a.VDB, corev1.EventTypeWarning, events.AuthParmsCopyFailed,
+			"Failed to copy auth parms to the pod '%s'", initiatorPod)
+	}
+	return err
+}
+
+// DestroyAuthParms will remove the auth parms file that was created in the pod
+func (a *Admintools) DestroyAuthParms(ctx context.Context, initiatorPod types.NamespacedName) {
```
roypaulin, line 73: I also moved this to the dispatcher abstraction. Does it make sense?

pkg/controllers/vdb/init_db.go
```diff
@@ -656,17 +600,13 @@ func (g *GenericDatabaseInitializer) hasCompatibleVersion(supportedVersion, even
 	return ctrl.Result{Requeue: true}
 }
 
-// genCommunalParmsMap takes a multiline string containing the parms already
-// set(in "parm = value" format) and returns the corresponding map.
-func genCommunalParmsMap(content string) map[string]string {
+// genLowerCaseKeysMap will generate a lowercase keys copy of
+// ConfigurationParams map. This is useful to detect duplication
+// between user-defined and operator-generated params.
+func (g *GenericDatabaseInitializer) genLowerCaseKeysMap() map[string]string {
```
spilchen, line 606: Could we just ensure the entries into the map always use a lowercase key? Regenerating the map with lowercase keys seems like wasted effort when we can just build the map with lowercase keys to begin with. Here is a suggestion on how to create a case insensitive map: https://stackoverflow.com/questions/11124705/how-to-create-a-case-insensitive-map-in-go

spilchen (commented):
Some comments for the init_db.go file.

PR #438, opened 2023-06-29
Always run e2e-vcluster with latest master

chinhtranvan:
The e2e-vcluster test suite needs to run with the latest version of the server. For this reason, when we run that test suite we will use a private image from the dockerhub repo with the lag lastest-master. 

We still build the v2 image to make sure nothing is broken. But since it uses an RPM from version 12.0.3 of the server, it doesn't have all of the functionality that the e2e-vcluster test suite needs. So, no e2e tests will run with it.


.github/workflows/build-images.yml
```diff
@@ -315,7 +315,7 @@ jobs:
     - name: Build and optionally push v2 server image
       if: ${{ inputs.v2_vertica_image == '' }}
       run: |
-        export VERTICA_IMG=${{ steps.v2_vertica_image.outputs.value }}
+        export VERTICA_IMG=verticadocker/vertica-k8s-private:latest-master
```
spilchen, line 318: Not sure if this is still WIP, but we must honour the image that's passed into e2e.yml. We should only use latest-master if it the e2e.yml was invoked through a push, PR or workflow dispatch and v2 image was empty.

chinhtranvan: Yes, It's ready for review now.

.github/workflows/build-images.yml
```diff
@@ -288,7 +288,7 @@ jobs:
         default: ghcr.io/${{ env.OWNER_LC }}/vertica-k8s:${{ github.sha }}-v2
         conditionals-with-values: |
           ${{ inputs.v2_vertica_image != '' }} => ${{ inputs.v2_vertica_image }}
-          ${{ github.event_name == 'pull_request' }} => vertica-k8s:kind-v2
+          ${{ github.event_name == 'pull_request' }} => verticadocker/vertica-k8s-private:latest-master
```
spilchen, line 291: We will use this image name when we build the image a few steps down. We need that build step to use some other image name. Otherwise, it will overwrite the image we want to use. Also, this doesn't handle the case when we push to main. That will use the ghcr.io image name from line 288.

spilchen (commented):
I haven't run this, but I think the changes need to be similar to this:
```
diff --git a/.github/actions/setup-e2e/action.yaml b/.github/actions/setup-e2e/action.yaml
index 4447c283..972e7433 100644
--- a/.github/actions/setup-e2e/action.yaml
+++ b/.github/actions/setup-e2e/action.yaml
@@ -29,7 +29,6 @@ runs:
         docker load --input nokeys-vertica-image/nokeys-vertica-image.tar
         docker load --input minimal-vertica-image/minimal-vertica-image.tar
         docker load --input operator-image/operator-image.tar
-        docker load --input v2-vertica-image/v2-vertica-image.tar
         docker load --input vlogger-image/vlogger-image.tar
         docker image ls -a
 
diff --git a/.github/workflows/build-images.yml b/.github/workflows/build-images.yml
index 904caf8d..4226650e 100644
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -285,18 +285,9 @@ jobs:
       id: v2_vertica_image
       with:
-        default: ghcr.io/${{ env.OWNER_LC }}/vertica-k8s:${{ github.sha }}-v2
+        default: verticadocker/vertica-k8s-private:latest-master
         conditionals-with-values: |
           ${{ inputs.v2_vertica_image != '' }} => ${{ inputs.v2_vertica_image }}
-          ${{ github.event_name == 'pull_request' }} => vertica-k8s:kind-v2
-
-    - name: Login to GitHub Container registry for non-PRs
-      uses: docker/login-action@v2
-      if: ${{ github.event_name != 'pull_request' && inputs.v2_vertica_image == '' || startsWith(inputs.v2_vertica_image, 'ghcr.io') }}
-      with:
-        registry: ghcr.io
-        username: ${{ github.repository_owner }}
-        password: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Login to Docker Hub
       uses: docker/login-action@v2
@@ -312,26 +303,16 @@ jobs:
       uses: ./.github/actions/download-rpm
       if: ${{ inputs.v2_vertica_image == '' }}
 
-    - name: Build and optionally push v2 server image
+    - name: Build the v2 image
       if: ${{ inputs.v2_vertica_image == '' }}
       run: |
-        export VERTICA_IMG=${{ steps.v2_vertica_image.outputs.value }}
+        # We never push the image we build here because the RPM we use is too
+        # old. The RPM needs to be from at least a 23.3.0 server. So, we build
+        # here only to test the creation of the image. We don't verify this
+        # image in the remaining CI. Pick a dummy image name so as not to
+        # overwrite the output build image that the e2e tests will use.
+        export VERTICA_IMG=vertica-k8s:kind-v2
         make docker-build-vertica-v2
-        if [ $GITHUB_EVENT_NAME != 'pull_request' ]
-        then
-          make docker-push-vertica
-        fi
-
-    - name: Save the image for consumption by dependent jobs (PRs only)
-      if: ${{ github.event_name == 'pull_request' }}
-      run: |
-        docker save ${{ steps.v2_vertica_image.outputs.value }} > v2-vertica-image.tar
-
-    - uses: actions/upload-artifact@v3
-      if: ${{ github.event_name == 'pull_request' }}
-      with:
-        name: v2-vertica-image
-        path: v2-vertica-image.tar
 
     - name: Do a local pull of the image if we didn't create it
       if: ${{ inputs.v2_vertica_image != '' }}
```

.github/workflows/build-images.yml
chinhtranvan: Agree! I aslo think about how to push the main. Will fix it as your suggestion!

chinhtranvan:
Ready for second round review! Could you take another look?

.github/workflows/build-images.yml
```diff
@@ -285,7 +285,7 @@ jobs:
       uses: spilchen/switch-case-action@v2
       id: v2_vertica_image
       with:
-        default: ghcr.io/${{ env.OWNER_LC }}/vertica-k8s:${{ github.sha }}-v2
+        default: verticadocker/vertica-k8s-private:latest-master
         conditionals-with-values: |
           ${{ inputs.v2_vertica_image != '' }} => ${{ inputs.v2_vertica_image }}
           ${{ github.event_name == 'pull_request' }} => vertica-k8s:kind-v2
```
spilchen, line 291: I don't think you applied everything from my suggestion. This line for instance should be removed. For PR's we want to use the default name set at 288. There was also a change needed in another file. Can you review the change I suggested again?

chinhtranvan: @spilchen I already tried your suggestion, but it looks like that we can't docker save the latest-master image 
```
Run docker save verticadocker/vertica-k8s-private:latest-master > v2-vertica-image.tar
Error response from daemon: reference does not exist
Error: Process completed with exit code 1.
```

.github/workflows/build-images.yml
```diff
@@ -285,10 +285,9 @@ jobs:
       uses: spilchen/switch-case-action@v2
       id: v2_vertica_image
       with:
-        default: ghcr.io/${{ env.OWNER_LC }}/vertica-k8s:${{ github.sha }}-v2
+        default: verticadocker/vertica-k8s-private:latest-master
```
spilchen, line 288: I think this needs to prefix with docker.io. There is other code in the workflow that looks at the prefix to know it has to log into dockerhub
```suggestion
        default: docker.io/verticadocker/vertica-k8s-private:latest-master
```

.github/workflows/build-images.yml
```diff
@@ -312,10 +311,15 @@ jobs:
...
       run: |
-        export VERTICA_IMG=${{ steps.v2_vertica_image.outputs.value }}
+        # We never push the image we build here because the RPM we use is too
+        # old. The RPM needs to be from at least a 23.3.0 server. So, we build
+        # here only to test the creation of the image. We don't verify this
+        # image in the remaining CI. Pick a dummy image name so as not to
+        # overwrite the output build image that the e2e tests will use.
+        export VERTICA_IMG=vertica-k8s:kind-v2
         make docker-build-vertica-v2
         if [ $GITHUB_EVENT_NAME != 'pull_request' ]
```
spilchen, line 324: We should remove this push because we aren't going to use this new image anywhere in the workflow.

.github/workflows/build-images.yml
chinhtranvan: Sure!

chinhtranvan:
As we discuss today, I will delete this PR.

spilchen:
Closing as these changes were made in #440

PR #442, opened 2023-07-04
Use extra config parms for VClusterOps create DB (#441)

roypaulin:
When calling create_db for admintools, we would create a file that has a bunch of config options in it. When using the vclusterOps library, these config parameters are passed as a map with the `ConfigurationParameters` option. This change will set up that option. This will allow us to create a database through vclusterOps with extra config parameters so that we can connect to a communal backend like S3 or use advanced options like S3 SSE.

Starts from #436


pkg/controllers/vdb/init_db.go
```diff
@@ -365,45 +332,26 @@ func (g *GenericDatabaseInitializer) getAzureAuthParmsContent(ctx context.Contex
...
-func (g *GenericDatabaseInitializer) getAdditionalConfigParmsContent(content string) string {
-	parmNames := genCommunalParmsMap(content)
-	parms := strings.Builder{}
+// setAdditionalConfigParms adds additional server config parameters
+// to the config parms map.
+func (g *GenericDatabaseInitializer) setAdditionalConfigParms() {
 	for k, v := range g.Vdb.Spec.Communal.AdditionalConfig {
 		// we lowercase parm names to catch duplications
 		// like AWSauth/awsauth. In case of duplicate names,
 		// we skip to the next parm.
```
spilchen, line 347: I think we can remove this comment now as the lowercase logic was moved to the CiMap

pkg/controllers/vdb/init_db.go
```diff
@@ -514,85 +462,81 @@ func (g *GenericDatabaseInitializer) getEnableHTTPS() string {
...
+	return ctrl.Result{}, nil
 }
 
-// getServerSideEncryptionAlgorithm returns an entry for S3ServerSideEncryption
-// if sse type is SSE-S3|SSE-KMS or for S3SseCustomerAlgorithm if SSE-C.
-func (g *GenericDatabaseInitializer) getServerSideEncryptionAlgorithm() string {
+// setServerSideEncryptionAlgorithm adds an entry to the config parms map for S3ServerSideEncryption,
+// if sse type is SSE-S3|SSE-KMS, or for S3SseCustomerAlgorithm, if sse type is SSE-C.
+func (g *GenericDatabaseInitializer) setServerSideEncryptionAlgorithm() {
 	if g.Vdb.IsSseC() {
```
spilchen, line 490: Can we change the `if {} if {} if {}` calls to a `switch` statement? Before we use to return once the first one succeeded. But now, we will check the next if.

spilchen (commented):
Looks good. Thanks for making the revisions from the other PR.

pkg/vadmin/create_db_at.go
```diff
@@ -76,3 +77,23 @@ func (a *Admintools) genCreateDBCmd(s *createdb.Parms) []string {
...
+}
+
+// GetInitiator returns the initiator pod name.
+func (d *DBCreator) GetInitiator() types.NamespacedName {
+	return d.Parms.Initiator
+}
+
+// LogFailure will log and record an event for an admintools -t create_db failure
+func (d *DBCreator) LogFailure(stdout string, err error) (ctrl.Result, error) {
+	return d.Admintools.logFailure("create_db", events.ReviveDBFailed, stdout, err)
```
spilchen, line 93: We are using the wrong event here:
```suggestion
	return d.Admintools.logFailure("create_db", events.CreateDBFailed, stdout, err)
```

spilchen (commented):
I think we are close. Just one minor comment.

spilchen (approved):

PR #450, opened 2023-07-06
Skip e2e-vcluster test suite only for forks

chinhtranvan:


chinhtranvan:
We skip this test suite on forks because it requires secrets that are only accessible through the vertica-kubernetes repo. Any branch run on the vertica/vertica-kubernetes repo needs to run the e2e-vcluster

spilchen:
Here is the workflow run that showed that we successfully run the e2e-vcluster test suite if e2e is started manually on a branch:
https://github.com/vertica/vertica-kubernetes/actions/runs/5477876655

spilchen (approved):

PR #451, opened 2023-07-07
Implement VClusterOps.FetchNodeState

le-an-vertica:
The uses the new VFetchNodeState API in the vclusterOps library when the restart reconciler needs state information about down nodes. This change is hidden behind the vclusterOps feature flag and isn't run by default.


pkg/vadmin/fetch_node_state_vc.go
```diff
@@ -27,5 +30,48 @@ import (
...
+
+	return stateMap, ctrl.Result{}, nil
+}
+
+func (v *VClusterOps) genFetchNodeStateOptions(s *fetchnodestate.Parms) vops.VFetchNodeStateOptions {
+	opts := vops.VFetchNodeStateOptionsFactory()
+
+	for k, v := range s.HostsNeeded {
+		if v {
+			opts.RawHosts = append(opts.RawHosts, k)
```
spilchen, line 62: I thought the opts.RawHost we pass in is the known up node. This should be the `s.InitiatorIP`.

pkg/vadmin/fetch_node_state_vc.go
```diff
@@ -27,5 +30,48 @@ import (
...
+	vopts := v.genFetchNodeStateOptions(&s)
+	nodesInfo, err := v.VFetchNodeState(&vopts)
+	if err != nil {
+		v.Log.Error(err, "failed to fetch node states")
+		return nil, ctrl.Result{}, err
+	}
+
+	// parse node states
+	stateMap := map[string]string{} // node name to state map
+	for _, nodeInfo := range nodesInfo {
```
spilchen, line 50: We should only return information for the hosts that the caller was interested. We can use the `s.HostsNeeded` map for that.

pkg/vadmin/fetch_node_state_vc_test.go
```diff
@@ -0,0 +1,71 @@
...
+	for i := 1; i <= 3; i++ {
+		nodeName := fmt.Sprintf("v_%s_node000%d", TestDBName, i)
+		expectedResults[nodeName] = "UP"
+	}
+	initiatorName := "v" + TestDBName + "_node0001"
+
+	It("should call vcluster-ops library with fetch_node_state task", func() {
+		dispatcher := mockVClusterOpsDispatcher()
+		actualResults, ctrlRes, err := dispatcher.FetchNodeState(ctx,
+			fetchnodestate.WithHost(initiatorName, TestInitiatorIP))
```
spilchen, line 66: The initiator and the requests hosts are separate arguments. We should have both options given. The hosts being a set of nodes that we think are down and the initiator is the machine we talk to.

pkg/vadmin/suite_test.go
```diff
@@ -119,6 +115,15 @@ func (m *MockVClusterOps) VerifyCommonOptions(options *vops.DatabaseOptions) err
 	return nil
 }
 
+// VerifyCommonOptions is used in vcluster-ops unit test for verifying the common options among all db ops
+func (m *MockVClusterOps) VerifyCommonOptions(options *vops.DatabaseOptions) error {
+	if *options.Name != TestDBName {
```
cchen-vertica, line 120: In the future, we will have db name in vcluster fetchNodeState. I think we can just simply assign the dbname to opts.Name in genFetchNodeStateOptions(). Then we don't need to change VerifyCommonOptions() here.

pkg/vadmin/fetch_node_state_vc.go
le-an-vertica: I just added `s.InitiatorIP` to `opts.RawHosts`. Please verify if this is what you suggested.

go.sum
```diff
@@ -278,6 +278,10 @@ github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/
 github.com/stretchr/testify v1.8.2 h1:+h33VjcLVPDHtOdpUCuF+7gSuG3yGIftsP1YvFihtJ8=
 github.com/vertica/vcluster v0.0.0-20230630151840-081eb662e260 h1:q2TQBIUvqgi96uNZ2YfNvEBTWTWhNbRRo1V8ThC4QM0=
 github.com/vertica/vcluster v0.0.0-20230630151840-081eb662e260/go.mod h1:zAO9nH73iAc40oZI+jd3sEZ3FRkrMLHWCvB0TIA4EHo=
+github.com/vertica/vcluster v0.0.0-20230705215918-5fe3c38815a8 h1:FZdMHXQUuA7N4SOEOXEoKX+7EnWb0n/p+JBCUVMfCGs=
```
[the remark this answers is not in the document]
le-an-vertica: Sure, I will get the newer version of vcluster

tests/e2e-vcluster/vcluster-ks-1/25-assert.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+apiVersion: v1
+kind: VerticaDB
+metadata:
+  name: v-vcluster-ks-1
+status:
+  subclusters:
+    - installCount: 3
+      addedToDBCount: 3
+      upNodeCount: 2
```
spilchen, line 22: I don't think we will able to pass this. This testcase does a stop db right after the create. But we don't have any restart capability. So, I don't think we can validate the fetch node in e2e just yet. My suggestion is to remove this file and we will validate restart, and thus fetch node, once we have more of the vclusterOps implemented.

spilchen (commented):
No issue with the code. Just a comment about the test.

tests/e2e-vcluster/vcluster-ks-1/25-assert.yaml
le-an-vertica: Thanks for this comment! I have removed `upNodeCount: 2`.

spilchen (approved):

PR #458, opened 2023-07-17
Allow use of GSM for communal credential secret

sivaalamp:
This adds a feature flag to get the secret for the communal credential from google secret manager (GSM) instead of from a k8s secret. 

It is off by default. You must set the following annotation in the VerticaDB CR:
```
kind: VerticaDB
metadata:
  name: v
  annotations:
    vertica.com/use-gcp-secret-manager: "true" 
spec:
  communal:
    credentialSecret: <full name of the secret in GSM>
...
```

The name of the GSM secret is taken from the existing `spec.communal.credentialSecret` field.


pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +279,55 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+	result, err := client.AccessSecretVersion(ctx, req)
+	if err != nil {
+		g.Log.Info("google: could not find default credentials")
+		return "", ctrl.Result{}, err
+	}
+
+	crc32c := crc32.MakeTable(crc32.Castagnoli)
+	checksum := int64(crc32.Checksum(result.Payload.Data, crc32c))
+	if checksum != *result.Payload.DataCrc32C {
+		return "", ctrl.Result{}, err
```
spilchen, line 303: err would be nil here. We should generate one:
```suggestion
		return "", ctrl.Result{}, fmt.Errorf("checksum failure, expected %d but got %d", *result.Payload.DataCrc32C, checksum)```

File: pkg/controllers/vdb/init_db.go [hunk 2]
```diff
@@ -269,12 +279,55 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+	}
+
+	var gs GcpSecret
+	err = json.Unmarshal([]byte(string(result.Payload.Data)), &gs)
+	if err != nil {
+		g.Log.Info("Permission 'secretmanager.versions.access' denied for resource to read the secret")
```
spilchen, line 309: We should use `g.Log.Error`

spilchen, line 309: Why do we consider this a permission issue? Couldn't we get this error if the JSON was in a different format (i.e. key values don't match or additional fields are included).

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +279,55 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+	}
+	defer client.Close()
+
+	req := &secretmanagerpb.AccessSecretVersionRequest{
+		Name: name,
+	}
+
+	result, err := client.AccessSecretVersion(ctx, req)
+	if err != nil {
+		g.Log.Info("google: could not find default credentials")
```
spilchen, line 296: We should use `g.Log.Error`

spilchen, line 296: What does the error message mean? What are default credentials

pkg/controllers/vdb/init_db.go
```diff
@@ -65,6 +70,11 @@ type GenericDatabaseInitializer struct {
 	ConfigurationParams *vtypes.CiMap
 }
 
+type GcpSecret struct {
```
spilchen, line 73: Lets not export this struct. Also, secret is too generic. Lets call out the fact that it's for credentials.
```suggestion
type gcpCredentialSecret struct {
```

pkg/controllers/vdb/init_db.go
```diff
@@ -65,6 +70,11 @@ type GenericDatabaseInitializer struct {
...
+	Accesskey string
+	Secretkey string
```
spilchen, line 75: Does this need to have specific field names, or could we have used something more generic like `map[string]string` instead?

pkg/controllers/vdb/init_db.go
```diff
@@ -65,6 +70,11 @@ type GenericDatabaseInitializer struct {
 	ConfigurationParams *vtypes.CiMap
 }
 
+type GcpSecret struct {
+	Accesskey string
```
spilchen, line 74: Depending on other comments, this may not apply anymore. But I'm including it here so I don't forget... since we use struct to unmarshal JSON, we should have JSON style comments attached with it.
```suggestion
	Accesskey string ``json:"accesskey"`
```

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +279,55 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+	}
+
+	crc32c := crc32.MakeTable(crc32.Castagnoli)
+	checksum := int64(crc32.Checksum(result.Payload.Data, crc32c))
+	if checksum != *result.Payload.DataCrc32C {
+		return "", ctrl.Result{}, err
+	}
+
+	var gs GcpSecret
+	err = json.Unmarshal([]byte(string(result.Payload.Data)), &gs)
```
spilchen, line 307: What data type is `result.Payload.Data`?

pkg/controllers/vdb/init_db.go
roypaulin, line 296: Where is it supposed to find the credentials? By credentials do you mean credentials for the GCP account?

pkg/meta/annotations.go
```diff
@@ -37,6 +37,12 @@ const (
 	// treated as a boolean.
 	VClusterOpsAnnotation     = "vertica.com/vcluster-ops"
 	VClusterOpsAnnotationTrue = "true"
+
+	// This is a feature flag for accessing the secrets configured in Google Secret Manager.
+	// This is set to true by default to let the verticadb running cluster to access the secrets.
+	// The value of this annotation is treated as a boolean.
+	GcpGsmAnnotation     = "vertica.com/gcpgsmaccess"
```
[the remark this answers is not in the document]
sivaalamp: Changed the annotation

pkg/meta/annotations.go
```diff
@@ -51,6 +57,12 @@ func UseVClusterOps(annotations map[string]string) bool {
 	return lookupBoolAnnotation(annotations, VClusterOpsAnnotation, false)
 }
 
+// UseGCPSecretManager returns true nk all admin commands should use the vclusterOps
+// library rather than admintools.
```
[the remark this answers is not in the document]
sivaalamp: Added the appropriate comments

pkg/meta/annotations.go
```diff
@@ -37,6 +37,12 @@ const (
 	// treated as a boolean.
 	VClusterOpsAnnotation     = "vertica.com/vcluster-ops"
 	VClusterOpsAnnotationTrue = "true"
+
+	// This is a feature flag for accessing the secrets configured in Google Secret Manager.
+	// This is set to true by default to let the verticadb running cluster to access the secrets.
```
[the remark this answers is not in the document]
sivaalamp: The default value is changed to false now and removed the comment about the same

pkg/controllers/vdb/init_db.go
```diff
@@ -17,16 +17,21 @@ package vdb
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"hash/crc32"
 	"regexp"
 	"strings"
 
+	gsm "cloud.google.com/go/secretmanager/apiv1"
```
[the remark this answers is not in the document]
sivaalamp: will check in the go.mod and go.sum files

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +279,55 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
 	g.ConfigurationParams.Set("KerberosEnableKeytabPermissionCheck", "0")
 }
 
+func (g *GenericDatabaseInitializer) setAuthFromGCSSecret(ctx context.Context) (string, ctrl.Result, error) {
+	name := "projects/otl-ot2-int-sb/secrets/new_vertica_secret/versions/1"
```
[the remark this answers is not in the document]
sivaalamp: Introduced the annotations to read the project id and secret name

pkg/controllers/vdb/init_db.go
sivaalamp: Added err as suggested

pkg/controllers/vdb/init_db.go
sivaalamp: Changed to Error instead of Info

pkg/controllers/vdb/init_db.go
sivaalamp: Changed to Error instead of Info

pkg/controllers/vdb/init_db.go
sivaalamp: This comment is obsolete now as I changed it to map[string]string{} and this struct is no more present

pkg/controllers/vdb/init_db.go
sivaalamp: Removed the struct and introduced map

pkg/controllers/vdb/init_db.go
sivaalamp: This comment is obsolete now as I changed it to map[string]string{} and this struct is no more present

pkg/controllers/vdb/init_db.go
sivaalamp: it is an uint8[]

pkg/controllers/vdb/init_db.go
sivaalamp: Updated with appropriate error message

sivaalamp: Yes.

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +274,58 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
 	g.ConfigurationParams.Set("KerberosEnableKeytabPermissionCheck", "0")
 }
 
+func (g *GenericDatabaseInitializer) setAuthFromGCSSecret(ctx context.Context) (string, ctrl.Result, error) {
+	name := "projects/" + g.Vdb.Annotations[meta.GcpProjectIDAnnotation] +
+		"/secrets/" + g.Vdb.Annotations[meta.GcpSecNameAnnontation] +
+		"/versions/" + g.Vdb.Annotations[meta.GcpSecVersionAnnotation]
```
spilchen, line 280: We shouldn't use annotations to come up with then secret name. We already have a field in the VerticaDB CR for the name and should use that: `spec.communal.credentialSecret`

pkg/controllers/vdb/init_db.go
spilchen: err is still nil here I believe.

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +274,58 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
 	g.ConfigurationParams.Set("KerberosEnableKeytabPermissionCheck", "0")
 }
 
+func (g *GenericDatabaseInitializer) setAuthFromGCSSecret(ctx context.Context) (string, ctrl.Result, error) {
```
spilchen, line 277: Let's drop `ctrl.Result` as an output parm. I think we set it to `ctrl.Result{}` each time.

pkg/controllers/vdb/init_db.go
spilchen: It is my understanding that the host where the operator runs has a profile setup to allow it access to the GSM.

pkg/controllers/vdb/init_db.go
spilchen: I think uint8 is the same as byte. Can we just pass in the payload without a cast?

pkg/controllers/vdb/init_db.go
sivaalamp: Making use of it now, so didn't change the return type

pkg/controllers/vdb/init_db.go
sivaalamp: Removed the new annotations and used the existing credentialSecret for GSM as well

pkg/controllers/vdb/init_db.go
spilchen: I know we talked about this on teams, but the code here looks wrong. I think my suggestion will solve it.

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +274,70 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+	name := g.Vdb.Spec.Communal.CredentialSecret
+	client, err := gsm.NewClient(ctx)
+	if err != nil {
+		return "", ctrl.Result{}, err
+	}
+	defer client.Close()
+
+	req := &secretmanagerpb.AccessSecretVersionRequest{
+		Name: name,
```
spilchen, line 286: Suggest removal of the `name` parameter:
```suggestion
		Name: g.Vdb.Spec.Communal.CredentialSecret,
```

config/samples/v1beta1_verticadb.yaml
```diff
@@ -15,12 +15,13 @@ apiVersion: vertica.com/v1beta1
 kind: VerticaDB
 metadata:
   name: verticadb-sample
+
```
spilchen, line 18: minor nit: can we just undo the changes in here entirely. I know semantically nothing changed, I would prefer if this file didn't change at all

pkg/meta/annotations.go
```diff
@@ -37,6 +37,11 @@ const (
 	// treated as a boolean.
 	VClusterOpsAnnotation     = "vertica.com/vcluster-ops"
 	VClusterOpsAnnotationTrue = "true"
+
+	// This is a feature flag for accessing the secrets configured in Google Secret Manager.
+	// The value of this annotation is treated as a boolean.
+	GcpGsmAnnotation     = "vertica.com/use-gcp-secret-manager"
+	GcpGsmAnnotationTrue = "true"
```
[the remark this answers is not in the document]
sivaalamp: These should be supplied in the yaml file to return true from UseGCPSecretManager so that it fetches gsm secrets.

pkg/controllers/vdb/init_db.go
sivaalamp: removed the name variable

pkg/controllers/vdb/init_db.go
sivaalamp: Was a copy paster error, corrected it now

pkg/controllers/vdb/init_db.go
sivaalamp: No casting is done

pkg/controllers/vdb/init_db.go
sivaalamp: Added the proper logs now

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +274,70 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+			"The communal credential secret '%s' does not have a key named '%s'", g.Vdb.Spec.Communal.CredentialSecret, cloud.CommunalAccessKeyName)
+		return "", ctrl.Result{Requeue: true}, nil
+	}
+
+	secretKey, ok := GcpCred[cloud.CommunalSecretKeyName]
+	if !ok {
+		g.VRec.Eventf(g.Vdb, corev1.EventTypeWarning, events.CommunalCredsWrongKey,
+			"The communal credential secret '%s' does not have a key named '%s'", g.Vdb.Spec.Communal.CredentialSecret, cloud.CommunalSecretKeyName)
+		return "", ctrl.Result{Requeue: true}, nil
+	}
```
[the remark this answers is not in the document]
sivaalamp: Not getting the auth value in proper readable format when the GcpCred is changed to map[string][]byte{}, so the code is retained

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +274,70 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+	if checksum != *result.Payload.DataCrc32C {
+		g.Log.Error(err, "Permission denied to read the secret")
+		return "", ctrl.Result{}, err
+	}
+
+	GcpCred := map[string]string{}
+
+	err = json.Unmarshal([]byte(string(result.Payload.Data)), &GcpCred)
+	if err != nil {
+		return "", ctrl.Result{}, fmt.Errorf("checksum failure, expected %d but got %d", *result.Payload.DataCrc32C, checksum)
```
spilchen, line 306: The error text does make sense here. Is this a copy+paste error?

pkg/controllers/vdb/init_db.go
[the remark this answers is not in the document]
spilchen: Can you provide more details here? Is one of the values base64 encoded?

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +274,70 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+	}
+
+	secretKey, ok := GcpCred[cloud.CommunalSecretKeyName]
+	if !ok {
+		g.VRec.Eventf(g.Vdb, corev1.EventTypeWarning, events.CommunalCredsWrongKey,
+			"The communal credential secret '%s' does not have a key named '%s'", g.Vdb.Spec.Communal.CredentialSecret, cloud.CommunalSecretKeyName)
+		return "", ctrl.Result{Requeue: true}, nil
+	}
+
+	auth := fmt.Sprintf("%s:%s", strings.TrimSuffix(string(accessKey), "\n"), strings.TrimSuffix(string(secretKey), "\n"))
```
spilchen, line 323: Why do you need to strip newlines from the secrets?

sivaalamp: the configuration parameter takes the auth value in terms of accesskey:secretkey, which forming the auth parameter to format in the required format if there are any new lines those needs to be trimmed. I already have checked, if there are new lines in the keys, nothing will happen to the string.

pkg/controllers/vdb/init_db.go
sivaalamp: Moved it already

pkg/meta/annotations.go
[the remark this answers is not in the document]
spilchen: But do we need this const defined in the code here? Where do we use it?

pkg/controllers/vdb/init_db.go
spilchen: But did you see a case where a newline was at the end of one of the keys?

spilchen:
@sivaalamp we are almost finished with the review. There are just two remaining comments to answer and/or address.

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +274,70 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+
 // setGCloudAuthParms adds the auth parms to the config parms map when we are
 // connecting to google cloud storage.
 func (g *GenericDatabaseInitializer) setGCloudAuthParms(ctx context.Context) (ctrl.Result, error) {
-	res, err := g.setAuth(ctx, "GCSAuth")
-	if verrors.IsReconcileAborted(res, err) {
-		return res, err
+	if meta.UseGCPSecretManager(g.Vdb.Annotations) {
+		auth, res, err := g.setAuthFromGCSSecret(ctx)
+		if err != nil {
```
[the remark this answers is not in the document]
sivaalamp: Added as suggested

pkg/meta/annotations.go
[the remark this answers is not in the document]
sivaalamp: Removed the GcpGsmAnnotationTrue constant, as we can directly supply value true from the yaml file

config/samples/v1beta1_verticadb.yaml
```diff
@@ -15,10 +15,16 @@ apiVersion: vertica.com/v1beta1
 kind: VerticaDB
 metadata:
   name: verticadb-sample
+  annotations:
```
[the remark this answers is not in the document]
sivaalamp: no change in this file now

pkg/controllers/vdb/init_db.go
```diff
@@ -269,12 +274,68 @@ func (g *GenericDatabaseInitializer) setKerberosAuthParms() {
...
+		return "", ctrl.Result{}, fmt.Errorf("data corruption detected, expected %d but got %d", *result.Payload.DataCrc32C, checksum)
+	}
+
+	GcpCred := map[string]string{}
+	err = json.Unmarshal(result.Payload.Data, &GcpCred)
+	if err != nil {
+		return "", ctrl.Result{}, err
+	}
+
+	accessKey, ok := GcpCred[cloud.CommunalAccessKeyName]
```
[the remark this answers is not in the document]
roypaulin: I am fine with adding a comment. I couldn't find an easy way to remove the duplication.

spilchen (approved):

PR #464, opened 2023-07-24
Add event logging for errors from vclusterops library

spilchen:
This is a continuation of the vclusterOps integration. The vclusterops library can now flow errors that follow the [RFC 7807 spec](https://datatracker.ietf.org/doc/html/rfc7807). This allows the operator to know about specific error without having to resort to regex based searching. 

We only handle some of the more popular create db failures for now. But this work here will allow us to easily add support for additional types that flow from the vclusterops library.

For instance, if you create a VerticaDB with vclusterops enabled, and the communal storage path given is not empty, we will now see a message like this:
```
Events:
  Type     Reason                  Age                   From                Message
  ----     ------                  ----                  ----                -------
  Normal   CreateDBStart           4m19s (x40 over 14m)  verticadb-operator  Starting create database
  Warning  CommunalPathIsNotEmpty  9m14s (x20 over 14m)  verticadb-operator  Communal storage is not empty
```

### DIFF
--- a/pkg/events/event.go
+++ b/pkg/events/event.go
@@ -34,7 +34,7 @@ const (
 	ReviveOrderBad                  = "ReviveOrderBad"
 	ObjectNotFound                  = "ObjectNotFound"
 	CommunalCredsWrongKey           = "CommunalCredsWrongKey" //nolint:gosec
-	CommunalEndpointIssue           = "S3EndpointIssue"
+	CommunalEndpointIssue           = "CommunalEndpointIssue"
 	S3BucketDoesNotExist            = "S3BucketDoesNotExist"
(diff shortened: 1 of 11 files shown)


pkg/events/event.go
```diff
@@ -34,7 +34,7 @@ const (
 	ReviveOrderBad                  = "ReviveOrderBad"
 	ObjectNotFound                  = "ObjectNotFound"
 	CommunalCredsWrongKey           = "CommunalCredsWrongKey" //nolint:gosec
-	S3EndpointIssue                 = "S3EndpointIssue"
+	CommunalEndpointIssue           = "S3EndpointIssue"
```
roypaulin, line 37: Shouldn't you also change the value of the constant?

PR #465, opened 2023-07-25
Implemented VClusterOps.AddNode() 

roypaulin:
Adds the logic for using VClusterOps library to add nodes to a subcluster.

### DIFF
--- a/pkg/vadmin/opts/addnode/opts.go
+++ b/pkg/vadmin/opts/addnode/opts.go
@@ -26,7 +26,7 @@ type Parms struct {
 	Hosts         []string
 	Subcluster    string
 	// will be removed after VER-88096
-	Nodes map[string]string
+	VNodeToHostMap map[string]string
 }
 
 type Option func(*Parms)
@@ -60,8 +60,8 @@ func WithSubcluster(subcluster string) Option {
 	}
 }
 
-func WithNodes(nodes map[string]string) Option {
+func WithVNodeToHostMap(vNodeToHostMap map[string]string) Option {
 	return func(s *Parms) {
-		s.Nodes = nodes
+		s.VNodeToHostMap = vNodeToHostMap
 	}
 }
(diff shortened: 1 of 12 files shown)


roypaulin:
I expect the tests to fail until vcluster is updated.

roypaulin:
If #464  is merged before this one, I will add events for add_node.

spilchen:
> I expect the tests to fail until vcluster is updated.

vcluster is updated now...

pkg/vadmin/opts/addnode/opts.go
```diff
@@ -25,6 +25,8 @@ type Parms struct {
 	InitiatorIP   string
 	Hosts         []string
 	Subcluster    string
+	// will be removed after VER-88096
+	Nodes map[string]string
```
spilchen, line 29: It may be temporary, but it would be good to include a more descriptive name. How about renaming it to something like this?
```suggestion
	VNodesToHostMap map[string]string
```

pkg/vadmin/opts/addnode/opts.go
```diff
@@ -57,3 +59,9 @@ func WithSubcluster(subcluster string) Option {
 		s.Subcluster = subcluster
 	}
 }
+
+func WithNodes(nodes map[string]string) Option {
```
spilchen, line 63: If we end up renaming `s.Nodes`, lets change this function name too.

spilchen (approved):
This looks good. There is some code in `prepLocalData` that is related to add node that I would like to see removed if we can:
```
	if vdb.IsDepotVolumePersistentVolume() && vdb.Spec.Local.IsDepotPathUnique() {
		rmCmds.WriteString(fmt.Sprintf("sudo chown dbadmin:verticadba -R %s/%s", paths.LocalDataPath, vdb.GetPVSubPath("depot")))
	}
```

I'll open a follow-on ticket to get that part cleaned up if we can.

PR #467, opened 2023-07-26
Log successful add nodes in the event log

spilchen:
This fixes a small regression in this release. We were not logging an event message when add nodes was successful.

Also, moving the pod facts invalidation for vclusterOps. Where it was before meant that we would always invalidate the pod facts cache, even for the case where the install reconciler does nothing.

### DIFF
--- a/pkg/controllers/vdb/install_reconciler.go
+++ b/pkg/controllers/vdb/install_reconciler.go
@@ -92,8 +92,6 @@ func (d *InstallReconciler) installForVClusterOps(ctx context.Context) (ctrl.Res
 		d.Log.Info("Requeue reconcile cycle because not all nodes have done the install for vclusterOps")
 		return ctrl.Result{Requeue: true}, nil
 	}
-	// Invalidate the pod facts cache since its out of date due to the install
-	d.PFacts.Invalidate()
 	return ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/vdb/install_reconciler.go
```diff
@@ -92,8 +92,6 @@ func (d *InstallReconciler) installForVClusterOps(ctx context.Context) (ctrl.Res
 		d.Log.Info("Requeue reconcile cycle because not all nodes have done the install for vclusterOps")
 		return ctrl.Result{Requeue: true}, nil
 	}
-	// Invalidate the pod facts cache since its out of date due to the install
```
roypaulin, line 95: Good catch!

pkg/controllers/vdb/install_reconciler.go
```diff
@@ -224,6 +222,8 @@ func (d *InstallReconciler) generateHTTPCerts(ctx context.Context) (bool, error)
 			if err != nil {
 				return false, errors.Wrap(err, fmt.Sprintf("failed to copy %s to the pod %s", fname, p.name))
 			}
+			// Invalidate the pod facts cache since its out of date due the https generation
+			d.PFacts.Invalidate()
```
roypaulin, line 226: Do you also want to invalidate podfacts here for admintools?

spilchen: It's not needed for AT because we invalidate elsewhere. But I don't think it hurts to invalidate here.

roypaulin (approved):

PR #473, opened 2023-07-31
Avoid using SQL to configure database after create

spilchen:
Traditionally when we create a database, we have run SQL commands to initialize certain aspects of the database. We were able to do this because the admintools -t create_db API has an option that let you specify an SQL file to run. With the vclusterOps integration, it doesn't have such a feature. Instead, the preferred way is to pass everything as config parameters. This takes advantage of new config parameters being added in 23.4.0. 

The new config parameters are:
- InitialDefaultSubclusterName: Controls the name of the initial default subcluster for new databases only. This has no effect if changed after the database has been created. The operator previously controlled this by doing an ALTER SUBCLUSTER SQL right after create
- InitialPreferredKSafe: Controls the initial value for preferred k-safe for new databases. This has no effect if changed after the database has been created. The operator previously controlled this by using the 'select set_preferred_ksafe()' meta-function after the database was created.

Note: EncryptSpreadComm was enabled via SQL, but a config parm for it always existed. However, it can only be set as a config parameter during create db on versions 23.3.0 or newer. So, that will continue to be enabled with SQL on older versions.

We cannot set EncryptSpreadComm=vertica for databases created with vclusterOps. There is a bug that is preventing the nodes from coming up. An internal ticket has been opened to fix the bug. In the mean time, I have removed the use of that setting the vcluster e2e testcases.


api/v1beta1/verticadb_types.go
```diff
@@ -1384,3 +1384,15 @@ func (v *VerticaDB) IsKnownDepotVolumeType() bool {
...
+// getFirstPrimarySubcluster returns the first primary subcluster defined in the vdb
+func (v *VerticaDB) GetFirstPrimarySubcluster() *Subcluster {
+	for i := range v.Spec.Subclusters {
+		sc := &v.Spec.Subclusters[i]
+		if sc.IsPrimary {
+			return sc
+		}
+	}
+	// We should never get here because the webhook prevents a vdb with no primary.
+	return &v.Spec.Subclusters[0]
```
roypaulin, line 1397: Why not return nil?

pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -160,6 +137,35 @@ func (c *CreateDBReconciler) preCmdSetup(ctx context.Context, initiatorPod types
...
+	// version, we can skip the remainder of the function.
+	vinf, ok := c.Vdb.MakeVersionInfo()
+	if ok && vinf.IsEqualOrNewer(vapi.DBSetupConfigParameters) {
+		return ctrl.Result{}, nil
+	}
+
+	// We include SQL to rename the default subcluster to match the name of the
+	// first subcluster in the spec -- any remaining subclusters will be added
+	// by DBAddSubclusterReconciler.
+	sc := c.getFirstPrimarySubcluster()
```
roypaulin, line 152: Can we add the remainder to a function?

roypaulin (commented):
Looks good! I just have a few comments.

api/v1beta1/verticadb_types.go
spilchen: This was copied over from some other code. But I think the rationale was it would avoid a panic in case someone did create a vdb without a primary. I'll change it to nil as I'd imagine other bad things would happen if you didn't have a primary.

pkg/controllers/vdb/createdb_reconciler.go
spilchen: Yep, makes sense

roypaulin (approved):

PR #474, opened 2023-07-31
Allow use of GSM for superuser password secret

sivaalamp:
This is similar to #458 except we read from GSM for the superuser password secret. We use the same feature flag as before. Here is a sample CR to enable it.

```
kind: VerticaDB
metadata:
  name: v
  annotations:
    vertica.com/use-gcp-secret-manager: "true" 
spec:
  superuserPasswordSecret: <full name of the secret in GSM>
...
```

The name of the GSM secret is taken from the existing `spec.superuserPasswordSecret` field.

Note, we no longer use the canary query for health probes as that depends on the superuser password being a k8s secret. Instead, we check if vertica is running by seeing if anyone is listening on the vertica client port.

### DIFF
--- a/changes/unreleased/Added-20230731-202319.yaml
+++ b/changes/unreleased/Added-20230731-202319.yaml
@@ -1,0 +1,5 @@
+kind: Added
+body: Allow use of GSM for superuser password secret
+time: 2023-07-31T20:23:19.559653595-03:00
+custom:
+  Issue: "474"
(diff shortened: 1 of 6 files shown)


spilchen:
@sivaalamp thanks for your contribution. I don't normally do this, but since we are short on time I opted to apply the code review comments myself. This will shorten the review cycle and allow us to get it in QA hands sooner.

spilchen (approved):

PR #479, opened 2023-08-11
Implemented VClusterOps.RemoveNode()

roypaulin:
Adds the logic for using VClusterOps library to remove nodes from a
subcluster.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -45,8 +45,6 @@ github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
 github.com/DATA-DOG/go-sqlmock v1.5.0 h1:Shsta01QNfFxHCfpW6YH2STWB0MudeXXEWMr20OEh60=
 github.com/DATA-DOG/go-sqlmock v1.5.0/go.mod h1:f/Ixk793poVmq4qj/V1dPUg2JEAKC73Q5eFN3EC/SaM=
-github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbtSwDGJws/X0=
-github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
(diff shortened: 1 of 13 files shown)


go.sum
```diff
@@ -123,6 +125,7 @@ github.com/go-openapi/swag v0.19.5/go.mod h1:POnQmlKehdgb5mhVOsnJFsivZCEZ/vjK9gh
 github.com/go-openapi/swag v0.19.14 h1:gm3vOOXfiuw5i9p5N9xJvfjvuofpyvLA9Wr6QfK5Fng=
 github.com/go-openapi/swag v0.19.14/go.mod h1:QYRuS/SOXUCsnplDa677K7+DxSOj6IPNl/eQntq43wQ=
 github.com/go-stack/stack v1.8.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/melR3HDY=
+github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0/go.mod h1:fyg7847qk6SyHyPtNmDHnmrv/HOrqktSC+C9fM+CJOE=
```
spilchen, line 128: Did you run `go mod tidy`?

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -241,7 +241,7 @@ func (o *ObjReconciler) checkForDeletedSubcluster(ctx context.Context) (ctrl.Res
 		// all pods in the subcluster.  If that isn't the case, we requeue to
 		// give those reconcilers a chance to do those actions.  Failure to do
```
spilchen, line 242: Can we update the comment? I would like to stop referring to admintools.conf. I think we can just remove the last sentence.

pkg/controllers/vdb/uninstall_reconciler.go
```diff
@@ -121,31 +123,16 @@ func (s *UninstallReconciler) uninstallPodsInSubcluster(ctx context.Context, sc
...
 		}
-
-		// Remove the installer indicator file so that we do an install if we then
-		// opt to scale out again.
-		cmd := s.genCmdRemoveInstallIndicator()
-		for _, pod := range podsToUninstall {
-			if _, _, err := s.PRunner.ExecInPod(ctx, pod.name, names.ServerContainer, cmd...); err != nil {
-				return ctrl.Result{}, fmt.Errorf("failed to call remove installer indicator file: %w", err)
-			}
+		if err != nil {
```
spilchen, line 134: We should use IsReconcileAborted
```suggestion
		if verrors.IsReconcileAborted(res, err) {
```

pkg/controllers/vdb/uninstall_reconciler.go
```diff
@@ -121,31 +123,16 @@ func (s *UninstallReconciler) uninstallPodsInSubcluster(ctx context.Context, sc
...
-		defer os.Remove(atConfTempFile)
+		var result ctrl.Result
+		var err error
 
-		if err := distributeAdmintoolsConf(ctx, s.Vdb, s.VRec, s.PFacts, s.PRunner, atConfTempFile); err != nil {
-			return ctrl.Result{}, err
+		if vmeta.UseVClusterOps(s.Vdb.Annotations) {
+			result, err = s.uninstallPodsInSubclusterForVClusterOps(ctx, podsToUninstall)
+		} else {
+			result, err = s.uninstallPodsInSubclusterForAdmintools(ctx, podsToUninstall)
```
spilchen, line 132: I think we should move this behind the vadmin/dispatcher abstraction. I would like all admintools stuff only in that package. I know we don't do that for install, but it's something I would like to change too. vadmin can't be given a PodFact, but I think we only need the IP and pod name from them, so hopefully it isn't too bad to create options for those.

pkg/controllers/vdb/uninstall_reconciler.go
```diff
@@ -156,6 +143,54 @@ func (s *UninstallReconciler) uninstallPodsInSubcluster(ctx context.Context, sc
...
 }
 
+// uninstallPodsInSubclusterForVClusterOps will call uninstall, for vclusterops, on a list
+// of pods that will be scaled down.
+func (s *UninstallReconciler) uninstallPodsInSubclusterForVClusterOps(ctx context.Context,
+	podsToUninstall []*PodFact) (ctrl.Result, error) {
+	cmd := s.genCmdRemoveHTTPTLSConfFile()
+	for _, pod := range podsToUninstall {
+		if _, _, err := s.PRunner.ExecInPod(ctx, pod.name, names.ServerContainer, cmd...); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to call remove installer indicator file: %w", err)
```
spilchen, line 153: Lets update the error message as its talking about the "installer indicator"

pkg/controllers/vdb/uninstall_reconciler.go
roypaulin: We need more than just ip and pod name for admintools. There are some functions like `distributeAdmintoolsConf()` or `findATBasePod()` that depend on PodFacts. Given that vadmin can't be given a PodFact, in order to move the whole thing under vadmin/dispatcher abstraction, we must refactor and move those functions to vadmin/at.go and has to be done simultaneously on `install` as it shares some of those functions with `uninstall`.

It can be done on this PR but I personally suggest to do it separately.

roypaulin: Yes I will open a ticket.

spilchen (approved):

PR #483, opened 2023-08-18
Implement VClusterOps.RestartNodes

spilchen:
This integration incorporates the vclusterops restart nodes functionality into the operator.

To get restart to work, I had to change the re-ip process. There were cases where, when using admintools, we had to re-ip hosts that had been installed but not yet added to the database. This was necessary to keep the hosts' IP addresses up to date in admintools.conf. This step is not necessary for vclusterops because there is no state file that needs to be synchronized.

### DIFF
--- a/pkg/vadmin/restart_node_vc.go
+++ b/pkg/vadmin/restart_node_vc.go
@@ -19,6 +19,10 @@ import (
 	"context"
 	"fmt"
 
+	vops "github.com/vertica/vcluster/vclusterops"
+	"github.com/vertica/vcluster/vclusterops/vstruct"
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/net"
 	"github.com/vertica/vertica-kubernetes/pkg/vadmin/opts/restartnode"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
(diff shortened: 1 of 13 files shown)


pkg/vadmin/restart_node_vc.go
```diff
@@ -28,7 +32,43 @@ import (
...
+	catPath := v.VDB.Spec.Local.GetCatalogPath()
+	su := vapi.SuperUser
+	honorUserInput := true
+	opts := vops.VRestartNodesOptions{
+		DatabaseOptions: vops.DatabaseOptions{
+			Name:           &v.VDB.Spec.DBName,
+			RawHosts:       []string{s.InitiatorIP},
+			Ipv6:           vstruct.MakeNullableBool(net.IsIPv6(s.InitiatorIP)),
+			IsEon:          vstruct.MakeNullableBool(v.VDB.IsEON()),
+			CatalogPrefix:  &catPath,
```
roypaulin, line 63: Do we still have to pass some of these options from db even though we get them from db in vclusterops?

pkg/vadmin/restart_node_vc.go
```diff
@@ -51,16 +49,12 @@ func (v *VClusterOps) RestartNode(ctx context.Context, opts ...restartnode.Optio
...
 
 func (v *VClusterOps) genRestartNodeOptions(s *restartnode.Parms, certs *HTTPSCerts) *vops.VRestartNodesOptions {
-	catPath := v.VDB.Spec.Local.GetCatalogPath()
 	su := vapi.SuperUser
 	honorUserInput := true
 	opts := vops.VRestartNodesOptions{
 		DatabaseOptions: vops.DatabaseOptions{
 			Name:           &v.VDB.Spec.DBName,
 			RawHosts:       []string{s.InitiatorIP},
-			Ipv6:           vstruct.MakeNullableBool(net.IsIPv6(s.InitiatorIP)),
```
roypaulin: I think we still need this option.

roypaulin (approved):

PR #484, opened 2023-08-21
Update VClusterOps.AddNode

roypaulin:
Now that we have a better way of getting node info from vclusterops, some refactoring was needed is order to get rid of some options that are not used anymore.


spilchen (commented):
Looks good. No major complaints. I guess you'll need to update this once the dependent vcluster change gets merged and sync'd.

PR #488, opened 2023-08-31
Implemented VClusterOps.RemoveSubcluster()

roypaulin:
Adds the logic for using VClusterOps library to remove a subcluster.

### DIFF
--- a/pkg/vadmin/remove_sc_vc.go
+++ b/pkg/vadmin/remove_sc_vc.go
@@ -17,15 +17,70 @@ package vadmin
 
 import (
 	"context"
-	"fmt"
+	"errors"
 
+	"github.com/vertica/vcluster/rfc7807"
+	vops "github.com/vertica/vcluster/vclusterops"
+	"github.com/vertica/vcluster/vclusterops/vstruct"
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/net"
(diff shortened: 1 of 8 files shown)


pkg/vadmin/remove_sc_vc.go
```diff
@@ -17,15 +17,70 @@ package vadmin
...
+
+func (v *VClusterOps) genRemoveSubclusterOptions(s *removesc.Parms, certs *HTTPSCerts) vops.VRemoveScOptions {
+	opts := vops.VRemoveScOptionsFactory()
+
+	// required options
+	opts.Name = &v.VDB.Spec.DBName
+	opts.SubclusterToRemove = &s.Subcluster
+
+	*opts.HonorUserInput = true
+	opts.RawHosts = append(opts.RawHosts, s.InitiatorIP)
```
spilchen, line 68: Is `opts.RawHosts` empty prior to this? Based on the comments in the PR, it sounds like we also will remove nodes, but I believe this is handled by a prior step in the reconcile. If so, I suggest we update the PR description and make this change to clarify this:

```suggestion
	opts.RawHosts = []string{s.InitiatorIP}
```

roypaulin: `remove_sc` also removes the nodes belonging to the sc to removed that's what I meant. But I will change it to just say `to remove a subcluster`

spilchen (approved):

PR #489, opened 2023-09-01
Remove NO_KEYS from v2 image

spilchen:
The v2 image will always be treated as the NO_KEYS version. Removing the option for checking the environment variable in that image. Treating all spots as if NO_KEYS=YES.

Also, you can now add additional arguments to the docker build for the vertica-k8s image by setting the VERTICA_ADDITIONAL_DOCKER_BUILD_OPTIONS environment variable.

For example:
```
export VERTICA_ADDITIONAL_DOCKER_BUILD_OPTIONS="--platform linux/amd64"
make docker-build-vertica
```

### DIFF
--- a/docker-vertica-v2/Dockerfile
+++ b/docker-vertica-v2/Dockerfile
@@ -66,7 +66,7 @@ RUN set -x \
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN set -x \
   && chown -R dbadmin:verticadba /home/dbadmin/ \
-  && rm -rf /opt/vertica/config/share/agent*;
+  && rm -rf /opt/vertica/config/share/agent*
 
 ##############################################################################################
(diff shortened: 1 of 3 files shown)


docker-vertica-v2/Dockerfile
```diff
@@ -73,9 +66,7 @@ RUN set -x \
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN set -x \
   && chown -R dbadmin:verticadba /home/dbadmin/ \
-  && if [[ ${NO_KEYS^^} == "YES" ]] ; then \
-    rm -rf /opt/vertica/config/share/agent*; \
-  fi
+  && rm -rf /opt/vertica/config/share/agent*;
```
roypaulin, line 69: It does not change anything but can you remove `;`

roypaulin (approved):
Looks good

PR #492, opened 2023-09-07
Add env var to run e2e tests with vclusterops

spilchen:
This adds a new environment variable that will control whether e2e tests are run with vclusterops enabled or not. This is controlled through kustomize. We can use this in the future to run a testcase in either deployment methods.

The default is to run with vclusterops disabled. To run with it enabled, run the following:
```
export VERTICA_DEPLOYMENT_METHOD=vclusterops
scripts/setup-kustomize.sh
<run your test>
```

Note, this works if an annotation is already present in the VerticaDB. I had to use some weird kustomize syntax to avoid clobbering other annotations.

fyi @roypaulin @cchen-vertica


cchen-vertica (approved):
Thanks for making this change.

roypaulin:
Does it append the vclusterops annotation if there is another annotation?

spilchen:
Yes it does. I followed the pattern as outlined in here:
https://github.com/kubernetes-sigs/kustomize/issues/1439

PR #495, opened 2023-09-08
Implement VClusterOps.DescribeDB

spilchen:
This adds to the vclusterops revive support by doing a prepass to collect the cluster_config.json, then modifying local paths to match the cluster config. We collect the cluster_config.json using the VReviveDB call with the DisplayOnly option. This pre-existed for admintools based deployments, but now we are hooking it up for vclusterops. 

The format for vclusterops is slightly different. So, I updated the reviveplanner package to have an admintools based parser and a vclusterops based parser.

Both parser implement a new interface: ClusterConfigParser. The admintools parser was moved to the pkg/reviveplanner/atparser package. The vclusterops parser was added in the pkg/reviveplanner/vcparser package.

The reviveplanner core logic works with either parser and was renamed from ATPlanner to Planner to show that it isn't tied to the admintools output.


pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -115,7 +115,8 @@ func (c *CreateDBReconciler) execCmd(ctx context.Context, initiatorPod types.Nam
 
 // preCmdSetup will generate the file we include with the create_db.
 // This file runs any custom SQL for the create_db.
-func (c *CreateDBReconciler) preCmdSetup(ctx context.Context, initiatorPod types.NamespacedName, podList []*PodFact) (ctrl.Result, error) {
+func (c *CreateDBReconciler) preCmdSetup(ctx context.Context, initiatorPod types.NamespacedName,
+	initiatorIP string, podList []*PodFact) (ctrl.Result, error) {
```
roypaulin, line 119: Where is `initiatorIP` used in `preCmdSetup`?

pkg/reviveplanner/vcparser/vc_parser.go
```diff
@@ -0,0 +1,109 @@
...
+// format of these should only change if a corresponding change is done in the
+// server. There are many fields in the cluster_config.json. We elected to only
+// include the fields that are needed by the revive planner.
+
+// Cluster is the top level JSON doc for the output for vcluster's ReviveDB
+// output when DisplayOnly is used.
+type Cluster struct {
+	Database        Database           `json:"Database"`
+	ShardCount      int                `json:"ShardCount"`
+	StorageLoctions []StorageLocations `json:"StorageLocation"`
```
roypaulin, line 39:
```suggestion
	StorageLocations []StorageLocations `json:"StorageLocation"`
```

pkg/reviveplanner/vcparser/vc_parser.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"encoding/json"
+	"path/filepath"
+
+	"github.com/vertica/vertica-kubernetes/pkg/reviveplanner/util"
+)
+
+type Parser struct {
+	Cluster Cluster
```
roypaulin, line 26: Why isn't there a `Log` like in admintools'?

roypaulin (commented):
Looks good!

pkg/controllers/vdb/createdb_reconciler.go
spilchen: `preCmdSetup` is part of an interface, so we need the parm. If we upgrade golangci-lint, we can enabled a linter that will pick this thing up. They recommend using `_` as the variable name. I can rename the parameter to that and upgrade the linter as a follow-on task.

pkg/reviveplanner/vcparser/vc_parser.go
spilchen: We had no reason to log a message (yet), so I never added it.

roypaulin (approved):

PR #496, opened 2023-09-08
Log event message for revive cluster lease error

spilchen:
This applies to the vclusterops implementation of revive. We will now log an event message if we hit an cluster lease error. We use the new ClusterLeaseNotExpired concrete type to know if we hit this error. This gets us closer in sync with the admintools handling of specific errors (see aterrors.go)

### DIFF
--- a/pkg/vadmin/vc_errors.go
+++ b/pkg/vadmin/vc_errors.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/vertica/vcluster/rfc7807"
+	"github.com/vertica/vcluster/vclusterops"
 	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
 	"github.com/vertica/vertica-kubernetes/pkg/events"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/vadmin/vc_errors.go
```diff
@@ -65,8 +78,16 @@ func (v *vcErrors) LogFailure(cmd string, err error) (ctrl.Result, error) {
...
-		return ctrl.Result{}, fmt.Errorf("failed command %s: %w", cmd, err)
+		return ctrl.Result{}, fmt.Errorf("failed command %s: %w", cmd, vproblem)
 	}
 	// All known errors we return with requeue set to true.
 	return ctrl.Result{Requeue: true}, nil
 }
+
+func (v *vcErrors) logClusterLeaseNotExpiredError(err *vclusterops.ClusterLeaseNotExpiredError) (ctrl.Result, error) {
+	v.Log.Info("vclusterOps command failed because the cluster lease was not expired", "msg", err.Error())
+	v.EVWriter.Eventf(v.VDB, corev1.EventTypeWarning, events.ReviveDBClusterInUse,
```
roypaulin, line 89: What about `ReviveDBNodeCountMismatch`, `ReviveDBPermissionDenied`, `ReviveDBNotFound`, are we going to get them from server through Rfc7807?

spilchen: It depends on where those errors originate. It's likely the last two originate from the https endpoint, so they will be rfc7807 errors. Not sure where the first one is detected.

spilchen: Some more details where these errors originate. The first one is detected in vclusterops, so we can return a specific error struct for that. I'll log this in the jira's we have for this work so we don't lose track.

ReviveDBNodeCountMismatch
[NMADownloadFileOp] nodes mismatch found on host 10.244.0.184: the number of the new nodes in --hosts is 4, but the number of the old nodes in description file is 3\ncould not find a host with a passing result"

ReviveDBPermissionDenied
[NMADownloadFileOp] HTTPS call failed on host 10.244.0.222
status code 500 returned from host 10.244.0.222: Error happened during executing vertica-download-file. Stdout: . Stderr:Initialization thread logged exception:
Could not copy file [s3://spilly/here/metadata/vertdb/cluster_config.json] to [/tmp/desc.json]: curlCode: 28, Timeout was reached                                                             Exiting process ( exit(1) ).. Return code: 1. Complete error info: exit status 1 

ReviveDBNotFound
[NMADownloadFileOp] HTTPS call failed on host 10.244.0.196\nstatus code 500 returned from host 10.244.0.196: Error happened during executing vertica-download-file. Stdout: . Stderr:Initialization thread logged exception:\nCould not copy file [/communal/0ae3e837-3167-402f-a8f7-316ffdfd65f3z/metadata/vertdb/cluster_config.json] to [/tmp/desc.json]: No such file or directory\nExiting process ( exit(1) ).. Return code: 1. Complete error info: exit status 1\n\ncould not find a host with a passing result"

roypaulin: The last 2 are from the nma right? not the embedded https server.

spilchen: The nma just copies out the result from the call to vertica-donwload-file. So, ultimately somewhere in the server this error is generated.

PR #502, opened 2023-09-12
Remove admintools from v2 vertica-k8s container

roypaulin:
In the v2 container, since we use vclusterops to do admin commands, we can remove admintools and its support applications and libraries (vbr, agent, scrutinize, python site packages).


roypaulin:
I updated github ci to test all e2e-udx but looks like for now we do not use the image that is built, but I can see that e2e-java failed. I wonder if you have already run e2e-udx on your current task?

scripts/setup-kustomize.sh
```diff
@@ -73,6 +73,9 @@ then
   source $USER_CONFIG_FILE
 fi
 
+if [ ! -z "${VERTICA_DEPLOYMENT_METHOD_OVERRIDE}" ]; then
+    VERTICA_DEPLOYMENT_METHOD=$VERTICA_DEPLOYMENT_METHOD_OVERRIDE
+fi
```
spilchen, line 78: I'm not sure if I like this. I understand why you did this because my instructions for using the new deployment method didn't work. 

I think to fix the earlier work I did we should:

1. Update tests/kustomize-defaults.cfg with this change so we don't set the variable.
```
# The type of deployment to use when running vertica admin commands. Valid
# values are: admintools or vclusterops. Uncomment this or set this environment
# in your shell to control the deployment type.
# VERTICA_DEPLOYMENT_METHOD=
```

2. Update the above line with following:
```suggestion
if [ -z "${VERTICA_DEPLOYMENT_METHOD}" ]; then
    VERTICA_DEPLOYMENT_METHOD=admintools
fi
```
3. Export VERTICA_DEPLOYMENT_METHOD in the e2e-udx.yml file

.github/workflows/e2e-udx.yml
```diff
@@ -59,6 +66,7 @@ jobs:
         export OPERATOR_IMG=${{ inputs.operator-image }}
         export VLOGGER_IMG=${{ inputs.vlogger-image }}
         export E2E_TEST_DIRS=tests/e2e-udx
+        export VERTICA_DEPLOYMENT_METHOD_OVERRIDE=${{ inputs.vertica-deployment-method }}
```
spilchen, line 69: See my other comment, but I think with those other fixes this can be rewritten like:
```suggestion
        export VERTICA_DEPLOYMENT_METHOD=${{ inputs.vertica-deployment-method }}
```

docker-vertica-v2/Dockerfile
```diff
@@ -161,7 +162,8 @@ RUN set -x \
   && tar -C / -Jxpf /tmp/s6-overlay-noarch.tar.xz \
   # Create a symlink to the rsync for use with vbr. This works around a problem
   # seen in some deployments where vbr cannot find rsync.
-  && ln -s /opt/vertica/bin/rsync /usr/bin/rsync
+  && ln -s /opt/vertica/bin/rsync /usr/bin/rsync \
+  && sh /tmp/cleanup.sh
 
 ENTRYPOINT [ "/init" ]
```
spilchen, line 169: Can you remove the `EXPOSE` for the agent port a few lines below?

spilchen, line 169: Can you update the LABEL a few lines below? The one `org.opencontainers.image.source` should point to the correct path. Also, we should add a label to indicate the deployment method (we would need one in the admintools image as well). Perhaps `vertica-deployment-method`?

docker-vertica-v2/packages/cleanup.sh
```diff
@@ -36,7 +36,7 @@ rm -r -f \
    /opt/vertica/oss/python*/lib/python*/idlelib
 
 # cleanup all test directories for packages under site-package
```
spilchen, line 38: Can we update this comment since we are not just deleting test now.

docker-vertica-v2/packages/cleanup.sh
```diff
@@ -36,7 +36,7 @@ rm -r -f \
    /opt/vertica/oss/python*/lib/python*/idlelib
```
spilchen, line 36: We can probably remove this line as it is redundant:
```
/opt/vertica/oss/python*/lib/python*/site-packages/pip*
```

docker-vertica-v2/packages/cleanup.sh
```diff
@@ -36,7 +36,7 @@ rm -r -f \
...
 
 # cleanup all test directories for packages under site-package
-find /opt/vertica/oss/python*/lib/python*/site-packages/ -type d -name "*[Tt]est" -exec rm -rf {} +
+rm -rf /opt/vertica/oss/python*/lib/python*/site-packages
```
spilchen, line 39: Can we include this with the larger `rm -r -f` above?

docker-vertica-v2/Dockerfile
```diff
@@ -101,6 +101,7 @@ ENV PATH "$PATH:/opt/vertica/bin:/opt/vertica/sbin"
 ENV DEBIAN_FRONTEND noninteractive
 
 COPY ./packages/init.d.functions /etc/rc.d/init.d/functions
+COPY ./packages/cleanup-initial.sh /tmp/cleanup.sh
```
spilchen, line 104: Can we retain the name of the file that we copy into the image? I think that helps clarity, especially since we already have a cleanup.sh.

docker-vertica-v2/packages/cleanup.sh
```diff
@@ -66,8 +66,21 @@ rm -rf \
...
+# removing admintools and supported libraries(vbr, agent, scrutinize)
+rm -rf \
+    /opt/vertica/bin/vbr* \
+    /opt/vertica/share/vbr \
+    /opt/vertica/bin/scrutinize \
+    /opt/vertica/agent \
+    /opt/vertica/config/logrotate/agent.logrotate \
+    /opt/vertica/sbin/vertica_agent* \
+    /opt/vertica/config/admintools* \
+    /opt/vertica/bin/admintools \
```
spilchen, line 80: A few other things that I think we can remove:
- `/opt/vertica/bin/adminTools`
- `/opt/vertica/sbin/update_vertica`
- `/opt/vertica/sbin/install_vertica`
- `/opt/vertica/config/apikeys.dat`

spilchen (commented):
This looks good. Thanks for the CI update. I guess you needed to do that to verify udx's with your new image.

spilchen:
> I updated github ci to test all e2e-udx but looks like for now we do not use the image that is built, but I can see that e2e-java failed. I wonder if you have already run e2e-udx on your current task?

This is a good point. We don't test with the v2 image we build because the RPM that we download is too old -- it doesn't the latest nma/https changes. I can build your image and push it to our private dockerhub account so that we can run it in the CI.

spilchen:
Try running the e2e test suite with vertica/vertica-k8s-private:roy-20230913-1-v2. The CI will have authority to pull this image.

roypaulin:
Can you push a new image?

spilchen:
Sure, try this image: vertica/vertica-k8s-private:roy-20230915-1-v2

.github/workflows/build-images.yml
```diff
@@ -285,7 +285,7 @@ jobs:
       uses: spilchen/switch-case-action@v2
       id: v2_vertica_image
       with:
-        default: docker.io/vertica/vertica-k8s-private:latest-master
+        default: docker.io/vertica/vertica-k8s-private:roy-20230915-1-v2
```
spilchen, line 288: We should back out this change.

spilchen (commented):
Just had one minor comment.

spilchen (approved):

PR #503, opened 2023-09-12
Enable e2e tests suites with vclusterops

roypaulin:
We are now starting to run  some e2e tests suites with vclusterops too. Some tests failed due to vclusterops bugs and they were moved to separate folders to be run with admintools only until the bugs are fixed.


.github/workflows/e2e.yml
```diff
@@ -110,9 +112,22 @@ jobs:
     secrets:
       DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
       DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+  
+  e2e-s3-vcluster:
+    if: ${{ ! github.event.pull_request.head.repo.fork  && (inputs.e2e_test_suites == 'vcluster s3' || inputs.e2e_test_suites == 'leg 2 vcluster' || inputs.e2e_test_suites == '')}}
```
spilchen, line 117: Why do we check for two test suites?
```suggestion
    if: ${{ ! github.event.pull_request.head.repo.fork  && (inputs.e2e_test_suites == 'vcluster s3' || inputs.e2e_test_suites == '')}}
```

.github/workflows/e2e-leg-2.yml
```diff
@@ -33,6 +36,10 @@ on:
         description: 'Name of the vertica server image'
         type: string
         required: false
+      vertica-deployment-method:
+        description: 'Back-end library(admintools|vclusterops)'
+        type: string
+        required: false
```
spilchen, line 42: minor: I suggest we use less generic language and use a choice. Same comment in the other workflow file.
```suggestion
      vertica-deployment-method:
        description: 'Vertica's deployment method'
        type: choice
        required: false
        default: admintools
        options:
        - admintools
        - vclusterops
```

spilchen (commented):
I like this approach. Just had some minor comments.

.github/workflows/e2e.yml
```diff
@@ -38,15 +38,17 @@ on:
...
+        - admintools leg 2
+        - admintools server upgrade
+        - admintools operator upgrade
+        - admintools udx
+        - admintools hostpath
+        - admintools enterprise 
+        - admintoold http server
+        - vcluster s3
+        - vcluster leg 2
         - vcluster
```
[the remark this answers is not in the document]
roypaulin: Yes, right after this one.

.github/workflows/e2e-leg-2.yml
spilchen, line 1: Since we will end up calling this workflow twice, can we chose a different artifact name in case of errors? If legs fail twice, one artifact will clobber the other one.

spilchen (approved):
Looks good.

PR #505, opened 2023-09-13
log event message for revive_db node count mismatch

cchen-vertica:
This checked revive_db node count mismatch error from vclusterops. When we get this kind of error, we will log the event "ReviveDBNodeCountMismatch".


spilchen (approved):
Looks good to me. No comments!

PR #507, opened 2023-09-14
Store name of NMA cert in environment variable

spilchen:
This is prep for when the NMA will directly read the cert from the secret store. We store the namespace and name of the secret in these environment variables: NMA_SECRET_NAMESPACE, NMA_SECRET_NAME

This applies only to the vclusterops based deployments.

This is for jira VER-88251

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -54,6 +54,10 @@ const (
 	CatalogPathEnv  = "CATALOG_PATH"
 	DepotPathEnv    = "DEPOT_PATH"
 	DatabaseNameEnv = "DATABASE_NAME"
+
+	// Environment variables that are set when deployed with vclusterops
+	NMASecretNamespaceEnv = "NMA_SECRET_NAMESPACE"
+	NMASecretNameEnv      = "NMA_SECRET_NAME"
 )
 
 // BuildExtSvc creates desired spec for the external service.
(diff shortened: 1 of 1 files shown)


pkg/builder/builder.go
```diff
@@ -575,9 +579,17 @@ func makeServerContainer(vdb *vapi.VerticaDB, sc *vapi.Subcluster) corev1.Contai
...
+			// mounted in the container.
 			{Name: "NMA_ROOTCA_PATH", Value: fmt.Sprintf("%s/%s", paths.HTTPServerCertsRoot, paths.HTTPServerCACrtName)},
 			{Name: "NMA_CERT_PATH", Value: fmt.Sprintf("%s/%s", paths.HTTPServerCertsRoot, corev1.TLSCertKey)},
 			{Name: "NMA_KEY_PATH", Value: fmt.Sprintf("%s/%s", paths.HTTPServerCertsRoot, corev1.TLSPrivateKeyKey)},
+			// New model is for the NMA to read the secrets directly from k8s.
+			// We provide the secret namespace and name for this reason. Once
+			// implemented we no longer need to provide the above environment
+			// variables.
+			{Name: NMASecretNamespaceEnv, Value: vdb.ObjectMeta.Namespace},
+			{Name: NMASecretNameEnv, Value: vdb.Spec.HTTPServerTLSSecret},
```
roypaulin, line 592: Are we going to use this with GSM too?

spilchen: Yes, eventually. The secret name will have a secret path reference like sm:// to denote the secret should be retrieved from GSM.

roypaulin (approved):

PR #508, opened 2023-09-15
Remove agent support

roypaulin:
Some backdoor support was added to the operator for the agent. This PR now  removes it.
The following changes were made:
- Remove agent reconciler and and related logics
- Update the Dockerfile and other files used to build vertica images
- Move back shared-svc to e2e-leg-2


spilchen (approved):
Looks good. I added a changie entry for the changelog since we had previously documented that we added the backdoor support.

PR #514, opened 2023-09-21
Change scope of operator from namespace to cluster

spilchen:
This changes the scope of the operator from being namespace scoped to cluster scoped. 

This is quite a big change, especially in the e2e tests. Here are some of the highlights:
- You can still can use make target to deploy/undeploy the operator as usual. The change is the location of the operator. It is being deployed in the verticadb-operator namespace by default, but you can override if you set the NAMESPACE variable.
- We no longer support DEPLOY_WITH=random. You need to specify what type of deployment you want (helm or olm). One of the test legs (e2e-leg-2) was converted to run olm tests. The rest of the legs use helm.
- Since a single operator can control many VerticaDB instances, I use additional concurrency when creating the operator. By default I use 5 go routines meaning we can work on 5 different VerticaDB CRs at the same time. This can be configured through new helm chart values.
- I got rid of some helm chart parameter that attempt to deal with allow non-cluster admins to install the operator. This is no longer needed because we can now deploy the operator once for the entire cluster. Gone are the settings for serviceNameOverride and skipRoleAndRoleBindingCreation. These are done unconditionally now when deploying the operator.
- One challenge I had to deal with was how to get access to operator config settings. Previously, the vertica pods had access to items from a ConfigMap that told us the version of the operator and how the operator was deployed (olm or helm). We still do this but because the ConfigMap is in a different namespace, we had to surface this in a different manner. The operator pod has this information as environment variables. When the operator creates the pods, it will look at the environment variables and create annotations based on their value. These annotations are then stored as files in a known location using the downward API; there is code in the server that fills in the dc_kubernetes_events table with information from these known locations.
- Another challenge was the operator creates the serviceAccount (and roles/rolebinding) for the vertica pod. Previously, we had just reused the same serviceAccount that the operator uses. But since serviceAccounts are namespace specific and the operator and the vertica pods can be in different namespaces, we needed another way. As a future item we will want the ability to specify a serviceAccount ahead of time. For now the operator will always generate one if it is missing.
- Cleaned up some of the e2e legs. Gone is the e2e-enterprise leg. And I added a new leg (e2e-leg-5) that will run all of the tests that need to start the operator in a specific mode. The rest of the legs depend on the operator to already be installed. Because of this, the new leg cannot run with any test parallelism. OLM tests, which we use to run at random will now only be run in e2e-leg-2. And I renamed the e2e-ezb test suite to e2e-leg-1 in the CI


pkg/controllers/vdb/serviceaccount_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+		VRec: vdbrecon,
+		Vdb:  vdb,
+		Log:  log.WithName("ServiceAccountReconciler"),
+	}
+}
+
+// Reconcile will ensure that a serviceAccount, role and rolebindings exists for
+// the vertica pods.
+func (s *ServiceAccountReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	rbacFinder := iter.MakeRBACFinder(s.VRec.Client, s.Vdb)
```
roypaulin, line 51: Can users still specify their own sa?

roypaulin:
Are you going to update README before the release. I guess doc work will follow.

roypaulin:
Will the make commands still work the same?

What about the e2e tests do we use one operator to control them all?

pkg/builder/labels_annotations.go
```diff
@@ -106,6 +107,15 @@ func MakeLabelsForSvcObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster, svcType st
 // included on new objects.
 func MakeAnnotationsForObject(vdb *vapi.VerticaDB) map[string]string {
 	annotations := make(map[string]string, len(vdb.Spec.Annotations))
+	// Surface operator config as annotations. This is picked up by the downward
+	// API and surfaced as files for the server to collect in the
+	// dc_kubernetes_events table.
+	if val, ok := os.LookupEnv("DEPLOY_WITH"); ok {
```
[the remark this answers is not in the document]
spilchen: No, it won't.

pkg/iter/sc_finder.go
```diff
@@ -144,17 +142,10 @@ func (m *SubclusterFinder) FindSubclusters(ctx context.Context, flags FindFlags)
 }
 
 // listObjectsOwnedByOperator will return all objects of a specific type that
-// are owned by the operator.  This includes objects like statefulsets or
-// service objects.  The type is derived from what kind of list is passed in.
-// We find objects the operator owns by using a set of labels that the operator
-// sets with each object it creates.
+// are owned by the operator.  This is just a wrapper to the common function
+// that implements this lookup.
 func (m *SubclusterFinder) listObjectsOwnedByOperator(ctx context.Context, list client.ObjectList) error {
```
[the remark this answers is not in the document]
spilchen: I'll remove it.

pkg/controllers/vdb/serviceaccount_reconciler.go
spilchen: For now they can't. But that was the follow-on story I mentioned in scrum.

spilchen:
> Are you going to update README before the release. I guess doc work will follow.

Good point. I can make some changes to the DEVELOPER.md. Doc will handle the public docs.

spilchen:
> Will the make commands still work the same?
> 
> What about the e2e tests do we use one operator to control them all?

Yes, none of the make commands change. You still do `make deploy` and `make undeploy` to bring up/down the operator. `make deploy` will start the operator in the verticadb-operator namespace by default.

The e2e tests had a major overhaul. Most of the testcases that deployed the operator simply removed that step. There was a class of tests that started the operator in specific modes. All of those were moved to a new leg (e2e-leg-5). That leg will assume each testcase will deploy and undeploy the cluster scoped operator.

DEVELOPER.md
```diff
@@ -99,7 +104,7 @@ To run Vertica in Kubernetes, we need to package Vertica inside a container. Thi
 By default, we create containers that are stored in the local docker daemon. The tag is either `latest` or, if running in a Kind environment, it is `kind`. You can control the container names by setting the following environment variables prior to running the make target.  
 
 - **OPERATOR_IMG**: Operator image name.
-- **VERTICA_IMG**: Vertica image name.
+- **VERTICA_IMG**: Vertica image name. Used interchangabely for v1 and v2 containers.
```
roypaulin, line 107: minor: interchangeably

roypaulin (approved):

PR #515, opened 2023-09-21
Deprecate the httpServerMode VerticaDB CR parameter

spilchen:
The operator used to handle the setup of Vertica's https service. It would create a config file during installation and a TLS secret to use for its certificates. It would also have a reconciler that would ensure the https service was running and would restart it if needed. However, all of this has been deprecated. The deployment method (vclusterops vs admintools) will now decide what setup is needed. 

The httpServerMode parameter used to control this, but it has been deprecated and all usage of it in the operator has been removed. 

The reconcile function to make sure that the embedded http service is running has also been removed. With vclusterops deployments, it must always be running. The health probes are planned to be changed to use an endpoint from the http service. So, if the service becomes unhealthy, the pod will be rescheduled. The operator will not need to do anything on its end to restart it. 

The e2e-http-server e2e leg now runs only with vclusterops and has been renamed to e2e-leg-6.


pkg/controllers/vdb/http_server.go
```diff
@@ -41,9 +39,3 @@ func hasCompatibleVersionForHTTPServer(vrec *VerticaDBReconciler, vdb *vapi.Vert
 	}
 	return true
 }
-
-// genHTTPServerCtrlQuery returns a query used to interact
-// with the http server.
-func genHTTPServerCtrlQuery(action string) string {
```
roypaulin, line 47: What do we need `httpServerMode ` for?

spilchen: Excellent question. I knew I had more to do when I opened this task. I added a commit that will deprecate that parameter. We now setup the TLS certs for nma solely based on the deployment method (vclusterops only). I had to move the e2e-http-server leg so its only run in vcluster mode. The leg was rename to e2e-leg-6 too.

roypaulin (approved):

PR #516, opened 2023-09-22
Set the JAVA_HOME environment variable in the Vertica v2 container

jizhuoyu:
Add ENV instruction in v2 vertica image Dockerfile to make JAVA_HOME env var available in all server containers

### DIFF
--- a/docker-vertica-v2/Dockerfile
+++ b/docker-vertica-v2/Dockerfile
@@ -98,6 +98,7 @@ COPY --from=builder /var/spool/cron/ /var/spool/cron/crontabs
 
 ENV PATH "$PATH:/opt/vertica/bin:/opt/vertica/sbin"
 ENV DEBIAN_FRONTEND noninteractive
+ENV JAVA_HOME "/usr"
 
 COPY ./packages/init.d.functions /etc/rc.d/init.d/functions
 COPY ./packages/cleanup-initial.sh /tmp/cleanup-initial.sh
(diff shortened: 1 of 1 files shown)


jizhuoyu:
Do we also want to remove these lines from Dockerfile?
```
  && if [[ ${MINIMAL^^} != "YES" ]] ; then \
    echo "JAVA_HOME=/usr" >> /etc/environment; \
  fi \
```

jizhuoyu:
Do we want to make the same changes to v1 image Dockerfile as well?

spilchen:
> Do we also want to remove these lines from Dockerfile?
> 
> ```
>   && if [[ ${MINIMAL^^} != "YES" ]] ; then \
>     echo "JAVA_HOME=/usr" >> /etc/environment; \
>   fi \
> ```

Yes, lets remove this part from the v2 image.

spilchen:
> Do we want to make the same changes to v1 image Dockerfile as well?

No. Let's keep the v1 image the same. We know the JAVA_HOME fix there works. This change should be targeted only for vclusterops deployments.

roypaulin:
We should move udx-java to e2e-udx for testing. Matt are you going to build a v2 image?

spilchen:
We could also do the udx-java changes as a follow-up. Once these changes are pushed to main the jenkins pipeline nightly run will rebuild the v2 image. So, we can do the follow-up after that is done.

spilchen (approved):

roypaulin (approved):

PR #517, opened 2023-09-23
Fix operator install issues during upgrade in vcluster mode

roypaulin:
When trying to upgrade and scale up at the same time, the operator got stuck at `OfflineUpgradeReconciler.InstallReconciler` preventing the operator from moving to the next actors to perform the scale up. To fix that, we relaxed `InstallReconciler` to not wait for all pods (in podfacts) to do the install before moving to the next actor.


scripts/setup-kustomize.sh
spilchen, line 1: Are the changes in this file needed? Or where they done to debug the testcase.

roypaulin: To debug and also for when the testcase will be run in vcluster mode.  But there is a separate jira to run e2e-server-upgrade in vcluster mode so I guess I can remove this.

spilchen: Yeah, lets remove it.

PR #519, opened 2023-09-27
Add v1 API for VerticaDB

spilchen:
This is the first phase in introducing a v1 API for VerticaDB. The old version, v1beta1, is deprecated, but will continue to work for now. However, you will receive a deprecation warning each time you use it. In a future release, v1beta1 will be removed altogether.

```
$ kubectl apply -f vdb.yaml
Warning: vertica.com/v1beta1 VerticaDB is deprecated, use vertica.com/v1 VerticaDB
verticadb.vertica.com/vc configured
```

In this PR, the two versions of the API are identical except for one field: spec.communal.hadoopConfig has been moved to spec.hadoopConfig. This change was made to prove that the conversion webhook is working properly. Follow-on work will move and rename other fields in the v1 API.

This PR also adds infrastructure to support two versions. A new conversion webhook has been added that will automatically convert between the v1beta1 and v1 versions. Subsequent PRs will start to make changes to the v1 API.

Note that you can still deploy the operator without a webhook. However, if you do, you must use the v1 API immediately. The VerticaDB controller now works exclusively with the v1 API, so you must use that version.

About half of the e2e testcases have been modified to use the new V1 API.

The CA bundle for the webhook now needs to be included in the CRD, since that is where the conversion webhook is configured. The operator will update this at runtime to handle the various ways the webhook TLS can be specified. The helm chart parameter webhook.caBundle has been removed, since it was already deprecated and there is little value in using the bundle specified there and applying it to the CRD conversion webhook config.


tests/e2e-udx/udx-cpp/setup-vdb/base/setup-vdb.yaml
```diff
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: vertica.com/v1beta1
+apiVersion: vertica.com/v1
```
roypaulin, line 14: How exactly can v1 and v1beta1 coexist? You moved some function from v1beta1/verticadb_types to v1/verticadb_types and you use the same images for e2e tests.

spilchen: The controller only works with v1. That's why some of the functions in v1beta1 could be moved out. There is a conversion function, which is called by a webhook to easily move between v1beta1 and v1 versions. If you are up for reading a bit I found this tutorial went over it quite well: https://book.kubebuilder.io/multiversion-tutorial/tutorial. I can also do a code walk-through.

api/v1beta1/verticadb_conversion.go
```diff
@@ -0,0 +1,398 @@
...
+		KerberosServiceName:    src.KerberosServiceName,
+		KerberosRealm:          src.KerberosRealm,
+		S3ServerSideEncryption: v1.ServerSideEncryptionType(src.S3ServerSideEncryption),
+		S3SseCustomerKeySecret: src.S3SseCustomerKeySecret,
+		AdditionalConfig:       src.AdditionalConfig,
+	}
+}
+
+// convertFromCommunal will convert from a v1 CommunalStorage to a v1beta1 version
+func convertFromCommunal(src *v1.CommunalStorage, hadoopConfig string) CommunalStorage {
```
roypaulin, line 265: How are you going to handle this when there will be more changes?

roypaulin (approved):

api/v1beta1/verticadb_conversion.go
spilchen: Yeah, good point. This file will need to be refactored as more changes are added.

PR #522, opened 2023-09-28
Pass communal storage path and params to re_ip and start_db

cchen-vertica:
Refreshed vlcusterops library, added communal storage path and communal storage params to options of re_ip and start_db, and enabled e2e test "revivedb-multi-sc".


pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -371,6 +387,13 @@ func (r *RestartReconciler) reipNodes(ctx context.Context, pods []*PodFact) (ctr
 	opts := []reip.Option{
 		reip.WithInitiator(r.InitiatorPod, r.InitiatorPodIP),
 	}
+	// If a communal path is set, include all of the EON parameters.
+	if r.Vdb.Spec.Communal.Path != "" {
```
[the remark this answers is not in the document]
spilchen: You can disregard this comment.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -390,6 +413,13 @@ func (r *RestartReconciler) restartCluster(ctx context.Context, downPods []*PodF
 	opts := []startdb.Option{
 		startdb.WithInitiator(r.InitiatorPod, r.InitiatorPodIP),
 	}
+	// If a communal path is set, include all of the EON parameters.
```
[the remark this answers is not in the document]
spilchen: You can disregard this comment.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -371,6 +377,20 @@ func (r *RestartReconciler) reipNodes(ctx context.Context, pods []*PodFact) (ctr
 	opts := []reip.Option{
 		reip.WithInitiator(r.InitiatorPod, r.InitiatorPodIP),
 	}
+	// If a communal path is set, include all of the EON parameters.
+	if r.Vdb.Spec.Communal.Path != "" {
+		// build communal storage params if there is not one
+		if r.ConfigurationParams == nil {
+			res, err := r.ConstructConfigParms(ctx)
+			if err != nil {
```
spilchen, line 385: We need to check res too. There is a helper for that.
```suggestion
			if verrors.IsReconcileAborted(res, err) {
```

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -390,6 +410,20 @@ func (r *RestartReconciler) restartCluster(ctx context.Context, downPods []*PodF
 	opts := []startdb.Option{
 		startdb.WithInitiator(r.InitiatorPod, r.InitiatorPodIP),
 	}
+	// If a communal path is set, include all of the EON parameters.
+	if r.Vdb.Spec.Communal.Path != "" {
+		// build communal storage params if there is not one
+		if r.ConfigurationParams == nil {
+			res, err := r.ConstructConfigParms(ctx)
+			if err != nil {
```
spilchen, line 418: Same comment here. Let's use the helper to check `res`
```suggestion
			if verrors.IsReconcileAborted(res, err) {
```

pkg/controllers/vdb/restart_reconciler.go
cchen-vertica: The res has been checked in lines 76 and 82 of ConstructConfigParms(). Are those checks not enough?

spilchen (approved):

pkg/controllers/vdb/restart_reconciler.go
cchen-vertica: After applying this change , the e2e test(https://github.com/vertica/vertica-kubernetes/actions/runs/6355160688/job/17267471737) "offline-upgrade-ks-0" keep failing on 75-wait-for-upgrade which said OfflineUpgradeInProgress should be true. Some res(when err == nil) might abort the restart reconciler after the offline upgrade. Need Matt's help to fix this.

cchen-vertica:
See the comment above for this e2e failure.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -390,6 +411,21 @@ func (r *RestartReconciler) restartCluster(ctx context.Context, downPods []*PodF
 	opts := []startdb.Option{
 		startdb.WithInitiator(r.InitiatorPod, r.InitiatorPodIP),
 	}
+	// If a communal path is set, include all of the EON parameters.
+	// Only for vclusterops as admintools relies on a config file.
```
spilchen, line 415: We are kind of breaking the abstraction by doing this. What is specific to `r.ConstructConfigParms` that doesn't work for admintools based deployments? If we do an upgrade with vclusterops enabled, will we not hit this issue?

roypaulin: The issue is that at step 65, we create a vdb with a bad endpoint, so step 75 fails because we try to retrieve  the non existent s3 creds secret. With vclusterops it found that No pods qualify for possible re-ip so it did not reach the ConstructConfigParms. So the question is why is the behavior with admintools different from vclusterops'.
Should we just modify step 75 in "offline-upgrade-ks-0".

spilchen: Can we try removing the `communal.credentialSecret` parameter in 65-setup-vdb-bad-endpoint.yaml? I think at the time the testcase was made, that field was required. But it is optional now.

PR #526, opened 2023-10-03
VCluster API Refresh

chinhtranvan:
This picks up the latest vclusterops package. The key change is it will hide sensitive config parms from the various logs. Originally, it was to use a new parameter to pass down config parms to create DB, but after a few revisions we opted to reuse the existing field, so this just becomes a refresh of the package and no API changes.


pkg/vadmin/create_db_vc.go
```diff
@@ -75,7 +75,7 @@ func (v *VClusterOps) genCreateDBOptions(s *createdb.Parms, certs *HTTPSCerts) v
 	}
 
 	// Additional configuration parameters for create db.
-	opts.ConfigurationParameters = s.ConfigurationParams
```
spilchen, line 78: I'm trying to understand the vclusterops changes again. We still have this parameter (`ConfigurationParameters`). Why would you want to use `ConfigurationParameters` vs `CommunalStorageParameters`? They both seem to do the same thing. In fact, we combine the maps when we call the bootstrap catalog op. We have masking for sensitive parms, but it doesn't seem like it matters where it comes from.

chinhtranvan: I agree that they both seem to do the same thing. AT is using both ConfigurationParameters and CommunalStorageParameters, and our vcluter mimic AT, there is some reasons e.g. For usability -- users understand better on the effect of the parameters they pass in

spilchen: Thanks Chinh

chinhtranvan:
@spilchen I just sync the latest vcluster repo

PR #527, opened 2023-10-03
Skip restart reconciler when DB init fails

chinhtranvan:
This is a bug fix for vclusterops deployments. When create_db or revive_db fails, the re-ip process in the restart_reconciler will be trigged and also fail. The solution is to skip the restart reconciler so that the create/revive can be attempted again. We will skip based on the DBInitialized status condition.


chinhtranvan:
I checked to ensure that the operator will skip restarting the reconciler. Based on the events I've observed, I haven't seen the operator process restarting the reconciler
Type     Reason          Age    From                Message
  ----     ------          ----   ----                -------
  Normal   CreateDBStart   5m42s  verticadb-operator  Starting create database
  Warning  CreateDBFailed  5m41s  verticadb-operator  Failed when calling VCreateDatabase

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -107,6 +107,15 @@ func (r *RestartReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctr
 		return ctrl.Result{}, err
 	}
 
+	// If create/revive db failed, we skip restarting the cluster for redoing create/revive db in vclusterops
+	if meta.UseVClusterOps(r.Vdb.Annotations) {
```
spilchen, line 111: Lets not make this specific to vcluster ops.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -107,6 +107,15 @@ func (r *RestartReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctr
 		return ctrl.Result{}, err
 	}
 
+	// If create/revive db failed, we skip restarting the cluster for redoing create/revive db in vclusterops
```
spilchen, line 110: This check should be done before we call `Collect`. Can we move it up?

chinhtranvan: Yes, will move it up

chinhtranvan:
@spilchen This is the result I try with my latest change, now we can skip the restart conciler, but we keep re-createdb.

  Warning  CreateDBFailed     45s                verticadb-operator  Failed when calling VCreateDatabase
  Normal   CreateDBStart      34s (x2 over 48s)  verticadb-operator  Starting create database
  Normal   CreateDBSucceeded  0s                 verticadb-operator  Successfully created database with subcluster 'pri'. It took 33.385842687s

chinhtranvan:
@spilchen  this is the result I achieved with my latest change. I can skip restarting the reconciler and continue to recreate the database
Warning  CreateDBFailed     45s                verticadb-operator  Failed when calling VCreateDatabase
 Normal   CreateDBStart      34s (x2 over 48s)  verticadb-operator  Starting create database
 Normal   CreateDBSucceeded  0s                 verticadb-operator  Successfully created database with subcluster 'pri'. It took 33.385842687s

pkg/controllers/vdb/init_db.go
```diff
@@ -58,7 +58,9 @@ func (g *GenericDatabaseInitializer) checkAndRunInit(ctx context.Context) (ctrl.
 		return ctrl.Result{}, err
 	}
 
-	if !g.PFacts.doesDBExist() {
+	// create a database if it doesn't exist, or redo the create/revive process if the database creation/revival fails
+	isSet, e := g.Vdb.IsConditionSet(vapi.DBInitialized)
+	if !g.PFacts.doesDBExist() || !isSet || e != nil {
```
spilchen, line 63: We aren't checking the error properly. Also, error variables should be called `err` where possible. The status condition check should be enough of a check that we don't need to use `doesDBExist`. There might be some UT pain with removing it, but I think it will be worth it to simplify the check.
```suggestion
	isSet, err := g.Vdb.IsConditionSet(vapi.DBInitialized)
	if err != nil {
	  return ctrl.Result{}, err
	}
	if !isSet {
```

pkg/controllers/vdb/createdb_reconciler_test.go
```diff
@@ -35,25 +35,6 @@ const TestPassword = "test-pw"
...
-		vdb.Status.Conditions = []vapi.VerticaDBCondition{{}, condition}
-
-		fpr := &cmds.FakePodRunner{}
-		pfacts := createPodFactsDefault(fpr)
-		dispatcher := vdbRec.makeDispatcher(logger, vdb, fpr, TestPassword)
-		r := MakeCreateDBReconciler(vdbRec, logger, vdb, fpr, pfacts, dispatcher)
-		Expect(r.Reconcile(ctx, &ctrl.Request{})).Should(Equal(ctrl.Result{}))
-		lastCall := fpr.Histories[len(fpr.Histories)-1]
-		Expect(lastCall.Command).ShouldNot(ContainElements("/opt/vertica/bin/admintools", "create_db"))
-	})
```
chinhtranvan, line 53: I remove this UT because it's not our behavior expect

pkg/controllers/vdb/restart_reconciler.go
chinhtranvan: As we discussed in the daily meeting, we will make this specific to vcluster ops and add a comment to explain.

spilchen (approved):

PR #528, opened 2023-10-04
Enable STRICT mTLS in istio testcase

spilchen:
This changes the istio testcase to enable STRICT mode for mTLS. There are some changes to the operator so that we can run in that mode with vclusterops deployments -- the headless service needs to include all of the ports used for communication, so I added the new ports needed for REST calls.

The istio testcase was moved to e2e-leg-5 because we need to redeploy the operator so that it has the istio sidecar attached to it.


tests/e2e-leg-5/istio/10-deploy-operator.yaml
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- command: sh -c "cd ../../.. && make undeploy-operator || true"
```
roypaulin, line 17: Now that one operator instance can manage multiple vdb instances why isn't this a problem?

spilchen: e2e-leg-5 is special. It only ever runs 1 testcase at a time. All of the tests that need to run the operator in a special way (e.g. webhook disabled, helm overrides, etc.) are in this leg.

roypaulin (approved):

PR #531, opened 2023-10-07
Rename v1 API fields

spilchen:
Since we have not yet released the v1 API, we are free to change it. I am making the following changes to the API:
- removed spec.ignoreClusterLease in favour of an annotation: vertica.com/ignore-cluster-lease
- removed spec.ignoreUpgradePath in favour of an annotation: vertica.com/ignore-upgrade-path
- removed spec.RestartTimeout in favour of an annotation: vertica.com/restart-timeout
- changes spec.temporarySubclusterRouting to be a pointer. This allows us to check for nil to see if this setting is used. And it won't show a default when you list the YAML manifest of a CR.

The v1beta1 API wasn't changed. The conversion webhook has been updated to convert between the two versions.


api/v1beta1/verticadb_conversion_test.go
```diff
@@ -0,0 +1,66 @@
...
+		Ω(v1VDB.Annotations[vmeta.IgnoreUpgradePathAnnotation]).Should(Equal(trueStrVal))
+		v1beta1VDB.Spec.IgnoreUpgradePath = false
+		Ω(v1beta1VDB.ConvertTo(&v1VDB)).Should(Succeed())
+		Ω(v1VDB.Annotations[vmeta.IgnoreUpgradePathAnnotation]).Should(BeEmpty())
+		v1VDB.Annotations[vmeta.IgnoreUpgradePathAnnotation] = trueStrVal
+		Ω(v1beta1VDB.ConvertFrom(&v1VDB)).Should(Succeed())
+		Ω(v1beta1VDB.Spec.IgnoreUpgradePath).Should(BeTrue())
+	})
+
+	It("should convert RestartTimeout", func() {
```
roypaulin, line 56: Are you going to add more testcase?

api/v1beta1/verticadb_conversion.go
```diff
@@ -103,59 +138,67 @@ func convertToSpec(src *VerticaDBSpec) v1.VerticaDBSpec {
...
+			Names:    src.TemporarySubclusterRouting.Names,
+			Template: convertToSubcluster(&src.TemporarySubclusterRouting.Template),
+		}
+	}
 	return dst
 }
 
 // convertFromSpec will convert from a v1 VerticaDBSpec to a v1beta1 version
-func convertFromSpec(src *v1.VerticaDBSpec) VerticaDBSpec {
+func convertFromSpec(src *v1.VerticaDB, srcSpec *v1.VerticaDBSpec) VerticaDBSpec {
```
roypaulin, line 151: Why not just pass `src`?

pkg/meta/annotations.go
```diff
@@ -51,29 +51,69 @@ const (
...
+	BuildRefAnnotation  = "vertica.com/buildRef"
+	// Annotation for the database's revive_instance_id
+	ReviveInstanceIDAnnotation = "vertica.com/revive-instance-id"
 )
 
 // IsPauseAnnotationSet will check the annotations for a special value that will
 // pause the operator for the CR.
 func IsPauseAnnotationSet(annotations map[string]string) bool {
-	return lookupBoolAnnotation(annotations, PauseOperatorAnnotation, false)
+	return lookupBoolAnnotation(annotations, PauseOperatorAnnotation)
```
roypaulin, line 81: Do we still need `AutoRestartVertica` if we have this?

api/v1beta1/verticadb_conversion.go
```diff
@@ -42,42 +44,75 @@ func (v *VerticaDB) ConvertFrom(srcRaw conversion.Hub) error {
...
 }
 
+// convertToAnnotations will create the annotations for a v1.VerticaDB CR taken
+// from a v1beta1.VerticaDB.
+func convertToAnnotations(src *VerticaDB) (newAnnotations map[string]string) {
+	newAnnotations = make(map[string]string, len(src.Annotations))
+	for k, v := range src.Annotations {
+		newAnnotations[k] = v
+	}
+	// Only include annotations for removed parameters if it is not the default
```
roypaulin, line 62: What do you mean by removed parameters?

roypaulin (commented):
Looks good!

pkg/meta/annotations.go
spilchen: Yes. Pause will not run any reconcilers. Where as auto restart vertica will just skilp the restart reconciler.

api/v1beta1/verticadb_conversion_test.go
spilchen: Good point. I think I need some tests for transientSubclusterRouting.

roypaulin (approved):

PR #440, opened 2023-06-30
Always run e2e-vcluster with latest master changes

chinhtranvan:
The e2e-vcluster test suite needs to run with the latest version of the server. For this reason, when we run that test suite we will use a private image from the dockerhub repo with the lag lastest-master. We also skip this test suite on forks because it requires secrets that are only accessible through the vertica-kubernetes repo.

We still build the v2 image to make sure nothing is broken. But since it uses an RPM from version 12.0.3 of the server, it doesn't have all of the functionality that the e2e-vcluster test suite needs. So, no e2e tests will run with it.

PR #560, opened 2023-10-25
CI changes for new k8s images

chinhtranvan:
There is some image changes for this PR
- remove the nokeys image from the pipeline and NO_KEYS env var usage
- GH actions to stop using latest-master for vcluster tests. Instead use full or minimal, depending on the leg.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -8,15 +8,15 @@ on:
         type: string
         required: true
       full_vertica_image:
-        description: 'Name of an existing full Vertica server image. If blank we will build one using the default name'
+        description: 'Name of an existing full Vertica server v2 image. If blank we will build one using the default name'
+        type: string
+        required: true
+      minimal_vertica_image:
+        description: 'Name of an existing minimal Vertica server v2 image. If blank we will build one using the default name'
         type: string
         required: true
       legacy_vertica_image:
         description: 'Name of an existing full image with admintools and SSH support. Leave blank to build one with the default name'
-        type: string
-        required: true
-      minimal_vertica_image:
-        description: 'Name of an existing minimal Vertica server image. If blank we will build one using the default name'
         type: string
         required: true
       vlogger_image:
@@ -42,12 +42,12 @@ on:
       full-vertica-image:
         description: "The image name of the full vertica server image"
         value: ${{ jobs.build-server-full.outputs.image }}
-      legacy-vertica-image:
-        description: "The image name of the full vertica server image, but with admintools and SSH support"
-        value: ${{ jobs.build-server-nokeys.outputs.image }}
       minimal-vertica-image:
         description: "The image name of the vertica server, but with optional software removed"
         value: ${{ jobs.build-server-minimal.outputs.image }}
+      legacy-vertica-image:
+        description: "The image name of the full vertica server image, but with admintools and SSH support"
+        value: ${{ jobs.build-server-nokeys.outputs.image }}
       vlogger-image:
         description: "The image name of the vertica logger sidecar"
         value: ${{ jobs.build-vlogger.outputs.image }}
@@ -67,12 +67,6 @@ jobs:
       image: ${{ steps.full_vertica_image.outputs.value }}
     steps:
 
-    - name: set lower case owner name
-      env:
-        OWNER: '${{ github.repository_owner }}'
-      run: |
-        echo "OWNER_LC=${OWNER,,}" >>${GITHUB_ENV}
-
     - name: Pick the name of the image
       uses: spilchen/switch-case-action@v2
       id: full_vertica_image
@@ -83,7 +77,7 @@ jobs:
 
     - name: Login to GitHub Container registry for non-PRs
       uses: docker/login-action@v2
-      if: ${{ github.event_name != 'pull_request' && inputs.full_vertica_image == '' || startsWith(inputs.full_vertica_image, 'ghcr.io') }}
+      if: ${{ github.event_name != 'pull_request' && startsWith(inputs.full_vertica_image, 'ghcr.io') }}
       with:
         registry: ghcr.io
         username: ${{ github.repository_owner }}
@@ -255,23 +249,17 @@ jobs:
       image: ${{ steps.minimal_vertica_image.outputs.value }}
     steps:
 
-    - name: set lower case owner name
-      env:
-        OWNER: '${{ github.repository_owner }}'
-      run: |
-        echo "OWNER_LC=${OWNER,,}" >>${GITHUB_ENV}
-
     - name: Pick the name of the image
       uses: spilchen/switch-case-action@v2
       id: minimal_vertica_image
       with:
-        default: docker.io/vertica/vertica-k8s-private:latest-master
+        default: docker.io/vertica/vertica-k8s-private:latest-minimal-master
         conditionals-with-values: |
           ${{ inputs.minimal_vertica_image != '' }} => ${{ inputs.minimal_vertica_image }}
 
     - name: Login to GitHub Container registry for non-PRs
       uses: docker/login-action@v2
-      if: ${{ github.event_name != 'pull_request' && inputs.minimal_vertica_image == '' || startsWith(inputs.minimal_vertica_image, 'ghcr.io') }}
+      if: ${{ github.event_name != 'pull_request' && startsWith(inputs.minimal_vertica_image, 'ghcr.io') }}
       with:
         registry: ghcr.io
         username: ${{ github.repository_owner }}
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -11,16 +11,12 @@ on:
         description: 'Name of an existing full Vertica server image. If blank we will build one using the default name'
         type: string
         required: true
-      nokeys_vertica_image:
-        description: 'Name of an existing Vertica server image with no keys inside. If blank we will build one using the default name'
+      legacy_vertica_image:
+        description: 'Name of an existing full image with admintools and SSH support. Leave blank to build one with the default name'
         type: string
         required: true
       minimal_vertica_image:
         description: 'Name of an existing minimal Vertica server image. If blank we will build one using the default name'
-        type: string
-        required: true
-      v2_vertica_image:
-        description: 'Name of a v2 Vertica server image. If blank we will build one using the default name'
         type: string
         required: true
       vlogger_image:
@@ -46,15 +42,12 @@ on:
       full-vertica-image:
         description: "The image name of the full vertica server image"
         value: ${{ jobs.build-server-full.outputs.image }}
-      nokeys-vertica-image:
-        description: "The image name of the vertica server image, but with no keys inside"
+      legacy-vertica-image:
+        description: "The image name of the full vertica server image, but with admintools and SSH support"
         value: ${{ jobs.build-server-nokeys.outputs.image }}
       minimal-vertica-image:
         description: "The image name of the vertica server, but with optional software removed"
         value: ${{ jobs.build-server-minimal.outputs.image }}
-      v2-vertica-image:
-        description: "The image name of the v2 vertica server image"
-        value: ${{ jobs.build-server-v2.outputs.image }}
       vlogger-image:
         description: "The image name of the vertica logger sidecar"
         value: ${{ jobs.build-vlogger.outputs.image }}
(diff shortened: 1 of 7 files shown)


.github/workflows/build-images.yml
```diff
@@ -356,10 +265,9 @@ jobs:
       uses: spilchen/switch-case-action@v2
       id: minimal_vertica_image
       with:
-        default: ghcr.io/${{ env.OWNER_LC }}/vertica-k8s:${{ github.sha }}-minimal
+        default: docker.io/vertica/vertica-k8s-private:latest-master
```
spilchen, line 268: Should this have a tag of `latest-minimal-master`?

.github/workflows/build-images.yml
```diff
@@ -84,10 +77,9 @@ jobs:
       uses: spilchen/switch-case-action@v2
```
spilchen, line 77: I can't make a comment on the exact line, but we set OWNER_LC a few lines that that we don't need anymore.

.github/workflows/build-images.yml
```diff
@@ -84,10 +77,9 @@ jobs:
...
       id: full_vertica_image
       with:
-        default: ghcr.io/${{ env.OWNER_LC }}/vertica-k8s:${{ github.sha }}
+        default: docker.io/vertica/vertica-k8s-private:latest-master
         conditionals-with-values: |
           ${{ inputs.full_vertica_image != '' }} => ${{ inputs.full_vertica_image }}
-          ${{ github.event_name == 'pull_request' }} => vertica-k8s:kind
 
     - name: Login to GitHub Container registry for non-PRs
       uses: docker/login-action@v2
```
spilchen, line 85: I can't make a comment on the exact line, but the next line when we check if we log into ghcr.io, we can remove the check for ''. We no longer default to a ghcr.io image. Something like this:
```
if: ${{ github.event_name != 'pull_request' && startsWith(inputs.full_vertica_image, 'ghcr.io') }}
```

.github/workflows/build-images.yml
```diff
@@ -243,103 +224,31 @@ jobs:
...
-        echo "Size: **$(docker inspect --format '{{.Size}}' ${{ steps.v2_vertica_image.outputs.value }} | numfmt --to=iec)**" >> $GITHUB_STEP_SUMMARY
-        echo "Image ID: **$(docker inspect --format '{{.ID}}' ${{ steps.v2_vertica_image.outputs.value }})**" >> $GITHUB_STEP_SUMMARY
-        echo "Digest: **$(IFS=":" read image tag <<< $(echo ${{ steps.v2_vertica_image.outputs.value }} | sed -e 's/^docker.io\///'); docker inspect --format='{{.RepoDigests}}' $image:$tag | sed 's:^.\(.*\).$:\1:' | tr " " "\n" | grep $image | cut -d'@' -f2 || echo "<none>")**" >> $GITHUB_STEP_SUMMARY
-        echo -n "Vertica Version: **" >> $GITHUB_STEP_SUMMARY
-        echo -n $(docker inspect --format '{{index .Config.Labels "vertica-version"}}' ${{ steps.v2_vertica_image.outputs.value }}) >> $GITHUB_STEP_SUMMARY
-        echo "**" >> $GITHUB_STEP_SUMMARY
-
   build-server-minimal:
     runs-on: ubuntu-latest
     outputs:
```
spilchen, line 254: I can't make a comment on the exact line, but a few lines down when we set `OWNER_LC`. That part can be removed.

.github/workflows/build-images.yml
```diff
@@ -356,10 +265,9 @@ jobs:
...
       id: minimal_vertica_image
       with:
-        default: ghcr.io/${{ env.OWNER_LC }}/vertica-k8s:${{ github.sha }}-minimal
+        default: docker.io/vertica/vertica-k8s-private:latest-master
         conditionals-with-values: |
           ${{ inputs.minimal_vertica_image != '' }} => ${{ inputs.minimal_vertica_image }}
-          ${{ github.event_name == 'pull_request' }} => vertica-k8s:kind-minimal
 
     - name: Login to GitHub Container registry for non-PRs
       uses: docker/login-action@v2
```
spilchen, line 273: The check in the line below we can update now that the default image isn't ghcr.io
```
if: ${{ github.event_name != 'pull_request' && startsWith(inputs.minimal_vertica_image, 'ghcr.io') }}
```

.github/workflows/e2e.yml
```diff
@@ -14,18 +14,14 @@ on:
         description: 'Name of an existing full vertica image. Leave blank to build one with the default name'
         type: string
         required: false
-      nokeys_vertica_image:
-        description: 'Name of an existing vertica image with no keys inside. Leave blank to build one with the default name'
+      legacy_vertica_image:
```
spilchen, line 17: Can you move the order of this down so it is after `minimal_vertica_image`?

.github/workflows/e2e.yml
```diff
@@ -14,18 +14,14 @@ on:
...
+        description: 'Name of an existing full image with admintools and SSH support. Leave blank to build one with the default name'
         type: string
         required: false
       minimal_vertica_image:
         description: 'Name of an existing minimal vertica image. Leave blank to build one with the default name'
```
spilchen, line 22: Can you update the description to indicate its a v2 image.

.github/workflows/e2e.yml
```diff
@@ -14,18 +14,14 @@ on:
         description: 'Name of an existing full vertica image. Leave blank to build one with the default name'
```
spilchen, line 14: Can you update the description to indicate its a v2 image.

.github/workflows/e2e.yml
```diff
@@ -176,7 +171,7 @@ jobs:
     with:
       vlogger-image: ${{ needs.build.outputs.vlogger-image }}
       operator-image: ${{ needs.build.outputs.operator-image }}
-      vertica-image: ${{ needs.build.outputs.v2-vertica-image }}
+      vertica-image: ${{ needs.build.outputs.minimal-vertica-image }}
```
spilchen, line 174: Move of the v2 replacements have been minimal. Can you change a few to be the full image? It would be nice if roughly half of the vcluster tests used the minimal and half used the full image.

.github/workflows/e2e.yml
```diff
@@ -277,8 +272,8 @@ jobs:
     with:
       vlogger-image: ${{ needs.build.outputs.vlogger-image }}
       operator-image: ${{ needs.build.outputs.operator-image }}
-      # We must pass the full image since the minimal image cannot run Java UDx's
-      vertica-image: ${{ needs.build.outputs.full-vertica-image }}
+      # pass the legacy vertica image
```
spilchen, line 275: Can you remove this comment? I don't think it adds much here.
```suggestion
```

spilchen (commented):
Looks good. I just had minor comments. It's too bad the e2e pipeline is slow today, we could have gotten a few runs in.

spilchen:
> run some e2e legs against a released vertica-k8s image (vertica/vertica-k8s:23.3.0-0)

I never saw a change for this. If so, it's fine to leave off of this PR. I think we have enough here.

chinhtranvan:
> > run some e2e legs against a released vertica-k8s image (vertica/vertica-k8s:23.3.0-0)
> 
> I never saw a change for this. If so, it's fine to leave off of this PR. I think we have enough here.

Yes, I will leave off of this PR

chinhtranvan:
The e2e test looks weird. I'm trying to understand why there is a 'no such file or directory' error when trying to open full-vertica-image/full-vertica-image.tar.

.github/workflows/build-images.yml
```diff
@@ -111,26 +97,16 @@ jobs:
...
-    - name: Save the image for consumption by dependent jobs (PRs only)
-      if: ${{ github.event_name == 'pull_request' }}
-      run: |
-        docker save ${{ steps.full_vertica_image.outputs.value }} > full-vertica-image.tar
-
-    - uses: actions/upload-artifact@v3
-      if: ${{ github.event_name == 'pull_request' }}
-      with:
-        name: full-vertica-image
-        path: full-vertica-image.tar
```
spilchen, line 133: This is needed so that the full image is available for later stages.

spilchen:
> The e2e test looks weird. I'm trying to understand why there is a 'no such file or directory' error when trying to open full-vertica-image/full-vertica-image.tar. Do we need to wait for JenKins to trigger GitHub CI?

There is a missing `actions/upload-artifact` step. Something like this is needed:
```
    - uses: actions/upload-artifact@v3
      if: ${{ github.event_name == 'pull_request' }}
      with:
        name: full-vertica-image
        path: full-vertica-image.tar
```

chinhtranvan:
> > The e2e test looks weird. I'm trying to understand why there is a 'no such file or directory' error when trying to open full-vertica-image/full-vertica-image.tar. Do we need to wait for JenKins to trigger GitHub CI?
> 
> There is a missing `actions/upload-artifact` step. Something like this is needed:
> 
> ```
>     - uses: actions/upload-artifact@v3
>       if: ${{ github.event_name == 'pull_request' }}
>       with:
>         name: full-vertica-image
>         path: full-vertica-image.tar
> ```

Oh, I remember why we don't need the path full-vertica-image.tar—we encountered this problem before. For v2-vertica-image, we won't push the image because the RPM we use is too old. Therefore, there's no need to save it under full-vertica-image/full-vertica-image.tar and load it later. We just need to pull the image. This is the result after I skip and load full-vertica-image.tar 
https://github.com/vertica/vertica-kubernetes/actions/runs/6649499000

There are two things I wonder 

- It passed successfully for image with tag - latest-master e2e tests, but not recognize latest-minimal-master in private docker-hub

```
Error response from daemon: manifest for vertica/vertica-k8s-private:latest-minimal-master not found: manifest unknown: manifest unknown
ERROR: image: "docker.io/vertica/vertica-k8s-private:latest-minimal-master" not present locally
```
- For the legacy images, it shows 
```
scripts/push-to-kind.sh -i 
scripts/push-to-kind.sh: option requires an argument -- i
ERROR: unrecognized option: -?
```
Do we need to wait for JenKins to trigger GitHub CI?

chinhtranvan:
> > > The e2e test looks weird. I'm trying to understand why there is a 'no such file or directory' error when trying to open full-vertica-image/full-vertica-image.tar. Do we need to wait for JenKins to trigger GitHub CI?
> > 
> > 
> > There is a missing `actions/upload-artifact` step. Something like this is needed:
> > ```
> >     - uses: actions/upload-artifact@v3
> >       if: ${{ github.event_name == 'pull_request' }}
> >       with:
> >         name: full-vertica-image
> >         path: full-vertica-image.tar
> > ```
> 
> Oh, I remember why we don't need the path full-vertica-image.tar—we encountered this problem before. For v2-vertica-image, we won't push the image because the RPM we use is too old. Therefore, there's no need to save it under full-vertica-image/full-vertica-image.tar and load it later. We just need to pull the image. Reference PR: https://github.com/vertica/vertica-kubernetes/pull/440/files
> 
> After skip loading the images, we passed all e2e vcluster tests, only have problems for AT tests, I see errors in AT tests scripts/push-to-kind.sh: option requires an argument -- i ERROR: unrecognized option: -?

Update: just fix the problem, build-e2e.yml

> > > The e2e test looks weird. I'm trying to understand why there is a 'no such file or directory' error when trying to open full-vertica-image/full-vertica-image.tar. Do we need to wait for JenKins to trigger GitHub CI?
> > 
> > 
> > There is a missing `actions/upload-artifact` step. Something like this is needed:
> > ```
> >     - uses: actions/upload-artifact@v3
> >       if: ${{ github.event_name == 'pull_request' }}
> >       with:
> >         name: full-vertica-image
> >         path: full-vertica-image.tar
> > ```
> 
> Oh, I remember why we don't need the path full-vertica-image.tar—we encountered this problem before. For v2-vertica-image, we won't push the image because the RPM we use is too old. Therefore, there's no need to save it under full-vertica-image/full-vertica-image.tar and load it later. We just need to pull the image. Reference PR: https://github.com/vertica/vertica-kubernetes/pull/440/files
> 
> After skip loading the images, we passed all e2e vcluster tests, only have problems for AT tests, I see errors in AT tests scripts/push-to-kind.sh: option requires an argument -- i ERROR: unrecognized option: -?

Just fix the problem: some previous changes in build-image.yml caused an empty Vertica Image. All comments have been addressed with successful e2e tests.

spilchen (approved):

PR #563, opened 2023-10-30
Make vclusterops the default deployment for v1 API

chinhtranvan:
This PR makes the change so that default deployments of the v1 API use vclusterops and continue defaulting to admintools-based deployments for v1beta1. Additionally, it sets the vertica.com/vcluster-ops to false for the v1beta1 webhook.

### DIFF
--- a/api/v1/helpers.go
+++ b/api/v1/helpers.go
@@ -73,10 +73,12 @@ func MakeVDB() *VerticaDB {
 			Kind:       VerticaDBKind,
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        nm.Name,
-			Namespace:   nm.Namespace,
-			UID:         "abcdef-ghi",
-			Annotations: make(map[string]string),
+			Name:      nm.Name,
+			Namespace: nm.Namespace,
+			UID:       "abcdef-ghi",
+			Annotations: map[string]string{
+				vmeta.VClusterOpsAnnotation: vmeta.VClusterOpsAnnotationFalse,
+			},
 		},
 		Spec: VerticaDBSpec{
 			AutoRestartVertica: true,
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -52,7 +52,7 @@ const (
 	S3Prefix                 = "s3://"
 	GCloudPrefix             = "gs://"
 	AzurePrefix              = "azb://"
-	ApiVersion               = "vertica.com/v1"
+	APIVersion               = "vertica.com/v1"
 )
 
 // hdfsPrefixes are prefixes for an HDFS path.
@@ -75,7 +75,8 @@ var _ webhook.Defaulter = &VerticaDB{}
 func (v *VerticaDB) Default() {
 	verticadblog.Info("default", "name", v.Name, "GroupVersion", GroupVersion)
 
-	v.APIVersion = ApiVersion
+	// default deployments of the v1 api use vclusterops
+	v.APIVersion = APIVersion
 	// setting the vertica.com/vcluster-ops to True
 	v.Annotations[vmeta.VClusterOpsAnnotation] = vmeta.VClusterOpsAnnotationTrue
 
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -74,9 +74,6 @@ var _ webhook.Defaulter = &VerticaDB{}
 func (v *VerticaDB) Default() {
 	verticadblog.Info("default", "name", v.Name, "GroupVersion", GroupVersion)
 
-	// setting the vertica.com/vcluster-ops to True
-	v.Annotations[vmeta.VClusterOpsAnnotation] = vmeta.VClusterOpsAnnotationTrue
-
 	// imagePullPolicy: if not set should default to Always if the tag in the image is latest,
 	// otherwise it should be IfNotPresent (set in verticadb_types.go)
 	if strings.HasSuffix(v.Spec.Image, ":latest") {
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -52,6 +52,7 @@ const (
 	S3Prefix                 = "s3://"
 	GCloudPrefix             = "gs://"
 	AzurePrefix              = "azb://"
+	ApiVersion               = "vertica.com/v1"
 )
 
 // hdfsPrefixes are prefixes for an HDFS path.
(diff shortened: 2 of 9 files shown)


chinhtranvan:
After adding CR v1 API to use vclusterops by default and adding v1beta1 to use AT, I observed numerous e2e failures but passed UT tests. @spilchen , one thing I am concerned about is that I remember you mentioning that we might have a lot of failures in UTs. I just want to reproduce the problem to identify UT failures

chinhtranvan:
After I add CR v1 api  to use vclusterops by default,  and add v1beta1 to use AT. I see a lot e2e failure, and pass UT tests. @spilchen ,one thing I concern I remember you mentioned that we will have a lot failures in UTs. I just want to reproduce the problem to see UT Failures

api/v1/verticadb_webhook.go
```diff
@@ -74,6 +74,9 @@ var _ webhook.Defaulter = &VerticaDB{}
 func (v *VerticaDB) Default() {
 	verticadblog.Info("default", "name", v.Name, "GroupVersion", GroupVersion)
 
+	// setting the vertica.com/vcluster-ops to True
+	v.Annotations[vmeta.VClusterOpsAnnotation] = vmeta.VClusterOpsAnnotationTrue
```
spilchen, line 78: One issue with this is that it will overwrite any value that was set for this annotation. Even if we have conditional logic, I don't think this is the correct spot for this.

api/v1beta1/verticadb_webhook.go
```diff
@@ -130,6 +130,9 @@ var _ webhook.Defaulter = &VerticaDB{}
 func (v *VerticaDB) Default() {
 	verticadblog.Info("default", "name", v.Name, "GroupVersion", GroupVersion)
 
+	// setting the vertica.com/vcluster-ops to false
+	v.Annotations[vmeta.VClusterOpsAnnotation] = vmeta.VClusterOpsAnnotationFalse
```
spilchen, line 134: Same comment here.

spilchen (commented):
A few additional changes that we need:
- the setting of the annotation should be done in verticadb_conversion.go, in the the function ConvertTo. This gets called each time we from v1beta1 -> v1. We only set the annotation if it isn't already set.
- the default value in pkg/meta/annotations should change the default value for vcluster-ops. This will ensure that all v1 APIs will default to vclusterops. This likely will break a bunch of UTs though.

chinhtranvan:
> A few additional changes that we need:
> 
> * the setting of the annotation should be done in verticadb_conversion.go, in the the function ConvertTo. This gets called each time we from v1beta1 -> v1. We only set the annotation if it isn't already set.
> * the default value in pkg/meta/annotations should change the default value for vcluster-ops. This will ensure that all v1 APIs will default to vclusterops. This likely will break a bunch of UTs though.

For second bullet point, I would like to confirm when you refer to the default value, are you suggesting that we need to change the default value for vcluster-ops so that all v1 APIs will default to it? Are you suggesting something like this:
```
VClusterOpsAnnotation = "vertica.com/v1"
```
I saw that only two unit tests were failed.

spilchen:
> > A few additional changes that we need:
> > 
> > * the setting of the annotation should be done in verticadb_conversion.go, in the the function ConvertTo. This gets called each time we from v1beta1 -> v1. We only set the annotation if it isn't already set.
> > * the default value in pkg/meta/annotations should change the default value for vcluster-ops. This will ensure that all v1 APIs will default to vclusterops. This likely will break a bunch of UTs though.
> 
> For second bullet point, I would like to confirm when you refer to the default value, are you suggesting that we need to change the default value for vcluster-ops so that all v1 APIs will default to it? Are you suggesting something like this:
> 
> ```
> VClusterOpsAnnotation = "vertica.com/v1"
> ```
> 
> I saw that only two unit tests were failed.

No. We have to change `UseVClusterOps` so that it returns true if the annotation isn't set. This will ensure that all v1 APIs default to true. v1beta1 should go through the conversion and have the annotation set for them to false.

pkg/meta/annotations.go
```diff
@@ -123,7 +123,16 @@ func IsPauseAnnotationSet(annotations map[string]string) bool {
...
+	// UseVClusterOps returns true if the annotation isn't set.
+	defaultValue := true
+	if val, ok := annotations[VClusterOpsAnnotation]; ok {
+		varAsBool, err := strconv.ParseBool(val)
+		if err != nil {
+			return false
+		}
+		return varAsBool
+	}
+	return defaultValue
```
chinhtranvan, line 135: @spilchen  I got 113 failed unit tests after this change. Is that what you meant by a bunch of UT failures you mentioned to me?

pkg/meta/annotations.go
```diff
@@ -123,7 +123,16 @@ func IsPauseAnnotationSet(annotations map[string]string) bool {
 // UseVClusterOps returns true if all admin commands should use the vclusterOps
 // library rather than admintools.
 func UseVClusterOps(annotations map[string]string) bool {
-	return lookupBoolAnnotation(annotations, VClusterOpsAnnotation)
+	// UseVClusterOps returns true if the annotation isn't set.
+	defaultValue := true
```
chinhtranvan, line 127: @spilchen I got 113 failed unit tests after this change. Is that what you meant by a bunch of test failures you mentioned to me?

spilchen: Yes :)

pkg/meta/annotations.go
```diff
@@ -123,7 +123,16 @@ func IsPauseAnnotationSet(annotations map[string]string) bool {
 // UseVClusterOps returns true if all admin commands should use the vclusterOps
 // library rather than admintools.
 func UseVClusterOps(annotations map[string]string) bool {
-	return lookupBoolAnnotation(annotations, VClusterOpsAnnotation)
```
spilchen, line 126: Can we just change `lookupBoolAnnotation` so that it accepts an argument to use for the default value?

pkg/meta/annotations.go
spilchen: I pushed a change to try and preserve the old unit test behaviour. Let me know what you think of it.

api/v1/helpers.go
```diff
@@ -73,10 +73,12 @@ func MakeVDB() *VerticaDB {
...
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        nm.Name,
-			Namespace:   nm.Namespace,
-			UID:         "abcdef-ghi",
-			Annotations: make(map[string]string),
+			Name:      nm.Name,
+			Namespace: nm.Namespace,
+			UID:       "abcdef-ghi",
+			Annotations: map[string]string{
+				vmeta.VClusterOpsAnnotation: vmeta.VClusterOpsAnnotationFalse,
```
chinhtranvan, line 80: Cool! This change keeps our unit tests aware that it still relies on admintool-based deployments, which can help avoid a bunch of failures defined in vcluster-based deployments.

chinhtranvan:
With this change, we are facing an error from Scrutinize that cannot find a database to target in the vdb-gen test in e2e-s3. I'm trying to fix it
```
Warning: No databses are defined! Please create a database to scrutinize.
Error: Scrutinize could not find a db to target. Cannot collect diagnostics
```

spilchen (approved):

PR #567, opened 2023-10-31
Removed transient nodes from node list when starting db

cchen-vertica:
Transient nodes are read-only secondary nodes so it cannot be started using vclusterops. As a result, I removed transient nodes from start node list. This change shouldn't affect existing features.

### DIFF
--- a/pkg/controllers/vdb/restart_reconciler.go
+++ b/pkg/controllers/vdb/restart_reconciler.go
@@ -160,7 +160,12 @@ func (r *RestartReconciler) reconcileCluster(ctx context.Context) (ctrl.Result, 
 		return ctrl.Result{Requeue: true}, nil
 	}
 
-	downPods := r.PFacts.findRestartablePods(r.RestartReadOnly, true)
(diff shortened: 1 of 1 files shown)


pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -436,7 +436,9 @@ func (r *RestartReconciler) restartCluster(ctx context.Context, downPods []*PodF
 		)
 	}
 	for i := range downPods {
-		opts = append(opts, startdb.WithHost(downPods[i].podIP))
+		if !downPods[i].isTransient {
```
spilchen, line 439: The `findRestartablePods` function has a filter for transient pods (the second parameter). This filter is included for cluster restarts, which makes sense because re-ip is necessary to get the nodes in the admintools.conf with the correct IPs. However, this is not necessary for vclusterops, as there is no admintools.conf. Therefore, I recommend modifying the call to `findRestartablePods` for vclusterops to exclude transient pods.

spilchen (approved):

PR #576, opened 2023-11-06
Change default subcluster type in V1 VerticaDB API

jizhuoyu:
- Change default subcluster type to `primary` always
- Remove the auto case conversion for the type string (so now when a user provides a type string with upper-case letters, e.g. `Primary`, an error will be reported)

Note: these changes apply to the v1 API. The v1beta1 API was not changed.

### DIFF
--- a/api/v1/verticadb_types.go
+++ b/api/v1/verticadb_types.go
@@ -602,12 +602,13 @@ type Subcluster struct {
 	// db.licenseSecret parameter.
 	Size int32 `json:"size"`
 
+	// +kubebuilder:default:=primary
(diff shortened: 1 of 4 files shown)


roypaulin:
Do we support multiple primaries? If no, what happens if there are more than one subcluster with omitted type?

jizhuoyu:
> Do we support multiple primaries? If no, what happens if there are more than one subcluster with omitted type?

We do support multiple primaries, indeed this default to `primary` behavior is the same as that in v1beta1 where we set `isPrimary` to `true` by default.

spilchen (approved):

PR #578, opened 2023-11-06
Added a new annotation: vertica.com/superuser-name

cchen-vertica:
Added a superuser-name annotation. This annotation can be used to customize Vertica superuser name. This should be used with vclusterops. When admintools is used, this annotation will be always set to "dbadmin".

### DIFF
--- a/.github/workflows/e2e-leg-4.yml
+++ b/.github/workflows/e2e-leg-4.yml
@@ -15,6 +15,9 @@ on:
       vertica-deployment-method:
         type: string
         required: false
+      vertica-superuser-name:
+        type: string
+        required: false
     secrets:
       DOCKERHUB_USERNAME:
         description: 'When working with images from docker.io, this is the username for login purposes'
@@ -79,7 +82,7 @@ jobs:
         if [ "${VERTICA_DEPLOYMENT_METHOD}" != "vclusterops" ]; then E2E_TEST_DIRS+=" tests/e2e-leg-4-at-only"; fi
         export VERTICA_SUPERUSER_NAME=${{ inputs.vertica-superuser-name }}
         if [ -z "${VERTICA_SUPERUSER_NAME}" -o "${VERTICA_DEPLOYMENT_METHOD}" != "vclusterops" ]; then 
-          VERTICA_SUPERUSER_NAME="dbadmin"
+          export VERTICA_SUPERUSER_NAME="dbadmin"
         fi
         # All helm deployments will use cert-manager to create the webhook cert
         export HELM_OVERRIDES="--set webhook.certSource=cert-manager"
--- a/.github/workflows/e2e-leg-4.yml
+++ b/.github/workflows/e2e-leg-4.yml
@@ -18,6 +18,7 @@ on:
       vertica-superuser-name:
         type: string
         required: false
+        default: myadmin
     secrets:
       DOCKERHUB_USERNAME:
         description: 'When working with images from docker.io, this is the username for login purposes'
--- a/.github/workflows/e2e-leg-4.yml
+++ b/.github/workflows/e2e-leg-4.yml
@@ -44,6 +44,11 @@ on:
         options:
         - admintools
         - vclusterops
+      vertica-superuser-name:
+        description: 'Vertica superuser name'
+        type: string
+        required: false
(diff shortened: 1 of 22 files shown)


pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -124,7 +124,7 @@ func (r *VerticaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	prunner := cmds.MakeClusterPodRunner(log, r.Cfg, passwd)
+	prunner := cmds.MakeClusterPodRunner(log, r.Cfg, vmeta.GetSuperuserName(vdb.Annotations), passwd)
```
spilchen, line 127: Can we use `vdb.GetVerticaUser`? It will make it easier to switch if we ever convert the annotation to a CR parm.

EDIT: This is a general comment as I see this pattern in quite a few places.

spilchen (commented):
Good approach. Just had one minor suggestion. Also, I think we should have a entry in the CHANGELOG for this. I'll add a changie entry to your PR.

.github/workflows/e2e-leg-4.yml
```diff
@@ -72,6 +77,10 @@ jobs:
         export E2E_TEST_DIRS="tests/e2e-leg-4"
         export VERTICA_DEPLOYMENT_METHOD=${{ inputs.vertica-deployment-method }}
         if [ "${VERTICA_DEPLOYMENT_METHOD}" != "vclusterops" ]; then E2E_TEST_DIRS+=" tests/e2e-leg-4-at-only"; fi
+        export VERTICA_SUPERUSER_NAME=${{ inputs.vertica-superuser-name }}
+        if [ -z "${VERTICA_SUPERUSER_NAME}" -o "${VERTICA_DEPLOYMENT_METHOD}" != "vclusterops" ]; then 
+          VERTICA_SUPERUSER_NAME="dbadmin"
```
spilchen, line 82: Should we export this?
```suggestion
          export VERTICA_SUPERUSER_NAME="dbadmin"
```

.github/workflows/e2e-leg-4.yml
```diff
@@ -72,6 +77,10 @@ jobs:
         export E2E_TEST_DIRS="tests/e2e-leg-4"
         export VERTICA_DEPLOYMENT_METHOD=${{ inputs.vertica-deployment-method }}
         if [ "${VERTICA_DEPLOYMENT_METHOD}" != "vclusterops" ]; then E2E_TEST_DIRS+=" tests/e2e-leg-4-at-only"; fi
+        export VERTICA_SUPERUSER_NAME=${{ inputs.vertica-superuser-name }}
```
spilchen, line 80: I don't think we are setting this correctly. When I look at your CI run I see dbadmin being used:
```
Vertica server image name: docker.io/vertica/vertica-k8s-private:latest-minimal-master
Base vertica server image name for upgrade tests: <not-set>
Vertica logger image name: vertica-logger:kind
Endpoint: https://minio.kuttl-e2e-s3/
Protocol: s3://
Communal Path Prefix: s3://nimbusdb/
Using private registry: NO
Add server mounts: NO
Deployment method: vclusterops
Vertica superuser name: dbadmin
```
https://github.com/vertica/vertica-kubernetes/actions/runs/6785933635/job/18448129113

The new parameter was added to `workflow_dispatch`. That gets called when you trigger just this workflow. There are a set of parameters for `workflow_call`. This is used when another workflow calls this workflow, which is what happens during a PR. Can you add the new parameter to `workflow_call`?

.github/workflows/e2e-leg-4.yml
```diff
@@ -15,6 +15,9 @@ on:
       vertica-deployment-method:
         type: string
         required: false
+      vertica-superuser-name:
+        type: string
+        required: false
```
spilchen, line 20: Should we set the default?
```suggestion
        required: false
        default: "myadmin"
```

cchen-vertica: I set the default value in workflow_dispatch.

spilchen (approved):

PR #581, opened 2023-11-08
Uninstall reconciler is a no-op for vclusterops

chinhtranvan:
This PR put a check at the beginning of the uninstall reconciler to be an early out for vcluster ops


pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -131,11 +132,17 @@ func (r *VerticaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	pfacts := MakePodFacts(r, prunner)
 	dispatcher := r.makeDispatcher(log, vdb, prunner, passwd)
 	var res ctrl.Result
-
+	const UninstallReconciler = "UninstallReconciler"
 	// Iterate over each actor
 	actors := r.constructActors(log, vdb, prunner, &pfacts, dispatcher)
 	for _, act := range actors {
-		log.Info("starting actor", "name", fmt.Sprintf("%T", act))
+		actName := fmt.Sprintf("%T", act)
```
spilchen, line 139: I realize this is WIP, but we should just add the `UseVClusterOps` check right in `uninstall_reconciler.go`. It is fine to call that and perform a no-op for vclusterops deployments.

roypaulin: Yes, like we have done for `ScheduleOnly` in `(s *UninstallReconciler) Reconcile()`

chinhtranvan:
@spilchen I look at closer some e2e test failure, and understand why this change can cause failure for these tests.
E..g k-safety-1-scaling at e2e-leg-1. This step scale up from 3 to 5 pods and scale down from 5 to 3.  The problem when we add nodes for pod3 and pod4, we occur an license issue error failure from add node step in HTTPSCreateNodeOp. Then we skip the uninstall reconciler, we will fail at obj_reconciler `Requeue since some pods still need db_remove_node and uninstall done.`. I look closer at obj_reconciler https://github.com/vertica/vertica-kubernetes/blob/main/pkg/controllers/vdb/obj_reconciler.go#L494, I see that `pod4.isInstalled =True` (because we skip uninstall reconciler).

spilchen:
> @spilchen I looked closer at some e2e test failures and understood why this change could cause failures for these tests. For example, in the k-safety-1-scaling at e2e-leg-1, this step involves increasing the number of pods to install Vertica from 3 to 5 and decreasing the number of pods to uninstall from 5 to 3. Scaling up is not an issue because we have makeInstallReconcile to handle that. The main issue arises during scale down. If we skip the uninstall reconciler, we cannot uninstall from 5 to 3 pods.
> 
> We skipped the uninstall reconciler, leading to a failure at obj_reconciler with the message `Requeue since some pods still need db_remove_node and uninstall done.` , I noticed that there are two variables to consider to get this error - pod.isInstalled or pod.DBExist https://github.com/vertica/vertica-kubernetes/blob/main/pkg/controllers/vdb/obj_reconciler.go#L49. If we skip uninstall reconciler for pod3 and pod4, pod.isInstalled is still True and we failed at step 10 scale down.

I think it is safe to skip that check in obj_reconciler for vclusterops deployments. We don't have any install state for vclusterops. I think Roy's change to have a static httpscerts.json file will probably mean we can skip the installer.

chinhtranvan:
> > @spilchen I looked closer at some e2e test failures and understood why this change could cause failures for these tests. For example, in the k-safety-1-scaling at e2e-leg-1, this step involves increasing the number of pods to install Vertica from 3 to 5 and decreasing the number of pods to uninstall from 5 to 3. Scaling up is not an issue because we have makeInstallReconcile to handle that. The main issue arises during scale down. If we skip the uninstall reconciler, we cannot uninstall from 5 to 3 pods.
> > We skipped the uninstall reconciler, leading to a failure at obj_reconciler with the message `Requeue since some pods still need db_remove_node and uninstall done.` , I noticed that there are two variables to consider to get this error - pod.isInstalled or pod.DBExist https://github.com/vertica/vertica-kubernetes/blob/main/pkg/controllers/vdb/obj_reconciler.go#L49. If we skip uninstall reconciler for pod3 and pod4, pod.isInstalled is still True and we failed at step 10 scale down.
> 
> I think it is safe to skip that check in obj_reconciler for vclusterops deployments. We don't have any install state for vclusterops. I think Roy's change to have a static httpscerts.json file will probably mean we can skip the installer.

After I skip that check in obj_reconciler for vclusterops deployments, I will pass e2e tests :)

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -491,10 +491,13 @@ func (o *ObjReconciler) checkIfReadyForStsUpdate(newStsSize int32, sts *appsv1.S
 		if !ok {
 			return ctrl.Result{}, fmt.Errorf("could not find pod facts for pod '%s'", pn)
 		}
-		if pf.isInstalled || pf.dbExists {
-			o.Log.Info("Requeue since some pods still need db_remove_node and uninstall done.",
-				"name", pn, "isInstalled", pf.isInstalled, "dbExists", pf.dbExists)
-			return ctrl.Result{Requeue: true}, nil
+		// Skip this check since there is no install state for vclusterops
+		if !vmeta.UseVClusterOps(o.Vdb.Annotations) {
```
spilchen, line 495: Lets move this check to the top of the function. No need to go into the loop if we are running vclusterops.

PR #582, opened 2023-11-09
vdb-gen supports adding annotation for deployment method

jizhuoyu:
- vdb-gen has a command line option to force a deployment method
- vdb-gen deduces default deployment method from server version in the live database (>= v24.1.0 use vclusterops)
- vdb-gen adds `SuperuserNameAnnotation` according to command line option `-user`
- modify vdb-gen e2e test to support this new command line option and support testing under 2 modes (AT and VC)


cmd/vdb-gen/main.go
```diff
@@ -73,8 +73,19 @@ func main() {
 		"If the communal backend is authenticated with Kerberos, use this parameter to pass in the contents of the krb5.keytab file")
 	flag.StringVar(&opts.DepotVolume, "depotvolume", "PersistentVolume",
 		"The type of volume to use for the depot. Allowable values will be: EmptyDir and PersistentVolume.")
+	flag.StringVar(&opts.DeploymentMethod, "deploymentmethod", "",
```
roypaulin, line 76: `deployment-method`

pkg/vdbgen/vdb.go
```diff
@@ -669,6 +677,21 @@ func (d *DBGenerator) setImage(ctx context.Context) error {
 	// hotfix 0 regardless of what hotfix was currently in use.
 	d.Objs.Vdb.Spec.Image = fmt.Sprintf("vertica/vertica-k8s:%s-0", version)
 
+	// Set proper annotation to ensure correct deployment method.
+	if _, exists := d.Objs.Vdb.Annotations[vmeta.VClusterOpsAnnotation]; !exists {
```
roypaulin, line 681: You can move this into its own function.

cmd/vdb-gen/main.go
```diff
@@ -73,8 +73,19 @@ func main() {
...
+		fmt.Sprintf("The cluster deployment method to use by the operator. Allowable values will be: %s and %s. If not specified "+
+			"a default deployment method will be deduced from the server version of the live database.",
+			vdbgen.DeploymentMethodAT, vdbgen.DeploymentMethodVC))
 	flag.Parse()
 
+	if opts.DeploymentMethod != "" &&
```
spilchen, line 82: Instead of setting up the flag via `flag.StringVar`, I think we can use `flag.Func` to implement this parse logic. Can you try that?

pkg/vdbgen/vdb.go
```diff
@@ -669,6 +677,21 @@ func (d *DBGenerator) setImage(ctx context.Context) error {
...
+		// command line option not provided, i.e. no forced deployment method, thus should
+		// determine deployment method based on running server version
+		verInfo, ok := vversion.MakeInfoFromStr("v" + version)
+		if !ok {
+			return errors.New("could not construct Info struct from the version string")
```
spilchen, line 686: It might be good to include the text of the version in the message.

tests/e2e-leg-4/vdb-gen/70-run-vdbgen.yaml
```diff
@@ -15,7 +15,7 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
   # We have already copied vdb-gen at step 45. Now we run it again, in the container.
-  - script: sh -c "kubectl -n $NAMESPACE exec v-vdb-gen-sc1-0 -- sh -c \"/tmp/run-vdb-gen.sh $NAMESPACE $(cd ../../.. && make echo-images | grep ^VERTICA_IMG= | cut -d'=' -f2) v-vdb-gen-revive2 $(kubectl get cm e2e -n $NAMESPACE -o jsonpath='{.data.caFileSecretName}')\" > /tmp/$NAMESPACE-vdb-gen.yaml"
+  - script: sh -c "kubectl -n $NAMESPACE exec v-vdb-gen-sc1-0 -- sh -c \"/tmp/run-vdb-gen.sh $NAMESPACE $(cd ../../.. && make echo-images | grep ^VERTICA_IMG= | cut -d'=' -f2) v-vdb-gen-revive2 $(kubectl get cm e2e -n $NAMESPACE -o jsonpath='{.data.caFileSecretName}') $VERTICA_DEPLOYMENT_METHOD\" > /tmp/$NAMESPACE-vdb-gen.yaml"
```
spilchen, line 18: What if `$VERTICA_DEPLOYMENT_METHOD` isn't set? We should have some default logic somewhere. The current logic for e2e tests is we'll use admintools if that var isn't set.

cmd/vdb-gen/main.go
```diff
@@ -60,19 +60,30 @@ func main() {
...
 		"If the communal backend is authenticated with Kerberos, use this parameter to pass in the contents of the krb5.keytab file")
-	flag.StringVar(&opts.DepotVolume, "depotvolume", "PersistentVolume",
+	flag.StringVar(&opts.DepotVolume, "depot-volume", "PersistentVolume",
 		"The type of volume to use for the depot. Allowable values will be: EmptyDir and PersistentVolume.")
+	flag.Func("deployment-method",
+		fmt.Sprintf("The cluster deployment method to use by the operator. Allowable values will be: %s and %s. If not specified "+
+			"a default deployment method will be deduced from the server version of the live database.",
+			vdbgen.DeploymentMethodAT, vdbgen.DeploymentMethodVC),
+		func(method string) error {
+			if method != vdbgen.DeploymentMethodAT && method != vdbgen.DeploymentMethodVC {
```
roypaulin, line 81: What if method is an empty string?

jizhuoyu: I don't think a user would be able to specify `-deployment-method` without providing an argument. If the user specified `-deployment-method ""` I think this should be treated as an error

jizhuoyu: image

roypaulin (approved):

PR #584, opened 2023-11-09
Remove https tls certs from v2 image

roypaulin:
https certs have been removed from the image and can now be generated by the vertica bootstrap catalog. This means we do not need the operator to generate the certs in vclusterops mode and can then skip install/uninstall for vclusterops.
A new option has been added to createdb to trigger https certs generation by vertica bootstrap catalog


roypaulin:
Needs the latest vclusterops.

pkg/builder/builder.go
```diff
@@ -126,7 +126,8 @@ func BuildHlSvc(nm types.NamespacedName, vdb *vapi.VerticaDB) *corev1.Service {
 }
 
 // buildConfigVolumeMount returns the volume mount for config.
-// If vclusterops flag is enabled we mount only /opt/vertica/config/https_certs
+// If vclusterops flag is enabled we mount only
+// /opt/vertica/config/node_management_agent.pid
```
spilchen, line 130: Is this true? It looks like we don't have any special logic for vclusterops now.

pkg/controllers/vdb/init_db.go
```diff
@@ -132,7 +133,15 @@ func (g *GenericDatabaseInitializer) checkPodList(podList []*PodFact) bool {
 		// - doesn't have the annotations that we use in the k8s Vertica DC
```
spilchen, line 133: Lets update the comment (one line above).

spilchen (approved):
Changes look good. I like how it cleans up a bunch of logic. There are a couple of minor comments that I trust you will fix, so I am approving.

roypaulin:
Thanks Matt. Do you know how I can get myself a V2 image built with the latest dockerfile?

pkg/builder/builder.go
roypaulin: Yes, if we don't persist /opt/vertica/config/node_management_agent.pid, nma will hit permission denied.
It is the same reason why we still need to have a volume for/opt/vertica/log

PR #585, opened 2023-11-09
Trim nodes in re_ip and start_db to avoid failures from wrong input

le-an-vertica:
This PR synced the new changes from `vclusterops` and set the "trim node" options to be true in `re_ip` and `start_db` to avoid possible wrong input in case the operators provide more hosts/nodes than needed.

### DIFF
--- a/pkg/vadmin/start_db_vc.go
+++ b/pkg/vadmin/start_db_vc.go
@@ -86,9 +86,10 @@ func (v *VClusterOps) genStartDBOptions(s *startdb.Parms, certs *HTTPSCerts) (vo
 	opts.CaCert = certs.CaCert
 	*opts.UserName = v.VDB.GetVerticaUser()
 	opts.Password = &v.Password
+	*opts.TrimHostList = true
 	*opts.HonorUserInput = true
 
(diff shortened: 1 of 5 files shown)


spilchen (approved):

pkg/vadmin/start_db_vc.go
```diff
@@ -89,6 +89,12 @@ func (v *VClusterOps) genStartDBOptions(s *startdb.Parms, certs *HTTPSCerts) (vo
 	*opts.HonorUserInput = true
 
 	// timeout option
-	opts.StatePollingTimeout = v.VDB.GetRestartTimeout()
+	opts.StatePollingTimeout = new(int)
```
roypaulin, line 92: Shouldn't we initialize this in vclusterops?

le-an-vertica: Yes, we should. I will create a small JIRA to track this. 
You may help me merge this and let QA test. I’m curious to know if there’s any other problems (which may happen in other steps other than re_ip or start_db) when running kill pod during remove_node.

PR #586, opened 2023-11-09
Added goroutine ID to reconciler log for debugging purpose

cchen-vertica:
This code change is only for debugging intermittent errors in "update-strategy" test of e2e-leg-2. After the issue is fixed, we should revert the changes in this PR since retrieving goroutine ID is not recommended in golang.

### DIFF
--- a/pkg/controllers/vdb/go_id_helper.go
+++ b/pkg/controllers/vdb/go_id_helper.go
@@ -1,0 +1,50 @@
+/*
+ (c) Copyright [2021-2023] Open Text.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ You may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+// Remove this file after VER-89903 is closed since the function
+// goid() inside this file is slow and not recommended by Go.
+package vdb
+
+import (
+	"bytes"
+	"errors"
+	"runtime"
+	"strconv"
+)
+
+const bufferSize = 32
+
+var (
+	goroutinePrefix = []byte("goroutine ")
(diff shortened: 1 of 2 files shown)


cchen-vertica:
> Given that it is temporary do you have to merge it?

Yes, this issue only can be reproduced in Github. We need to merge it and wait for that issue to resurface. Then we can get more information from Github log.

spilchen:
Where did you read that retrieving the goroutine ID is not recommended? I would like to have a permanent solution to identify the goroutines in the log messages.

cchen-vertica:
> Where did you read that retrieving the goroutine ID is not recommended? I would like to have a permanent solution to identify the goroutines in the log messages.

Go removed GoID(). Here is the go document: https://go.dev/doc/faq#no_goroutine_id. If each goroutine in the operator only executes the function Reconcile() once, we can generate a UUID for every Reconcile() call.

spilchen:
> > Where did you read that retrieving the goroutine ID is not recommended? I would like to have a permanent solution to identify the goroutines in the log messages.
> 
> Go removed GoID(). Here is the go document: https://go.dev/doc/faq#no_goroutine_id. If each goroutine in the operator only executes the function Reconcile() once, we can generate a UUID for every Reconcile() call to simulate a goroutine id.

Thanks for the background. This is good to know. I like your idea of assigning a UUID each time the Reconcile function is called. At least then you can trace the calls for a reconcile more easily.

cchen-vertica:
> > > Where did you read that retrieving the goroutine ID is not recommended? I would like to have a permanent solution to identify the goroutines in the log messages.
> > 
> > 
> > Go removed GoID(). Here is the go document: https://go.dev/doc/faq#no_goroutine_id. If each goroutine in the operator only executes the function Reconcile() once, we can generate a UUID for every Reconcile() call to simulate a goroutine id.
> 
> Thanks for the background. This is good to know. I like your idea of assigning a UUID each time the Reconcile function is called. At least then you can trace the calls for a reconcile more easily.

The controller has tagged a UUID to each reconcile. We can get it from context. However, I found we had 5 fixed goroutines that can run multiple reconciles. I'd like to use the current code(go_id_helper.go) for debugging the e2e test error first. The reason of using goroutineID is that we can know if the redundant create_db reconcile is executed in the same goroutine or not. After the error is fixed, I can change goroutineID to reconcilerID.

roypaulin (approved):

PR #588, opened 2023-11-15
Add developer CI test for server log rotate

chinhtranvan:
There are some changes in this PR:
- remove cron from the image
- remove setup of logrotate config files from the image
- update the operator's install reconciler to avoid logrotate setup for 24.1.0 database (we still need to keep it for older ones)
- add a test to GitHub CI to verify in-server log rotate
- manually verify the upgrade scenario. When we upgrade to 24.1.0 we should set the config parm set to indicate in-database log rotate is active.


roypaulin:
The new config param is not specific to vclusterops so I think we should also update `docker-vertica`

chinhtranvan:
> The new config param is not specific to vclusterops so I think we should also update `docker-vertica`

Thank you for your comments. The thing I wonder about is whether, if we update docker-vertica to remove the log rotate setup and cron, we will encounter a failure at install_reconciler` with the error message: cp: cannot stat '/home/dbadmin/logrotate/logrotate/admintool.logrotate': No such file or directory. This is because some older database versions need to use the logrotate setup. Docker-vertica is used in versions 11.0.1 or higher, but this configuration parameter is used from version 24.1.0+

tests/e2e-leg-6/verify-server-logrotate/10-assert.yaml
```diff
@@ -0,0 +1,23 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-verify-server-logrotate
+status:
+  containerStatuses:
+    - name: test
+      state:
+        terminated:
+          exitCode: 0  # If ksafe value is wrong, exit code will be 5
```
roypaulin, line 23: Update the comment.

tests/e2e-leg-6/verify-server-logrotate/expected-output/logrotateTimerServiceOut.txt
```diff
@@ -0,0 +1,63 @@
...
+(1 row)
+
+     node_name     | success | max_size  | max_age  |                               log_file                                | need_rotation 
+-------------------+---------+-----------+----------+-----------------------------------------------------------------------+---------------
+ v_vertdb_node0001 | t       | 104857600 | 7        | /data/vertdb/v_vertdb_node0001_catalog/vertica.log                    | f
+ v_vertdb_node0001 | t       | 104857600 | 7        | /data/vertdb/v_vertdb_node0001_catalog/UDxLogs/UDxFencedProcesses.log | f
+ v_vertdb_node0001 | t       |      1024 | 7        | /data/vertdb/v_vertdb_node0001_catalog/vertica.log                    | t
+ v_vertdb_node0001 | t       |      1024 | 7        | /data/vertdb/v_vertdb_node0001_catalog/UDxLogs/UDxFencedProcesses.log | f
+ v_vertdb_node0001 | t       |      1024 | 00:00:01 | /data/vertdb/v_vertdb_node0001_catalog/vertica.log                    | t
+ v_vertdb_node0001 | t       |      1024 | 00:00:01 | /data/vertdb/v_vertdb_node0001_catalog/UDxLogs/UDxFencedProcesses.log | f
```
roypaulin, line 44: Why does max_age has a different format in the first 4 rows?

tests/e2e-leg-6/verify-server-logrotate/sql-input/logrotate.sql
```diff
@@ -0,0 +1,35 @@
+-- Cleanup the DC table
```
roypaulin, line 1: Can you first check the config param `EnableLogrotate`? So that if it is not 1 it fails here. 
If one query  returns an error are we still going to run the rest of them? If yes can we exit once one query fails?

roypaulin:
Did you:
- manually verify the upgrade scenario?
- Can you also manually try you logrotate script with a 24.1 admintools deployment(with an  image you build with the latest Dockerfile

tests/e2e-leg-6/verify-server-logrotate/21-assert.yaml
```diff
@@ -0,0 +1,19 @@
+# (c) Copyright [2021-2023] Open Text.
```
roypaulin, line 1: can you make this step 96

tests/e2e-leg-6/verify-server-logrotate/20-delete-cr.yaml
```diff
@@ -0,0 +1,18 @@
+# (c) Copyright [2021-2023] Open Text.
```
roypaulin, line 1: Can you make this step 95

tests/e2e-leg-6/verify-server-logrotate/expected-output/logrotateTimerServiceOut.txt
chinhtranvan: because we set LogRotateMaxAge to 1sec to rotate log

tests/e2e-leg-6/verify-server-logrotate/sql-input/logrotate.sql
chinhtranvan: 1. We don't need to do it; select hurry_service('SYSTEM', 'LogRotate', 60) will handle it. If we don't set the config parameter EnableLogrotate to 1, it will fail immediately at hurry_service.
2. Yes, I can use one more if-else condition to run queries. If any queries fail, we can return an error.

roypaulin:
It fails because all those tests use an old vertica version which relies on what was removed from docker-vertica. So yes let's not change docker-vertica. Sorry you wasted a lot of time because of that.

tests/e2e-leg-6/verify-server-logrotate/expected-output/logrotateTimerServiceOut.txt
roypaulin: Then why is it 7 in the first 4 rows?

roypaulin:
Did you verify the upgrade scenario mentioned on jira?

tests/e2e-leg-6/verify-server-logrotate/expected-output/logrotateTimerServiceOut.txt
chinhtranvan: because we use default config param first, default Age is 7

chinhtranvan: Link Ref: https://git.verticacorp.com/projects/VER/repos/server/pull-requests/19596/diff#vertica/SQLTest/expected/logrotate_timer_service_test.out

roypaulin:
Let's also add a `README.txt

chinhtranvan:
> Did you:
> 
> * manually verify the upgrade scenario?
> * Can you also manually try you logrotate script with a 24.1 admintools deployment(with an  image you build with the latest Dockerfile



> Did you verify the upgrade scenario mentioned on jira?

1. manually try the logrotate script with a 24.1 admintools, I run script successfully, I can connect to vsql to schedule task to rotate
2. For verify upgrade scenario, I try with  `upgradePolicy: Offline` with the image is pulled from rpm 24.1.0, I still connect to vsql to schedule task to rotate. I'm not sure if this verify is correct
This is my yaml file

kind: VerticaDB
metadata:
  name: vc
spec:
  image: vertica-k8s:24.1.0
  imagePullPolicy: IfNotPresent
  communal:
    path: "/communal"
    includeUIDInPath: true
  upgradePolicy: Offline
  kSafety: "1"
  local:
    dataPath: /data
    catalogPath: /catalog
    depotPath: /depot
  subclusters:
    - name: pri
      size: 3
  volumes:
  - name: hostpath
    hostPath:
      path: /host
  volumeMounts:
  - name: hostpath
    mountPath: /communal
    subPath: Linux64/Test/k8s

vertdb=> select hurry_service('SYSTEM', 'LogRotate', 60);
        hurry_service         
------------------------------
 Service hurried and finished
(1 row)

tests/e2e-leg-6/verify-server-logrotate/README.txt
```diff
@@ -0,0 +1,12 @@
+This is a test suite that verifies the server logrotate
+Design: https://confluence.verticacorp.com/pages/viewpage.action?spaceKey=DEV&title=Spec+for+logrotate+in+the+Server
```
roypaulin, line 2: We do not put confluence link on github. Also this not need to be this big we just want a summary of what the test does(see other README.txt

roypaulin (approved):
I am approving this (assuming you address my minor comment) but I will wait before merging it. I just want to ask Matt about the v1 image first.

spilchen (approved):

PR #590, opened 2023-11-16
Make spread encryption the default for v1 API create_db

cchen-vertica:
For v1 api, set 'vertica' as the default value of encryptSpreadComm: when encryptSpreadComm is set to an empty string or not set, the operator will also enable spread encryption when creating a database. Add another value "disabled" to encryptSpreadComm for the user to explicitly disable spread encryption.


pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -175,7 +175,7 @@ func (c *CreateDBReconciler) generatePostDBCreateSQL(ctx context.Context, initia
 	// config parm in the create db call.
 	if c.Vdb.Spec.EncryptSpreadComm != vapi.EncryptSpreadCommDisabled && ok && vinf.IsOlder(vapi.SetEncryptSpreadCommAsConfigVersion) {
 		sb.WriteString(fmt.Sprintf(`alter database default set parameter EncryptSpreadComm = '%s';
-		`, c.Vdb.Spec.EncryptSpreadComm))
+		`, vapi.EncryptSpreadCommWithVertica))
```
roypaulin, line 178: Why did you do this? What if a user wants to disable spread encryption(if that even makes sense)?

pkg/controllers/vdb/init_db_test.go
```diff
@@ -147,10 +147,13 @@ var _ = Describe("init_db", func() {
 		res, err := g.ConstructConfigParms(ctx)
 		ExpectWithOffset(1, err).Should(Succeed())
 		ExpectWithOffset(1, res).Should(Equal(ctrl.Result{}))
-		Expect(g.ConfigurationParams.Size()).Should(Equal(1))
+		Expect(g.ConfigurationParams.Size()).Should(Equal(2))
 		v, ok := g.ConfigurationParams.Get("InitialDefaultSubclusterName")
 		Expect(ok).Should(BeTrue())
 		Expect(v).Should(Equal(vdb.Spec.Subclusters[0].Name))
+		v, ok = g.ConfigurationParams.Get("EncryptSpreadComm")
```
roypaulin, line 154: I think you can create a new test function for spread encryption where you can directly call `setEncryptSpreadCommConfigIfNecessary()` and have a few scenarios by changing `EncryptSpreadComm` and vertica version.

tests/e2e-leg-1/autoscale-by-pod/setup-vdb/base/setup-vdb.yaml
```diff
@@ -36,3 +36,4 @@ spec:
   imagePullSecrets: []
   volumes: []
   volumeMounts: []
+  encryptSpreadComm: disabled
```
roypaulin, line 39: Why didn't you do the same for other e2e tests?

pkg/controllers/vdb/createdb_reconciler.go
cchen-vertica: The code here is for enabling spread encryption. The if statement above makes sure the user does not want to disable spread encryption. When encryptSpreadComm is an empty string, 'c.Vdb.Spec.EncryptSpreadComm' won't give the correct value to the ddl, but 'vapi.EncryptSpreadCommWithVertica' will.

tests/e2e-leg-1/autoscale-by-pod/setup-vdb/base/setup-vdb.yaml
cchen-vertica: I think encrypting spread channel will not affect other e2e tests. And since "enable" is the default behavior now, we can test the default behavior in other e2e tests. 

It turns out the admintool e2e tests are using old Vertica version(12.0.3) that does not support spread channel encryption without a db restart. Do you think should we update the Vertica version for the admintool e2e tests or should I disable spread channel encryption for the admintool e2e tests?

pkg/controllers/vdb/config.go
```diff
@@ -582,3 +582,12 @@ func genUnsupportedVerticaVersionEventMsg(feature, supportedVersion string) stri
 	prefix := "The engine (%s) doesn't have the required change to setup"
 	return fmt.Sprintf("%s %s. You must be on version %s or greater", prefix, feature, supportedVersion)
 }
+
+// getEncryptSpreadComm will return "vertica" if encryptSpreadComm is set to
+// an empty string, otherwise return the value of encryptSpreadComm
+func (g *ConfigParamsGenerator) getEncryptSpreadComm() string {
```
spilchen, line 588: Can we move this to VeritcaDB? And rename it so that it's exported (i.e. `GetEncryptSpreadComm`). There is a file (helpers.go) that has these kind of functions.

spilchen (commented):
I suggest we update the broken e2e testcases to explicitly disable spread encryption.

spilchen (approved):

PR #591, opened 2023-11-17
webhook validates deployment type changes

jizhuoyu:
- deployment type cannot change from vclusterops to admintools (the webhook will report this as an error)
- deployment type can change from admintools to vclusterops, assuming image is also changed from v1 to v2 (verified manually by `kubectl edit`)


roypaulin (approved):
Looks good! there is an e2e-test for webhook (vdb-webhook) but I don't think this one fits there. so I am approving.

PR #592, opened 2023-11-20
Change upgrade path rules for new server versions

roypaulin:
With the new server versions, the minor is changing quite frequently. Each release is essentially a minor version change. And that is happening every quarter. So we added a rule so that the x.4.0 cannot be skipped but intermediate ones can. For instance. If someone is on 23.4.0, they can directly to 24.4.0 skipping 24.1.0, 24.2.0 and 24.3.0

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -30,6 +30,11 @@ type Info struct {
 	Components        // The same version as VdbVer but broken down into individual components
 }
 
+const (
+	LTSMinor = 4
+	LTSPatch = 0
+)
+
 // UpgradePaths has all of the vertica releases supported by the operator.  For
 // each release, the next release that must be upgrade too.  Use this map to
 // know if a new version is the next supported version by Vertica.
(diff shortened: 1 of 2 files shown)


pkg/version/version.go
```diff
@@ -49,6 +49,12 @@ var UpgradePaths = map[Components]Info{
 	{12, 0, 2}: {"v23.3.x", Components{23, 3, 0}},
```
spilchen, line 49: We want to allow 12.0.x to go directly to the next LTS (23.4.0).

pkg/version/version.go
```diff
@@ -143,7 +156,7 @@ func (i *Info) IsValidUpgradePath(targetVer string) (ok bool, failureReason stri
 		// version of vertica that came out after the version of this operator.
```
spilchen, line 156: Since our release versions are more predictable, we can generalize this logic better. Basically if we are on x.y.0, if y < 4 we can go to x.4.0, if y == 4, we can go to x+1.4.0.

pkg/version/version.go
```diff
@@ -49,6 +49,12 @@ var UpgradePaths = map[Components]Info{
 	{12, 0, 2}: {"v23.3.x", Components{23, 3, 0}},
 	{12, 0, 3}: {"v23.3.x", Components{23, 3, 0}},
 	{12, 0, 4}: {"v23.3.x", Components{23, 3, 0}},
+	// From here upwards the x.4.0 release (4th quarter release)
+	// is the LTS version so all releases will have the closest x.4.0
+	// as their next release they can be upgraded to.
+	// example:
+	//	- {23, 4, 0}: {"v24.4.x", Components{24, 4, 0}}
+	//	- {24, 1, 0}: {"v24.4.x", Components{24, 4, 0}}
```
spilchen, line 57: Can we add some UT to make sure we got the logic correctly? Some cases to confirm:
- 12.0.3 -> 12.0.4 (Allowed)
- 12.0.3 -> 23.3.0 (Allowed)
- 12.0.3 -> 23.4.0 (Allowed)
- 12.0.3 -> 24.4.0 (Blocked)
- 12.0.3 -> 24.1.0 (Blocked)
- 23.3.0 -> 23.4.0 (Allowed)
- 23.3.0 -> 24.2.0 (Blocked)
- 23.3.0 -> 24.3.0 (Blocked)

pkg/version/version.go
```diff
@@ -109,6 +115,13 @@ func (i *Info) IsEqualExceptPatch(other *Info) bool {
 	return other.VdbMajor == i.VdbMajor && other.VdbMinor == i.VdbMinor
 }
 
+// IsLowerExceptPatch returns true if the receiver's major/minor version
+// is lower than the given version.
+func (i *Info) IsLowerExceptPatch(other *Info) bool {
```
spilchen, line 120: Prefer to use newer/older terms like other function. Can you rename this?
```suggestion
func (i *Info) IsOlderExceptPatch(other *Info) bool {
```

spilchen (approved):

pkg/version/version.go
spilchen: Yes, I guess the map really isn't needed anymore for new releases.

PR #593, opened 2023-11-20
Remove v1beta1 webhook

chinhtranvan:
This PR simply removes the v1beta1 webhook so that when we apply a v1 API, it only applies the v1 webhook


chinhtranvan:
I checked Zhuoyi's manifest.yml file. I tested it 30 times, and it didn't show any errors from v1beta1. The errors only came from the v1 webhook.
```
The VerticaDB "v-wrong-checkverticavdb1-scs21-scwe213s" is invalid: 
* spec.subclusters[0].Type: Invalid value: "ppp": subcluster type is invalid. A valid case-sensitive type a user can specify is "primary" or "secondary". ("transient" is a valid type that should only be set by the operator during online upgrade)
* spec.subclusters: Invalid value: []v1.Subcluster{v1.Subcluster{Name:"sc1", Size:3, Type:"ppp", ImageOverride:"", NodeSelector:map[string]string(nil), Affinity:v1.Affinity{NodeAffinity:(*v1.NodeAffinity)(nil), PodAffinity:(*v1.PodAffinity)(nil), PodAntiAffinity:(*v1.PodAntiAffinity)(nil)}, PriorityClassName:"", Tolerations:[]v1.Toleration(nil), Resources:v1.ResourceRequirements{Limits:v1.ResourceList(nil), Requests:v1.ResourceList(nil), Claims:[]v1.ResourceClaim(nil)}, ServiceType:"ClusterIP", ServiceName:"sc1", ClientNodePort:0, VerticaHTTPNodePort:0, ExternalIPs:[]string(nil), LoadBalancerIP:"", ServiceAnnotations:map[string]string(nil)}}: there must be at least one primary subcluster
```

spilchen (approved):
Looks good!

PR #594, opened 2023-11-20
Decide k-safety value on the number of primary nodes

le-an-vertica:
Currently, the operator's webhook calculates the k-safety value based on the number of all nodes rather than the number of primary nodes.

This PR will correct this as the server calculates the k-safety value based on the number of primary nodes.

In addition, to back port the `v1beta1`, we introduced `RelaxKSafetyCheckAnnotation` to keep the old behaviour for `v1beta1` deployments.


api/v1/verticadb_webhook.go
```diff
@@ -380,10 +381,23 @@ func (v *VerticaDB) validateKsafety(allErrs field.ErrorList) field.ErrorList {
 
 func (v *VerticaDB) getClusterSize() int {
 	sizeSum := 0
-	for i := range v.Spec.Subclusters {
-		sc := &v.Spec.Subclusters[i]
-		sizeSum += int(sc.Size)
+
+	if v.Annotations[vmeta.RelaxKSafetyCheckAnnotation] == strconv.FormatBool(true) {
```
roypaulin, line 385: I know this is still a drag but this is not how you should deal with annotations. See how it is done for other annotations in annotations.go

api/v1beta1/verticadb_types.go
```diff
@@ -217,6 +217,15 @@ type VerticaDBSpec struct {
...
 	KSafety KSafetyType `json:"kSafety,omitempty"`
 
+	// +kubebuilder:default:=false
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// When set to true, the webhook will calculate the k-safety value
+	// based on the number of primary nodes in the cluster;
+	// otherwise, the calculation will be based on the number of all nodes
+	// in the cluster.
+	RelaxKSafetyCheck bool `json:"relaxKSafetyCheck,omitempty"`
```
roypaulin, line 227: you don't need to add a new field to the CR spec, only the annotation, see how we do with vcluster-ops annotation.

pkg/meta/annotations.go
```diff
@@ -83,6 +83,12 @@ const (
 	KSafetyAnnotation   = "vertica.com/k-safety"
 	KSafetyDefaultValue = "1"
 
+	// When set to true, the webhook will calculate the k-safety value
+	// based on the number of primary nodes in the cluster;
+	// otherwise, the calculation will be based on the number of all nodes
+	// in the cluster.
+	RelaxKSafetyCheckAnnotation = "vertica.com/relax-k-safety-check"
```
roypaulin, line 90: This is where you define the function to check the annotation, see other book annotations as example.

pkg/meta/annotations.go
```diff
@@ -83,6 +83,12 @@ const (
 	KSafetyAnnotation   = "vertica.com/k-safety"
 	KSafetyDefaultValue = "1"
 
+	// When set to true, the calculation will be based on the number of all nodes
+	// in the cluster;
+	// otherwise, the webhook will calculate the k-safety value
+	// based on the number of primary nodes in the cluster;
+	RelaxKSafetyCheckAnnotation = "vertica.com/relax-k-safety-check"
```
spilchen, line 90: Could we rename the annotation so that the default is true? 
```suggestion
	StrictKSafetyCheckAnnotation = "vertica.com/strict-k-safety-check"
```

spilchen (approved):

PR #595, opened 2023-11-21
Run legs against previously released vertica image

chinhtranvan:
This PR adds two previously images,` vertica/vertica-k8s:23.4.0-0-minimal` and `vertica/vertica-k8s:12.0.4-0-minimal`, in the available e2e-leg*.yml files. Additionally, a parameter `artifact_suffix` is introduced to avoid duplicating the artifact file for each run of the leg.

### DIFF
--- a/.github/workflows/e2e-leg-1.yml
+++ b/.github/workflows/e2e-leg-1.yml
@@ -12,7 +12,7 @@ on:
       vertica-image:
         type: string
         required: false
-      artifact-version:
+      artifact-suffix:
         type: string
         required: false
       vertica-deployment-method:
@@ -39,8 +39,8 @@ on:
         description: 'Name of the vertica server image'
         type: string
         required: false
-      artifact-version:
-        description: 'Version of the artifact file'
+      artifact-suffix:
+        description: 'Version of the suffix artifact file'
         type: string
         required: false
       vertica-deployment-method:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-leg-1.yml
```diff
@@ -79,5 +86,5 @@ jobs:
     - uses: actions/upload-artifact@v3
       if: failure()
       with:
-        name: logs-e2e-leg-1-${{ inputs.vertica-deployment-method }}
+        name: logs-e2e-leg-1-${{ inputs.vertica-deployment-method }}-${{ inputs.artifact-version }}
```
spilchen, line 89: If we don't specify this new parm, the artifact name seems a little odd (e.g. logs-e2e-leg-1-admintools-). Could we call it artifact-suffix then just add it on without a dash? This would allow us to omit this parameter and still come up with a good name.
```suggestion
        name: logs-e2e-leg-1-${{ inputs.vertica-deployment-method }}${{ inputs.artifact-suffix }}
```

.github/workflows/e2e.yml
```diff
@@ -111,6 +111,20 @@ jobs:
...
+  
+  e2e-leg-1-admintools-previous-release:
+    if: ${{ inputs.e2e_test_suites == 'all' || inputs.e2e_test_suites == 'admintools leg 1' || inputs.e2e_test_suites == '' }}
+    needs: [build] 
+    uses: ./.github/workflows/e2e-leg-1.yml
+    with:
+      vlogger-image: ${{ needs.build.outputs.vlogger-image }}
+      operator-image: ${{ needs.build.outputs.operator-image }}
+      vertica-image: "vertica/vertica-k8s:23.4.0-0-minimal"
+      artifact-version: "23.4.0-release"
```
spilchen, line 123: Suggest changing this if you accept my other comment:
```suggestion
      artifact-suffix: "-23.4.0-release"
```

.github/workflows/e2e-leg-1.yml
chinhtranvan: Yes, It will be better, You're right, the artifact name seems a little odd if we don't specify the new param.

spilchen (approved):

PR #596, opened 2023-11-21
Operator mounts NMA certs in a volume by default

jizhuoyu:
- The operator respects the `vertica.com/mount-nma-certs` annotation when it is provided
- When it is not provided the operator assumes its value is `true` which means the NMA will read certs from a mounted secret volume (and proper env vars will be available in the db node containers)
- `serviceaccount_reconciler` skips role/rolebinding lookup and creation if `vertica.com/mount-nma-certs` is `true` or not provided (which will default to `true`), as that role is merely serving the purpose of granting the `get` and `list` permissions for secrets, which is only required when the NMA actually reads certs from k8s secrets directly.

### DIFF
--- a/pkg/builder/builder_test.go
+++ b/pkg/builder/builder_test.go
@@ -279,6 +279,13 @@ var _ = Describe("builder", func() {
 		Expect(NMACertsVolumeExists(vdb, ps.Volumes)).Should(BeTrue())
 		Expect(NMACertsVolumeMountExists(&c)).Should(BeTrue())
 		Expect(NMACertsEnvVarsExist(vdb, &c)).Should(BeTrue())
+		// test default value (which should be true)
+		delete(vdb.Annotations, vmeta.MountNMACerts)
+		ps = buildPodSpec(vdb, &vdb.Spec.Subclusters[0])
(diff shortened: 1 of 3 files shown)


jizhuoyu:
One observation I had is that, for a particular `vdb`, once we created the role/rolebinding in a reconcile loop (because the annotation indicates that no NMA certs should be mounted), even if later on we edit the `vdb` to flip the value of the annotation to `true` and let whole cluster to be reconstructed, it seems like the role will still exist and rolebinding will still be bound to the service account. Should we worry about this? @spilchen

spilchen (approved):

spilchen:
> One observation I had is that, for a particular `vdb`, once we created the role/rolebinding in a reconcile loop (because the annotation indicates that no NMA certs should be mounted), even if later on we edit the `vdb` to flip the value of the annotation to `true` and let whole cluster to be reconstructed, it seems like the role granting access to secrets will still exist and rolebinding will still be bound to the service account. Should we worry about this? @spilchen

I’m not worried about it. I don’t expect it to be a common occurrence to change when the NMA certs are mounted.

PR #598, opened 2023-11-22
Add feature flag to control whether NMA runs in monolithic or sidecar container

jizhuoyu:
- Add feature flag `vertica.com/run-nma-in-monolithic-container` which controls running mode of NMA, i.e. whether the NMA process runs in a monolithic container together with the main vertica process (which is the current behavior) or it runs in a sidecar container within the same pod as the server container (which will be supported in the future)
- The default value of this feature flag would be `false`
- Add logic in `setup-kustomize.sh` to explicitly set the flag to true for existing e2e tests to work

### DIFF
--- a/pkg/meta/annotations.go
+++ b/pkg/meta/annotations.go
@@ -48,9 +48,17 @@ const (
 	// This is a feature flag for mounting NMA certs as a secret volume in server containers
 	// if deployment method is vclusterops. When set to true the NMA reads certs from this mounted
 	// volume, when set to false it reads certs directly from k8s secret store.
-	MountNMACerts      = "vertica.com/mount-nma-certs"
-	MountNMACertsTrue  = "true"
-	MountNMACertsFalse = "false"
+	MountNMACertsAnnotation      = "vertica.com/mount-nma-certs"
+	MountNMACertsAnnotationTrue  = "true"
+	MountNMACertsAnnotationFalse = "false"
+
+	// This is a feature flag for running NMA process in a monolithic container together with
+	// vertica main process. When set to true the NMA process runs in a monolithic container
+	// together with vertica main process, when set to false it runs in a separate sidecar
+	// container.
+	RunNMAInMonolithicContainerAnnotation      = "vertica.com/run-nma-in-monolithic-container"
+	RunNMAInMonolithicContainerAnnotationTrue  = "true"
+	RunNMAInMonolithicContainerAnnotationFalse = "false"
 
 	// This is a feature flag for accessing the secrets configured in Google Secret Manager.
 	// The value of this annotation is treated as a boolean.
@@ -145,7 +153,11 @@ func UseVClusterOps(annotations map[string]string) bool {
 // UseNMACertsMount returns true if the NMA reads certs from the mounted secret
 // volume rather than directly from k8s secret store.
 func UseNMACertsMount(annotations map[string]string) bool {
-	return lookupBoolAnnotation(annotations, MountNMACerts, true /* default value */)
+	return lookupBoolAnnotation(annotations, MountNMACertsAnnotation, true /* default value */)
+}
+
+func RunNMAInMonolithicContainerMode(annotations map[string]string) bool {
+	return lookupBoolAnnotation(annotations, RunNMAInMonolithicContainerAnnotation, false /* default value */)
 }
 
 // UseGCPSecretManager returns true if access to the communal secret should go through
--- a/pkg/builder/builder_test.go
+++ b/pkg/builder/builder_test.go
@@ -267,20 +267,20 @@ var _ = Describe("builder", func() {
 	It("should mount or not mount NMA certs volume according to annotation", func() {
 		vdb := vapi.MakeVDBForHTTP("v-nma-tls-abcde")
 		vdb.Annotations[vmeta.VClusterOpsAnnotation] = vmeta.VClusterOpsAnnotationTrue
-		vdb.Annotations[vmeta.MountNMACerts] = vmeta.MountNMACertsFalse
+		vdb.Annotations[vmeta.MountNMACertsAnnotation] = vmeta.MountNMACertsAnnotationFalse
 		ps := buildPodSpec(vdb, &vdb.Spec.Subclusters[0])
 		c := makeServerContainer(vdb, &vdb.Spec.Subclusters[0])
 		Expect(NMACertsVolumeExists(vdb, ps.Volumes)).Should(BeFalse())
 		Expect(NMACertsVolumeMountExists(&c)).Should(BeFalse())
 		Expect(NMACertsEnvVarsExist(vdb, &c)).Should(BeTrue())
-		vdb.Annotations[vmeta.MountNMACerts] = vmeta.MountNMACertsTrue
+		vdb.Annotations[vmeta.MountNMACertsAnnotation] = vmeta.MountNMACertsAnnotationTrue
 		ps = buildPodSpec(vdb, &vdb.Spec.Subclusters[0])
 		c = makeServerContainer(vdb, &vdb.Spec.Subclusters[0])
 		Expect(NMACertsVolumeExists(vdb, ps.Volumes)).Should(BeTrue())
 		Expect(NMACertsVolumeMountExists(&c)).Should(BeTrue())
 		Expect(NMACertsEnvVarsExist(vdb, &c)).Should(BeTrue())
 		// test default value (which should be true)
-		delete(vdb.Annotations, vmeta.MountNMACerts)
+		delete(vdb.Annotations, vmeta.MountNMACertsAnnotation)
 		ps = buildPodSpec(vdb, &vdb.Spec.Subclusters[0])
 		c = makeServerContainer(vdb, &vdb.Spec.Subclusters[0])
 		Expect(NMACertsVolumeExists(vdb, ps.Volumes)).Should(BeTrue())
--- a/pkg/controllers/vdb/verticadb_controller.go
+++ b/pkg/controllers/vdb/verticadb_controller.go
@@ -173,6 +173,9 @@ func (r *VerticaDBReconciler) constructActors(log logr.Logger, vdb *vapi.Vertica
 		// Always start with a status reconcile in case the prior reconcile failed.
 		MakeStatusReconciler(r.Client, r.Scheme, log, vdb, pfacts),
 		MakeMetricReconciler(r, log, vdb, prunner, pfacts),
+		// Check NMA running mode and report error when configured to run in a sidecar
+		// as this is not supported yet
+		MakeNMARunningModeReconciler(r, log, vdb),
 		// Report any pods that have low disk space
 		MakeLocalDataCheckReconciler(r, vdb, pfacts),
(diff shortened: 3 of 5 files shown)


pkg/meta/annotations.go
```diff
@@ -48,9 +48,17 @@ const (
...
-	MountNMACertsFalse = "false"
+	MountNMACertsAnnotation      = "vertica.com/mount-nma-certs"
+	MountNMACertsAnnotationTrue  = "true"
+	MountNMACertsAnnotationFalse = "false"
+
+	// This is a feature flag for running NMA process in a monolithic container together with
+	// vertica main process. When set to true the NMA process runs in a monolithic container
+	// together with vertica main process, when set to false it runs in a separate sidecar
+	// container.
+	RunNMAInMonolithicContainerAnnotation      = "vertica.com/run-nma-in-monolithic-container"
```
roypaulin, line 59: I would suggest simply `run-nma-as-sidecar`, then the default would be `true`

pkg/controllers/vdb/nmarunningmode_reconciler.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+}
+
+// Reconcile will verify the running mode of NMA
+func (n *NMARunningModeReconciler) Reconcile(_ context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	err := n.verifyNMARunningMode()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	return ctrl.Result{}, nil
```
roypaulin, line 53: This is not needed, you can just `return ctrl.Result{}, err`, and remove `if` above

pkg/controllers/vdb/nmarunningmode_reconciler_test.go
```diff
@@ -0,0 +1,61 @@
...
+var _ = Describe("k8s/nmarunningmode_reconcile", func() {
+	ctx := context.Background()
+
+	It("should fail the reconclier if we configure the NMA to run in a sidecar container", func() {
+		vdb := vapi.MakeVDB()
+		vdb.Spec.Subclusters[0].Size = 1
+		vdb.Annotations[vmeta.RunNMAInMonolithicContainerAnnotation] = vmeta.RunNMAInMonolithicContainerAnnotationTrue
+		test.CreateVDB(ctx, k8sClient, vdb)
+		defer test.DeleteVDB(ctx, k8sClient, vdb)
+		// test.CreatePods(ctx, k8sClient, vdb, test.AllPodsRunning)
```
roypaulin, line 39: Let's remove these comments.

pkg/meta/annotations.go
spilchen: I agree with Roy. I like that annotation name better.

pkg/controllers/vdb/nmarunningmode_reconciler.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Verify whether the NMA is configured to run in a monolithic container and report error if not so
+func (n *NMARunningModeReconciler) verifyNMARunningMode() error {
+	if !vmeta.RunNMAInMonolithicContainerMode(n.Vdb.Annotations) {
+		n.VRec.Eventf(n.Vdb, corev1.EventTypeWarning, events.NMAInSidecarNotSupported, "Running NMA in a sidecar container is not supported yet")
+		return fmt.Errorf("running NMA in a sidecar container is not supported yet")
```
spilchen, line 60: I think we need to make it very clear what the user needs to do if they see this error. Can we include mention of the annotation they should set?
```suggestion
		n.VRec.Eventf(n.Vdb, corev1.EventTypeWarning, events.NMAInSidecarNotSupported, "Running NMA in a sidecar container is not supported yet. Set the %s annotation to %s", RunNMAInSidecarAnnotation, RunNMAInSidecarAnnotationFalse)
		return fmt.Errorf("running NMA in a sidecar container is not supported yet. Set the %s annotation to %s", RunNMAInSidecarAnnotation, RunNMAInSidecarAnnotationFalse)
```

pkg/controllers/vdb/nmarunningmode_reconciler.go
```diff
@@ -0,0 +1,63 @@
...
+func MakeNMARunningModeReconciler(vdbrecon *VerticaDBReconciler, log logr.Logger, vdb *vapi.VerticaDB) controllers.ReconcileActor {
+	return &NMARunningModeReconciler{
+		VRec: vdbrecon,
+		Vdb:  vdb,
+		Log:  log.WithName("NMARunningModeReconciler"),
+	}
+}
+
+// Reconcile will verify the running mode of NMA
+func (n *NMARunningModeReconciler) Reconcile(_ context.Context, _ *ctrl.Request) (ctrl.Result, error) {
```
spilchen, line 48: We should no-op if not running vclusterops.

scripts/setup-kustomize.sh
```diff
@@ -155,6 +155,9 @@ patches:
     - op: replace
       path: /spec/communal/path
       value: ${COMMUNAL_PATH_PREFIX}${TESTCASE_NAME}
+    - op: add
+      path: /metadata/annotations/vertica.com~1run-nma-in-monolithic-container
+      value: "true"
```
spilchen, line 160: I'm thinking maybe we should set this only if deploying with vclusterops.

jizhuoyu:
> Can we add some logic to verticadb_conversion.go to set the annotation so that we monolithic containers is the default? This way v1beta1 will default to monolithic container and v1 defaults to sidecar container.

Could you be more specific about this?
In any case, if this annotation (`vertica.com/run-nma-in-sidecar`) is explicitly set to any value, we don't change this annotation during conversion?
Then if the annotation is not present in v1beta1 CR, we have the following scenarios.
From v1beta1 to v1:
- if `VClusterOpsAnnotation` is `false` or not present
  - we don't set this annotation, because this has no effects for admintools deployment
- if `VClusterOpsAnnotation` is `true`
  - we set this annotation to be `false`, as that is supposed to be the default for the original v1beta1 CR
 From v1 to v1beta1:
- if `VClusterOpsAnnotation` is `false`
  - we don't set this annotation, because this has no effects for admintools deployment
- if `VClusterOpsAnnotation` is `true` or not present
  - we set this annotation to be `true`, as that is supposed to be the default for the original v1 CR

pkg/meta/annotations.go
jizhuoyu: done

pkg/controllers/vdb/nmarunningmode_reconciler.go
jizhuoyu: done

pkg/controllers/vdb/nmarunningmode_reconciler_test.go
jizhuoyu: done

scripts/setup-kustomize.sh
jizhuoyu: done

pkg/controllers/vdb/nmarunningmode_reconciler.go
jizhuoyu: done

pkg/controllers/vdb/nmarunningmode_reconciler.go
jizhuoyu: done

spilchen:
> > Can we add some logic to verticadb_conversion.go to set the annotation so that we monolithic containers is the default? This way v1beta1 will default to monolithic container and v1 defaults to sidecar container.
> 
> Could you be more specific about this? In any case, if this annotation (`vertica.com/run-nma-in-sidecar`) is explicitly set to any value, we don't change this annotation during conversion? Then if the annotation is not present in v1beta1 CR, we have the following scenarios.
> 
> **From v1beta1 to v1:**
> 
> * if `VClusterOpsAnnotation` is `false` or not present
>   
>   * we don't set this annotation, because this has no effects for admintools deployment
> * if `VClusterOpsAnnotation` is `true`
>   
>   * we set this annotation to be `false`, as that is supposed to be the default for the original v1beta1 CR
> 
> **From v1 to v1beta1:**
> 
> * if `VClusterOpsAnnotation` is `false`
>   
>   * we don't set this annotation, because this has no effects for admintools deployment
> * if `VClusterOpsAnnotation` is `true` or not present
>   
>   * we set this annotation to be `true`, as that is supposed to be the default for the original v1 CR
> 
> Do these sound OK?

Yes, this sounds good. Just need to set the annotation, if missing, during the conversion Iwebhook.

jizhuoyu:
> > > Can we add some logic to verticadb_conversion.go to set the annotation so that we monolithic containers is the default? This way v1beta1 will default to monolithic container and v1 defaults to sidecar container.
> > 
> > 
> > Could you be more specific about this? In any case, if this annotation (`vertica.com/run-nma-in-sidecar`) is explicitly set to any value, we don't change this annotation during conversion? Then if the annotation is not present in v1beta1 CR, we have the following scenarios.
> > **From v1beta1 to v1:**
> > 
> > * if `VClusterOpsAnnotation` is `false` or not present
> >   
> >   * we don't set this annotation, because this has no effects for admintools deployment
> > * if `VClusterOpsAnnotation` is `true`
> >   
> >   * we set this annotation to be `false`, as that is supposed to be the default for the original v1beta1 CR
> > 
> > **From v1 to v1beta1:**
> > 
> > * if `VClusterOpsAnnotation` is `false`
> >   
> >   * we don't set this annotation, because this has no effects for admintools deployment
> > * if `VClusterOpsAnnotation` is `true` or not present
> >   
> >   * we set this annotation to be `true`, as that is supposed to be the default for the original v1 CR
> > 
> > Do these sound OK?
> 
> Yes, this sounds good. Just need to set the annotation, if missing, during the conversion webhook.

done

api/v1beta1/verticadb_conversion.go
```diff
@@ -123,6 +131,14 @@ func convertFromAnnotations(src *v1.VerticaDB) (newAnnotations map[string]string
 		}
 		newAnnotations[key] = val
 	}
+	_, VClusterOpsAnnotationOK := src.Annotations[vmeta.VClusterOpsAnnotation]
+	_, RunNMAInSidecarAnnotationOK := src.Annotations[vmeta.RunNMAInSidecarAnnotation]
+	// If the VClusterOpsAnnotation annotation is set to true or not present, set the RunNMAInSidecarAnnotation
+	// to true if it's not present so that CRs converted from v1 will still run NMA in a sidecar container, which is
+	// the default behavior of v1
+	if (!VClusterOpsAnnotationOK || vmeta.UseVClusterOps(src.Annotations)) && !RunNMAInSidecarAnnotationOK {
```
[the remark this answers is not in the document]
jizhuoyu: yeah I originally thought it would be clearer if we explicitly check whether the annotation exists rather than relying on the behavior of `UseVClusterOps()` with respect to default value. Given that the behavior of `UseVClusterOps()` is expected to be stable in the long term (v1 API will always by default use vclusterops), it makes sense to not check explicitly here.

spilchen (approved):

PR #600, opened 2023-11-22
Use k8s.io package to manage VerticaDB status condition

roypaulin:
In the v1beta1 API we added our own status condtions to the VerticaDB. Now in v1, we want to use the Condition struct from `k8s.io/apimachinery/pkg/apis/meta/v1`. This PR is to remove v1.VerticaDBCondition and all related code in favor of `metav1.Condition`. It will make it easier to manage status condition with provided methods


roypaulin:
I expect some tests to fail due to the conditions slice not sorted anymore. I am going to update them.

pkg/controllers/vdb/stopdb_reconciler.go
```diff
@@ -82,7 +81,7 @@ func (s *StopDBReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl
 		// Clear the condition now that we stopped the cluster.  We rely on the
 		// restart reconciler that follows this to bring up vertica.
 		err = vdbstatus.UpdateCondition(ctx, s.VRec.Client, s.Vdb,
-			vapi.VerticaDBCondition{Type: vapi.VerticaRestartNeeded, Status: corev1.ConditionFalse},
+			vapi.MakeCondition(vapi.VerticaRestartNeeded, metav1.ConditionFalse, vapi.UpComingRestart),
```
spilchen, line 84: Suggest "StopCompleted" for the reason.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -91,13 +92,13 @@ func MakeRestartReconciler(vdbrecon *VerticaDBReconciler, log logr.Logger,
...
 		err := vdbstatus.UpdateCondition(ctx, r.VRec.Client, r.Vdb,
-			vapi.VerticaDBCondition{Type: vapi.AutoRestartVertica, Status: corev1.ConditionFalse},
+			vapi.MakeCondition(vapi.AutoRestartVertica, metav1.ConditionFalse, vapi.AutoRestartVerticaDisabled),
 		)
 		return ctrl.Result{}, err
 	}
 
 	err := vdbstatus.UpdateCondition(ctx, r.VRec.Client, r.Vdb,
-		vapi.VerticaDBCondition{Type: vapi.AutoRestartVertica, Status: corev1.ConditionTrue},
+		vapi.MakeCondition(vapi.AutoRestartVertica, metav1.ConditionTrue, vapi.AutoRestartVerticaEnabled),
```
spilchen, line 101: Suggest "Detected" as the reason too.

spilchen (commented):
Thanks for doing this. It nice to see the condition logic get simplified.

pkg/controllers/vdb/restart_reconciler.go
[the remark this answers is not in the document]
roypaulin: Yes if status condition did not change this is what will be executed in `SetStatusCondition()`:
```
existingCondition.Reason = newCondition.Reason
existingCondition.Message = newCondition.Message
existingCondition.ObservedGeneration = newCondition.ObservedGeneration
```
But that is not a problem as it remains `DeepEqual` to vdb.Status

spilchen (approved):
Looks good!

roypaulin:
I ended up changing a lot of files in the e2e tests. Let me know if it looks good to you.

spilchen:
> I ended up changing a lot of files in the e2e tests. Let me know if it looks good to you.

Is there a reason why we couldn't just check the upgrade conditions for upgrade tests? You have added checks for AutoRestartVertica and DBInitialized. That seems like overkill.

roypaulin:
> > I ended up changing a lot of files in the e2e tests. Let me know if it looks good to you.
> 
> Is there a reason why we couldn't just check the upgrade conditions for upgrade tests? You have added checks for AutoRestartVertica and DBInitialized. That seems like overkill.

Hum you're right. I just wanted to make them as similar as possible to what was there before, but yes we only need to check the upgrade conditions. I will make some changes.

PR #602, opened 2023-11-23
Handle missing versions as an error

spilchen:
The controller will obtain the Vertica version and save it as an annotation in the VerticaDB object. This is expected to be completed before the create db operation. This will allow us to take the appropriate actions based on the Vertica version we are using. However, this is not always reliable due to the caching of the Kubernetes client. Previously, we tolerated missing versions, but I am now changing some of these to errors in cases where we need the version.

This was causing an issue when enabling spread encryption during create db. In older Vertica versions, or if the version is missing, the server would restart. By handling missing versions as an error, we will take the correct action. I believe this will also improve the reliability of the e2e tests.

I am also removing some temporary code that included the Go routine ID with each reconcile. I am generating a UUID and using that instead. This is a more permanent solution.


pkg/vadmin/re_ip_at.go
```diff
@@ -116,11 +119,14 @@ func (a *Admintools) genReIPCommand() []string {
 	// some nodes are up.  This was done to support doing a reip while there are
 	// read-only secondary nodes.
 	vinf, ok := a.VDB.MakeVersionInfo()
+	if !ok {
+		return nil, fmt.Errorf("could not find version from vdb")
+	}
 	if ok && vinf.IsEqualOrNewer(vapi.ReIPAllowedWithUpNodesVersion) {
```
roypaulin, line 125: We don't need `ok`  here anymore.

pkg/controllers/vdb/imageversion_reconciler.go
```diff
@@ -82,8 +83,7 @@ func (v *ImageVersionReconciler) Reconcile(ctx context.Context, _ *ctrl.Request)
 
 	vinf, ok := v.Vdb.MakeVersionInfo()
 	if !ok {
-		// Version info is not in the vdb.  Fine to skip.
-		return ctrl.Result{}, nil
+		return ctrl.Result{}, errors.New("version not found in the VerticaDB")
```
roypaulin, line 86: Let's output the same  error message everywhere we check if the version is in the vdb.

pkg/controllers/vdb/config.go
```diff
@@ -115,6 +117,19 @@ func (g *ConfigParamsGenerator) GetConfigParms() *vtypes.CiMap {
 	return g.ConfigurationParams
 }
 
+// setup will initialize parms in the ConfigParamsGenerator struct
+func (g *ConfigParamsGenerator) setup() error {
+	if g.ConfigurationParams == nil {
+		g.ConfigurationParams = vtypes.MakeCiMap()
+	}
+	var ok bool
+	g.VInf, ok = g.Vdb.MakeVersionInfo()
```
roypaulin, line 126: I thought we already checked this on `createdb_reconciler`.

spilchen: We use `ConfigParamsGenerator` outside of create db. So, it is best not to rely on that.

pkg/controllers/vdb/imageversion_reconciler.go
spilchen: I'll create a variant of `MakeVersionInfo` that returns an error called `MakeVersionInfoCheck`

roypaulin (approved):

PR #603, opened 2023-11-24
Use /version endpoint for vertica health probes

roypaulin:
Before this, for health check probes, we were using a combination of canary query (select 1) and checking if the client port is being listened on. Now we can use the `/version` http endpoint that does not need any authentication so we don't need to mount the superuser password anymore.


roypaulin:
Oops forgot that this is vclusterops only.

pkg/builder/builder.go
```diff
@@ -642,10 +657,27 @@ func makeCanaryQueryProbe(vdb *vapi.VerticaDB) *corev1.Probe {
 	}
 }
 
+// getHTTPServerVersionEndpointProbe returns an HTTPGet probe if vclusterops
+// is enabled and server version is newer or equal to v23.4
+func getHTTPServerVersionEndpointProbe(vdb *vapi.VerticaDB) *corev1.Probe {
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		// if the server version is older than v23.4 then
+		// we return nil
+		vdbVer, ok := vdb.MakeVersionInfo()
```
spilchen, line 666: I'm not sure if we will ever have the version the first time. The version isn't known until the pod is running and this needs to be setup before the pod is running. Rather than rely on a specific version, lets just tie it too vclusterops. If vclusterops is enabled we will use this new health probe. That is more or less a check for 24.1.0.

pkg/builder/builder.go
```diff
@@ -62,6 +62,8 @@ const (
 	NMAKeyEnv             = "NMA_KEY_PATH"
 	NMASecretNamespaceEnv = "NMA_SECRET_NAMESPACE"
 	NMASecretNameEnv      = "NMA_SECRET_NAME"
+
+	HTTPServerVersionPath = "/v1/version"
```
spilchen, line 66: We probably don't need to export this. Suggest a rename:
```suggestion
	httpServerVersionPath = "/v1/version"
```

pkg/builder/builder.go
```diff
@@ -681,11 +726,7 @@ func makeStartupProbe(vdb *vapi.VerticaDB) *corev1.Probe {
 
 // makeLivenessProbe will return the Probe object to use for the liveness probe.
 func makeLivenessProbe(vdb *vapi.VerticaDB) *corev1.Probe {
-	// We check if the TCP client port is open. We used this approach,
-	// rather than issuing 'select 1' like readinessProbe because we need
-	// to minimize variability. If the livenessProbe fails, the pod is
-	// rescheduled. So, it isn't as forgiving as the readinessProbe.
-	probe := makeVerticaClientPortProbe()
+	probe := makeDefaultLivenessProbe(vdb)
```
spilchen, line 729: Will we be able to use the overrides to change the probe back to checking the TCP client port? Similar question to change back to canary query.

pkg/builder/builder_test.go
```diff
@@ -173,9 +173,27 @@ var _ = Describe("builder", func() {
 		Expect(c.StartupProbe.PeriodSeconds).Should(Equal(NewPeriodSeconds))
 	})
 
-	It("should not mount superuser password if probe's overridden", func() {
+	It("shoul have all probes use the htpp version endpoint", func() {
```
spilchen, line 176: spelling:
```suggestion
	It("should have all probes use the htpp version endpoint", func() {
```

pkg/builder/builder_test.go
```diff
@@ -192,7 +210,16 @@ var _ = Describe("builder", func() {
 		}
 		c := buildPodSpec(vdb, &vdb.Spec.Subclusters[0])
 		Expect(isPasswdIncludedInPodInfo(vdb, &c)).Should(BeFalse())
+
 		vdb.Spec.StartupProbeOverride = nil
+
+		// case 2: if in vclusterops mode and version >= v23.4
```
spilchen, line 216: We can update this since we aren't going to check the version now.
```suggestion
		// case 2: if in vclusterops mode
```

pkg/builder/builder.go
roypaulin: Yes we will. I will add a unit test.

roypaulin: > Yes we will. I will add a unit test.

But I guess we will need to mount the superuser password in all cases.

spilchen: No you shouldn't have to. The builder should be smart enough to only mount the superuser password if it is required in the probe (see `probeContainsSuperuserPassword`)

spilchen (approved):

PR #605, opened 2023-11-28
Run scrutinize for vclusterops deployments

spilchen:
For vclusterops deployments, the server team developed a new method for scrutinize using vcluster. This uses the new scrutinize method for those deployments. We had a collection script that we run at the end of e2e tests. This has been updated. We also call this script in a few testcases to verify that scrutinize runs correctly.

### DIFF
--- a/scripts/capture-scrutinize.sh
+++ b/scripts/capture-scrutinize.sh
@@ -23,21 +23,32 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 REPO_DIR=$(dirname $SCRIPT_DIR)
 NS=$(kubens)
 HOST_OP_DIR=$REPO_DIR/int-tests-output
+EXIT_ON_ERROR=
+
+source $SCRIPT_DIR/logging-utils.sh
 
 function usage() {
-    echo "usage: $(basename $0) [-n <namespace_prefix>]"
+    echo "usage: $(basename $0) [-n <namespace_prefix>] [-o <dir>] [-x]"
     echo
     echo "Options:"
     echo "  -n <namespace_prefix>   Collect scrutinize only for VerticaDB that "
     echo "                          have a namespace matching this prefix."
+    echo "  -o <output-dir>         Directory to store scrutinize output in."
+    echo "  -x                      Return an error and exit if scrutinize fails"
     exit 1
 }
 
-while getopts "n:h" opt
+while getopts "n:ho:x" opt
 do
     case $opt in
         n)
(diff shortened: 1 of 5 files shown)


scripts/capture-scrutinize.sh
```diff
@@ -49,34 +60,106 @@ do
...
-        if [[ -n $pods ]]
+        if [[ -z $pods ]]
         then
-            set -o xtrace
+            logWarning "No pods running. Skip collection"
+            continue
+        fi
+
+        # Depending on the deployment we have two scrutinize methods. The
+        # scrutinize standalone only works widh admintools based deployments.
```
roypaulin, line 94:
```suggestion
        # scrutinize standalone only works with admintools based deployments.
```

tests/e2e-leg-6/nma-certs-mount/55-run-scrutinize.yaml
```diff
@@ -0,0 +1,17 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: cd ../../..  && scripts/capture-scrutinize.sh -o /tmp/scrutinize-op -x -n $NAMESPACE
```
roypaulin, line 17: Why do we run it twice in `nma-certs-mount`

scripts/capture-scrutinize.sh
```diff
@@ -49,34 +60,106 @@ do
...
+done
+
+logInfo "Save off the OLM operator log if deployed"
+if kubectl get ns olm 2> /dev/null
+then
+    kubectl logs -n olm -l app=olm-operator --tail=-1 > $HOST_OP_DIR/olm-operator.log
+fi
+
+logInfo "Collect scrutinize for VerticaDB clusters"
 for ns in $NS
```
roypaulin, line 79: It is not too important but is it possible to break some parts into functions to make it more readable?

roypaulin (commented):
Looks good!

tests/e2e-leg-6/nma-certs-mount/55-run-scrutinize.yaml
spilchen: This testcase runs the nma in two modes (getting secrets from mount points and getting secrets from k8s), so it was important to verify that scrutinize works in both modes.

roypaulin (approved):

PR #606, opened 2023-11-28
Enable e2e-server-upgrade with vcluster

roypaulin:
Now we have all the tools to run e2e-server-upgrade with vclusterops. For now we a picking a static v2 image as base image but once we start building a v2 image in CI we are going to use a logic similar as admintools'.

### DIFF
--- a/.github/workflows/e2e-server-upgrade.yml
+++ b/.github/workflows/e2e-server-upgrade.yml
@@ -12,6 +12,9 @@ on:
       vertica-image:
         type: string
         required: false
+      vertica-deployment-method:
+        type: string
+        required: false
     secrets:
       DOCKERHUB_USERNAME:
         description: 'When working with images from docker.io, this is the username for login purposes'
@@ -33,6 +36,14 @@ on:
         description: 'Name of the vertica server image'
         type: string
         required: false
+      vertica-deployment-method:
+        description: 'Vertica deployment method'
+        type: choice
+        required: false
+        default: admintools
+        options:
+        - admintools
+        - vclusterops
 
 jobs:
 
@@ -61,6 +72,7 @@ jobs:
         export VERTICA_IMG=${{ inputs.vertica-image }}
         export OPERATOR_IMG=${{ inputs.operator-image }}
         export VLOGGER_IMG=${{ inputs.vlogger-image }}
+        export VERTICA_DEPLOYMENT_METHOD=${{ inputs.vertica-deployment-method }}
         export BASE_VERTICA_IMG=$(scripts/guess-server-upgrade-base-image.sh $VERTICA_IMG)
         echo "Upgrading server from image: $BASE_VERTICA_IMG"
         scripts/run-k8s-int-tests.sh -s -e tests/external-images-server-upgrade-ci.txt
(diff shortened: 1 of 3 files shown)


spilchen (commented):
This all looks good. Can you try and see if e2e-leg-5-at-only/online-upgrade-down-cluster can be moved to e2e-leg-5?

spilchen:
The latest changes look good. In your last run, some of the tests fail intermittently. We may need to run them in AT only and open jira tasks for each to dig into their failures.

I peeked at one (offline-upgrade-ks-0) and it was repeatedly hitting this vclusterops error:
verticadb-operator verticadb-operator-controller-manager-6474d657c7-wm2gl manager 2023-11-29T00:50:33.207Z  ERROR   controllers.VerticaDB.RestartNode   fail to provide a non-existent node name v_vd_node0002  {"verticadb": "kuttl-test-close-toad/v-base-upgrade", "reconcile-uuid": "34806e65-2d36-4e38-b060-15465689b781"}

roypaulin:
> The latest changes look good. In your last run, some of the tests fail intermittently. We may need to run them in AT only and open jira tasks for each to dig into their failures.
> 
> I peeked at one (offline-upgrade-ks-0) and it was repeatedly hitting this vclusterops error: verticadb-operator verticadb-operator-controller-manager-6474d657c7-wm2gl manager 2023-11-29T00:50:33.207Z ERROR controllers.VerticaDB.RestartNode fail to provide a non-existent node name v_vd_node0002 {"verticadb": "kuttl-test-close-toad/v-base-upgrade", "reconcile-uuid": "34806e65-2d36-4e38-b060-15465689b781"}

I think even before that `removenode` fails:
`controllers.VerticaDB	Event logging	{"eventtype": "Warning", "reason": "RemoveNodesFailed", "message": "Failed when calling database remove node"}`

roypaulin:
I am going to isolate those 2 tests and open tickets.

roypaulin:
For `online-upgrade-block-downgrade` the issue is that the image used in step 20 is not compatible with vclusterops so we hit `WrongImage` instead of invalidUpgradePath at step 30.

spilchen:
> For `online-upgrade-block-downgrade` the issue is that the image used in step 20 is not compatible with vclusterops so we hit `WrongImage` instead of invalidUpgradePath at step 30.
> 
> edit: `vertica/vertica-k8s:11.1.1-0-minimal` his is what we use. Maybe we can write a script that calls `patch-image-in-vdb.sh` with a different image depending on the deployment type, that would be better than addding a new env var.
> 
> Something like:
> 
> ```
> if vclusterops
>    patch-image-in-vdb.sh vertica/vertica-k8s-private:somethingolder
> else
>    patch-image-in-vdb.sh vertica/vertica-k8s:11.1.1-0-minimal
> fi
> ```
> 
> Or we simply update patch-image-in-vdb.sh to use a static image for vclusterops for now

I think we still need 11.1.1-0 for some e2e runs because we still run against 12.0.4. I like the idea of using a different image for vclusterops, but it has to be a public image, which there isn't any. That will have to wait until we publish the v2 image for 24.2.0. Can we just permanently leave this in an *-at-only test suite?

spilchen (approved):
This

roypaulin:
I will let you merge. There is still one expected test, maybe you know what to do.

PR #607, opened 2023-11-28
VCluster API Refresh: replace VRestartNode with VStartNode

le-an-vertica:
This PR synced the API changes on the vclusterops side, which renamed VRestartNode to VStartNode and other "start node" related functions.

### DIFF
--- a/pkg/vadmin/restart_node_vc.go
+++ b/pkg/vadmin/restart_node_vc.go
@@ -30,9 +30,9 @@ import (
 // lost cluster quorum. The IP given for each vnode may not match the current IP
 // in the vertica catalogs.
 func (v *VClusterOps) RestartNode(ctx context.Context, opts ...restartnode.Option) (ctrl.Result, error) {
-	v.setupForAPICall("RestartNode")
+	v.setupForAPICall("StartNode")
 	defer v.tearDownForAPICall()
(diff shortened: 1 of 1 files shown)


pkg/vadmin/restart_node_vc.go
```diff
@@ -30,9 +30,9 @@ import (
 // lost cluster quorum. The IP given for each vnode may not match the current IP
 // in the vertica catalogs.
 func (v *VClusterOps) RestartNode(ctx context.Context, opts ...restartnode.Option) (ctrl.Result, error) {
```
le-an-vertica, line 32: I kept this because this function implements an interface that is also used by the Admintools operator.

roypaulin: The question is @spilchen now that we are moving away from `admintools` should we replace `restartnode*` with `startnode*`

roypaulin (approved):

pkg/vadmin/restart_node_vc.go
spilchen: Possibly. But I think we can fix that up in the future. There are already cases where the operator dispatcher interface name doesn't match the vclusterops API name.

spilchen (approved):

PR #611, opened 2023-11-30
Deploy operator with single YAML manifest

spilchen:
Now that our operator is cluster scoped, we can deploy it using simple `kubectl apply -f ...` commands. This adds that manifest to our release artifacts. This will be published on GitHub in our 2.0.0 release. Once released the command will be:

```
# Install CRDs
kubectl apply --server-side=true --force-conflicts -f https://github.com/vertica/vertica-kubernetes/releases/latest/download/verticadbs.vertica.com-crd.yaml
kubectl apply --server-side=true --force-conflicts -f https://github.com/vertica/vertica-kubernetes/releases/latest/download/verticaautoscalers.vertica.com-crd.yaml
kubectl apply --server-side=true --force-conflicts -f https://github.com/vertica/vertica-kubernetes/releases/latest/download/eventtriggers.vertica.com-crd.yaml
# Deploy the operator
kubectl apply -f https://github.com/vertica/vertica-kubernetes/releases/latest/download/operator.yaml
```

### DIFF
--- a/changes/unreleased/Added-20231130-121009.yaml
+++ b/changes/unreleased/Added-20231130-121009.yaml
@@ -1,0 +1,5 @@
+kind: Added
+body: Deploy the operator with kubectl command
+time: 2023-11-30T12:10:09.870483929-04:00
+custom:
+  Issue: "611"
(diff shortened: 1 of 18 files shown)


roypaulin:
I get the idea but what about all the parameters that are usually set on helm charts?

spilchen:
> I get the idea but what about all the parameters that are usually set on helm charts?

The manifest just has the default values for those. That is good enough to deploy the operator. If you need a customized deployment, helm is still the way to go.

PR #613, opened 2023-11-30
Fix intermittent auto-restart-vertica test failure

roypaulin:
That test will create an EventTrigger that will run when the AutoRestartVertica status condition goes to False. We change the status condition, but we don't always see the EventTrigger creating the job. 
Each time it fails, looking at the log, you can see that although AutoRestartVertica is False(checked with kubectl), the value used for comparison by eventtrigger is true: `{"et": "kuttl-test-novel-toad/hello-world", "expected": "False", "found": "True", "refObjectName": "v-auto-restart-vertica", "condition": "AutoRestartVertica"}`, so we are somehow picking the wrong value. I suspect it is because we are using a `v1beta1` vdb object to get the status from k8s so instead I used a `v1`.


spilchen (commented):
Thanks for digging into this. I suspected it was related to the API version change but I didn't have time to dig further. 

I noticed the testcase checks for status conditions by including them in the assert. That could be prone to errors as it it depends on a strict order. Should we do some conversions to use 'kubectl wait' like you did in the step 16?

roypaulin:
> Thanks for digging into this. I suspected it was related to the API version change but I didn't have time to dig further.
> 
> I noticed the testcase checks for status conditions by including them in the assert. That could be prone to errors as it it depends on a strict order. Should we do some conversions to use 'kubectl wait' like you did in the step 16?

Yes I will update that step.

PR #614, opened 2023-12-01
README link updates

rs-vertica:
Previously, we relied on redirects from old doc version URLs to new URLs. This replaces old link URLs so we do not rely on redirects.


spilchen (approved):
Thanks for updating.

PR #617, opened 2023-12-01
Fix error communal storage path is invalid

roypaulin:
We include the UID in the communal path to generate a unique path for each new instance of vdb. To do that, we append the UID to the path this way `path/uid`. The problem is if the path ends with `/` like `s3://vertica-fleeting/k8s/`, we end up with 2 slashes like `s3://vertica-fleeting/k8s//uid` which will fail vclusterops communal storage path check with error: `communal storage path is invalid: use an absolute local path or a correct remote url path`.
So I removed the trailing slash from the path before appending the UID.

### DIFF
--- a/api/v1/helpers.go
+++ b/api/v1/helpers.go
@@ -189,17 +189,17 @@ func (v *VerticaDB) GetPVSubPath(subPath string) string {
 
 // GetDBDataPath get the data path for the current database
 func (v *VerticaDB) GetDBDataPath() string {
-	return fmt.Sprintf("%s/%s", v.Spec.Local.DataPath, v.Spec.DBName)
+	return fmt.Sprintf("%s/%s", strings.TrimSuffix(v.Spec.Local.DataPath, "/"), v.Spec.DBName)
 }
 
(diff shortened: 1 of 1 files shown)


spilchen (approved):

api/v1/helpers.go
```diff
@@ -189,17 +189,17 @@ func (v *VerticaDB) GetPVSubPath(subPath string) string {
 
 // GetDBDataPath get the data path for the current database
 func (v *VerticaDB) GetDBDataPath() string {
-	return fmt.Sprintf("%s/%s", v.Spec.Local.DataPath, v.Spec.DBName)
+	return fmt.Sprintf("%s/%s", strings.TrimSuffix(v.Spec.Local.DataPath, "/"), v.Spec.DBName)
```
jizhuoyu, line 192: Have we considered Go `path` [lib](https://pkg.go.dev/path/filepath#Join)?

roypaulin: `strings` is fine in this case and there is not a corresponding method for this in `path/filepath`.

PR #620, opened 2023-12-05
Add webhook check for svc names

jizhuoyu:
- check that vdb name is a valid k8s service object name because vdb name is the headless service name
- check that external service name (concatenated by vdb name, `-`, and subcluster `serviceName`) is valid

note: service names must conform to [rfc-1035-label-names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names)


api/v1/verticadb_webhook.go
```diff
@@ -320,6 +322,43 @@ func (v *VerticaDB) validateEndpoint(allErrs field.ErrorList) field.ErrorList {
 	return allErrs
 }
 
+func (v *VerticaDB) hasValidSvcName(allErrs field.ErrorList) field.ErrorList {
+	vdbName := v.ObjectMeta.Name
+	// vdb name will be used as is for headless svc name,
+	// thus match regex for service name (DNS-1035 label)
+	r := regexp.MustCompile(`^[a-z]([\-a-z0-9]{0,61}[a-z0-9])?$`)
+	if !r.MatchString(vdbName) {
```
roypaulin, line 330: You do not need to validate vdb name. Resource names(ObjectMeta.Name) are already checked against a set of [rules](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names)

api/v1/verticadb_webhook.go
```diff
@@ -320,6 +322,43 @@ func (v *VerticaDB) validateEndpoint(allErrs field.ErrorList) field.ErrorList {
...
+		// invalid vdb name does not necessarily imply invalid external service name
+		// but usually invalid vdb name results in invalid external service name for
+		// all subclusters, so prompt users to fix vdb name first and do not check
+		// external service names for now
+		return allErrs
+	}
+	// check external svc names
+	for i := range v.Spec.Subclusters {
+		sc := &v.Spec.Subclusters[i]
+		extSvcName := v.ObjectMeta.Name + "-" + sc.GetServiceName()
```
roypaulin, line 345: There is already `hasValidDomainName` that checks that sc names are valid, so I think here you should just check `sc.ServiceName` and skip if it is empty.

api/v1/verticadb_webhook.go
```diff
@@ -320,6 +322,43 @@ func (v *VerticaDB) validateEndpoint(allErrs field.ErrorList) field.ErrorList {
...
+		// vdb name will be used together with subcluster serviceName for external svc name,
+		// thus match regex for service name (DNS-1035 label)
+		r := regexp.MustCompile(`^[a-z]([\-a-z0-9]{0,61}[a-z0-9])?$`)
```
roypaulin, line 348: There is already `IsValidSubclusterName` that does this so may be we can rename that function and use it whenever we need to check if a string respect k8s object naming convention.

api/v1/verticadb_webhook.go
```diff
@@ -320,6 +322,43 @@ func (v *VerticaDB) validateEndpoint(allErrs field.ErrorList) field.ErrorList {
...
+		fieldPrefix := field.NewPath("spec").Child("subclusters").Index(i)
```
roypaulin, line 349: In `hasValidDomainName` we iterate over subclusters and check each name. Maybe we can extend that function to also check `sc.ServiceName` so that we don't have to create a separate function.

jizhuoyu: I'm not sure about this one. If we choose to check `sc.ServiceName` in `hasValidDomainName` we'll need to rename this function, as we're not checking "domain" name for `sc.ServiceName` but [DNS-1035 label](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names). And because vdb name and `sc.ServiceName` are related in the validation logic here (I choose not to check `sc.ServiceName` if we already found that vdb name is invalid, i.e. unsuitable for a headless service name, because in many cases an invalid vdb name results in an invalid external service name as well, given that the former is a prefix in the latter), I think maybe it's better to keep the validation logic of both headless service name as well as external service name here.

api/v1/verticadb_webhook.go
roypaulin: No you can try yourself, "1v-" is going to fail.

roypaulin: Object.meta is already being validated against this regex `'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')`

roypaulin: In that case create a separate function to validate vdb.

api/v1/verticadb_webhook.go
roypaulin: You can have a separate function where you check vdb name and make sure it respect the rules for a service. In `hasValidDomainName` when you iterate over subclusters, and for each, if subcluster name is valid, you can call a function that checks serviceName. Both `sc.Name` and `sc.ServiceName` are part of subcluster so it makes sense to check them on the same iteration and we cannot keep adding a function each time we need to check a new field(unless it is necessary) and of course we could change ``hasValidDomainName` function name as it is not accurate given that it is only used for subcluster name.

jizhuoyu: I see your point, but what if the user created an invalid vdb name like "1v" which would make all external service names invalid? In such a case the user will be prompted many errors: one for vdb name which is intended, and one for each of the subcluster serviceName because the concatenated external service name is invalid even if sc.serviceName is valid. What it be confusing to users that even if there is nothing wrong with sc.serviceName it is still reported that sc.serviceName is invalid (they'll eventually figured it out in the error message that it is the concatenated string that's invalid so the root cause is vdb name)?

jizhuoyu: > I see your point, but what if the user created an invalid vdb name like "1v" which would make all external service names invalid? In such a case the user will be prompted many errors: one for vdb name which is intended, and one for each of the subcluster `serviceName` because the concatenated external service name is invalid even if `sc.serviceName` is valid. Would it be confusing to users that even if there is nothing wrong with `sc.serviceName` it is still reported that `sc.serviceName` is invalid (they'll eventually figure it out in the error message that it is the concatenated string that's invalid so the root cause is vdb name)?

One potential solution might be adding a check for vdb name right before we loop over subclusters and check `serviceName` in `hasValidDomainName` (I think I can rename this to `hasValidScAndExtSvcName`), if the vdb name itself is invalid we don't check `serviceName` because users should fix vdb name first rather than be confused by more error messages due to the invalid vdb name. Of course we would always check and report vdb name in a function (I think we can name it `hasValidHlSvcName`) that is parallel to `hasValidScAndExtSvcName`

roypaulin: You can have a function `IsValidVDBName` that returns a bool where you have the logic to check vdb name. If it returns  `false` you exit early. All of that in `hasValidDomainName`.

```
hasValidDomainName() {
   if !isvalidVDBName {
       ...
       return allerrs
    }
    for i := range v.Spec.Subclusters {
          sc := &v.Spec.Subclusters[i]
           if IsValidSubclusterName {
                if !IsValidSvcName(svcName){
                  .......
                 }
                continue
           }
            err := field.Invalid(field.NewPath("spec").Child("subcluster").Index(i).Child("name"),
				v.Spec.Subclusters[i],
				"is not a valid domain name")
    }
}
```
This is what I had in mind.

api/v1/verticadb_webhook.go
roypaulin: I like the idea to have a function for each type of regex that would be called IsValidSubcluster, IsValidSvcName, IsvalidVDBName, etc...

jizhuoyu: yes, that's what I was thinking

api/v1/verticadb_webhook.go
```diff
@@ -400,13 +400,46 @@ func (v *VerticaDB) getClusterSize() int {
...
+func (v *VerticaDB) hasValidSvcAndScName(allErrs field.ErrorList) field.ErrorList {
+	// check headless svc names
+	vdbName := v.ObjectMeta.Name
+	// vdb name will be used as is for headless svc name,
+	// thus match regex for service name (DNS-1035 label)
+	isValidHlSvcName := IsValidServiceName(vdbName)
+	if !isValidHlSvcName {
+		err := field.Invalid(field.NewPath("metadata").Child("name"),
+			v.ObjectMeta.Name,
+			fmt.Sprintf("vdb name is used as the headless service name, and hence must match regex '%s'",
```
roypaulin, line 412:
```suggestion
			fmt.Sprintf("vdb name must match regex '%s'",
```
Let's keep it simple

api/v1/verticadb_webhook.go
```diff
@@ -439,13 +400,46 @@ func (v *VerticaDB) getClusterSize() int {
 	return sizeSum
 }
 
-func (v *VerticaDB) hasValidDomainName(allErrs field.ErrorList) field.ErrorList {
+func (v *VerticaDB) hasValidSvcAndScName(allErrs field.ErrorList) field.ErrorList {
+	// check headless svc names
+	vdbName := v.ObjectMeta.Name
```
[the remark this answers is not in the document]
jizhuoyu: will do

api/v1/verticadb_webhook.go
```diff
@@ -400,13 +400,46 @@ func (v *VerticaDB) getClusterSize() int {
...
+			err := field.Invalid(fieldPrefix.Child("name"),
+				sc.GenCompatibleFQDN(),
+				fmt.Sprintf("subcluster name is not valid, change it so that it matches regex '%s'", RFC1123DNSSubdomainNameRegex))
+			allErrs = append(allErrs, err)
+		}
+		if !isValidHlSvcName {
+			// invalid vdb name does not necessarily imply invalid external service name
+			// but usually invalid vdb name results in invalid external service name for
+			// all subclusters, so prompt users to fix vdb name first and do not check
+			// external service names for now
```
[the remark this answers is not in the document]
jizhuoyu: Yeah what I want to express is an invalid vdb name does not necessarily mean an invalid external service name, for instance `v-` is an invalid name, but a generated external service name `v--sc` is a valid service name. But I agree that no matter the external service name will be valid given an invalid vdb name we could always skip the check here.

api/v1/verticadb_webhook.go
```diff
@@ -400,13 +400,46 @@ func (v *VerticaDB) getClusterSize() int {
...
+			continue
+		}
+		// check external svc names
+		extSvcName := v.ObjectMeta.Name + "-" + sc.GetServiceName()
```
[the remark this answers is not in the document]
jizhuoyu: will do

api/v1/verticadb_webhook.go
```diff
@@ -400,13 +400,46 @@ func (v *VerticaDB) getClusterSize() int {
...
+		// vdb name will be used together with subcluster serviceName for external svc name,
+		// thus match regex for service name (DNS-1035 label)
+		if !IsValidServiceName(extSvcName) {
+			err := field.Invalid(fieldPrefix.Child("serviceName"),
+				sc.GetServiceName(),
+				fmt.Sprintf("subcluster serviceName is prefixed by vdb name %q and \"-\" to generate external service name %q,"+
+					" which must match regex '%s', modify the serviceName and/or vdb name so that external service name matches the regex",
```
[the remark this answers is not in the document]
jizhuoyu: makes sense, will do

api/v1/verticadb_webhook.go
jizhuoyu: will do

PR #629, opened 2023-12-11
Prevent pod removal if db still exists at node

spilchen:
This is specific to vclusterops. We have a check in the object reconciler that makes sure we don't remove a pod that is still part of the vertica cluster. We must run DB remove node logic. We wrongly skipped this check for vcluster. This adds the check back for vcluster, but adjusts it because there is no installed state for that deployment.

This also moves the k-safety-0-scaling testcase to a test suite that runs in admintools and vclusterops deployments.


roypaulin (approved):
Looks good!

PR #631, opened 2023-12-12
Fix restart reconciler requeue logic for vclusterops

spilchen:
This is to fix intermittent failures in the pending-pod e2e testcase when run in vclusterops mode. The restart reconciler needs to requeue if pods that it needs aren't running. We were too aggressive with vclusterops because we triggered on installed state, which may or may set correctly. This changes the check for vclusterops and only looks at dbExists state. It also correctly sets the installed state for vclusterops in all cases.

### DIFF
--- a/pkg/controllers/vdb/podfacts.go
+++ b/pkg/controllers/vdb/podfacts.go
@@ -470,6 +470,12 @@ func getPathToVerifyCatalogExists(pf *PodFact) string {
 func (p *PodFacts) checkIsInstalled(_ context.Context, vdb *vapi.VerticaDB, pf *PodFact, gs *GatherState) error {
 	pf.isInstalled = false
 
+	// VClusterOps don't have an installed state, so we can handle that without
+	// checking if the pod is running.
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		return p.checkIsInstalledForVClusterOps(pf)
+	}
+
 	scs, ok := vdb.FindSubclusterStatus(pf.subclusterName)
 	if ok {
 		// Set the install indicator first based on the install count in the status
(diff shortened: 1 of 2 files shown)


pkg/controllers/vdb/podfacts.go
```diff
@@ -995,14 +999,22 @@ func (p *PodFacts) countRunningAndInstalled() int {
 	})
 }
 
-// countInstalledAndNotRestartable returns number of installed pods that aren't yet restartable
-func (p *PodFacts) countInstalledAndNotRestartable() int {
+// countNonRunningPodsNeededByRestart returns number of pods that aren't yet
+// running but the restart reconciler needs to handle them.
+func (p *PodFacts) countNonRunningPodsNeededByRestart(vclusterOps bool) int {
```
roypaulin, line 1004: I think this function name is not very accurate as we are not only counting non running pods

pkg/controllers/vdb/podfacts.go
```diff
@@ -995,14 +999,22 @@ func (p *PodFacts) countRunningAndInstalled() int {
...
 	return p.countPods(func(v *PodFact) int {
-		// We don't count non-running pods that aren't yet managed by the parent
-		// sts.  The sts needs to be created or sized first.
-		// We need the pod to have the DC table annotations since the DC
+		// Non-restartable pods are pods that aren't yet running, or don't have
+		// the necesseary DC table annotations, but need to be handled by the
```
roypaulin, line 1007:
```suggestion
		// the necessary DC table annotations, but need to be handled by the
```

pkg/controllers/vdb/podfacts.go
spilchen: Good point. Maybe countNotRestartablePods()?

roypaulin (approved):

PR #632, opened 2023-12-13
Allow upgrade from admintools to vcluster

chinhtranvan:
After upgrading from Admintools 23.4 to VClusterOPs, the HTTPSPollNodeStateOp in startDB times out and indicates 0 host(s) up. This issue arises due to the absence of httpCerts setup in AT for versions below 24.1. This PR generates httpCerts setup to enable HTTP service for vclusterOps


roypaulin:
There is a reconciler `httpservercertgen_reconciler` that generates cert secret if one does not exist but is only run with vclusterops. You also need to update it otherwise each time `NMATLSSecret is not specified you will get a `secret not found` error.

roypaulin:
Also take a closer look at https://github.com/vertica/vertica-kubernetes/pull/584 and see what has been changed/removed and needs to be brought back, like `podfacts.go`

chinhtranvan:
I test it manually, I can upgrade successfully with this change.
 Type    Reason                   Age   From                Message
  ----    ------                   ----  ----                -------
  Normal  CreateDBStart            11m   verticadb-operator  Starting create database
  Normal  CreateDBSucceeded        10m   verticadb-operator  Successfully created database with subcluster 'pri'. It took 1m6.8127633s
  Normal  UpgradeStart             119s  verticadb-operator  Vertica server upgrade has started.
  Normal  ClusterRestartStarted    40s   verticadb-operator  Starting restart of the cluster
  Normal  ClusterRestartSucceeded  15s   verticadb-operator  Successfully restarted the cluster and it took 25s
  Normal  UpgradeSucceeded         13s   verticadb-operator  Vertica server upgrade has completed successfully.  New image is 'vertica-k8s:24.1.0-v

pkg/controllers/vdb/httpservercertgen_reconciler.go
```diff
@@ -54,10 +53,6 @@ func MakeHTTPServerCertGenReconciler(vdbrecon *VerticaDBReconciler, log logr.Log
 // Reconcile will create a TLS secret for the http server if one is missing
 func (h *HTTPServerCertGenReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
 	const PKKeySize = 2048
-	// Early out if the NMA isn't going to be used.
-	if !vmeta.UseVClusterOps(h.Vdb.Annotations) {
-		return ctrl.Result{}, nil
-	}
```
chinhtranvan, line 60: As discussed in the scrum, we can't change the order for HTTPServerCertGenReconciler. We will leave it to generate HTTPServerCertGenReconciler for all AT database versions.

pkg/controllers/vdb/install_reconciler.go
```diff
@@ -266,6 +305,11 @@ func (d *InstallReconciler) genCreateConfigDirsScript(p *PodFact) (string, error
 	}
 
 	if !vinf.IsEqualOrNewer(vapi.InDatabaseLogRotateMinVersion) {
+		if !p.dirExists[paths.HTTPTLSConfDir] {
```
roypaulin, line 308: This block has nothing to do with logrotate so let's move it out.

pkg/controllers/vdb/install_reconciler.go
```diff
@@ -101,6 +102,17 @@ func (d *InstallReconciler) installForAdmintools(ctx context.Context) (ctrl.Resu
 		// hosts to the config, we have to know that a re_ip will succeed.
 		d.addHostsToATConf,
 	}
+	// generateHTTPCerts setup is only required for versions before 24.1.0 of the database.
+	vinf, err := d.Vdb.MakeVersionInfoCheck()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if vinf.IsOlder(vapi.VcluseropsAsDefaultDeploymentMethodMinVersion) {
```
roypaulin, line 110: @Matt I think the operator should either always generate the certs or do it if image version >= "v23.4.0" for admintools. Because before this change, the operator never generated certs for any admintools deployment. So for an upgrade case like 24.1.0(admintools)-->24.2.0(vclusterops) we are still going to need certs.

pkg/controllers/vdb/uninstall_reconciler.go
spilchen, line 1: I don't think the change to the uninstall reconciler is needed. We can leave the file around if need be.

spilchen (approved):

roypaulin (approved):

PR #633, opened 2023-12-13
Online upgrade to account for terminating pods

spilchen:
The controller was not handling terminating pods correctly. This could lead to problems during online upgrades, such as not properly restarting primaries before taking down secondaries.

The issue occurs when the operator deletes pods with the old image. If the pod is still running immediately after the deletion, the operator may skip the restart and move on to the next phase of the upgrade. This is a small window because the termination grace period is 0 seconds, but we still see it occasionally in the e2e runs.

To test this issue, I added a new annotation (vertica.com/termination-grace-period-seconds) that makes it easier to hit this timing scenario. This annotation causes the pod to be in the termination step for a longer period of time.

The solution was to check for pod termination in podfacts and treat the pod as not running if it is set.

### DIFF
--- a/pkg/controllers/vdb/podfacts.go
+++ b/pkg/controllers/vdb/podfacts.go
@@ -69,9 +69,9 @@ type PodFact struct {
 	// true means the pod exists in k8s.  false means it hasn't been created yet.
 	exists bool
 
-	// true means the pod has been bound to a node, and all of the containers
-	// have been created. At least one container is still running, or is in the
-	// process of starting or restarting.
+	// true means the pod has been bound to a node, not in the process of being
+	// process, and all of the containers have been created. At least one
+	// container is still running, or is in the process of starting or restarting.
 	isPodRunning bool
 
 	// true means the statefulset exists and its size includes this pod.  The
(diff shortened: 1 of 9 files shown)


pkg/controllers/vdb/podfacts.go
```diff
@@ -69,9 +69,9 @@ type PodFact struct {
 	// true means the pod exists in k8s.  false means it hasn't been created yet.
 	exists bool
 
-	// true means the pod has been bound to a node, and all of the containers
-	// have been created. At least one container is still running, or is in the
-	// process of starting or restarting.
+	// true means the pod has been bound to a node, not in the process of being
+	// process, and all of the containers have been created. At least one
```
roypaulin, line 73: This sentence does not read well.

pkg/meta/annotations.go
```diff
@@ -140,6 +140,9 @@ const (
 	// annotation is not provided the default value "dbadmin" will be used.
 	SuperuserNameAnnotation   = "vertica.com/superuser-name"
 	SuperuserNameDefaultValue = "dbadmin"
+
+	// Annotation to control the termination grace period for vertica pods.
+	TerminationGracePeriodSecondsAnnotaton = "vertica.com/termination-grace-period-seconds"
```
roypaulin, line 145: Is there any use for this annotation other than reproducing that issue?

roypaulin (commented):
Looks good!

pkg/meta/annotations.go
spilchen: Yes, only for test purposes. There are no plans to externalize this annotation.

roypaulin (approved):

PR #636, opened 2023-12-15
Read secrets from Amazon Secrets Manager

spilchen:
This is a refactor of the cloud package. I am taking out the MultiSourceSecretFetcher from the package and moving it to a new package. The new package will have less dependencies, like vapi, ctrl, and logr. The goal is that this new package will be included other products like vcluster and the nma. This will cut down on code duplication, as the similar logic is needed in these other products.

I also added support to read secrets from AWS Secrets Manager. The secret name must have a special path reference followed by the full ARN of the secret in the following formate: awssm://<secret-arn>. This can be used for the following secrets: communal.credentialSecret, nmaTLSCert and passwordSecret.


pkg/cloud/secret.go
```diff
@@ -17,34 +17,31 @@ package cloud
...
+type ControllerSecretFetcher struct {
 	client.Client
 	Log      logr.Logger
 	EVWriter events.EVWriter
 	VDB      *vapi.VerticaDB
 }
 
 // Fetch reads the secret from a secret store. The contents of the secret is successful.
-func (m *MultiSourceSecretFetcher) Fetch(ctx context.Context, secretName types.NamespacedName) (map[string][]byte, error) {
+func (m *ControllerSecretFetcher) Fetch(ctx context.Context, secretName types.NamespacedName) (map[string][]byte, error) {
```
roypaulin, line 44: minor
```suggestion
func (c *ControllerSecretFetcher) Fetch(ctx context.Context, secretName types.NamespacedName) (map[string][]byte, error) {
```

pkg/cloud/secret.go
```diff
@@ -17,34 +17,31 @@ package cloud
...
 )
 
-// MultiSourceSecretFetcher is secret reader that handles retrival from
-// different sources such as Kubernetes secret store and Google Secrets Manager
-// (GSM).
-type MultiSourceSecretFetcher struct {
+// ControllerSecretFetcher is secret reader designed for the verticadb
+// controller. It can handle retrival from different sources, such as Kubernetes
+// secret store, Google Secrets Manager (GSM), etc.
+type ControllerSecretFetcher struct {
```
roypaulin, line 36: Isn't `VerticaDBSecretFetcher` more accurate?

roypaulin (commented):
Looks good!

pkg/cloud/secret.go
spilchen: I like this name.

pkg/secrets/fetcher.go
```diff
@@ -0,0 +1,120 @@
...
+// out the different secret stores.
+type NotFoundError struct {
+	msg string
+}
+
+func (e *NotFoundError) Error() string {
+	return e.msg
+}
+
+// Fetch reads the secret from a secret store. The contents of the secret is successful.
```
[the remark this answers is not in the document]
spilchen: I missed a word. I updated it to be: The contents of the secret is returned on success.

roypaulin (approved):

PR #640, opened 2023-12-18
Add restore parameters to v1 API

jizhuoyu:
- add `RestorePointPolicy` type in `Spec`
  - `Archive` must be specified if a restore is intended
  - one of `ID` and `Index` must be specified (`Index` assumed to be 1-based)
- webhook rules to validate `RestorePointPolicy`
- revivedb_reconciler checks whether restore is supported given the server version and deployment method
  - must use vclusterops and has server version greater than v24.2.0


api/v1/verticadb_webhook.go
```diff
@@ -211,6 +211,18 @@ func (v *VerticaDB) hasValidInitPolicy(allErrs field.ErrorList) field.ErrorList
 	case CommunalInitPolicyCreate:
 	case CommunalInitPolicyCreateSkipPackageInstall:
 	case CommunalInitPolicyRevive:
+		if v.IsRestoreConfig() && !v.Spec.RestorePoint.IsValidRestorePointPolicy() {
```
roypaulin, line 214: I suggest you check `RestorePoint` in its own function and there you can just exit earlier if the initPolicy doesn't support it.

Also the error msg is too long.

api/v1/verticadb_types.go
```diff
@@ -341,6 +346,8 @@ const (
 	CommunalInitPolicyCreate = "Create"
 	// The database in the communal path will be initialized in the VerticaDB
 	// through a revive_db.  The communal path must have a preexisting database.
+	// This option could also be used to initialize the database by restoring
+	// from a database archive if restorePoint field is properly specified.
 	CommunalInitPolicyRevive = "Revive"
```
roypaulin, line 351: I thought we were going to add a new initPolicy option `Restore`.

api/v1/verticadb_types.go
```diff
@@ -354,6 +361,27 @@ const (
 	CommunalInitPolicyCreateSkipPackageInstall = "CreateSkipPackageInstall"
 )
 
+// RestorePointPolicy is used to locate the exact archive and restore point within archive
+// when a database restore is intended
+type RestorePointPolicy struct {
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced"
```
roypaulin, line 367: `RestorePoint` in VerticaDBSpec already has the xDescriptor `ui:advanced` marker so you don't need to add it again to each of its subfield. `// +operator-sdk:csv:customresourcedefinitions:type=spec` is enough.

api/v1/verticadb_types.go
```diff
@@ -354,6 +361,27 @@ const (
...
+	// +kubebuilder:validation:Optional
+	// Name of the restore archive to use for bootstapping.
+	// This name refers to an object in the database.
+	// This must be specified if initPolicy is Revive and a restore is intended.
+	Archive string `json:"archive,omitempty"`
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced"
```
roypaulin, line 373:
```suggestion
	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number"
```

api/v1/helpers.go
```diff
@@ -572,3 +572,11 @@ func (v *VerticaDB) GetEncryptSpreadComm() string {
 func (v *VerticaDB) IsKSafetyCheckStrict() bool {
 	return vmeta.IsKSafetyCheckStrict(v.Annotations)
 }
+
+func (r *RestorePointPolicy) IsValidRestorePointPolicy() bool {
```
roypaulin, line 576: Can you add a description for each of these functions?

pkg/controllers/vdb/revivedb_reconciler.go
```diff
@@ -96,6 +101,18 @@ func (r *ReviveDBReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ct
 	return g.checkAndRunInit(ctx)
 }
 
+func (r *ReviveDBReconciler) isRestoreSupported() error {
```
roypaulin, line 104: I suggest `hasCompatibleVersionForRestorePoint()`

pkg/version/version.go
```diff
@@ -92,6 +92,33 @@ func (i *Info) IsEqualOrNewer(inVer string) bool {
 	return true
 }
 
+// IsNewer returns true if the version in the Vdb is newer
+// than the given version
+func (i *Info) IsNewer(inVer string) bool {
```
roypaulin, line 97: There is a lot of code duplication with  `IsEqualOrNewer`

roypaulin (commented):
I see this is still a draft but given that you have requested my review I am adding a few comments.

api/v1/version.go
```diff
@@ -56,6 +56,8 @@ const (
 	VcluseropsAsDefaultDeploymentMethodMinVersion = "v24.1.0"
 	// Starting in v24.1.0, we use server logrotate and not depend on cron job
 	InDatabaseLogRotateMinVersion = "v24.1.0"
+	// Starting in v24.2.0 (not including), restoring from a restore point in archive is supported.
```
spilchen, line 59: We will support restore in 24.2.0.

api/v1/verticadb_types.go
```diff
@@ -117,6 +117,11 @@ type VerticaDBSpec struct {
 	// options are to create a new database or revive an existing one.
 	InitPolicy CommunalInitPolicy `json:"initPolicy"`
 
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced"
+	// The restore point policy specifies which restore point among the archive to restore from
+	RestorePoint RestorePointPolicy `json:"restorePoint,omitempty"`
```
spilchen, line 123: Lets make this a pointer. That way we can set it to nil if not needed.

pkg/controllers/vdb/revivedb_reconciler.go
```diff
@@ -96,6 +101,18 @@ func (r *ReviveDBReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ct
...
+	if vinf, ok := r.Vdb.MakeVersionInfo(); !vmeta.UseVClusterOps(r.Vdb.Annotations) ||
```
spilchen, line 105: Let's use `MakeVersionInfoCheck`. That way if we can't find the version in the vdb, we will return a specific error, rather than the restore error below.

pkg/controllers/vdb/revivedb_reconciler.go
```diff
@@ -96,6 +101,18 @@ func (r *ReviveDBReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ct
...
+		!ok || !vinf.IsNewer(vapi.RestoreUnsupportedMaxVersion) {
+		r.VRec.Eventf(r.Vdb, corev1.EventTypeWarning, events.ReviveDBRestoreUnsupported,
+			fmt.Sprintf("Restoring from a restore point is unsupported in ReviveDB given the current server version and deployment method, "+
```
spilchen, line 108: If we are generating the same error message for both, lets generate it once and use it as a string. We can use `VRec.Event` and `errors.New` if the message is already created.

api/v1/verticadb_types.go
```diff
@@ -117,6 +117,11 @@ type VerticaDBSpec struct {
 	// options are to create a new database or revive an existing one.
 	InitPolicy CommunalInitPolicy `json:"initPolicy"`
 
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced"
+	// The restore point policy specifies which restore point among the archive to restore from
```
spilchen, line 122: In future we may use part of the RestorePoint struct to know what archive to create a save point in. It won't necessarily only be used for bootstraping a database with a restore. Can we change the language slightly?
```suggestion
	// Specifies the restore point details to use with this instance of the VerticaDB.
```

pkg/controllers/vdb/revivedb_reconciler.go
```diff
@@ -80,6 +80,11 @@ func (r *ReviveDBReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	// Check if restoring from a restore point is supported
+	if err := r.isRestoreSupported(); r.Vdb.IsRestoreConfig() && err != nil {
```
spilchen, line 84: Should we revert the check? Check if its a restore, then validate that we support it. Otherwise, we will always generate the event message and error, but can choose to ignore it if not doing a restore.

api/v1/version.go
jizhuoyu: Isn't it true that
> 24.2.0 will be blocked too albeit temporarily.

?

api/v1/verticadb_webhook.go
jizhuoyu: I can move the logic into its own function to make it more clear and readable, though it won't necessarily let us exit earlier as anyway all the checks in the webhook will be executed.

api/v1/verticadb_types.go
jizhuoyu: As discussed, we would like to proceed with a simpler design where we reuse `Revive` for the purpose of restore at this moment. We can revisit this design choice later if necessary.

pkg/version/version.go
jizhuoyu: will reduce duplication

api/v1/helpers.go
```diff
@@ -572,3 +572,15 @@ func (v *VerticaDB) GetEncryptSpreadComm() string {
...
+// IsValidRestorePointPolicy will true if the RestorePointPolicy is properly specified,
+// i.e. having a non-empty archive, and either a valid index or a valid id (but not both)
+func (r *RestorePointPolicy) IsValidRestorePointPolicy() bool {
+	return r != nil && r.Archive != "" && ((r.Index > 0) != (r.ID != ""))
+}
+
+// IsRestoreConfig will return whether the vdb is configured to initialize by reviving from
+// a restore point in an archive
+func (v *VerticaDB) IsRestoreConfig() bool {
+	return v.Spec.InitPolicy == CommunalInitPolicyRevive && v.Spec.RestorePoint != nil && v.Spec.RestorePoint.Archive != ""
```
roypaulin, line 585:
```suggestion
	return v.Spec.InitPolicy == CommunalInitPolicyRevive && v.Spec.RestorePoint != nil
```
We are checking archive later so do we need this here. With this the webhook will never let the user know that their CR is wrong because Archive is empty.

api/v1/verticadb_webhook.go
```diff
@@ -223,6 +224,20 @@ func (v *VerticaDB) hasValidInitPolicy(allErrs field.ErrorList) field.ErrorList
 	return allErrs
 }
 
+func (v *VerticaDB) hasValidRestorePolicy(allErrs field.ErrorList) field.ErrorList {
+	if v.IsRestoreConfig() && !v.Spec.RestorePoint.IsValidRestorePointPolicy() {
+		err := field.Invalid(field.NewPath("spec").Child("restorePoint"),
+			v.Spec.RestorePoint,
+			fmt.Sprintf("restorePoint is invalid. When initPolicy is set to %s and restorePoint.archive is non-empty, "+
```
roypaulin, line 231: Instead of one big msg, how about we split it according to the failure reason?

api/v1/version.go
jizhuoyu: I copied this from JIRA ticket description. As discussed, we will now set v24.2.0 as the min version that supports restore.

pkg/version/version.go
```diff
@@ -30,6 +30,14 @@ type Info struct {
 	Components        // The same version as VdbVer but broken down into individual components
 }
 
+type ComparisonResult int
+
+const (
+	CompareEqual   ComparisonResult = iota // CompareEqual represents the state where versions are equal
+	CompareLarger                          // CompareLarger represents the state where the first version is larger
+	CompareSmaller                         // CompareSmaller represents the state where the first version is smaller
```
spilchen, line 38: Lets avoid exporting these const values. Can you begin them with lowercase?

roypaulin: So we are keeping these changes?

jizhuoyu: changed to lowercase

spilchen (approved):

pkg/version/version.go
spilchen: I was okay with the refactor.

PR #641, opened 2023-12-18
Allow upgrade through a deployment change

spilchen:
If upgrading from an admintools deployment to a vclusterops deployment, this change is needed if doing that via an online upgrade. We can't handle an online upgrade in this fashion because the NMA isn't running in pods still doing an admintools style of deployment. We have to take down any pods running admintools deployments. This will make the online upgrade offline, but at leaste then we can proceed.

### DIFF
--- a/pkg/controllers/vdb/onlineupgrade_reconciler.go
+++ b/pkg/controllers/vdb/onlineupgrade_reconciler.go
@@ -468,7 +468,26 @@ func (o *OnlineUpgradeReconciler) checkVersion(ctx context.Context, sts *appsv1.
 		}
 		return &PodFact{}, false
 	}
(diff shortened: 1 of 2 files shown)


roypaulin:
How will this trigger offlineupgrade? Or are just doing onlineupgrade in an offlineupgrade fashion?

spilchen:
> How will this trigger offlineupgrade? Or are just doing onlineupgrade in an offlineupgrade fashion?

It will be more like the latter. We don't know until we are in the middle of an online upgrade that we cannot proceed. So, we simply take down the pods for the secondaries that are running admintools. The rest of the upgrade will function much like an offline upgrade.

PR #644, opened 2023-12-20
Fix upgrades across deployments when ignoring upgrade paths

spilchen:
In a previous change, we transition from an online to an offline upgrade if you attempt do to an upgrade across deployments. Across deployments means you are changing from admintools to vclusterops.  The original fix worked, but we mistakenly bypass it if we are also ignoring upgrade paths. This change ensures we handle things properly if also ignoring upgrade paths.


roypaulin (approved):
Looks good!

PR #645, opened 2023-12-20
Generate prometheus_metrics_latest.json

spilchen:
This adds a file to the repo called prometheus_metrics_latest.json. It is a summary of all of the metrics exposed by the operator. The purpose of this file is for doc writers to use it to include details about the metrics found in the operator.

There is an e2e test (metrics-no-auth) that will ensure this file stays up to date.

@rs-vertica I included some metrics in the json that the operator inherits from the SDK. Here is some more detail about those metrics here: https://book.kubebuilder.io/reference/metrics-reference. It looks like it is just a summary of the help text. I can omit those and just include the ones we add. We don't have any control over the help text. So, if the ones we inherit have any spelling and/or grammatical errors then we would need to manually fix them. Let me know if you want me to omit some.

### DIFF
--- a/prometheus_metrics_latest.json
+++ b/prometheus_metrics_latest.json
@@ -1,0 +1,306 @@
+{
+    "generatedWith": "parse_prometheus_metrics.py",
+    "info": "Recreate by extracting output in kuttl test 'kubectl test --test metrics-no-auth'",
+    "metrics": [
+        {
+            "description": "Number of currently used workers per controller",
+            "name": "controller_runtime_active_workers",
+            "type": "gauge"
+        },
+        {
+            "description": "Maximum number of concurrent reconciles per controller",
+            "name": "controller_runtime_max_concurrent_reconciles",
+            "type": "gauge"
+        },
+        {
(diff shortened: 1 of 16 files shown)


rs-vertica (approved):
I looked over the additional metrics, and I think we can include them. There  weren't any types, and the descriptions more or less follow the same format that we use for our metrics, so I don't think there is a consistency issue. Plus, it will be easier on users to just get them all in one place rather than following links.

prometheus_metrics_latest.json
```diff
@@ -0,0 +1,306 @@
+{
+    "generatedWith": "parse_prometheus_metrics.py",
+    "info": "Recreate by extracting output in kuttl test 'kubectl test --test metrics-no-auth'",
+    "metrics": [
+        {
+            "description": "Number of currently used workers per controller",
```
rs-vertica, line 6: Add a period to the end of each description.

spilchen: Some of these we don't have control over the text. But I was able to add a period at the end of each description in the script that pulls the metrics.

PR #650, opened 2024-01-02
Check dependent objects at the end of a reconcile

spilchen:
This issue closes a timing loophole in which the removal of a dependent object does not always trigger a reconcile. This was observed occasionally in the use-own-serviceaccount e2e test. We originally relied on watches that we set up with kube-apiserver. However, there appears to be a small window of time in which the removal of a watched object does not trigger a reconcile.


roypaulin (approved):
Looks good! minor: can you add a description to some of these functions?

PR #652, opened 2024-01-03
VCluster API refresh: allow create db to proceed if db is already running

spilchen:
This builds on a change made in vclusterops. It can now emit a specific error if it discovers that the database is already running. We add a feature flag to tolerate this (the default is true). This can allow us to proceed if we are in an environment with few resources available, which can cause instability during the creation process. If we have previously progressed far enough that Vertica is running, we will skip the create db step and proceed to the next reconciler. This will update the status condition to indicate that the database has been initialized.

### DIFF
--- a/pkg/meta/annotations.go
+++ b/pkg/meta/annotations.go
@@ -145,6 +145,16 @@ const (
 
 	// Annotation to control the termination grace period for vertica pods.
 	TerminationGracePeriodSecondsAnnotaton = "vertica.com/termination-grace-period-seconds"
+
+	// During the create database process, if we discover that Vertica is
+	// already running, we can either treat this as an error or continue.
+	// Continuing may be the best option if we have already progressed far
+	// enough in the create database process to have created the catalog. We can
+	// then continue to start any nodes that may be down in order to bring the
+	// cluster online.
+	FailCreateDBIfVerticaIsRunningAnnotation      = "vertica.com/fail-create-db-if-vertica-is-running"
+	FailCreateDBIfVerticaIsRunningAnnotationTrue  = "true"
+	FailCreateDBIfVerticaIsRunningAnnotationFalse = "false"
 )
 
 // IsPauseAnnotationSet will check the annotations for a special value that will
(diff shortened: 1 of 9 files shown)


pkg/vadmin/restart_node_vc.go
```diff
@@ -54,21 +54,17 @@ func (v *VClusterOps) RestartNode(ctx context.Context, opts ...restartnode.Optio
...
-			Cert:           certs.Cert,
-			CaCert:         certs.CaCert,
-			UserName:       &su,
-			Password:       &v.Password,
-			HonorUserInput: &honorUserInput,
-		},
-		Nodes: s.RestartHosts,
-	}
-	// timeout option
+	opts := vops.VStartNodesOptionsFactory()
```
spilchen, line 57: Restructured code here so that we call the factor function for opts. This was needed to properly initialize a new field added to the opts.

pkg/vadmin/create_db_vc.go
```diff
@@ -47,6 +49,21 @@ func (v *VClusterOps) CreateDB(ctx context.Context, opts ...createdb.Option) (ct
...
 	vdb, err := v.VCreateDatabase(&vopts)
 	if err != nil {
+		// If it is determined that vertica is already running, we may allow the
+		// failure to continue. This could be a result of a previous failed
+		// create db that was able to partially complete but ultimately failed
+		// before updating the status condition to indicate that the create db
+		// was finished. By ignoring the error this time, we will be able to
+		// update the status condition and start any nodes that may be down in
+		// order to bring the cluster online.
+		dbIsRunningError := &vops.DBIsRunningError{}
```
roypaulin, line 59: What do you mean by partially complete? Do you mean on the operator side?

pkg/meta/annotations.go
```diff
@@ -238,6 +248,12 @@ func GetTerminationGracePeriodSeconds(annotations map[string]string) int {
 	return lookupIntAnnotation(annotations, TerminationGracePeriodSecondsAnnotaton)
 }
 
+// FailCreateDBIfVerticaIsRunning returns how to handle failures during create
+// db if vertica is found to be running.
+func FailCreateDBIfVerticaIsRunning(annotations map[string]string) bool {
```
roypaulin, line 253: Can you tell what returning true means?

roypaulin (commented):
Looks good!

pkg/vadmin/create_db_vc.go
spilchen: No, this is the server side. The cases we saw is that create db fails during the package installation. If we got far enough to bring up a vertica process, chances are the database is fine to use.

roypaulin (approved):

PR #653, opened 2024-01-04
Upgrade path detection will now only prevent downgrades

spilchen:
This operator update removes logic that prevented upgrades if the user skipped too many releases. This change is consistent with a recent update to the Vertica documentation, which no longer requires users to upgrade through all intermediate releases.

The old documentation stated that "Vertica upgrades must be performed in a linear fashion, with each release upgraded before the next." However, this requirement has been removed. Users can now upgrade to any supported release, regardless of whether they have skipped intermediate releases.

The only restriction that the operator still enforces is that users cannot downgrade.


roypaulin (commented):
You must also `IgnoreUpgradePathAnnotation`  now that we can skip any release.

roypaulin (approved):

PR #654, opened 2024-01-05
Support for running the NMA as a sidecar

roypaulin:
This adds initial support in the operator to be able to run the NMA as a sidecar. This is the default way of deploying the statefulset with the v1 VerticaDB API. This requires a server version of at least 24.2.0. When running 24.1.0 you must set this annotation in the VerticaDB: vertica.com/run-nma-in-sidecar: "false". In 24.2.0 or newer, this annotation can be omitted. 

We now properly setup the statefulset for the sidecar deployment. The same image is used for both the NMA and server containers. The NMA container is the first container in the pod spec template. This was done because we want it to be the default one when entering via 'kubectl exec'. This is important when collecting scrutinize as that must be run in the NMA container.

It originally had some container changes, like removing the s6 init process, but we had to back that out as we need to continue to support running the monolithic container in 24.2.0 for the time being.

We added a temporary hack to account for the vertica process turning into a daemon. We add a sleep infinity after running the vertica start command. This ensures the pod never exits under normal circumstances. This will be removed in a subsequent PR once the server side changes are fixed.


api/v1/verticadb_webhook.go
```diff
@@ -1011,6 +1013,20 @@ func (v *VerticaDB) checkImmutableDeploymentMethod(oldObj *VerticaDB, allErrs fi
 	return allErrs
 }
 
+// checkImmutableNMADeploymentMethod will check if the nma deployment type is changing from
+// sidecar to monolithic
+func (v *VerticaDB) checkImmutableNMADeploymentMethod(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {
```
roypaulin, line 1018: I am not sure if we want any kind of check but I added this because I thought it should not be possible to go from sidecar to monolithic

docker-vertica-v2/Dockerfile
```diff
@@ -121,23 +118,7 @@ ENV VERTICA_STARTUP_LOG_DUPLICATE "/proc/1/fd/1"
 
 COPY ./packages/init.d.functions /etc/rc.d/init.d/functions
```
roypaulin, line 119: What is this for?

docker-vertica-v2/Dockerfile
```diff
@@ -121,23 +118,7 @@ ENV VERTICA_STARTUP_LOG_DUPLICATE "/proc/1/fd/1"
...
-COPY s6-rc.d/ /etc/s6-overlay/s6-rc.d/
-
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-RUN set -x \
-  # Untar the init program that was downloaded earlier
-  && tar -C / -Jxpf /tmp/s6-overlay-x86_64.tar.xz \
-  && tar -C / -Jxpf /tmp/s6-overlay-noarch.tar.xz
-
-ENTRYPOINT [ "/init" ]
+ENTRYPOINT [ "/opt/vertica/bin/node_management_agent" ]
```
roypaulin, line 121: I set this as default entrypoint. But do we want it to br possible to use it as a monolithic container(it will useful if the image is used outside of k8s)?

pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -231,6 +228,9 @@ func (r *VerticaDBReconciler) constructActors(log logr.Logger, vdb *vapi.Vertica
 		MakeObjReconciler(r, log, vdb, pfacts, ObjReconcileModeAll),
 		// Set version info in the annotations and check that it is the minimum
 		MakeImageVersionReconciler(r, log, vdb, prunner, pfacts, false),
+		// Check NMA running mode and report error when not supported by
+		// the server version
+		MakeNMARunningModeReconciler(r, log, vdb),
```
roypaulin, line 233: Is there something in particular we want to do in this reconciler? For now I just check the version and the annotations

pkg/builder/builder.go
```diff
@@ -1160,6 +1215,46 @@ func buildCanaryQuerySQL(vdb *vapi.VerticaDB) string {
 	return fmt.Sprintf("vsql %s -c 'select 1'", passwd)
 }
 
+// buildVerticaStartCommand returns the vertica start command that
+// will serve as entrypoint to the server container
+func buildVerticaStartCommand() []string {
+	startCmd := fmt.Sprintf("/opt/vertica/bin/vertica --startup-conf %s", paths.StartupConfFile)
```
roypaulin, line 1221: I am going to update it when the issue on the server is fixed.

pkg/builder/builder.go
```diff
@@ -620,7 +650,29 @@ func makeServerContainer(vdb *vapi.VerticaDB, sc *vapi.Subcluster) corev1.Contai
...
+	envVars := buildNMAEnvVars(vdb)
+	envVars = append(envVars, corev1.EnvVar{
+		Name:  NMALogPath,
+		Value: StdOut,
+	})
+
+	return corev1.Container{
+		Image:           pickImage(vdb, sc),
+		ImagePullPolicy: vdb.Spec.ImagePullPolicy,
+		Name:            names.NMAContainer,
```
roypaulin, line 672: What about the resource for this container? How to we want to set it? I did not want to just set the same resource as the vertica container because I think they have different needs.

roypaulin:
I would like to get some early comments.

pkg/builder/builder.go
```diff
@@ -140,7 +145,36 @@ func buildConfigVolumeMount(vdb *vapi.VerticaDB) corev1.VolumeMount {
 	}
 }
 
-// buildVolumeMounts returns the volume mounts to include in the sts pod spec
+// buildServerVolumeMounts returns the volume mounts to include
+// in the server container
+func buildServerVolumeMounts(vdb *vapi.VerticaDB) []corev1.VolumeMount {
+	volMnts := buildVolumeMounts(vdb)
```
spilchen, line 151: We mount a subpath to scrutinize at /tmp/scrutinize. I think we should do that only in the nma container. We should not run scrutinize in the server container. We might not even have the NMA TLS certs mounted, which scrutinize uses.

spilchen, line 151: If we don't mount /tmp/scrutinize in the server container, then we will need to update scripts/capture-scrutinize.sh. We need to run scrutinize in the NMA container if that exists.

EDIT: probably no change here if the nma is the first container. That will default to the nma anyway.

pkg/builder/builder.go
```diff
@@ -838,6 +890,9 @@ func makeServerSecurityContext(vdb *vapi.VerticaDB) *corev1.SecurityContext {
 // makeContainers creates the list of containers to include in the pod spec.
 func makeContainers(vdb *vapi.VerticaDB, sc *vapi.Subcluster) []corev1.Container {
 	cnts := []corev1.Container{makeServerContainer(vdb, sc)}
+	if vdb.IsSideCarDeploymentEnabled() {
```
spilchen, line 893: I think we may want the NMA to be the first container. The first container is the default one if you do a `kubectl exec` without specifying a container. This will come in handy if you want to run scrutinize. As it should be the nma container that you run this in.

pkg/controllers/vdb/verticadb_controller.go
spilchen: We can probably remove this reconciler and move its logic into ImageVersionReconciler.

pkg/names/names.go
```diff
@@ -24,8 +24,13 @@ import (
...
 
 const (
-	ServerContainer      = "server"
-	ServerContainerIndex = 0
+	ServerContainer = "server"
+	NMAContainer    = "nma"
+)
+
+const (
+	ServerContainerIndex = iota
```
spilchen, line 32: I guess if we set the nma as the first container, we will need to pick ServerContainerIndex based on the deployment.

pkg/controllers/vdb/upgrade.go
```diff
@@ -203,9 +204,13 @@ func (i *UpgradeManager) updateImageInStatefulSets(ctx context.Context) (int, er
 func (i *UpgradeManager) updateImageInStatefulSet(ctx context.Context, sts *appsv1.StatefulSet) (bool, error) {
 	stsUpdated := false
 	// Skip the statefulset if it already has the proper image.
-	if sts.Spec.Template.Spec.Containers[names.ServerContainerIndex].Image != i.Vdb.Spec.Image {
+	cnts := sts.Spec.Template.Spec.Containers
+	if !i.Vdb.CheckIfContainersHaveVDBImage(cnts, names.ServerContainerIndex) {
 		i.Log.Info("Updating image in old statefulset", "name", sts.ObjectMeta.Name)
 		sts.Spec.Template.Spec.Containers[names.ServerContainerIndex].Image = i.Vdb.Spec.Image
+		if i.Vdb.IsSideCarDeploymentEnabled() {
+			sts.Spec.Template.Spec.Containers[names.NMAContainerIndex].Image = i.Vdb.Spec.Image
```
spilchen, line 212: We update both images together. Is there a case where the server and NMA will use different images? I am trying to understand why we have the `CheckIfContainersHaveVDBImage`. Shouldn't checking the server container be sufficient.

pkg/controllers/vdb/podfacts.go
```diff
@@ -431,7 +431,7 @@ func (p *PodFacts) genGatherScript(vdb *vapi.VerticaDB, pf *PodFact) string {
 		echo -n 'vnodeName: '
 		cd %s/%s/v_%s_node????_catalog 2> /dev/null && basename $(pwd) | rev | cut -c9- | rev || echo ""
 		echo -n 'verticaPIDRunning: '
-		[[ $(pgrep ^vertica) ]] && echo true || echo false
+		[[ $(pgrep -f "^.*vertica\s-D") ]] && echo true || echo false
```
spilchen, line 434: Why is this change needed?

spilchen (commented):
This is really good. I like how we worked around the server issue for now. I created VER-91286 to address that problem.

pkg/controllers/vdb/podfacts.go
roypaulin: Because the server container container starts there is already a vertica process polling for the config file, so the node was always reported up. This is to search for `/opt/vertica/bin/vertica -D***` which is the actual start command. This is just a work around to allow me to progress.

pkg/controllers/vdb/upgrade.go
roypaulin: Maybe I just overthought. I was thinking about upgrade but even in that case the new container will have the same image. I will make it simpler.

tests/e2e-server-upgrade-at-only/offline-upgrade-ks-0/setup-vdb-bad-endpoint/customized-overlay/kustomization.yaml
```diff
@@ -19,8 +19,5 @@ patches:
     - op: add
       path: /metadata/annotations/vertica.com~1vcluster-ops
       value: "true"
-    - op: add
```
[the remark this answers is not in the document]
roypaulin: got you!

scripts/setup-kustomize.sh
```diff
@@ -192,9 +192,6 @@ EOF
     - op: add
       path: /metadata/annotations/vertica.com~1vcluster-ops
       value: "true"
-    - op: add
```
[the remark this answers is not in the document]
roypaulin: I will add a new env.

pkg/builder/builder.go
roypaulin: I missed that part, thanks!

pkg/builder/builder.go
roypaulin: I will have nma be the first container.

pkg/names/names.go
roypaulin: Yes, but given that nma is sidecar will be the permanent solution, it make sense to have it as 1st container.

roypaulin: Yes I mean I can add a logic that picks the correct index based on the deployment type.

pkg/controllers/vdb/podfacts.go
spilchen: This brings up an interesting case. How are we able to determine if vertica process is ready for connections now. We can't do this by checking the process. I guess check for -D works as long as the vertica process runs as a daemon. Maybe we should check the HTTPS endpoint? We can call /v1/version. If that returns HTTP code 200, then vertica is up.

roypaulin: That works too and is consistent with the health probe. What I did here is just expanding the regex to match `/opt/vertica/bin/vertica -D /data/vertdb/v_vertdb_node0001_catalog***`(-D stands for db-path) but not `/opt/vertica/bin/vertica --stattup-config ***`

spilchen: I think we should check endpoint in this PR. Otherwise, it will be something we need to fix when we change the vertica process to stop running as a daemon.

roypaulin: I agree, I will make the change

pkg/builder/builder.go
```diff
@@ -46,16 +46,18 @@ const (
 	VerticaClusterCommPort  = 5434
 	SpreadClientPort        = 4803
 	NMAPort                 = 5554
+	StdOut                  = "/proc/1/fd/1"
```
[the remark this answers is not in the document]
roypaulin: Yes I have the same issue. We can already create a ticket. If I find a quick fix while working on this, we will close it.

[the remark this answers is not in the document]
spilchen: I guess the same issue happens with the server container. We no longer print out the contents of the startup.log.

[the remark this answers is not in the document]
roypaulin: Yes, it prints out spread log.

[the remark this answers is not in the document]
spilchen: I created VER-91291

pkg/builder/builder.go
```diff
@@ -1160,6 +1215,46 @@ func buildCanaryQuerySQL(vdb *vapi.VerticaDB) string {
...
+	return []corev1.EnvVar{
+		// The NMA will read the secrets directly from the secret store.
+		// We provide the secret namespace and name for this reason.
+		{Name: NMASecretNamespaceEnv, Value: vdb.ObjectMeta.Namespace},
+		{Name: NMASecretNameEnv, Value: vdb.Spec.NMATLSSecret},
+	}
+}
+
+func getShareProcessNamespace(vdb *vapi.VerticaDB) bool {
+	return vdb.IsSideCarDeploymentEnabled()
```
spilchen, line 1255: Regarding the stdout issue with NMA and server, lets try running without this set. This may cause those to work. I assume we needed this when talking to EOM folks, but lets find the actual use case before setting it.

scripts/setup-kustomize.sh
```diff
@@ -130,8 +141,32 @@ if [ -n "$PRIVATE_REG_SERVER" ]; then echo "YES"; else echo "NO"; fi
 echo -n "Add server mounts: "
 if [ -n "$USE_SERVER_MOUNT_PATCH" ]; then echo "YES"; else echo "NO"; fi
 echo "Deployment method: $VERTICA_DEPLOYMENT_METHOD"
+echo "NMA running mode: $NMA_RUNNING_MODE"
 echo "Vertica superuser name: $VERTICA_SUPERUSER_NAME"
 
+if [ "$NMA_RUNNING_MODE" == "sidecar" ]
+then
+    cp $RESTART_SANITY/sidecar-40-assert.yaml $RESTART_SANITY/40-assert.yaml
```
roypaulin, line 149: I know this does not look good but kuttl does not support partial comparison of slices in asserts so I did not have many options

roypaulin (commented):
Making nma the 1st container lead to a lot of changes in the tests

roypaulin:
@spilchen ready for another round

pkg/controllers/vdb/podfacts.go
```diff
@@ -566,8 +570,14 @@ func (p *PodFacts) checkForSimpleGatherStateMapping(_ context.Context, _ *vapi.V
 	pf.localDataSize = gs.LocalDataSize
 	pf.localDataAvail = gs.LocalDataAvail
 	pf.admintoolsExists = gs.AdmintoolsExists
-	// If the vertica process is running, then the database is UP. This is
-	// consistent with the liveness probe, which goes a bit further and checks
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		// In vclusterops mode, we call an HTTPS endpoint.
+		// If that returns HTTP code 200, then vertica is up
+		pf.upNode = gs.UpNode
+		return nil
```
spilchen, line 577: I suggest make the upnode check a function rather than exiting early here. If we add new checks in the future, we will need to work around this early exit. So, it could be prone to bugs.

pkg/controllers/vdb/podfacts.go
```diff
@@ -566,8 +570,14 @@ func (p *PodFacts) checkForSimpleGatherStateMapping(_ context.Context, _ *vapi.V
...
+	}
+	// For admintools, if the vertica process is running, then the database is UP.
+	// This is consistent with the liveness probe, which goes a bit further and checks
 	// if the client port is opened. If the vertica process dies, the liveness
 	// probe will kill the pod and we will be able to do proper restart logic.
 	// At one point, we ran a query against the nodes table. But it became
```
spilchen, line 583: Can we go through this file and look at the usage of VerticaPIDRunning? For cases where we use it as a proxy for the db being up, we should just check upNode state.

pkg/controllers/vdb/imageversion_reconciler.go
```diff
@@ -99,7 +99,37 @@ func (v *ImageVersionReconciler) Reconcile(ctx context.Context, _ *ctrl.Request)
...
+
+// Verify whether the NMA is configured to run in sidecar container
+func (v *ImageVersionReconciler) verifyNMARunningMode(vinf *version.Info) error {
+	// versions below 24.2.0 cannot be used to run NMA in a sidecar container
+	if v.Vdb.IsNMASideCarDeploymentEnabled() {
+		if vinf.IsEqualOrNewer(vapi.NMAInSideCarDeploymentMinVersion) {
+			return nil
+		}
+		v.VRec.Eventf(v.Vdb, corev1.EventTypeWarning, events.NMAInSidecarNotSupported,
+			"The NMA version cannot be used for nma sidecar deployment")
```
spilchen, line 118: Can we include the version that does support sidecar?

pkg/names/names.go
```diff
@@ -24,8 +24,14 @@ import (
 )
 
 const (
-	ServerContainer      = "server"
-	ServerContainerIndex = 0
+	ServerContainer = "server"
+	NMAContainer    = "nma"
+	Index0          = 0
```
spilchen, line 29: Can we rename this to `FirstContainer`? I think this is only used when wanted to reference index [0]

pkg/names/names.go
```diff
@@ -24,8 +24,14 @@ import (
...
+)
+
+const (
+	NMAContainerIndex = iota
+	ServerContainerIndex
```
spilchen, line 34: Can we avoid externalizing these outside of the package? External users should be able to use `GetServerContainerIndex`. So suggest renames of: `nmaContainerIndex` and `serverContainerIndex`.

scripts/setup-kustomize.sh
spilchen: I agree this is ugly. What if added a feature flag in the operator via an annotation that let you control the index of the server container. Then for the tests that have a strict dependency that server is first we use that feature flag. This flag would only be used in test purposes and won't be externalized.

roypaulin:
>I agree this is ugly. What if added a feature flag in the operator via an annotation that let you control the index of the server container. Then for the tests that have a strict dependency that server is first we use that feature flag. This flag would only be used in test purposes and won't be externalized.

That won't help much, the issue is kuttl that doesn't suppport partial comparison of slices, which means that when `nmainsidecar` is enabled, we also need to add the nma container to the asserts even though we are only interested in the server. Otherwise you get a slice length mismatch error.

roypaulin:
> Were you going to add some defaults in this PR?

For the resource I set the values you suggested but I could not create the db. Maybe not enough resource on my env? I am not going to set any default resource for nma and on a follow-on jira we are going to add an annotation(as you suggested) to set it.

spilchen:
> > I agree this is ugly. What if added a feature flag in the operator via an annotation that let you control the index of the server container. Then for the tests that have a strict dependency that server is first we use that feature flag. This flag would only be used in test purposes and won't be externalized.
> 
> That won't help much, the issue is kuttl that doesn't suppport partial comparison of slices, which means that when `nmainsidecar` is enabled, we also need to add the nma container to the asserts even though we are only interested in the server. Otherwise you get a slice length mismatch error.

This is only for 3 testcases right? What if we changed those testcases so they run in vclusterops 24.2.0 only (essentially move them to e2e-leg-6). Then we don't need to worry about the case where there is only 1 container.

roypaulin:
> This is only for 3 testcases right? What if we changed those testcases so they run in vclusterops 24.2.0 only (essentially move them to e2e-leg-6). Then we don't need to worry about the case where there is only 1 container.

Yes, I was not sure if we wanted to do that. I will move them to e2e-leg-6.

pkg/controllers/vdb/podfacts.go
roypaulin: Now I kept this because in `checkIfNodeIsDoingStartup`  when `verticaPIDRunning` is false, we return immediately. Without this, as soon as the pod is running i`verticaPIDRunning`  but that's just the start cmd polling the file and I do not want to set `verticaPIDRunning` based on that. Maybe we can keep it until we fix the daemon jira.

roypaulin:
Still making some changes

pkg/controllers/vdb/podfacts.go
spilchen: Okay. Leave a comment with the jira then if you are going to address this later. I had another comment about using verticaPIDRunning less. It would be nice if it was used only for admintools deployments.

roypaulin: We can. There is only one place we use it. I just need to figure out to not use it for vclusterops

roypaulin:
Now 3 legs run with `nmainsidecar` enabled. If any test fails it would be because of some last minute change and I will address it.

spilchen (approved):
Looks good. I'll merge this later today. I want to get a few fixes into main for the 2.0.1 release first. This change will go in the release after 2.0.1.

spilchen:
> Looks good. I'll merge this later today. I want to get a few fixes into main for the 2.0.1 release first. This change will go in the release after 2.0.1.

Actually there are few fixes that are needed for 2.0.1 that won't get in today. I think this is sufficiently gated that it won't cause issues for older releases. So, I'll put this in later today before we have all of the 2.0.1 fixes.

PR #658, opened 2024-01-08
implement status condition transitions for vertica restore points query cr

chinhtranvan:
In the VerticaRestorePointsQuery CRD we have status conditions to know when the query has started and when it has finished.

  conditions:
  - lastTransitionTime: <ts>
    status: "True"
    type: Querying
  - lastTransitionTime: <ts>
    status: "True"
    type: QueryComplete

This PR can add the status condition logic as follows:
- skip reconcile if QueryComplete is true
- set Querying status condition prior to calling vclusterops API
- clear Querying status condition after calling API. Include an error message in the status condition if the API fails.
- set the QueryComplete if the vclusterops API suceeded


api/v1beta1/verticarestorepointsquery_types.go
```diff
@@ -47,8 +48,54 @@ const (
 // VerticaRestorePointsQueryStatus defines the observed state of VerticaRestorePointsQuery
 type VerticaRestorePointsQueryStatus struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=status
-	// RestorePoints used to list out the available restore points.
-	RestorePoints string `json:"restorePoints"`
+	// Conditions for VerticaRestorePointsQuery
+	Conditions []VerticaRestorePointsQueryCondition `json:"conditions,omitempty"`
```
spilchen, line 52: We should use the `metav1.Condition` type rather than creating our own. This allows us to use the condition functions in the operator-sdk package. We use this in the v1 VerticaDB API (search for metav1.Condition).

pkg/controllers/vrpq/verticarestorepointsquery_controller.go
```diff
@@ -102,9 +102,11 @@ func (r *VerticaRestorePointsQueryReconciler) SetupWithManager(mgr ctrl.Manager)
 // constructActors will a list of actors that should be run for the reconcile.
 // Order matters in that some actors depend on the successeful execution of
 // earlier ones.
-func (r *VerticaRestorePointsQueryReconciler) constructActors(_ *vapi.VerticaRestorePointsQuery,
-	_ logr.Logger) []controllers.ReconcileActor {
+func constructActors(r *VerticaRestorePointsQueryReconciler, vrpq *vapi.VerticaRestorePointsQuery,
```
spilchen, line 105: Why did you move this function out of the `VerticaRestorePointsQueryReconciler` struct?

pkg/vadmin/restore_points_query_at.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package vadmin
+
+import (
+	"context"
+)
+
+func (a *Admintools) ListRestorePoints(_ context.Context) error {
+	a.Log.Info("start restore points query")
```
spilchen, line 23: I don't think we will ever implement this in the admintools side. Lets return an error if this is ever called.

pkg/controllers/vrpq/listrestorepoints_reconciler.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/go-logr/logr"
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/controllers"
+	"github.com/vertica/vertica-kubernetes/pkg/vadmin"
+	vrpqstatus "github.com/vertica/vertica-kubernetes/pkg/vrpqstatus"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type ListRestorePointsQueryReconciler struct {
```
spilchen, line 30: We kind of repeat the name in the package. Maybe just call it the `QueryReconciler`?

pkg/controllers/vrpq/listrestorepoints_reconciler.go
```diff
@@ -0,0 +1,86 @@
...
+		v.Vrpq.Status.Conditions[vapi.QueryCompleteIndex].Status == corev1.ConditionTrue {
+		return ctrl.Result{}, nil
+	}
+
+	return ctrl.Result{}, v.setListRestorePointsQueryConditions(ctx, req)
+}
+
+// setListRestorePointsQueryConditions will update the status condition before and after calling
+// list restore points api
+func (v *ListRestorePointsQueryReconciler) setListRestorePointsQueryConditions(ctx context.Context, _ *ctrl.Request) error {
```
spilchen, line 58: We should have a unit test for this reconciler.

spilchen (commented):
We should include a simple e2e test. Even just to wait on the new status condition after creating the CR.

pkg/controllers/vrpq/verticarestorepointsquery_controller.go
chinhtranvan: oops, Will move it back

pkg/controllers/vrpq/listrestorepoints_reconciler_test.go
```diff
@@ -0,0 +1,74 @@
...
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/cmds"
+	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
+	"github.com/vertica/vertica-kubernetes/pkg/test"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const TestPassword = "test-pw"
+
+var _ = Describe("stopdb_reconcile", func() {
```
spilchen, line 33: copy/paste error?
```suggestion
var _ = Describe("query_reconciler", func() {
```

pkg/vadmin/restore_points_query_at.go
```diff
@@ -0,0 +1,25 @@
...
+
+package vadmin
+
+import (
+	"context"
+	"errors"
+)
+
+func (a *Admintools) ListRestorePoints(_ context.Context) error {
+	return errors.New("ListRestorePoints is not supported in the admintools side")
```
spilchen, line 24:
```suggestion
	return errors.New("ListRestorePoints is not supported for admintools deployments")
```

pkg/vrpqstatus/status.go
```diff
@@ -0,0 +1,71 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Update(ctx context.Context, clnt client.Client, log logr.Logger, vrpq *vapi.VerticaRestorePointsQuery,
+	updateFunc func(*vapi.VerticaRestorePointsQuery) error) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		// Always fetch the latest to minimize the chance of getting a conflict error.
+		nm := types.NamespacedName{Namespace: vrpq.Namespace, Name: vrpq.Name}
+		if err := clnt.Get(ctx, nm, vrpq); err != nil {
+			log.Info("VerticaRestorePointsQuery resource not found.  Ignoring since object must be deleted")
```
spilchen, line 37: This should be done if the IsNotFound() function returned true. We can't assume all errors are not found errors.

pkg/controllers/vrpq/listrestorepoints_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+		v1.MakeCondition(vapi.Querying, metav1.ConditionTrue, ""))
+	if err != nil {
+		return err
+	}
+
+	errAPI := q.Dispatcher.ListRestorePoints(ctx)
+	// Include an error message in the status condition if the API fails.
+	if errAPI != nil {
+		q.VRec.Log.Info("Fail to call Vcluster list restore points API")
+		return errAPI
```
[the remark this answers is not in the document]
chinhtranvan: If we want to clear the Querying status condition before we exit, I believe we should move lines 74-79 before the API call fails, as we want to clear it regardless of the API result.

chinhtranvan:
> We should include a simple e2e test. Even just to wait on the new status condition after creating the CR.

@spilchen I apply a manifest yaml file 
```
kind: VerticaRestorePointsQuery
metadata:
  name: archive-query
spec:
  verticaDBName: vert
  archiveName: backup 

```
Then I use ` k describe VerticaRestorePointsQuery archive-query` and get this info 
```
Name:         archive-query
Namespace:    test
Labels:       <none>
Annotations:  <none>
API Version:  vertica.com/v1beta1
Kind:         VerticaRestorePointsQuery
Metadata:
  Creation Timestamp:  2024-01-10T04:34:57Z
  Generation:          1
  Resource Version:    14783
  UID:                 77670a09-d6e6-4fcd-b298-00d4ba0a6803
Spec:
  Archive Name:     backup
  Vertica DB Name:  vert
Status:
  Conditions:
    Last Transition Time:  2024-01-10T04:35:02Z
    Message:               
    Reason:                UnKnown
    Status:                True
    Type:                  Querying
    Last Transition Time:  2024-01-10T04:35:02Z
    Message:               
    Reason:                UnKnown
    Status:                True
    Type:                  QueryComplete
Events:                    <none>
```
I will open another Jira to add this simple e2e tests, since my brain doesn't work well to figure out the way to add this e2e test.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -0,0 +1,76 @@
...
+	// set Querying status condition prior to calling vclusterops API
+	err := vrpqstatus.UpdateCondition(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+		v1.MakeCondition(vapi.Querying, metav1.ConditionTrue, ""))
+	if err != nil {
+		return err
+	}
+
+	// clear Querying status condition
+	err = vrpqstatus.UpdateCondition(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+		v1.MakeCondition(vapi.Querying, metav1.ConditionUnknown, ""))
```
spilchen, line 68: Any reason why we didn't pick `metav1.ConditionalFalse` to turn off the status condition.  Unknown is reserved for "k8s can't decide if a resource is in the condition or not"

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -0,0 +1,76 @@
...
+// Temporarily, runListRestorePoints will not call the ListRestorePoints API
+// since the dispatcher is not set up yet
+func (q *QueryReconciler) runListRestorePoints(ctx context.Context, _ *ctrl.Request) error {
+	// set Querying status condition prior to calling vclusterops API
+	err := vrpqstatus.UpdateCondition(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+		v1.MakeCondition(vapi.Querying, metav1.ConditionTrue, ""))
+	if err != nil {
+		return err
+	}
+
```
spilchen, line 65: Maybe add a comment to indicate the API call should be done here

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: yes, will update to `metav1.ConditionalFalse`

spilchen:
Latest changes look good. As discussed in scrum, lets add the e2e test to this PR.

tests/e2e-leg-6/verify-vrpq-conditions/setup-vrpq/base/setup-vrpq.yaml
spilchen, line 1: I don't think we need kustomize for this. We don't do any kind of substitution. Lets put this CR right in step 3 and avoid making the kustomize call.

tests/e2e-leg-6/verify-vrpq-conditions/03-setup.yaml
```diff
@@ -11,7 +11,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: kuttl.dev/v1betat1
+apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
-  - command: bash -c "kustomize build setup-vrpq/overlay | kubectl -n $NAMESPACE apply -f - "
+  - script: |
```
spilchen, line 17: You don't need to use kubectl here. Just replace the kuttl.dev/v1beta1 YAML manifest with your new CR.

spilchen (approved):

PR #667, opened 2024-01-16
Remove 'sleep infinity' hack for NMA sidecar support

spilchen:
When we did the first drop of the NMA sidecar support (#654), we had to added a 'sleep infinity' to the vertica command because that process would always daemonize itself. With a recent server change that prevents vertica from becoming a daemon, we can now remove the hack.

This also adds health probes to the NMA container by checking the /v1/health endpoint.

A couple of other minor changes went into this PR:
- a new annotation was added (`vertica.com/skip-deployment-check`) to avoid checking if the image was built for admintools or vclusterops deployments. This is useful if using a build that can do either types of deployments.
- set more environment variables in the NMA container that were set in the server. The main driver for this is it allows you to run vsql without having to specify the username.

### DIFF
--- a/.github/workflows/e2e-udx.yml
+++ b/.github/workflows/e2e-udx.yml
@@ -71,6 +71,7 @@ jobs:
         export VLOGGER_IMG=${{ inputs.vlogger-image }}
         export E2E_TEST_DIRS=tests/e2e-udx
         export VERTICA_DEPLOYMENT_METHOD=${{ inputs.vertica-deployment-method }}
+        export NMA_RUNNING_MODE=sidecar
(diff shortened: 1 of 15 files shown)


roypaulin (commented):
Looks good! I wonder if you should also take a look at the restart logic. I made a small change there on my PR. Do we still need that?

spilchen:
> Looks good! I wonder if you should also take a look at the restart logic. I made a small change there on my PR. Do we still need that?

yes, the restart logic had to be updated. Thanks for pointing that out. The restart test now runs clean.

roypaulin (approved):

PR #673, opened 2024-01-25
Add custom print columns for VerticaRestorePointsQuery

chinhtranvan:
This PR adds custom print columns for the new VerticaRestorePointsQuery CR, so that a kubectl get vrpq command tells you if the query has completed.

### DIFF
--- a/pkg/controllers/vrpq/query_reconciler.go
+++ b/pkg/controllers/vrpq/query_reconciler.go
@@ -1,0 +1,144 @@
+/*
+ (c) Copyright [2021-2023] Open Text.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ You may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package vrpq
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	v1 "github.com/vertica/vertica-kubernetes/api/v1"
+	vapi "github.com/vertica/vertica-kubernetes/api/v1beta1"
+	"github.com/vertica/vertica-kubernetes/pkg/controllers"
+	verrors "github.com/vertica/vertica-kubernetes/pkg/errors"
+	restorepointsquery "github.com/vertica/vertica-kubernetes/pkg/vadmin/opts/restorepointsquery"
+	config "github.com/vertica/vertica-kubernetes/pkg/vdbconfig"
+	vrpqstatus "github.com/vertica/vertica-kubernetes/pkg/vrpqstatus"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type QueryReconciler struct {
+	VRec           *VerticaRestorePointsQueryReconciler
+	Vrpq           *vapi.VerticaRestorePointsQuery
+	Log            logr.Logger
+	InitiatorPod   types.NamespacedName // The pod that we run admin commands from
+	InitiatorPodIP string               // The IP of the initiating pod
+	config.ConfigParamsGenerator
+}
+
+func MakeRestorePointsQueryReconciler(r *VerticaRestorePointsQueryReconciler, vrpq *vapi.VerticaRestorePointsQuery,
+	log logr.Logger) controllers.ReconcileActor {
+	return &QueryReconciler{
+		VRec: r,
+		Vrpq: vrpq,
+		Log:  log.WithName("QueryReconciler"),
+		ConfigParamsGenerator: config.ConfigParamsGenerator{
+			VRec: r,
+			Log:  log.WithName("QueryReconciler"),
+		},
+	}
+}
+
+func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctrl.Result, error) {
+	// no-op if QueryComplete is true
+	isSet := q.Vrpq.IsStatusConditionTrue(vapi.Querying)
+	if isSet {
+		return ctrl.Result{}, nil
+	}
+	// collect information from a VerticaDB.
+	if res, err := q.collectInfoFromVdb(ctx); verrors.IsReconcileAborted(res, err) {
+		return res, err
+	}
+
+	return ctrl.Result{}, q.runListRestorePoints(ctx, req)
+}
+
+// fetch the VerticaDB and collect access information to the communal storage for the VerticaRestorePointsQuery CR,
+func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result, error) {
+	vdb := &v1.VerticaDB{}
+	res := ctrl.Result{}
+
+	opts := []restorepointsquery.Option{
+		restorepointsquery.WithInitiator(q.InitiatorPod, q.InitiatorPodIP),
+	}
+
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		var e error
+		if res, e = fetchVDB(ctx, q.VRec, q.Vrpq, vdb); verrors.IsReconcileAborted(res, e) {
+			return e
+		}
+		q.Vdb = vdb
+		// build communal storage params if there is not one
+		if q.ConfigurationParams == nil {
+			res, e = q.ConstructConfigParms(ctx)
+			if verrors.IsReconcileAborted(res, e) {
+				return e
+			}
+		}
+		// extract out the communal and config information to pass down to the vclusterops API.
+		opts = append(opts,
+			restorepointsquery.WithCommunalPath(vdb.GetCommunalPath()),
+			restorepointsquery.WithConfigurationParams(q.ConfigurationParams.GetMap()),
+		)
+
+		return nil
+	})
+
+	return res, err
+}
+
+// setListRestorePointsQueryConditions will update the status condition before and after calling
+// list restore points api
+// Temporarily, runListRestorePoints will not call the ListRestorePoints API
+// since the dispatcher is not set up yet
+func (q *QueryReconciler) runListRestorePoints(ctx context.Context, _ *ctrl.Request) error {
+	// update query message prior to calling vclusterops API
+	err := vrpqstatus.UpdateMessageStatus(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+		"Query is still in progress")
+	if err != nil {
+		return err
+	}
+
+	// set Querying status condition prior to calling vclusterops API
+	err = vrpqstatus.UpdateCondition(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+		v1.MakeCondition(vapi.Querying, metav1.ConditionTrue, ""))
+	if err != nil {
+		return err
+	}
+
+	// API should be called to proceed here
+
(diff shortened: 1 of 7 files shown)


chinhtranvan:
After trying `kubectl get vrpq` with this change, I obtained this information
```
NAME   AGE   MESSAGE
vc     10s   Query has completed successfully
```

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -108,8 +108,15 @@ func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result,
 // Temporarily, runListRestorePoints will not call the ListRestorePoints API
 // since the dispatcher is not set up yet
 func (q *QueryReconciler) runListRestorePoints(ctx context.Context, _ *ctrl.Request) error {
```
spilchen, line 110: Can we update the query_reconciler_test.go to validate these new codepaths?

spilchen (commented):
Can we update the e2e testcase to check the new state?

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -108,8 +108,15 @@ func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result,
...
+	// update query message prior to calling vclusterops API
+	err := vrpqstatus.UpdateMessageStatus(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+		"Query is still in progress")
```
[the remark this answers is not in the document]
chinhtranvan: I agree with you that we should add a defer statement to always reset the status. However, I'm uncertain about the message content. Should we name it 'reset query'? My concern is that if the defer call sets the state for both success and failure, when we execute 'kubectl get vrpq', it will display:
```                                                                                                                                           
"NAME   AGE     MESSAGE
  vc     5m21s   reset query
``
We won't be able to distinguish whether it was successful or failed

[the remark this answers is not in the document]
Comment (jizhuoyu) [reply pkg/controllers/vrpq/query_reconciler.go:113]: We could always set the `Querying` condition to `False` (if it is not already) when we exit?
As for the `Message/State` I think we could set its value depending on its current value? If its current value is "Query is still in progress" it means the query hasn't been successfully completed yet, and we could say something like "Query failed". If its current value is "Query has completed successfully" we have no need to update it.

File: pkg/controllers/vrpq/query_reconciler.go [hunk 3]
```diff
@@ -108,8 +108,15 @@ func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result,
 // Temporarily, runListRestorePoints will not call the ListRestorePoints API
```
jizhuoyu, line 108: I'm wondering what is `setListRestorePointsQueryConditions` in the above comment

pkg/vrpqstatus/status_test.go
```diff
@@ -188,4 +188,22 @@ var _ = Describe("status", func() {
 		Expect(vrpq.Status.Conditions[0].LastTransitionTime).ShouldNot(Equal(origTime))
 	})
 
+	It("should update the message status", func() {
```
jizhuoyu, line 191: If we add a message for a failed query, e.g. "Query failed", we could also add a test here

jizhuoyu (commented):
Left some thoughts and suggestions.

pkg/controllers/vrpq/query_reconciler.go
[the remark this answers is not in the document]
roypaulin: Isn't `defer` more suitable for cleanup operations rather than modifying values? It can make the code more complex. Although you can have `defer` call a function pass a parameter(maybe an error) and do what you have to do, but we would need to change the returned value in the defer call. Any reason why we cannot set success/failure whithin the function body based on the API call result?

pkg/vrpqstatus/status_test.go
chinhtranvan: No need to update here; statust_test.go only tests whether the function updates the message or not. Regardless of the information contained in the message, it will always be updated. We can update in the query_reconciler_test.go

pkg/controllers/vrpq/query_reconciler.go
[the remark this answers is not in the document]
chinhtranvan: Should we include "Query failed" at this PR? I wonder it depends on calling API, and we have a plan to call vcluster API next jira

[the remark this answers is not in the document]
chinhtranvan: @jizhuoyu I wonder what current value are you mentioning now? `Query is still in progress` and `Query has completed successfully` are messages we update in message field. As @spilchen mentioned, there are 3 messages we can update  `Querying`, `Queried successful` and `Query failed`. 

@roypaulin @spilchen since we are currently discussing whether to use a deferred function or not, I will wait until we reach an agreement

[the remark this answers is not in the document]
jizhuoyu: @chinhtranvan I mean you could always first get the current value of message/state field via the k8s client, right? And you could then update the field accordingly.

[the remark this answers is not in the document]
jizhuoyu: @chinhtranvan This only makes sense if we want to use a single `defer` statement to handle all error paths within the function body, though.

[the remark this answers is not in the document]
spilchen: I see the concern with using the defer. I think I would feel better if there were less error code paths in `runListRestorePoints`. Could we combine the updates to the status condition and the state in the same call? Then it really cuts down on the error flows. We kind of already set the Querying status condition in lock step with the state. So, it would make sense to update them at the same time.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -103,13 +103,20 @@ func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result,
...
+	if err != nil {
+		return err
+	}
+
 	// set Querying status condition prior to calling vclusterops API
-	err := vrpqstatus.UpdateCondition(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+	err = vrpqstatus.UpdateCondition(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
 		v1.MakeCondition(vapi.Querying, metav1.ConditionTrue, ""))
 	if err != nil {
 		return err
```
spilchen, line 122: Since we have no error codepath yet, can we add a comment (and jira) to line 125 (`// API should be called to proceed here`) so we don't forget.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -124,6 +131,13 @@ func (q *QueryReconciler) runListRestorePoints(ctx context.Context, _ *ctrl.Requ
...
 	}
 
+	// update query message after calling vclusterops API successfully
+	err = vrpqstatus.UpdateMessageStatus(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+		"Queried successful")
```
spilchen, line 136: I know I had suggested "Queried successful" :smile: , but seeing its use I wonder if "Query successful" would be better?

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -103,28 +108,29 @@ func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result,
...
 	}
 
 	// API should be called to proceed here
+	// If we receive a failure result from the API, a state message and condition need to be updated
 
 	// clear Querying status condition
-	err = vrpqstatus.UpdateCondition(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
-		v1.MakeCondition(vapi.Querying, metav1.ConditionFalse, ""))
+	err = vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+		v1.MakeCondition(vapi.Querying, metav1.ConditionFalse, "CompletedOrFailed"), querying)
```
chinhtranvan, line 128: I decided to use the message 'CompletedOrFailed' when Go complained that 'Completed|Failed' does not match the regular expression pattern ^A-Za-z?$.`

chinhtranvan:
For e2e test case verify server logrotate,  it appears to include 'editor.log' for rotation. Prior to that, it only performs log rotation for 'vertica.log'
``
  node_name     | success | max_size  | max_age |                               log_file                                | need_rotation 
-------------------+---------+-----------+---------+-----------------------------------------------------------------------+---------------
 v_vertdb_node0001 | t       | 104857600 | 7       | /data/vertdb/v_vertdb_node0001_catalog/vertica.log                    | f
 v_vertdb_node0001 | t       | 104857600 | 7       | /data/vertdb/v_vertdb_node0001_catalog/UDxLogs/UDxFencedProcesses.log | f
 v_vertdb_node0001 | t       | 104857600 | 7       | /data/vertdb/v_vertdb_node0001_catalog/editor.log                     | f
 v_vertdb_node0001 | t       |      1024 | 7       | /data/vertdb/v_vertdb_node0001_catalog/vertica.log                    | t
 v_vertdb_node0001 | t       |      1024 | 7       | /data/vertdb/v_vertdb_node0001_catalog/UDxLogs/UDxFencedProcesses.log | f
 v_vertdb_node0001 | t       |      1024 | 7       | /data/vertdb/v_vertdb_node0001_catalog/editor.log                     | t
(6 rows)
```
I test with my local machine , it run successfuly and don't include editor.log. I add the new change to fix the e2e test in this PR

File: tests/e2e-leg-6/verify-server-logrotate/expected-output/logrotateTimerServiceOut.txt
Comment (spilchen) [inline tests/e2e-leg-6/verify-server-logrotate/expected-output/logrotateTimerServiceOut.txt:1]: We may need to apply this change to main as I think it affects it there to. Although we need the testcase to continue to run against a 24.1.0 server. This is probably fine for now to get your PR clean.

File: pkg/controllers/vrpq/query_reconciler.go (hunk 7 above)
Comment (spilchen) [reply pkg/controllers/vrpq/query_reconciler.go:128]: Sorry, I wasn't clear. I meant we would pick Completed or Failed, depending on the result of the action. Can you change it to Completed?

File: pkg/controllers/vrpq/query_reconciler.go [hunk 8]
```diff
@@ -33,6 +33,11 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+const (
+	querying     = "Querying"
+	successQuery = "Query successful"
```
spilchen, line 38: Suggest including state to denote that they are used in the state status field.
```suggestion
	stateQuerying     = "Querying"
	stateSuccessQuery = "Query successful"
```

pkg/vrpqstatus/status.go
```diff
@@ -60,14 +60,17 @@ func Update(ctx context.Context, clnt client.Client, log logr.Logger, vrpq *vapi
...
-// UpdateCondition will update a condition status
-// This is a no-op if the status condition is already set.  The input vdb will
+// UpdateConditionAndState will update a condition and state status
+// This is a no-op if the status condition is already set.  The input vrpq will
 // be updated with the status condition.
-func UpdateCondition(ctx context.Context, clnt client.Client, log logr.Logger,
-	vrpq *vapi.VerticaRestorePointsQuery, condition *metav1.Condition) error {
-	// refreshConditionInPlace will update the status condition in vdb.  The update
+func UpdateConditionAndState(ctx context.Context, clnt client.Client, log logr.Logger,
+	vrpq *vapi.VerticaRestorePointsQuery, condition *metav1.Condition, msg string) error {
```
spilchen, line 67: message was the old name of the field. Lets update that.
```suggestion
	vrpq *vapi.VerticaRestorePointsQuery, condition *metav1.Condition, state string) error {
```

tests/e2e-leg-6/verify-server-logrotate/expected-output/logrotateTimerServiceOut.txt
chinhtranvan: Then, I need to open a PR in the main branch and apply this change?

PR #674, opened 2024-01-25
Assign resources to the NMA sidecar container

spilchen:
The NMA sidecar container now has default resources assigned to it in the pod spec. Some default values have been selected: 1CPU and 250MB of memory. These are for requests only. I intentionally left off the limit because of the burstable nature of the NMA container.

The resources are only set if the subcluster has defined resources. This ensures that the container can be run in low-resource environments (for testing purposes).

New annotations have been added to facilitate resource tuning. The new annotations are as follows:
vertica.com/nma-resources-limits-memory
vertica.com/nma-resources-limits-cpu
vertica.com/nma-resources-requests-memory
vertica.com/nma-resources-requests-cpu

Lastly, an existing e2e test case was updated. The test case was moved to the e2e-leg-7, as it assumes the use of the NMA sidecar and can only run on 24.2.0.


pkg/meta/annotations.go
```diff
@@ -159,6 +164,20 @@ const (
 	// image is built for that (and vice-versa). This annotation allows you to
 	// skip that check.
 	SkipDeploymentCheckAnnotation = "vertica.com/skip-deployment-check"
+
+	// Set of annotations that you can use to control the resources of the NMA
+	// sidecar. The actual annoation name is:
```
roypaulin, line 169:
```suggestion
	// sidecar. The actual annotation name is:
```

pkg/meta/annotations.go
```diff
@@ -266,6 +285,42 @@ func GetSkipDeploymentCheck(annotations map[string]string) bool {
...
 }
 
+// GetNMASidecarResource is used to retrieve a specific resource for the NMA
+// sidecar. If any parsing error occurs, the default value is returned.
+func GetNMASidecarResource(annotations map[string]string, resourceName corev1.ResourceName) resource.Quantity {
+	annotationName := GenNMASidecarResourceAnnotationName(resourceName)
+	defVal := DefaultSidecarResource[resourceName]
+	quantityStr := lookupStringAnnotation(annotations, annotationName, defVal.String())
+	if quantityStr == "" {
+		return resource.Quantity{}
```
roypaulin, line 295: Why and we returning defVal here? Are you treating resource-annotation not set different from resource-annotation=""?

pkg/meta/resources.go
```diff
@@ -26,7 +26,7 @@ var (
 	DefaultSidecarResource = map[corev1.ResourceName]resource.Quantity{
 		corev1.ResourceLimitsCPU:      resource.MustParse("1"),
```
roypaulin, line 27: Is it 1000 millicores?

spilchen: Yes, this is 1000 millicore. I believe just specifying 1 is equivalent to 1000m.

pkg/meta/annotations.go
spilchen: Yes, if someone set the annotation but leaves it blank, we aren't going to use the default. I'll add a comment.

api/v1/verticadb_webhook.go
```diff
@@ -895,6 +896,20 @@ func (v *VerticaDB) hasValidPodSecurityContext(allErrs field.ErrorList) field.Er
 	return allErrs
 }
 
+func (v *VerticaDB) hasValidNMAResourceLimit(allErrs field.ErrorList) field.ErrorList {
+	nmaMemoryLimit := vmeta.GetNMAResource(v.Annotations, v1.ResourceLimitsMemory)
+	if nmaMemoryLimit.IsZero() { // Zero implies it isn't set.
+		return allErrs
+	}
+	if vmeta.MinNMAMemoryLimit.Cmp(nmaMemoryLimit) == 1 {
```
roypaulin, line 904: Isn't there a min limit for cpu too?

spilchen: No. The memory is the important min to enforce because we cannot run things like catalog editor without at least 1GB of memory.

PR #676, opened 2024-01-26
fix verify-server-logrotate test

chinhtranvan:
This PR is just a quick fix for e2e verify-server-logrotate in leg 6, as after 24.1.0 server, we also rotate the editor.log


spilchen (approved):
Thanks for fixing this!

spilchen:
Once in main, you should be able to merge it into vnext. I just updated vnext so you should only have to pull this commit in. No need for a PR when doing a simple merge.

chinhtranvan:
> Once in main, you should be able to merge it into vnext. I just updated vnext so you should only have to pull this commit in. No need for a PR when doing a simple merge.

Sure, after this PR is merged to main branch, I will merge main into vnext with this commit

PR #683, opened 2024-01-29
Allow NMA health probes to be overridden via annotations

spilchen:
This adds support for a bunch of annotations to control various settings in the NMA health probes. The format is:
```
vertica.com/nma-<probe-name>-probe-<field-name>
```

Where `<probe-name>` is one of: 
- readiness
- startup
- liveness

The `<field-name>` is one of:
- success-threshold
- failure-threshold
- period-seconds
- timeout-seconds
- initial-delay-seconds


pkg/meta/annotations.go
```diff
@@ -339,6 +356,27 @@ func GenNMAResourcesAnnotationName(resourceName corev1.ResourceName) string {
...
 
+// GenNMAHealthProbeAnnotationName returns the name of the annotation for a specific health probe field.
+func GenNMAHealthProbeAnnotationName(probeName, field string) string {
+	return fmt.Sprintf("vertica.com/nma-%s-probe-%s", probeName, field)
+}
+
+// GetNMAHealthProbeOverride returns the value of a NMA health probe annotation.
+// If the annotation isn't set, or its value doesn't convert to an int, then
+// (0,false) is returned.
+func GetNMAHealthProbeOverride(annotations map[string]string, probeName, field string) (int32, bool) {
```
roypaulin, line 367: So the user can set any numerical value? There is no threshold for any of these fields?

spilchen: We can probably reject negative values. But there is no upper bound on the values. I'll add a check for negativity.

roypaulin (approved):

PR #685, opened 2024-01-30
Check cluster quorum in restart_reconciler

roypaulin:
Before this, to choose between restart_node and restart_cluster, we simply checked the number of non-readonly up nodes. In many occasions, in vclusterops mode, it led to  a failed restart_node call before restart_cluster. Now we make sure to still have cluster quorum before choosing restart_node path.


pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -128,6 +128,16 @@ func (r *RestartReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctr
 		r.Vdb.Spec.InitPolicy != vapi.CommunalInitPolicyScheduleOnly {
 		return r.reconcileCluster(ctx)
 	}
+	// If at least one non-readonly is up when cluster does not have quorum,
+	// we requeue to give that node(s) time to go down before restarting the entire
+	// cluster
+	if !r.PFacts.doesDBHaveQuorum() {
```
spilchen, line 134: One issue with this is that it doesn't account for pods that exist but aren't yet part of the database. Consider you start with a 3 node system, then grow the primaries to like 9 nodes. But before we do the add node, we get into here. We may think 6 of the 9 nodes are down and wrongly assume we have lost cluster quorum. I think we need to push this check much lower. There is a spot where we fetch cluster state (fetchClusterNodeState) to prune out nodes that aren't down yet in spread. Perhaps that check needs to expand to allow us to check for cluster quorum?

spilchen (commented):
I suggest we add a changie entry too. Maybe something like: "Restart node when we have lost cluster quorum may cause the operator to restart"

pkg/controllers/vdb/restart_reconciler.go
roypaulin: Yes I agree.

roypaulin: I added 2 new commits. The older one does what you suggested. On the newer  one I only count a primary if dbExists  to avoid the above situation . Wouldn't that be enough to account for pods that aren't yet part of the database?

Otherwise you can review the older commit and I will keep that as a solution.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -354,6 +354,27 @@ func (r *RestartReconciler) removePodsWithClusterUpState(ctx context.Context, po
...
+	totalPrimaryCount := 0
+	upPrimaryCount := 0
+	for _, pod := range r.PFacts.Detail {
+		if !pod.isPrimary {
+			continue
+		}
+		state, ok := clusterState[pod.vnodeName]
+		if ok {
+			totalPrimaryCount++
+			if state == vadmin.StateUp {
```
spilchen: I think we need to use the clusterState to know if a node exists in the catalog, but we should use the podfacts for up/down. We can come to the conclusion a bit quicker than spread does. So, it can be likely to see a pod that is down but the cluster state says is up.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -336,12 +326,22 @@ func (r *RestartReconciler) restartPods(ctx context.Context, pods []*PodFact) (c
...
+// pods that aren't considered down removed. It exists early if there is no
+// cluster quorum
 func (r *RestartReconciler) removePodsWithClusterUpState(ctx context.Context, pods []*PodFact) ([]*PodFact, ctrl.Result, error) {
 	clusterState, res, err := r.fetchClusterNodeStatus(ctx, pods)
 	if verrors.IsReconcileAborted(res, err) {
 		return nil, res, err
 	}
+	// If cluster does not have quorum, we requeue to
+	// restart the entire cluster
+	if !r.doesDBHaveQuorum(clusterState) {
```
spilchen: We are kind of abusing this function. For example, this new check isn't to remove pods with cluster up state. I suggest we do a bit of a refactor. Maybe pass in the clusterState into this function and move the fetchClusterNodeStatus to a calling function.

pkg/vadmin/restart_node_vc.go
```diff
@@ -46,6 +47,11 @@ func (v *VClusterOps) RestartNode(ctx context.Context, opts ...restartnode.Optio
 	vcOpts := v.genStartNodeOptions(&s, certs)
 	err = v.VStartNodes(vcOpts)
 	if err != nil {
+		reIPNoClusterQuorumError := &vops.ReIPNoClusterQuorumError{}
```
spilchen, line 50: I think we should log a message here to say we got the error. This is usually done in `vcErrors.LogFailure`. Can we move this logic into there?

spilchen (commented):
I prefer to the older commit. I'm still a little concerned about relying only on the podfatcts. If we get it wrong, then we don't restart at all. That's bad. Plus, I don't know if we are protected during scale down. We have a pending delete flag, but we still don't know if the host has been removed from the catalog yet. It could be either.

roypaulin:
> I prefer to the older commit. I'm still a little concerned about relying only on the podfatcts. If we get it wrong, then we don't restart at all. That's bad. Plus, I don't know if we are protected during scale down. We have a pending delete flag, but we still don't know if the host has been removed from the catalog yet. It could be either.

Now I understand the risk on relying on podfacts, Thanks! I will address your comments.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -275,11 +275,24 @@ func (r *RestartReconciler) reconcileNodes(ctx context.Context) (ctrl.Result, er
 
 // restartPods restart the down pods using an admin command
 func (r *RestartReconciler) restartPods(ctx context.Context, pods []*PodFact) (ctrl.Result, error) {
-	// Reduce the pod list according to the cluster node state
-	downPods, res, err := r.removePodsWithClusterUpState(ctx, pods)
+	// we want to fetch the state of down pods and primary pods. The letter will
+	// allow us to check cluster quorum and abort restart_node if needed
+	podsToFetch := combinePodsWithoutDuplicates(r.PFacts.findPrimaryPods(), pods)
```
spilchen, line 280: This logic in here is getting complicated. We may be able to simplify this by just getting the cluster node status for all pods. Basically don't do any filtering in FetchNodeState. We are given a map, which we can do fast lookups on, so there isn't a large gain to having the map filtered.

roypaulin: I was just worried about having a map too big. But FetchNodeState() already returns the entire cluster state so there is no benefits. I will remove that.

Now I am wondering why we have a `HostsNeeded` map in fetchnode config.

spilchen: It was probably an optimization that didn't make sense to do, which was probably my doing (:smile:).

spilchen (approved):

PR #686, opened 2024-01-31
 Add empty controller and API for VerticaScrutinize

roypaulin:
This PR build out the stub for an empty controller to handle the VerticaScrutinize API. For now The operator can observe the new API, initiate a reconciliation iteration, and take no action, as we have set nil for the actors during this implementation phase

### DIFF
--- a/config/crd/patches/webhook_in_verticascrutinizes.yaml
+++ b/config/crd/patches/webhook_in_verticascrutinizes.yaml
@@ -1,0 +1,8 @@
+# The following patch enables a conversion webhook for the CRD
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: verticascrutinizes.vertica.com
+spec:
+  conversion:
+    strategy: None
--- a/pkg/controllers/vscr/verticascrutinize_controller.go
+++ b/pkg/controllers/vscr/verticascrutinize_controller.go
@@ -38,9 +38,9 @@ type VerticaScrutinizeReconciler struct {
 	EVRec  record.EventRecorder
 }
 
-//+kubebuilder:rbac:groups=vertica.com,resources=verticascrutinizes,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=vertica.com,resources=verticascrutinizes/status,verbs=get;update;patch
-//+kubebuilder:rbac:groups=vertica.com,resources=verticascrutinizes/finalizers,verbs=update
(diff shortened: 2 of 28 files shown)


pkg/controllers/vscr/verticascrutinize_controller.go
```diff
@@ -0,0 +1,107 @@
...
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the VerticaScrutinize object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
+func (r *VerticaScrutinizeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("vrpq", req.NamespacedName)
```
spilchen, line 55:
```suggestion
	log := r.Log.WithValues("vscr", req.NamespacedName)
```

config/crd/patches/webhook_in_verticascrutinizes.yaml
```diff
@@ -0,0 +1,8 @@
+# The following patch enables a conversion webhook for the CRD
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: verticascrutinizes.vertica.com
```
spilchen, line 5: I don't like the plural form of the CR. Can it be verticascrutinizers instead of verticascrutinizes? I think there are ways to control the plural form when calling `kubebuilder create api`

scripts/config-transformer.sh
```diff
@@ -33,6 +33,7 @@ mv $TEMPLATE_DIR/verticadbs.vertica.com-crd.yaml $CRD_DIR
 mv $TEMPLATE_DIR/verticaautoscalers.vertica.com-crd.yaml $CRD_DIR
 mv $TEMPLATE_DIR/eventtriggers.vertica.com-crd.yaml $CRD_DIR
 mv $TEMPLATE_DIR/verticarestorepointsqueries.vertica.com-crd.yaml $CRD_DIR
+mv $TEMPLATE_DIR/verticascrutinizes.vertica.com-crd.yaml $CRD_DIR
```
spilchen, line 36: I wonder if we can make this more future proof for new CRDs by using a wildcard?
```suggestion
mv $TEMPLATE_DIR/*-crd.yaml $CRD_DIR
```

spilchen (commented):
This is a good start! I just had a few minor comments.

spilchen (approved):

PR #688, opened 2024-02-01
Chinh/implement controller logic for vertica restore points query

chinhtranvan:
This PR is going to call the vclusterops API to perform the restore points query. It will copy the result of the query into the CRs status fields


chinhtranvan:
Even this is a draft PR, I would like to get early comments for this PR

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,18 +88,45 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+
+	passwd, err := q.GetSuperuserPassword(ctx, q.Log)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	prunner := cmds.MakeClusterPodRunner(q.Log, q.VRec.Cfg, q.Vdb.GetVerticaUser(), passwd)
+	pfacts := MakePodFacts(q.VRec, prunner)
+	dispatcher := q.makeDispatcher(q.Log, q.Vdb, prunner, passwd)
+
+	if e := pfacts.Collect(ctx, q.Vdb); e != nil {
```
chinhtranvan, line 105: I got error in the unit tests, "failed to copy and execute the gather script: could not execute: pod vertica-sample-defaultsubcluster-0 does not have a host assigned". It's from this line https://github.com/vertica/vertica-kubernetes/blob/main/pkg/cmds/exec.go#L260. I'm wondering if I've missed any steps for setting up pods after creating the database.

pkg/vadmin/restore_points_query_vc.go
```diff
@@ -17,9 +17,62 @@ package vadmin
...
+	v.Log.Info("Setup restore point options", "hosts", opts.Hosts)
+
+	opts.Ipv6 = vstruct.MakeNullableBool(net.IsIPv6(s.InitiatorIP))
+	opts.ConfigurationParameters = s.ConfigurationParams
+
+	// auth options
+	opts.Key = certs.Key
+	opts.Cert = certs.Cert
+	opts.CaCert = certs.CaCert
+	*opts.UserName = v.VDB.GetVerticaUser()
```
spilchen, line 74: Why do we need the user/password to get then restore points? I thought it just talked to the NMA.

pkg/vadmin/restore_points_query_vc.go
```diff
@@ -17,9 +17,62 @@ package vadmin
...
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-func (v *VClusterOps) ListRestorePoints(_ context.Context) error {
-	v.Log.Info("start restore points query")
-	return nil
+// RestartNode will restart a subset of nodes. Use this when vertica has not
+// lost cluster quorum. The IP given for each vnode may not match the current IP
+// in the vertica catalogs.
+func (v *VClusterOps) ListRestorePoints(ctx context.Context, opts ...restorepoints.Option) (ctrl.Result, error) {
```
spilchen, line 32: We use ListRestorePoints when we are going to call ShowRestorePoints. I think we need some consistency here. Wherever we have `list`, we should use `show` instead.

roypaulin:
I have too many observations so let's talk after the standup.

pkg/controllers/vrpq/podfacts.go
spilchen, line 1: Duplicating podfacts is a little too heavy for this controller. I believe, we just need to collect this so that we can find a pod to execute the vclusterops API. I think it would be a lot simpler if we just used the iter package. There is a function that returns statefulsets (`FindStatefulSets`). You can then use that to read each pod, stopping when you find one that has the NMA container in ready state.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,18 +90,38 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
 		return res, err
 	}
 
-	return ctrl.Result{}, q.runListRestorePoints(ctx, req)
+	dispatcher := q.makeDispatcher(q.Log, q.Vdb, q.PRunner, q.Password)
```
spilchen, line 93: I don't think we need the password. Lets see if we can get by not passing that in. Perhaps make it a pointer so we can easily tell when its not available (and return an error when it should).

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,18 +90,38 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+	// Create a TLS secret for the NMA service
+	err := q.generateCertsForNMA(ctx)
```
spilchen, line 95: I don't understand why we are generating a cert here. We need to reuse the one the NMA already has present.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,18 +90,38 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+		q.Log.Info("No pods running")
+		return ctrl.Result{}, nil
+	}
+	hostList := q.PFacts.findExpectedNodeIps()
+
+	opts := []restorepointsquery.Option{}
+	// extract out the communal and config information to pass down to the vclusterops API.
+	opts = append(opts,
+		restorepointsquery.WithInitiator(pf.name, pf.podIP),
+		restorepointsquery.WithHosts(hostList),
```
spilchen, line 114: Why do we need a split between initiator and host? Can we just provide one  that acts as the initiator and host list. We only need to talk to one host to get this information.

pkg/vrpqstatus/status.go
```diff
@@ -65,6 +65,14 @@ func Update(ctx context.Context, clnt client.Client, log logr.Logger, vrpq *vapi
 // be updated with the status condition.
 func UpdateConditionAndState(ctx context.Context, clnt client.Client, log logr.Logger,
 	vrpq *vapi.VerticaRestorePointsQuery, condition *metav1.Condition, state string) error {
+	// Check if the condition is already set
```
spilchen, line 68: I am confused by this change. If I recall, this was done because we would endlessly loop by updating the status condition continuously. One thing I don't understand is that once the query is done, we should skip the reconcile. Wouldn't that avoid updating the status condition?

Also, this is a large PR. I would suggest pulling this out and handling it separately.

PR #689, opened 2024-02-01
Remove s6-overlay and NMA sidecar annotation

spilchen:
This removes from the v2 server container the s6-overlay init process. Now that we have full support for running the NMA in a sidecar, this systemd-style process isn't needed. One advantage of this change is it allows us run in OpenShift with the restricted SCC. This makes it easier to deploy Vertica in OpenShift because this is the default SCC. This will be available for consumption in the v24.2.0 server to be released in April 2024.

Also removed in this change is the `vertica.com/run-nma-in-sidecar` annotation. This previously was needed to correctly setup the pod spec. But the operator can automatically determine this. It defaults to using the NMA sidecar for all vclusterOps deployments. If it finds itself on 24.1.0, then it will adjust the statefulset so that the NMA and server are run in a single monolithic container.

This also will address different use cases as it relates to NMA deployment and the Vertica version.
- If vclusterOps is requested on an image prior to 24.1.0, then the container will not start. Both the NMA container, which starts node_management_agent directly, and the server container, which starts vertica with the `--startup-conf` setting, will fail immediately. We cannot run anything in these pods to even get the version. We added a new reconciler in the vdb controller that will check if the pod is in a crash loop backoff. And if so it will log an event message under the assumption its because we are trying to run vclusterOps on an old image. 
- If vclusterOps is used with a 24.1.0 image, the server container won't start because it is missing the `--startup-conf` setting that is new in 24.2.0. The pod facts collection will now use the NMA container if present for these cases.


pkg/events/event.go
```diff
@@ -83,7 +83,7 @@ const (
 	MgmtFailedDiskFull              = "MgmtFailedDiskfull"
 	LowLocalDataAvailSpace          = "LowLocalDataAvailSpace"
 	WrongImage                      = "WrongImage"
-	NMAInSidecarNotSupported        = "NMAInSidecarNotSupported"
+	NMADeploymentIncompatibilty     = "NMADeploymentIncompatibility"
```
roypaulin, line 86: Is it used anywhere?

roypaulin (commented):
Looks good!

roypaulin:
Why did you move some tests?

spilchen:
> Why did you move some tests?

When you start a 24.1.0 server, we initially deploy with the assumption that there is a NMA sidecar. It isn't until the pods are running that we can check the version. So, we then tear down the statefulset and rebuild it without a sidecar. There was one test (mount-certs) that was sensitive to this. It would run a kubectl exec in the server container, which may or may not be running. I could have added more checks into that e2e test I think. But at the time, I opted to move it to leg 7, which runs on 24.2.0 only. Leg 3 was getting light on tests, so I then moved another test into it.

pkg/controllers/vdb/podfacts.go
```diff
@@ -161,6 +161,12 @@ type PodFact struct {
 
 	// The in-container path to the catalog. e.g. /catalog/vertdb/v_node0001_catalog
 	catalogPath string
+
+	// true if the pod's spec includes a sidecar to run the NMA
+	hasNMASidecar bool
```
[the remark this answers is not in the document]
spilchen: I'd prefer to keep the original name since we should only camel case on word boundaries (i.e. the word is "sidecar" not "side car")

pkg/controllers/vdb/crashloop_reconciler.go
```diff
@@ -0,0 +1,119 @@
...
+					"Image cannot be used for vclusterOps deployments. Change the deployment by changing the %s annotation",
+					vmeta.VClusterOpsAnnotation)
+				// Don't bother checking anymore pods as this setting is global for the CR.
+				return
+			}
+		}
+	}
+}
+
+func (c *CrashLoopReconciler) findNMAContainerStatus(pod *corev1.Pod) *corev1.ContainerStatus {
```
roypaulin, line 112: I see around some functions related to k8s, so I wonder if at some point a separate package will be needed, especially to avoid code duplication.

pkg/events/event.go
spilchen: No. I will remove it.

spilchen:
I hit an issue upgrading from 23.4 to 24.2. So, there are few more changes to upgrade to handle that. I also added a new e2e leg (8) that does this upgrade path to verify this change.

pkg/controllers/vdb/crashloop_reconciler.go
spilchen: Yes, we will have to keep an eye on it. I already moved this function out to the bulder package (even before I saw this) to use the same logic somewhere else.

roypaulin (approved):

PR #691, opened 2024-02-01
Configure createdb timeout by annotation vertica.com/createdb-timeout

jizhuoyu:
This adds an annotation to be able to control the timeout we wait during create db for the nodes to come up. The annotation value is: `vertica.com/createdb-timeout`.

### DIFF
--- a/api/v1/helpers.go
+++ b/api/v1/helpers.go
@@ -396,6 +396,11 @@ func (v *VerticaDB) GetRestartTimeout() int {
 	return vmeta.GetRestartTimeout(v.Annotations)
 }
 
+// GetCreateDBNodeStartTimeout returns the timeout value for createdb node startup
+func (v *VerticaDB) GetCreateDBNodeStartTimeout() int {
+	return vmeta.GetCreateDBNodeStartTimeout(v.Annotations)
+}
+
 // IsNMASideCarDeploymentEnabled returns true if the conditions to run NMA
 // in a sidecar are met
 func (v *VerticaDB) IsNMASideCarDeploymentEnabled() bool {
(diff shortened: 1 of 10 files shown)


spilchen (approved):
Looks good

api/v1/helpers.go
```diff
@@ -396,6 +396,11 @@ func (v *VerticaDB) GetRestartTimeout() int {
 	return vmeta.GetRestartTimeout(v.Annotations)
 }
 
+// GetCreateDBNodeStartTimeout returns the timeout value for createdb node startup
+func (v *VerticaDB) GetCreateDBNodeStartTimeout() int {
```
roypaulin, line 400: What about negative values?

spilchen: I think a negative value would be okay. It should cause vclusterops to not wait for the polling. Although I have some doubts that vclusterops actually honours the timeout specified. I added some comments in the jira so that we investigate that part. But when/if it does use the timeout it will treat any negative number like its no timeout. There isn't any other way to turn off the polling.

roypaulin (approved):

spilchen:
Lets see if there is a vclusterOps change before merging this.

api/v1/helpers.go
jizhuoyu: As discussed, we'll add a webhook check to reject negative values.

api/v1/verticadb_webhook.go
```diff
@@ -910,6 +911,17 @@ func (v *VerticaDB) hasValidNMAResourceLimit(allErrs field.ErrorList) field.Erro
 	return allErrs
 }
 
+func (v *VerticaDB) hasValidCreateDBTimeout(allErrs field.ErrorList) field.ErrorList {
```
roypaulin, line 914: Looks good!

spilchen (commented):
I'm good with the latest changes. We just need to pull in the latest vclusterOps once that change has been merged in.

spilchen:
The vcluster change with your timeout fix has been sync'd. You can now pull in the latest vclusterops package here to get that change.

jizhuoyu:
> The vcluster change with your timeout fix has been sync'd. You can now pull in the latest vclusterops package here to get that change.

Done.

PR #693, opened 2024-02-02
Implement controller logic for VerticaRestorePointsQuery

chinhtranvan:
As we discussed, this task is divided into two parts. This PR completed the first part, which involved calling the vclusterops API to perform the restore points query. A follow-up task to copy the result of the query into the CRs status fields will be processed later


pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -62,7 +73,7 @@ func MakeRestorePointsQueryReconciler(r *VerticaRestorePointsQueryReconciler, vr
 
 func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctrl.Result, error) {
 	// no-op if QueryComplete is true
-	isSet := q.Vrpq.IsStatusConditionTrue(vapi.Querying)
+	isSet := q.Vrpq.IsStatusConditionTrue(vapi.QueryComplete)
```
chinhtranvan, line 76: @spilchen This is the culprit that causes us to endlessly loop into verticadb.log. `Vapi.Querying` is updated to false to clear the querying status condition. It will no longer be true. I just updated to `vapi.QueryComplete`, and I don't see the endlessly looped behavior anymore

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,18 +82,45 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
 // fetch the VerticaDB and collect access information to the communal storage for the VerticaRestorePointsQuery CR,
 func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result, error) {
 	vdb := &v1.VerticaDB{}
 	res := ctrl.Result{}
-
-	opts := []restorepointsquery.Option{
-		restorepointsquery.WithInitiator(q.InitiatorPod, q.InitiatorPodIP),
-	}
-
 	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
spilchen, line 124: I may have missed this before, but we only need the retry for updates. We are just fetching vdb state. So, I think we need the `RetryOnConflict`.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,18 +82,45 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+		podIP = pod.Status.PodIP
+		break
+	}
+
+	// setup dispatcher for vclusterops API
+	passwd, err := q.GetSuperuserPassword(ctx, q.Log, q.Vdb)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	prunner := cmds.MakeClusterPodRunner(q.Log, q.VRec.Cfg, q.Vdb.GetVerticaUser(), passwd)
```
spilchen, line 107: We don't need a pod runner since the dispatcher is going to be vclusterOps only.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -134,3 +179,37 @@ func (q *QueryReconciler) runListRestorePoints(ctx context.Context, _ *ctrl.Requ
...
 		v1.MakeCondition(vapi.QueryComplete, metav1.ConditionTrue, "Completed"), stateSuccessQuery)
 }
+
+// makeDispatcher will create a Dispatcher object based on the feature flags set.
+func (q *QueryReconciler) makeDispatcher(log logr.Logger, vdb *v1.VerticaDB, prunner cmds.PodRunner,
+	passwd string) vadmin.Dispatcher {
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		return vadmin.MakeVClusterOps(log, vdb, q.VRec.GetClient(), passwd, q.VRec, vadmin.SetupVClusterOps)
+	}
+	return vadmin.MakeAdmintools(log, vdb, prunner, q.VRec, q.OpCfg.DevMode)
```
spilchen, line 189: We don't need the admintools case. Lets just return an error if UseVClusterOps is not true. This will allow us to not require the pod runner to be passed in.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -134,3 +179,37 @@ func (q *QueryReconciler) runListRestorePoints(ctx context.Context, _ *ctrl.Requ
 	return vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
 		v1.MakeCondition(vapi.QueryComplete, metav1.ConditionTrue, "Completed"), stateSuccessQuery)
 }
+
+// makeDispatcher will create a Dispatcher object based on the feature flags set.
+func (q *QueryReconciler) makeDispatcher(log logr.Logger, vdb *v1.VerticaDB, prunner cmds.PodRunner,
+	passwd string) vadmin.Dispatcher {
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		return vadmin.MakeVClusterOps(log, vdb, q.VRec.GetClient(), passwd, q.VRec, vadmin.SetupVClusterOps)
```
spilchen, line 187: The password isn't needed since we our API is going to talk strictly to the NMA. We should make the password field a pointer so that we will not that its not used rather than a dbadmin account without a password being set.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -134,3 +179,37 @@ func (q *QueryReconciler) runListRestorePoints(ctx context.Context, _ *ctrl.Requ
...
+	}
+	return vadmin.MakeAdmintools(log, vdb, prunner, q.VRec, q.OpCfg.DevMode)
+}
+
+// GetSuperuserPassword returns the superuser password if it has been provided
+func (q *QueryReconciler) GetSuperuserPassword(ctx context.Context, log logr.Logger,
```
spilchen, line 193: I think we can remove this because we aren't going to need the dbadmin password.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -96,32 +134,39 @@ func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result,
...
 	return res, err
 }
 
-// runListRestorePoints will update the status condition before and after calling
-// list restore points api
-// Temporarily, runListRestorePoints will not call the ListRestorePoints API
-// since the dispatcher is not set up yet
-func (q *QueryReconciler) runListRestorePoints(ctx context.Context, _ *ctrl.Request) error {
+// runShowRestorePoints will update the status condition and state before and after calling
+// showrestorepoints vclusterops api
```
spilchen, line 144: This could be updated.
```suggestion
// runShowRestorePoints call the vclusterOps API to get the restore points.
```

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -96,32 +134,39 @@ func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result,
...
+			err = vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+				v1.MakeCondition(vapi.Querying, metav1.ConditionFalse, "Failed"), stateFailedQuery)
+			if err != nil {
+				return err
+			}
+			return errRun
+		}
+	}
+	q.VRec.Eventf(q.Vrpq, corev1.EventTypeNormal, events.ShowRestorePointsSucceeded,
+		"Successfully showed restore point with database '%s'. It took %s", q.Vdb.Spec.DBName, time.Since(start))
```
spilchen, line 169:
```suggestion
		"Successfully queried restore points in %s", time.Since(start).Truncate(time.Second))
```

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -96,32 +134,39 @@ func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result,
...
+	// call showRestorePoints vcluster API
+	q.VRec.Eventf(q.Vrpq, corev1.EventTypeNormal, events.ShowRestorePointsStarted,
+		"Starting show restore points")
+	start := time.Now()
+	if res, errRun := dispatcher.ShowRestorePoints(ctx, opts...); verrors.IsReconcileAborted(res, errRun) {
+		if errRun != nil {
+			err = vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+				v1.MakeCondition(vapi.Querying, metav1.ConditionFalse, "Failed"), stateFailedQuery)
+			if err != nil {
+				return err
```
spilchen, line 163: Lets join the two errors rather than losing one.
```suggestion
				errRun = errors.Join(errRun, err)
```

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -96,32 +134,39 @@ func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result,
...
+			}
+			return errRun
+		}
+	}
+	q.VRec.Eventf(q.Vrpq, corev1.EventTypeNormal, events.ShowRestorePointsSucceeded,
+		"Successfully showed restore point with database '%s'. It took %s", q.Vdb.Spec.DBName, time.Since(start))
 
 	// clear Querying status condition
 	err = vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
```
spilchen, line 172: Could we merge the two UpdateConditionAndState calls into one? We may need to allow multiple status conditions to be set.

pkg/controllers/vrpq/query_reconciler_test.go
```diff
@@ -41,12 +44,46 @@ var _ = Describe("query_reconcile", func() {
...
+		vdb.Spec.NMATLSSecret = secretName
+
 		createS3CredSecret(ctx, vdb)
 		defer deleteCommunalCredSecret(ctx, vdb)
 		test.CreateVDB(ctx, k8sClient, vdb)
 		defer test.DeleteVDB(ctx, k8sClient, vdb)
+		test.CreatePods(ctx, k8sClient, vdb, test.AllPodsRunning)
+		defer test.DeletePods(ctx, k8sClient, vdb)
+
+		caCert, err := security.NewSelfSignedCACertificate(2048)
```
spilchen, line 81: Why are we creating a cert for the unit test? If you need the NMA TLS secret, you can use `CreateFakeTLSSecret` in the test package.

pkg/test/helpers.go
```diff
@@ -197,6 +198,21 @@ func BuildTLSSecret(vdb *vapi.VerticaDB, name, key, cert, rootca string) *corev1
 	return secret
 }
 
+func CreateFakePemTLSSecret(ctx context.Context, vdb *vapi.VerticaDB, c client.Client, name string, caCert, cert security.Certificate) {
```
spilchen, line 201: Lets use `CreateFakeTLSSecret` instead.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,18 +82,45 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+	finder := iter.MakeSubclusterFinder(q.VRec.Client, q.Vdb)
+	pods, err := finder.FindPods(ctx, iter.FindExisting)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// find a pod to execute the vclusterops API
+	var podIP string
+	for i := range pods.Items {
+		if i != names.GetNMAContainerIndex() {
```
spilchen, line 94: `i` and the NMA container index are not comparable. They track different things.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,18 +82,45 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+			continue
+		}
+		pod := &pods.Items[i]
+		podIP = pod.Status.PodIP
```
spilchen, line 98: Finding a pod doesn't mean it is running. There is other state we should be looking at. Is it running, is the NMA container ready. Those are two things we need to check before using this pod as the host to communicate with.

pkg/test/helpers.go
chinhtranvan: It complains `failed to show restore points: fail to show restore points: execute NMAHealthOp failed, details: fail to load HTTPS certificates, details tls: failed to find any PEM data in certificate input` if I use CreateFakeTLSSecret "https://github.com/vertica/vertica-kubernetes/actions/runs/7752956063/job/21143979568"

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +79,108 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+		if pod.Status.Phase == podRunning {
+			for j := range pod.Status.ContainerStatuses {
+				// check NMA container ready
+				if pod.Status.ContainerStatuses[j].Ready {
+					podIP = pod.Status.PodIP
+					break
+				}
+			}
+		}
+		if podIP != "" {
```
roypaulin, line 101: I don't think this is needed, There is already a `break` at line 96. You probably want to check at the end of the loop if a pod satisfying the above conditions was found.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +79,108 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+	}
+
+	// find a pod to execute the vclusterops API
+	var podIP string
+	for i := range pods.Items {
+		pod := &pods.Items[i]
+		if pod.Status.Phase == podRunning {
+			for j := range pod.Status.ContainerStatuses {
+				// check NMA container ready
+				if pod.Status.ContainerStatuses[j].Ready {
```
roypaulin, line 95: You need to check the container name first to make sure it is indeed the `nma`(pod.Status.ContainerStatuses[j].Name == names.NMAContainer`.

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: A break at line 96 exits just the first loop. A break here exits the second loop. I mean, if we find a pod satisfying the above conditions, we don't need to check all pods in the second loop.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +79,108 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
 	return vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
 		v1.MakeCondition(vapi.QueryComplete, metav1.ConditionTrue, "Completed"), stateSuccessQuery)
 }
+
+// makeDispatcher will create a Dispatcher object based on the feature flags set.
+func (q *QueryReconciler) makeDispatcher(log logr.Logger, vdb *v1.VerticaDB,
+	passwd *string) (vadmin.Dispatcher, error) {
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		var password string
+		if passwd != nil {
```
roypaulin, line 180: I don't think we need this check if we already know we are not going to need the dbadmin password.

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: Yes, I realized that too. Actually, I have a question: Do we only have an NMA container if we define the annotation vertica.com/run-nma-in-sidecar: "true"? What if we don't define this annotation and we only have a server container? If we include the condition pod.Status.ContainerStatuses[j].Name == names.NMAContainer, we will not find any pods that satisfy the above conditions.

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: Just think, if I make it into a function, it would be better. We can return the value and avoid using break statements.

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: Do we need to check to ensure that it's an NMA container? I realized that it has an NMA container if we define the annotation vertica.com/run-nma-in-sidecar: "true". Otherwise, it will be a server container. A server container still has NMA certs to run vclusterops

pkg/test/helpers.go
chinhtranvan: As discussed in the scrum meeting, we should create a 'dummy' function to avoid calling vclusterOps for the unit tests.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +80,106 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+	return "", fmt.Errorf("can't find a pod to call the vclusterOps API")
+}
+
+// makeDispatcher will create a Dispatcher object based on the feature flags set.
+func (q *QueryReconciler) makeDispatcher(log logr.Logger, vdb *v1.VerticaDB,
+	_ *string) (vadmin.Dispatcher, error) {
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		return vadmin.MakeVClusterOps(log, vdb, q.VRec.GetClient(), "", q.VRec, vadmin.SetupVClusterOps), nil
+	}
+	return nil, fmt.Errorf("ShowRestorePoints is not supported for admintools deployments")
```
spilchen, line 184: I think this is the first error we would see if we tried to run this against an Admintools deployment. We should surface an event and no error for that case. I opened a follow-on jira (VER-91846). Nothing to be done in this PR.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +80,106 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+		}
+	}
+	return "", fmt.Errorf("can't find a pod to call the vclusterOps API")
+}
+
+// makeDispatcher will create a Dispatcher object based on the feature flags set.
+func (q *QueryReconciler) makeDispatcher(log logr.Logger, vdb *v1.VerticaDB,
+	_ *string) (vadmin.Dispatcher, error) {
+	if vmeta.UseVClusterOps(vdb.Annotations) {
+		return vadmin.MakeVClusterOps(log, vdb, q.VRec.GetClient(), "", q.VRec, vadmin.SetupVClusterOps), nil
```
spilchen, line 182: Lets leave a comment saying why we pass in an empty string for the password.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +80,106 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
-		v1.MakeCondition(vapi.Querying, metav1.ConditionFalse, "Completed"), stateQuerying)
-	if err != nil {
-		return err
+	// call showRestorePoints vcluster API
+	q.VRec.Eventf(q.Vrpq, corev1.EventTypeNormal, events.ShowRestorePointsStarted,
+		"Starting show restore points")
+	start := time.Now()
+	if res, errRun := dispatcher.ShowRestorePoints(ctx, opts...); verrors.IsReconcileAborted(res, errRun) {
+		if errRun != nil {
+			err = vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
```
spilchen, line 146: We should log an event if the query failed.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +80,106 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
-	// clear Querying status condition
-	err = vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
-		v1.MakeCondition(vapi.Querying, metav1.ConditionFalse, "Completed"), stateQuerying)
-	if err != nil {
-		return err
+	// call showRestorePoints vcluster API
+	q.VRec.Eventf(q.Vrpq, corev1.EventTypeNormal, events.ShowRestorePointsStarted,
+		"Starting show restore points")
+	start := time.Now()
+	if res, errRun := dispatcher.ShowRestorePoints(ctx, opts...); verrors.IsReconcileAborted(res, errRun) {
```
spilchen, line 144: Is there any case where `ShowRestorePoints` will return a `ctrl.Result` with a requeue? We don't do anything if that occurs. I'd like to remove that from `ShowRestorePoints` if we don't have any reason to return one.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +80,106 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+				errRun = errors.Join(errRun, err)
+			}
+			return errRun
+		}
 	}
+	q.VRec.Eventf(q.Vrpq, corev1.EventTypeNormal, events.ShowRestorePointsSucceeded,
+		"Successfully queried restore points in %s", time.Since(start).Truncate(time.Second))
 
 	// set the QueryComplete if the vclusterops API succeeded
 	return vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
```
spilchen, line 158: We don't clear the Querying status condition if successful.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +80,106 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
-// fetch the VerticaDB and collect access information to the communal storage for the VerticaRestorePointsQuery CR,
-func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result, error) {
+// fetch the VerticaDB and collect access information to the communal storage for the VerticaRestorePointsQuery CR
+func (q *QueryReconciler) collectInfoFromVdb(ctx context.Context) (res ctrl.Result, err error) {
 	vdb := &v1.VerticaDB{}
-	res := ctrl.Result{}
-
-	opts := []restorepointsquery.Option{
-		restorepointsquery.WithInitiator(q.InitiatorPod, q.InitiatorPodIP),
+	if res, err = fetchVDB(ctx, q.VRec, q.Vrpq, vdb); verrors.IsReconcileAborted(res, err) {
```
spilchen, line 114: Why do we fetch the vdb? We should already have one present.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -71,66 +80,106 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, req *ctrl.Request) (ctr
...
+		pod := &pods.Items[i]
+		if pod.Status.Phase == podRunning {
+			for j := range pod.Status.ContainerStatuses {
+				if pod.Status.ContainerStatuses[j].Ready && pod.Status.ContainerStatuses[j].Name == names.NMAContainer {
+					return pod.Status.PodIP, nil
+				}
+			}
+		}
+	}
+	return "", fmt.Errorf("can't find a pod to call the vclusterOps API")
```
spilchen, line 175: We shouldn't surface this as an error. We should write a message to the log saying we couldn't find any pod to run the query at and requeue the iteration.

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: oh I think you had a previous comments `Could we merge the two UpdateConditionAndState calls into one? We may need to allow multiple status conditions to be set?`

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: This is the one present from https://jira.verticacorp.com/jira/browse/VER-90383. I just remove RetryOnConflict based on your previous comments

pkg/controllers/vrpq/suite_test.go
```diff
@@ -135,3 +139,52 @@ func deleteSecret(ctx context.Context, vdb *v1.VerticaDB, secretName string) {
 	Expect(k8sClient.Get(ctx, nm, secret)).Should(Succeed())
 	Expect(k8sClient.Delete(ctx, secret)).Should(Succeed())
 }
+
+// mockVClusterOpsDispatchWithCustomSetup is like mockVClusterOpsDispatcher,
+// except you provide your own setup API function.
+func mockVClusterOpsDispatcherWithCustomSetup(vdb *v1.VerticaDB,
```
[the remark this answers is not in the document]
chinhtranvan: I wonder we can't reuse the code even we export this function. I can create a  `helper.go` and move it to this function. If you have a better place, please let me know

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -141,44 +141,53 @@ func (q *QueryReconciler) runShowRestorePoints(ctx context.Context, dispatcher v
...
-		if errRun != nil {
-			err = vrpqstatus.UpdateConditionAndState(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
-				v1.MakeCondition(vapi.Querying, metav1.ConditionFalse, "Failed"), stateFailedQuery)
-			if err != nil {
-				errRun = errors.Join(errRun, err)
-			}
-			return errRun
+	errRun := dispatcher.ShowRestorePoints(ctx, opts...)
+	if errRun != nil {
+		q.VRec.Eventf(q.Vrpq, corev1.EventTypeNormal, events.ShowRestorePointsFailed, "Failed when calling show restore points")
```
spilchen, line 146: We can just use `Event` and not `Eventf` because we aren't formatting any parms.

pkg/controllers/vrpq/query_reconciler.go
spilchen: You added it back, but now we have 2 status condition updates back to back. I was hoping to include this in a single status update. Maybe defer this until your next PR when you have to update the status with the results that come back. All 3 things, can be done in a single status update call.

pkg/controllers/vrpq/query_reconciler.go
spilchen, line 146: We should use `corev1.EventTypeWarning` instead of `corev1.EventTypeNormal` because its a failure condition.

pkg/controllers/vrpq/query_reconciler.go
spilchen, line 1: What do we use `QueryReconciler.OpCfg` for?

chinhtranvan: will remove it

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: As discussed, fetchVDB is still necessary because we require gathering information from VerticaDB; specifically, we need to retrieve fields in Vdb

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: Thank you for this information. Done!

pkg/controllers/vrpq/query_reconciler.go
chinhtranvan: Sure, will keep in mind and fix it in part 2 for this PR

chinhtranvan:
@spilchen All comments have been addressed. As we discussed in the scrum, I will wait for your changes to be updated and merged into 'vnext' in order to resolve the failures in the e2e tests

spilchen (approved):
Thanks for doing the revisions. vnext has been merged up. You should be able to pull that into your branch for the e2e tests to succeed.

chinhtranvan:
> Thanks for doing the revisions. vnext has been merged up. You should be able to pull that into your branch for the e2e tests to succeed.

Thank you for informing me. I observed that the scorecard test failed when you merged 'main' into 'vnext' https://github.com/vertica/vertica-kubernetes/actions/runs/7788337652. The error message reads: 'Unable to find image 'registry:2' locally. Docker: Error response from daemon: unauthorized: authentication required.' It appears that we are missing the image

spilchen:
> > Thanks for doing the revisions. vnext has been merged up. You should be able to pull that into your branch for the e2e tests to succeed.
> 
> Thank you for informing me. I observed that the scorecard test failed when you merged 'main' into 'vnext' https://github.com/vertica/vertica-kubernetes/actions/runs/7788337652. The error message reads: 'Unable to find image 'registry:2' locally. Docker: Error response from daemon: unauthorized: authentication required.' It appears that we are missing the image

That's not an image we build. So, I'm guessing its a flaky network connection. I'll rerun that job once the workflow is done.

PR #697, opened 2024-02-09
VCluster API Refresh

chinhtranvan:
This includes a few updates to the vcluster API. Some notable changes:
- New API to install the packages (`VInstallPackages`)
- Update to function signature for `VReviveDB`; we are returning a new parameter, but it doesn't apply to the operators usage
- Reduced the noise in the log when polling for up nodes

### DIFF
--- a/pkg/vadmin/revive_db_vc.go
+++ b/pkg/vadmin/revive_db_vc.go
@@ -42,7 +42,7 @@ func (v *VClusterOps) ReviveDB(ctx context.Context, opts ...revivedb.Option) (ct
 	s.Make(opts...)
 
 	vcOpts := v.genReviveDBOptions(&s, certs)
-	_, err = v.VReviveDatabase(vcOpts)
+	_, _, err = v.VReviveDatabase(vcOpts)
 	if err != nil {
 		return v.logFailure("VReviveDatabase", events.ReviveDBFailed, err)
 	}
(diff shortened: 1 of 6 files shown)


pkg/vadmin/revive_db_vc.go
```diff
@@ -42,7 +42,7 @@ func (v *VClusterOps) ReviveDB(ctx context.Context, opts ...revivedb.Option) (ct
 	s.Make(opts...)
 
 	vcOpts := v.genReviveDBOptions(&s, certs)
-	_, err = v.VReviveDatabase(vcOpts)
+	_, _, err = v.VReviveDatabase(vcOpts)
```
roypaulin, line 45: I am curious about what the `VCoordinationDatabase` is used for in vcluster.

pkg/vadmin/revive_db_vc_test.go
```diff
@@ -31,11 +31,11 @@ const (
 )
 
 // mock version of VReviveDatabase() that is invoked inside VClusterOps
-func (m *MockVClusterOps) VReviveDatabase(options *vops.VReviveDatabaseOptions) (string, error) {
+func (m *MockVClusterOps) VReviveDatabase(options *vops.VReviveDatabaseOptions) (string, *vops.VCoordinationDatabase, error) {
 	// verify basic options
 	err := m.VerifyDBNameAndIPv6(&options.DatabaseOptions)
 	if err != nil {
-		return "", err
+		return "", &vops.VCoordinationDatabase{}, err
```
roypaulin, line 38: can we just return nil?

chinhtranvan: Sure

pkg/vadmin/revive_db_vc.go
chinhtranvan: Ah, in the ReviveDB process. I think Le wants to save this information in the VCoordinationDatabase. With this information, he can write it into the config file

chinhtranvan:
@spilchen Could you merge this PR so that I can merge the main branch into vnext? We need to include some updates for vnext with this change.

PR #698, opened 2024-02-11
Ensure httpstls.json is generated for databases migrating to v24.1.0

spilchen:
This fixes an issue where we didn't generate the httptls.json file for databases migrated to 24.1.0. Without this file generated, it is possible that the HTTPS service doesn't come up, which is needed by vclusterOps in 24.1.0+ deployments.

We can avoid generating this file in the following circumstances:
- Databases created in 24.1.0+ will have default certs generated for HTTPS during the bootstrap-catalog call.
- Database that have setup TLS certs in the catalog. See the docs for that here: https://docs.vertica.com/24.1.x/en/admin/managing-db/https-service/#certificate-authentication

The operator won't know about the second case. We have introduced a new annotation for the first case. When `vertica.com/https-tls-conf-generation` is set to false, we will avoid generation. When the operator creates a database on a 24.1.0 database it will automatically set this annotation.


pkg/vk8s/meta_updater.go
```diff
@@ -0,0 +1,89 @@
...
+		if err := cl.Get(ctx, nm, obj); err != nil {
+			if errors.IsNotFound(err) {
+				return nil
+			}
+			return err
+		}
+
+		annotationsChanged := false
+		objAnnotations := obj.GetAnnotations()
+		for k, v := range chgs.NewAnnotations {
```
roypaulin, line 51: As labels and annotations are both map, does it make sense to create a new function to use this logic for both?

pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -199,7 +202,27 @@ func (c *CreateDBReconciler) generatePostDBCreateSQL(ctx context.Context, initia
...
-func (c *CreateDBReconciler) postCmdCleanup(_ context.Context) (ctrl.Result, error) {
+func (c *CreateDBReconciler) postCmdCleanup(ctx context.Context) (ctrl.Result, error) {
+	pf, ok := c.findPodToRunInit()
+	if !ok {
+		return ctrl.Result{}, errors.New("could not find a PodFact for create db's post cmd cleanup")
+	}
+	// The generation of httpstls.json is influenced by the DBInitialized status
+	// condition. Now that has changed, we need to set an annotation to continue
+	// getting the same behavior. Since the default behavior is to generate the
+	// file, we need to set an annotation if we didn't generate the file yet.
```
roypaulin, line 213: For some reason this last part is a little bit confusing to me.

roypaulin (approved):
Looks good!

pkg/vk8s/meta_updater.go
spilchen: Yes, this is a good idea.

pkg/controllers/vdb/createdb_reconciler.go
spilchen: Once we have set the DBInitialized status then the function IsHTTPSTlsConfGenerationEnabled() will return true. It doesn't know what version was used to create the database, so it defaults to creating the httpstls.json file. This annotation allows us to keep skipping the generation of that file. It's not perfect because you could to a revive in the future, and if you don't set this annotation, then we will generate the file. But I think its better to generate the file and not need it than the opposite.

roypaulin (approved):

PR #700, opened 2024-02-13
Block access to VerticaRestorePointsQuery for incompatible VerticaDBs

chinhtranvan:
This PR adds some checks for the new VerticaRestorePointsQuery CR. The VerticaDB should be running with a minimum version of 24.2.0 and should be deployed with vclusterOps. We don't support admintools deployments


pkg/events/event.go
```diff
@@ -53,6 +53,8 @@ const (
 	ClusterRestartSucceeded         = "ClusterRestartSucceeded"
 	ShowRestorePointsStarted        = "ShowRestorePointsStarted"
 	ShowRestorePointsFailed         = "ShowRestorePointsFailed"
+	IncompatibleDB                  = "IncompatibleDB"
```
roypaulin, line 56: I suggest `RestoreFromRestorePointNotSupported` or `RestoreNotSupported`.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -79,10 +78,32 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.
...
+	// check version for Vdb, the minimim version should be 24.2.0
+	vinf, err := q.Vdb.MakeVersionInfoCheck()
 	if err != nil {
-		return ctrl.Result{}, err
+		return ctrl.Result{}, nil
+	}
+	if !vinf.IsEqualOrNewer(v1.RestoreSupportedMinVersion) {
+		q.VRec.Event(q.Vrpq, corev1.EventTypeWarning, events.IncompatibleDB, "Incompatibility with the database")
+		err = vrpqstatus.Update(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+			[]*metav1.Condition{v1.MakeCondition(vapi.QueryComplete, metav1.ConditionTrue, "IncompatibleDB")}, stateFailedQuery, nil)
```
roypaulin, line 89: I don't think it makes sense to set `QueryComplete` when `Querying has not started. Also I do not think we want to retry if version does not support Restore or vclusterops is disabled. So maybe we can add an extra condition that indicates that meet all the requirements to successfully perform a `Query` and only when it is true can we start`Querying`. Or we can add a special reason to `Querying` to express what I just described. What do you think @spilchen ?

spilchen: I agree. We shouldn't retry if things aren't compatible because chances are they will be incompatible next reconcile iteration too. 

Maybe add a QueryReady status condition. We can then set this to false, with a reason depending on what we find.

pkg/controllers/vrpq/query_reconciler.go
spilchen, line 1: If we add a new status condition, maybe we can move those checks to a new reconcile object. The logic in here is getting quite complex, so it would help to have that separation.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -79,10 +78,32 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.
...
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// setup dispatcher for vclusterops API
+	dispatcher := q.makeDispatcher(q.Log, q.Vdb, nil /*password*/)
+	if dispatcher == nil {
```
spilchen, line 98: I'm not a fan of the of this side affect (i.e. if dispatch cannot be made, then it must mean its admintools deployments). Lets be explicit with our check.

pkg/controllers/vrpq/precheckquery_reconciler.go
```diff
@@ -0,0 +1,85 @@
...
+	vdb := &v1.VerticaDB{}
+	nm := names.GenNamespacedName(q.Vrpq, q.Vrpq.Spec.VerticaDBName)
+	if res, err := vk8s.FetchVDB(ctx, q.VRec, q.Vrpq, nm, vdb); verrors.IsReconcileAborted(res, err) {
+		return res, err
+	}
+
+	// check version for Vdb, the minimim version should be 24.2.0
+	vinf, err := vdb.MakeVersionInfoCheck()
+	if err != nil {
+		return ctrl.Result{}, nil
```
spilchen, line 62: Lets return the error
```suggestion
		return ctrl.Result{}, err
```

pkg/controllers/vrpq/precheckquery_reconciler.go
```diff
@@ -0,0 +1,85 @@
...
+ limitations under the License.
+*/
+
+package vrpq
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	v1 "github.com/vertica/vertica-kubernetes/api/v1"
```
spilchen, line 22: Can we rename this package import to be `vapi`? That's what we have done in most places when importing this.

pkg/controllers/vrpq/precheckquery_reconciler.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// check version for Vdb, the minimim version should be 24.2.0
+	vinf, err := vdb.MakeVersionInfoCheck()
+	if err != nil {
+		return ctrl.Result{}, nil
+	}
+	if !vinf.IsEqualOrNewer(v1.RestoreSupportedMinVersion) {
+		q.VRec.Event(q.Vrpq, corev1.EventTypeWarning, events.RestoreNotSupported, "Incompatibility with the database")
+		err = vrpqstatus.Update(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+			[]*metav1.Condition{v1.MakeCondition(vapi.QueryReady, metav1.ConditionFalse, "IncompatibleDB")}, stateFailedQuery, nil)
```
spilchen, line 67: We change the state to be "Query failed". That implies we want the query. What about "Incompatible"?

pkg/controllers/vrpq/precheckquery_reconciler.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Should be deployed with vclusterops, not supported for admintools deployments
+	if !vmeta.UseVClusterOps(vdb.Annotations) {
+		q.VRec.Event(q.Vrpq, corev1.EventTypeWarning, events.AdmintoolsNotSupported,
+			"ShowRestorePoints is not supported for admintools deployments")
+		err = vrpqstatus.Update(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+			[]*metav1.Condition{v1.MakeCondition(vapi.QueryReady, metav1.ConditionFalse, "AdmintoolsNotSupported")}, stateFailedQuery, nil)
```
spilchen, line 79: Suggest for state message, we use "Incompatible" instead of `stateFailedQuery`

pkg/controllers/vrpq/precheckquery_reconciler.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/vertica/vertica-kubernetes/pkg/names"
+	"github.com/vertica/vertica-kubernetes/pkg/vk8s"
+	vrpqstatus "github.com/vertica/vertica-kubernetes/pkg/vrpqstatus"
+	corev1 "k8s.io/api/core/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type PrecheckQueryReconciler struct {
```
spilchen, line 37: Suggest calling this `VDBVerifyReconciler` to match the name used for the scrutinize controller. Please update the file name too.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -80,10 +79,7 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.
 	}
```
spilchen, line 79: We want to check the new status condition in here. QueryReady is not true, then we exit and perform as a no-op.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -80,10 +79,7 @@ func (q *QueryReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.
...
 
 	// setup dispatcher for vclusterops API
-	dispatcher, err := q.makeDispatcher(q.Log, q.Vdb, nil /*password*/)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
+	dispatcher := q.makeDispatcher(q.Log, q.Vdb, nil /*password*/)
```
spilchen, line 82: We still want to return an error here. We should never see the error, but its good defensive programming to handle the case where admintools deployments is set in here.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -132,7 +128,8 @@ func (q *QueryReconciler) runShowRestorePoints(ctx context.Context, dispatcher v
 	opts []showrestorepoints.Option) (err error) {
 	// set Querying status condition ,state and restore points prior to calling vclusterops API
 	err = vrpqstatus.Update(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
-		[]*metav1.Condition{v1.MakeCondition(vapi.Querying, metav1.ConditionTrue, "Started")}, stateQuerying, nil)
+		[]*metav1.Condition{v1.MakeCondition(vapi.QueryReady, metav1.ConditionTrue, "Started"),
```
spilchen, line 131: We shouldn't be touching QueryReady in this reconciler. That should only be modified in the proceeding reconciler.

pkg/controllers/vrpq/precheckquery_reconciler.go
```diff
@@ -0,0 +1,85 @@
...
+	if !vmeta.UseVClusterOps(vdb.Annotations) {
+		q.VRec.Event(q.Vrpq, corev1.EventTypeWarning, events.AdmintoolsNotSupported,
+			"ShowRestorePoints is not supported for admintools deployments")
+		err = vrpqstatus.Update(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+			[]*metav1.Condition{v1.MakeCondition(vapi.QueryReady, metav1.ConditionFalse, "AdmintoolsNotSupported")}, stateFailedQuery, nil)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{}, nil
```
spilchen, line 84: If we get here then we should set the QueryReady status condition.

pkg/controllers/vrpq/precheckquery_reconciler.go
```diff
@@ -0,0 +1,85 @@
...
+		return res, err
+	}
+
+	// check version for Vdb, the minimim version should be 24.2.0
+	vinf, err := vdb.MakeVersionInfoCheck()
+	if err != nil {
+		return ctrl.Result{}, nil
+	}
+	if !vinf.IsEqualOrNewer(v1.RestoreSupportedMinVersion) {
+		q.VRec.Event(q.Vrpq, corev1.EventTypeWarning, events.RestoreNotSupported, "Incompatibility with the database")
```
spilchen, line 65: I feel like the message doesn't convey why we aren't compatible.
```suggestion
		q.VRec.Eventf(q.Vrpq, corev1.EventTypeWarning, events.RestoreNotSupported, "The Vertica version '%s' doesn't support in-database restore points", vinf.VdbVer)
```

pkg/controllers/vrpq/vdbverify_reconciler.go
```diff
@@ -0,0 +1,92 @@
...
+			"ShowRestorePoints is not supported for admintools deployments")
+		err = vrpqstatus.Update(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+			[]*metav1.Condition{vapi.MakeCondition(v1beta1.QueryReady, metav1.ConditionFalse, "AdmintoolsNotSupported")}, stateIncompatibleDB, nil)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+	return ctrl.Result{}, vrpqstatus.Update(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
+		[]*metav1.Condition{vapi.MakeCondition(v1beta1.QueryReady, metav1.ConditionTrue, "Started")}, stateQuerying, nil)
```
spilchen, line 91: The reason for the QueryReady status condition is done at the point. So suggestion this:
```suggestion
		[]*metav1.Condition{vapi.MakeCondition(v1beta1.QueryReady, metav1.ConditionTrue, "Complete")}, stateQuerying, nil)
```

spilchen (approved):
Looks good. I just had one minor suggestion.

spilchen:
@chinhtranvan I see you are having issues with the online-upgrade-with-transient test. Zhuyou is having issues with it to in his PR. I added a commit ([664eb66](https://github.com/vertica/vertica-kubernetes/pull/703/commits/664eb6693539c01ef2e205f8e78d8dac36c7ea92)) to make it less flaky. If that works you may want to merge in his changes when it goes into vnext.

chinhtranvan:
> @chinhtranvan I see you are having issues with the online-upgrade-with-transient test. Zhuyou is having issues with it to in his PR. I added a commit ([664eb66](https://github.com/vertica/vertica-kubernetes/pull/703/commits/664eb6693539c01ef2e205f8e78d8dac36c7ea92)) to make it less flaky. If that works you may want to merge in his changes when it goes into vnext.

Yes, I will let @jizhuoyu merge first, then I merge vnext to my branch. Thank you for this information

PR #701, opened 2024-02-13
Avoid deploying the operator with default label

### LINKED ISSUE #694: Migration to v24.1 issues
Hello,

We had some Vertica clusters on Openshift, running version v23.x, and we wanted to update them to v24.1.

We encountered some issues, that I want to share here : 

- Operator is installed with OLM (with Openshift OperatorHub), and we had to uninstall Operator v1 first on every namespaces, then install v2 cluster wide.
With Openshift, operator installs in "openshift-operators" namespace, and the 3 services created (controller-manager, metrics and webhooks) use the label selector "control-plane=controller-manager".
But, there is some other pods (like ArgoCD), who use this label and the services route to both Vertica and ArgoCD controller, which results in error in Verica controller.
I had to manually update those services to add label selector "app.kubernetes.io/name=vertica-db-operator"

- Our clusters were installed with httpServerMode=Disabled with resulted that Vertica never created https certs for HTTPServer.
With v23 it is not a problem due to controller using admintools, but with v24, controller expect HTTPServer to be running to send commands and to check Vertica liveness (in K8S probes).
I had to first enable HTTPServer with v23 and Operator v1, to make it generate certificates, then I can update to operator v2, then to Vertica v24.
Could the operator generate certs and httpstls.json file if they are missing during migration ? Actually I can see in the code it only generate them during DB bootstrap, not migration.

- In order to run Vertica v23 cluster, using CRD v1beta and Operator v2, the operator converts the CRD to v1, and generate a ServiceAccount, with a randomized name, as we can't set ServiceAccountName in v1beta CRD.
On Openshift, we had to wait that Operator generate ServiceAccount, then create a RoleBinging to give privileged scc to Vertica Pods.
So this can't be done with infra as code tools like ArgoCD, as the ServiceAccount name is not predictable.
Can the v1beta CRD be updated to add ServiceAccountName, so it can be taken in account during conversion ?

And I have a last question :
- Is the config directory (containing certs for https server) backuped up to communal storage ?
Because if we lose a node and his local volume, config dir would be lost, and I don't think operator would recreate httpstls.json file if it it missing.

spilchen:
Any operator that is built with the operator-sdk framework will have default selector labels added for the operator like this:
```
  control-plane: controller-manager
```

When the operator is deployed in the same namespace as other operator, that also continue using the default label, then the service object for the operator fails to work. The service object is used by the webhook, so it will route webhook requests to the wrong pod.

This change is to use the following label instead:
```
  control-plane: verticadb-operator
```

I also took the opportunity to rename the operator objects. The deployment object was renamed from `verticadb-operator-controller-manager` to `verticadb-operator-manager`.

Closes #694

### DIFF
--- a/tests/e2e-leg-5/helm-nameoverride/15-assert.yaml
+++ b/tests/e2e-leg-5/helm-nameoverride/15-assert.yaml
@@ -15,7 +15,7 @@ apiVersion: v1
 kind: Pod
 metadata:
   labels:
-    control-plane: controller-manager
+    control-plane: verticadb-operator
 status:
   phase: Running
 ---
@@ -32,7 +32,7 @@ metadata:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: matt-controller-manager
+  name: matt-manager
 ---
 apiVersion: v1
 kind: ConfigMap
--- a/tests/e2e-leg-5/helm-nameoverride/15-assert.yaml
+++ b/tests/e2e-leg-5/helm-nameoverride/15-assert.yaml
@@ -15,7 +15,7 @@ apiVersion: v1
 kind: Pod
 metadata:
   labels:
-    control-plane: verticadb-operator
+    control-plane: matt
 status:
   phase: Running
 ---
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -567,7 +567,7 @@ By default, the memory profiler is disabled. To enable it, add a parameter when 
 1. Use `kubectl edit` to open the running deployment for editing:
 
    ```shell
-   kubectl edit deployment verticadb-operator-controller-manager
+   kubectl edit deployment verticadb-operator-manager
    ```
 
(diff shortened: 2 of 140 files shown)


tests/e2e-leg-5/helm-nameoverride/15-assert.yaml
```diff
@@ -15,7 +15,7 @@ apiVersion: v1
 kind: Pod
 metadata:
   labels:
-    control-plane: controller-manager
+    control-plane: matt
```
roypaulin, line 18: Is this on purpose?

roypaulin (commented):
Looks good!

tests/e2e-leg-5/helm-nameoverride/15-assert.yaml
spilchen: Actulally it is. With the change of the control-plane label, it can now be templated in the helm chart. So, in this testcase, which sets the `nameOverride` parameter, the label will be set to the value of that parameter.

roypaulin (approved):

PR #702, opened 2024-02-13
Fixes Istio proxy sidecar initialization issue

spilchen:
Starting in istio 1.17, the istio proxy is now injected as the first container. This breaks some assumptions in the operator code. It assumes that any sidecar is injected at the end and the first two containers are the nma and server. This changes fixes that by referencing containers by their names rather than index. I removed all of the helpers based on index and instead get them by name.

### DIFF
--- a/changes/unreleased/Fixed-20240213-111303.yaml
+++ b/changes/unreleased/Fixed-20240213-111303.yaml
@@ -1,0 +1,5 @@
+kind: Fixed
+body: Fix when istio proxy sidecar is injected as the first container
+time: 2024-02-13T11:13:03.768289581-04:00
+custom:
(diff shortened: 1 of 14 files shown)


roypaulin (approved):
Looks good! can you also merge this into vnext?

roypaulin:
There is a test that fails with `Error response from daemon`. Zhuoyu also got that error.

spilchen:
> There is a test that fails with `Error response from daemon`. Zhuoyu also got that error.

Yes, seems like an issue with our docker account. We no longer have pull access to our private repository where we keep all the server images. I have a PR opened (#704) to use a different repository for the time being.

PR #706, opened 2024-02-15
Run scrutinize in its own pod

roypaulin:
When a VerticaScrutinize resource is created, the operator:
  - collects some info from the specified VerticaDB in order to construct vcluster scrutinize command arguments
  - builds the scrutinize podspec. That pod will shares the same serviceaccount, podsecuritycontext, image as vertica pods.
  - adds some env vars to the podspec needed by nma and also 2 new env vars(`PASSWORD_NAMESPACE`, `PASSWORD_NAME`) to allow vcluster scrutinize to read the db password from secret.

Once the pod is created, a new reconciler(`PodPollingReconciler`) will wait for scrutinize to finish and update status condition


roypaulin:
I am not sure about using the same serviceaccount and podsecuritycontext on verticapods and scrutinize pod. Were you planning for scrutinize to have its own?

pkg/builder/builder.go
```diff
@@ -420,13 +445,21 @@ func buildVolumes(vdb *vapi.VerticaDB) []corev1.Volume {
...
+		secrets.IsK8sSecret(vdb.Spec.NMATLSSecret) {
+		vols = append(vols, buildNMACertsSecretVolume(vdb))
+	}
 	if vscr.Spec.Volume == nil {
-		return buildDefaultScrutinizeVolume()
+		vols = append(vols, buildDefaultScrutinizeVolume())
+		return vols
 	}
-	return *vscr.Spec.Volume
+	vols = append(vols, *vscr.Spec.Volume)
```
spilchen, line 461: If someone provides their own volume, how do we ensure the volume mount will use it? I think its required that any custom volume must be called `scrutinize`

pkg/builder/builder.go
```diff
@@ -787,18 +822,21 @@ func makeNMAContainer(vdb *vapi.VerticaDB, sc *vapi.Subcluster) corev1.Container
...
+func makeScrutinizeInitContainer(vscr *v1beta1.VerticaScrutinize, vdb *vapi.VerticaDB, args []string) corev1.Container {
+	cnt := corev1.Container{
+		Image:        vdb.Spec.Image,
+		Name:         names.ScrutinizeInitContainer,
+		Command:      buildScrutinizeCmd(args),
+		VolumeMounts: buildScrutinizeVolumeMounts(vscr, vdb),
 		Resources:    vscr.Spec.Resources,
+		Env:          buildCommonEnvVars(vdb),
 	}
+	cnt.Env = append(cnt.Env, buildNMAEnvVars(vdb)...)
```
spilchen, line 834: Reusing `buildNMAEnvVars` is fined for scrutinize, but it might be clear if we also rename the function. This isn't the env vars for the NMA anymore as its used by scrutinize too. How about `buildNMATLSCertsEnvVar`?

pkg/controllers/vscr/podpolling_reconciler.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	if !cntStatus.Ready {
+		if cntStatus.State.Terminated != nil {
+			p.VRec.Eventf(p.Vscr, corev1.EventTypeWarning, events.VclusterOpsScrutinizeFailed,
+				"Vcluster scrutinize run failed")
+			cond := v1.MakeCondition(v1beta1.ScrutinizeCollectionFinished, metav1.ConditionTrue, events.VclusterOpsScrutinizeFailed)
+			return ctrl.Result{}, vscrstatus.UpdateCondition(ctx, p.VRec.Client, p.Vscr, cond)
+		}
+		if cntStatus.State.Running != nil {
+			p.VRec.Eventf(p.Vscr, corev1.EventTypeWarning, events.VclusterOpsScrutinizeInProgress,
```
spilchen, line 82: I don't think an event is needed if we are just going to continue to poll. Maybe just a log message.

pkg/controllers/vscr/podpolling_reconciler.go
```diff
@@ -0,0 +1,114 @@
...
+		return ctrl.Result{}, fmt.Errorf("could not find scrutinize container status")
+	}
+	if !cntStatus.Ready {
+		if cntStatus.State.Terminated != nil {
+			p.VRec.Eventf(p.Vscr, corev1.EventTypeWarning, events.VclusterOpsScrutinizeFailed,
+				"Vcluster scrutinize run failed")
+			cond := v1.MakeCondition(v1beta1.ScrutinizeCollectionFinished, metav1.ConditionTrue, events.VclusterOpsScrutinizeFailed)
+			return ctrl.Result{}, vscrstatus.UpdateCondition(ctx, p.VRec.Client, p.Vscr, cond)
+		}
+		if cntStatus.State.Running != nil {
```
spilchen, line 81: We want to requeue if the container is waiting too (`cntStatus.State.Waiting`)

pkg/builder/builder.go
```diff
@@ -695,17 +728,19 @@ func buildPodSpec(vdb *vapi.VerticaDB, sc *vapi.Subcluster) corev1.PodSpec {
...
-func buildScrutinizePodSpec(vscr *v1beta1.VerticaScrutinize) corev1.PodSpec {
+func buildScrutinizePodSpec(vscr *v1beta1.VerticaScrutinize, vdb *vapi.VerticaDB, args []string) corev1.PodSpec {
 	termGracePeriod := int64(scrutinizeTermGracePeriod)
 	return corev1.PodSpec{
 		NodeSelector:                  vscr.Spec.NodeSelector,
 		Affinity:                      GetK8sAffinity(vapi.Affinity(vscr.Spec.Affinity)),
 		Tolerations:                   vscr.Spec.Tolerations,
-		InitContainers:                makeScrutinizeInitContainers(vscr),
+		InitContainers:                makeScrutinizeInitContainers(vscr, vdb, args),
 		Containers:                    []corev1.Container{makeScrutinizeMainContainer(vscr)},
```
spilchen, line 738: The main container continues to use busybox. Is that the best container to use? I noticed it doesn't have bash, curl or any package manager. Should we find an image that does have these things? Are there other things it needs? Maybe an annotation to allow the default image to be overridden too.

pkg/builder/builder.go
```diff
@@ -82,6 +82,10 @@ const (
 	// Scrutinize constants
 	scrutinizeMountName       = "scrutinize"
 	scrutinizeTermGracePeriod = 0
+	// Environment variables to allow vcluster scrutinize to get the password
+	// from a secret
+	superuserPasswordSecretNamespaceEnv = "SUPERUSER_PASSWD_NAMESPACE"
```
spilchen, line 87: Does scrutinize already read this env var? I can't seem to find it.

pkg/controllers/vscr/podpolling_reconciler.go
```diff
@@ -0,0 +1,114 @@
+/*
+Copyright [2021-2023] Open Text.
+Licensed under the Apache License, Version 2.0 (the "License");
```
spilchen, line 3: This copyright notice looks different then other files. Its missing some blank lines?

spilchen: I see this in other files too (podpolling_reconciler_test.go)

pkg/controllers/vdb/serviceaccount_reconciler.go
```diff
@@ -73,10 +73,14 @@ func (s *ServiceAccountReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
...
-	// mounted volume or non-k8s secret store.
+	// There are some cases where there is no need to create the role and rolebinding:
+	//	- NMA reads certs from mounted volume or non-k8s secret store
+	//	- no password secret is specified or the password secret is a non-k8s secret.
+	//    This case refers to vclusterOps scrutinize which may need to read the db
+	//	  password directly from the API
 	if vmeta.UseVClusterOps(s.Vdb.Annotations) &&
-		(vmeta.UseNMACertsMount(s.Vdb.Annotations) || !secrets.IsK8sSecret(s.Vdb.Spec.NMATLSSecret)) {
+		(vmeta.UseNMACertsMount(s.Vdb.Annotations) || !secrets.IsK8sSecret(s.Vdb.Spec.NMATLSSecret)) &&
+		(s.Vdb.Spec.PasswordSecret == "" || !secrets.IsK8sSecret(s.Vdb.Spec.PasswordSecret)) {
```
spilchen, line 83: Another approach is to mount the password secret in the pod, then have scrutinize read the password from the file that is mounted. I think we want to do this for other vcluster commands anyway (we have a jira ticket). Can you add a comment here that we should not create role/rolebinding for access to password secret once VER-90362 is addressed?

spilchen (commented):
This looks good. I deployed the new CR on my machine and it was able to run scrutinize just fine. It's pretty slick.

pkg/builder/builder.go
```diff
@@ -292,10 +296,31 @@ func buildStartupConfVolumeMount() corev1.VolumeMount {
...
+	return volMnts
+}
+
+// buildScrutinizeSharedVolumeMount returns the volume mount shared by the scrutinize
+// init container and the main container
+func buildScrutinizeSharedVolumeMount(vscr *v1beta1.VerticaScrutinize) corev1.VolumeMount {
 	return corev1.VolumeMount{
 		Name:      getScrutinizeVolumeMountName(vscr),
 		MountPath: paths.ScrutinizeTmp,
+		SubPath:   "scrutinize",
```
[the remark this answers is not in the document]
roypaulin: I don't know why I thought it was not a good idea. That's the only reason I had a mount for the log file. Will  make the changes.

pkg/builder/builder.go
roypaulin: There is a function `getScrutinizeVolumeMountName()` that returns the name of the custom volume if any or "scrutinize".

pkg/builder/builder.go
roypaulin: No that will part of a separate PR to add the logic to read this. But scrutinize already reads NMA certs from secret so I will reuse the logic.

pkg/controllers/vscr/podpolling_reconciler.go
roypaulin: Hum I will take a closer look. Mind giving an example of a file that has the "correct" copyright statement?

pkg/builder/builder.go
spilchen: Got it. Suggest we rename this to be: PASSWORD_SECRET_NAMESPACE and PASSWORD_SECRET_NAME

pkg/controllers/vscr/podpolling_reconciler.go
spilchen: pkg/controllers/vdb/actor.go

pkg/builder/builder.go
spilchen: Maybe rockylinux:8?

pkg/controllers/vscr/vdbverify_reconciler.go
```diff
@@ -93,6 +94,16 @@ func (s *VDBVerifyReconciler) checkVersionAndDeploymentType(ctx context.Context)
 			"The server version %s does not have scrutinize support through vclusterOps", ver)
 		return s.updateScrutinizeReadyCondition(ctx, metav1.ConditionFalse, events.VclusterOpsScrutinizeNotSupported)
 	}
+
+	if vinf.IsOlder(v1.ScrutinizeDBPasswdInSecretMinVersion) {
+		ver, _ := s.Vdb.GetVerticaVersionStr()
+		s.VRec.Eventf(s.Vscr, corev1.EventTypeWarning, events.VclusterOpsScrutinizePasswdInSecretNotSupported,
+			"The server version %s does not support vclusterOps scrutinize reading db password from secret", ver)
```
spilchen, line 101: Can we shorten this to say it only runs with 24.2.0?
```suggestion
			"The server version %s is not supported with VerticaScrutinize. The minimum server version it supports is %s.", ver, v1.ScrutinizeDBPasswdInSecretMinVersion)
```

pkg/controllers/vscr/vdbverify_reconciler.go
```diff
@@ -93,6 +94,16 @@ func (s *VDBVerifyReconciler) checkVersionAndDeploymentType(ctx context.Context)
 			"The server version %s does not have scrutinize support through vclusterOps", ver)
 		return s.updateScrutinizeReadyCondition(ctx, metav1.ConditionFalse, events.VclusterOpsScrutinizeNotSupported)
 	}
+
+	if vinf.IsOlder(v1.ScrutinizeDBPasswdInSecretMinVersion) {
+		ver, _ := s.Vdb.GetVerticaVersionStr()
+		s.VRec.Eventf(s.Vscr, corev1.EventTypeWarning, events.VclusterOpsScrutinizePasswdInSecretNotSupported,
```
spilchen, line 100: Suggest we reuse the event. It all means the same thing, we don't support the deployment.
```suggestion
		s.VRec.Eventf(s.Vscr, corev1.EventTypeWarning, events.VclusterOpsScrutinizeNotSupported,
```

spilchen (approved):
Looks great. I had a few minor suggestions but approving as I trust they will be applied.

PR #709, opened 2024-02-16
Create webhook for VerticaRestorePointQuery CR

chinhtranvan:


tests/e2e-leg-6/verify-vrpq-conditions/20-assert.yaml
spilchen, line 1: I'm a little confused about this rename. Are you sure this is suppose to be an assert file? It looks like we have logic in it to create a vrpq. It should only be checking what was done (or not done) in 20-invalid-startTimestamp.yaml.

spilchen (commented):
Thanks for adding this in. A few minor comments in addition to what I already added:

- There should be an update to PROJECT to indicate that we have a webhook now for the CR.
- Unrelated to this PR, but I noticed it while playing with vrpq. Can we remove `omitempty` from VerticaRestorePointsQueryStatus.RestorePoints? I created a vrpq that didn't return anything and that field wasn't there. I think it will be clear if we have it even if it is empty. That is what most people are going to look for when they create the CR.

tests/e2e-leg-6/verify-vrpq-conditions/40-create-vrpq-with-start-timestamp-filter.yaml
```diff
@@ -15,5 +15,5 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
   - script: |-
-      export STARTTIMESTAMP=$(kubectl get vrpq archive-query -o json -n $NAMESPACE | jq -r '.status.restorePoints[] | select(.index==1 and .archive=="db") | .timestamp')
+      export STARTTIMESTAMP="$(kubectl get vrpq archive-query -o json -n $NAMESPACE | jq -r '.status.restorePoints[] | select(.index==1 and .archive=="db") | .timestamp')"
```
chinhtranvan, line 18: @spilchen  There is one thing I've noticed: when I run it without double quotes on my local machine, it doesn't work. It freezes, possibly due to the timestamp containing a space. I'm unsure why it passes in the GitHub CI but not on my local machine. After discussing with @jizhuoyu , we concluded that adding double quotes would make it more persistent and reliable

tests/e2e-leg-6/verify-vrpq-conditions/20-assert.yaml
chinhtranvan: As discussed in the scrum, we should add one more step filter for vrpq

spilchen:
Can we remove omitempty from VerticaRestorePointsQueryStatus.RestorePoints? I think that will make it clear when we create the CR and no results are returned.

spilchen:
> Can we remove omitempty from VerticaRestorePointsQueryStatus.RestorePoints? I think that will make it clear when we create the CR and no results are returned.

Never mind, I noticed that it's on the server side.

chinhtranvan:
> > Can we remove omitempty from VerticaRestorePointsQueryStatus.RestorePoints? I think that will make it clear when we create the CR and no results are returned.
> 
> Never mind, I noticed that it's on the server side.

yes, I already opened a PR on the server side, you can take a look it https://git.verticacorp.com/projects/VER/repos/server/pull-requests/20274/overview

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -146,7 +147,7 @@ func (q *QueryReconciler) runShowRestorePoints(ctx context.Context, dispatcher v
 	opts []showrestorepoints.Option) (err error) {
 	// set Querying status condition ,state and restore points prior to calling vclusterops API
 	err = vrpqstatus.Update(ctx, q.VRec.Client, q.VRec.Log, q.Vrpq,
-		[]*metav1.Condition{vapi.MakeCondition(v1beta1.Querying, metav1.ConditionTrue, "Started")}, stateQuerying, nil)
```
spilchen, line 149: I'm guessing this change is needed because the restorePoints list is required now? Did you try setting the optional marker?
```
```

chinhtranvan: Oops, why didn't I consider this setting? Yes, this setting is better. We don't need to update nil to vops.RestorePoints{}, and it still returns an empty status for Restore Points in the CR

spilchen (approved):
Thanks for making the final revisions

PR #714, opened 2024-02-20
Write event when admintools is deployed with a v24.2.0 server

spilchen:
The v24.2.0 server image, when built for vclusterops, will not work if admintools deployment is selected.  The container will fail to run. This will surface an event message if we detect this is the case. This is tied into the crashloop_reconciler.


pkg/vk8s/container.go
```diff
@@ -53,6 +54,35 @@ func GetServerImage(cnts []corev1.Container) (string, error) {
...
+	return nil
+}
+
+// HasCreateContainerError returns true if the container cannot start because
+// there is no command specified. This can happen if the image doesn't have a
+// command to automatically run. This typically means the image is designed for 24.2.0+.
+func HasCreateContainerError(containerStatus *corev1.ContainerStatus) bool {
+	return containerStatus.State.Waiting != nil &&
+		containerStatus.State.Waiting.Reason == "CreateContainerError" &&
+		strings.Contains(containerStatus.State.Waiting.Message, "no command specified")
```
roypaulin, line 83: I guess there is no other reasons for `CreateContainerError`

roypaulin (commented):
Looks good!

roypaulin (approved):

PR #718, opened 2024-02-21
Set resources for scrutinize main container

roypaulin:
this give users a way to set the main container resources through annotations. The main container resources remain unset unless the resources are set for the init container too. 

The annotations are as follow:
```
vertica.com/scrutinize-main-container-resources-limits-cpu
vertica.com/scrutinize-main-container-resources-limits-memory
vertica.com/scrutinize-main-container-resources-requests-cpu
vertica.com/scrutinize-main-container-resources-requests-memory
```


pkg/vk8s/container.go
```diff
@@ -43,6 +54,15 @@ func getNamedContainer(cnts []corev1.Container, cntName string) *corev1.Containe
 	return nil
 }
 
+func getNamedContainerStatus(cntStatuses []corev1.ContainerStatus, cntName string) *corev1.ContainerStatus {
```
spilchen, line 57: A similar function `findContainerStatus`, was added in main, but it wasn't in vnext yet. I have just merged main into vnext. Can you merge up and use the new function?

spilchen (commented):
I like the code reuse in this. Just had a few minor comments.

pkg/vk8s/container.go
roypaulin: I saw that function but I would like it to be `findContainerStatus(cntStatuses []corev1.ContainerStatus, cntName string)` to have some flexibility because scrutinitize init container status belongs to `pod.Status.InitContainerStatuses` so I can't use `findContainerStatus` as it is.

spilchen: I'm okay with changing the function signature.

PR #719, opened 2024-02-27
Allow namespace scoped operator

spilchen:
This change allows for deploying the operator at the namespace scope, only for Helm deployments.  When deployed this way, the operator will watch only custom resources within its deployment namespace.  Multiple namespace deployments are possible.

By default, the webhook is deployed with the operator.  However, with multiple namespace scoped deployments in a cluster, the webhook can only be deployed once.

There are two ways to handle this:
- Set `webhook.enable` to false for any subsequent operator deployment
- Deploy only the webhook (`controllers.enable` set to false) and deploy each operator with the webhook disabled (`webhook.enable` set to false)

A new Helm chart parameter, `controllers.scope`, has been added to control the scope of the operator, with valid values of "cluster" or "namespace" and a default of "cluster."

The e2e-leg-2 has been converted to deploy the operator with namespace scope.

All command-line options used by the operator have been replaced with environment variables read from a ConfigMap.  This change was made when adding the namespace scope and controllers disable options, as this method is more extensible.

This change removes most of the logging Helm chart parameters, leaving only `logging.level`.  Those using the logging Helm chart parameters can set appropriate entries in the ConfigMap to retain the old behavior.

Finally, the 'make run' option has been removed.  The operator can now only be run by running a container within Kubernetes, as it was too difficult to set up a ConfigMap and environment variables to run the operator.


pkg/opcfg/config.go
```diff
@@ -28,93 +29,236 @@ import (
...
+// GetIsControllersEnabled returns true if the controllers for each custom
+// resource will start. If this is false, then the manager will just act as a
+// webhook (if enabled).
+func GetIsControllersEnabled() bool {
+	return lookupBoolEnvVar("CONTROLLERS_ENABLED", envMustExist)
+}
+
+// GetAreControllersNamespaceScoped returns true if the controllers only watch a
+// single namespace.
+func GetAreControllersNamespaceScoped() bool {
```
roypaulin, line 46: This function's name sounds strange. Why not just `AreControllersNamespaceScoped`?

pkg/opcfg/config.go
```diff
@@ -28,93 +29,236 @@ import (
...
-	DefaultMaxFileAge                     = 7
-	DefaultMaxFileRotation                = 3
-	DefaultLevel                          = "info"
-	DefaultDevMode                        = true
-	DefaultVerticaDBConcurrency           = 5
-	DefaultVerticaAutoscalerDBConcurrency = 1
-	DefaultEventTriggerDBConcurrency      = 1
+	// Helper consts for the mustExist parameter for the lookup*EnvVar functions.
+	envMustExist   = true
+	envCanNotExist = false
```
roypaulin, line 206: Can you add a comment for each of these consts?

pkg/opcfg/config.go
```diff
@@ -28,93 +29,236 @@ import (
...
+}
+
+// GetLoggingLevel returns the logging level to use. Logging levels are: debug,
+// info, warn, error.
+func getLoggingLevel() string {
+	return lookupStringEnvVar("LOG_LEVEL", envMustExist)
+}
+
+// GetIsDebugLoggingEnabled returns true if the debug logging level is selected.
+func GetIsDebugLoggingEnabled() bool {
```
roypaulin, line 119: Why do you add the `Get` prefix?

pkg/opcfg/config.go
```diff
@@ -28,93 +29,236 @@ import (
...
+	return lookupStringEnvVar("WEBHOOK_CERT_SECRET", envMustExist)
+}
+
+// GetOperatorNamespace retrieves the namespace that the operator is running in
+func GetOperatorNamespace() string {
+	return lookupStringEnvVar("OPERATOR_NAMESPACE", envMustExist)
+}
+
+// GetLeaderElectionID returns the name to use for leader election. This ensures
+// that the operator can only once in a namespace.
```
roypaulin, line 178: I think there is a missing word in this sentence.

pkg/opcfg/config.go
```diff
@@ -28,93 +29,236 @@ import (
...
+	if !found {
+		if mustExist {
+			dieIfNotFound(envName)
+		}
+		return ""
+	}
+	return valStr
+}
+
+// lookupBoolEnvVar will look for an environment variable and return its value
```
roypaulin, line 240:
```suggestion
// lookupIntEnvVar will look for an environment variable and return its value
```

pkg/opcfg/config.go
```diff
@@ -28,93 +29,236 @@ import (
...
+// as if it's a boolean. Any errors will stop the manager.
```
roypaulin, line 241: I think the comment too must be updated.

pkg/opcfg/config.go
```diff
@@ -124,49 +268,52 @@ func (o *OperatorConfig) getEncoderConfig(devMode bool) zapcore.EncoderConfig {
...
+		MaxAge:     getLoggingMaxFileAge(), // days
 	}
 	return zapcore.AddSync(lumberJackLogger)
 }
 
-// getZapcoreLevel takes the level as string and returns the corresponding
-// zapcore.Level. If the string level is invalid, it returns the default
-// level
-func (o *OperatorConfig) getZapcoreLevel() zapcore.Level {
+// getZapcoreLevel return the logging level to use for the logging. Levels are
```
roypaulin, line 309: The end of the comment is missing.

roypaulin (commented):
First round. I am still looking at other files.

pkg/opcfg/config.go
spilchen: Good point. I think since all of the other exported functions in this file start with `Get`. I kept it up. But this one makes sense to leave off though.

pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -73,8 +71,6 @@ type VerticaDBReconciler struct {
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=persistentvolumeclaims,verbs=get;list;watch;update
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
-// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations,verbs=get;list;watch;update;patch
-// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;list;watch;update;patch
```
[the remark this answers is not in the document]
spilchen: Good question. I created config/rbac/webhook_config_clusterrole.yaml and config/rbac/webhook_config_clusterrolebinding.yaml with these permissions. I did this because its easier to template. We don't want these when not deploying the operator. So, its easy to add this templating in if they are in a separate file.

pkg/vadmin/at.go
```diff
@@ -45,11 +46,11 @@ func (a *Admintools) execAdmintools(ctx context.Context, initiatorPod types.Name
 	// Dump relevant contents of the admintools.conf before and after the
 	// admintools calls. We do this for PD purposes to see what changes occurred
 	// in the file.
-	if a.DevMode {
+	if opcfg.GetIsDebugLoggingEnabled() {
```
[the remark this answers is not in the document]
spilchen: We used devMode as a proxy for extra debugging. I think its clearer if we use the log level. I should have done this initially.

scripts/template-helm-chart.sh
```diff
@@ -203,4 +202,27 @@ cat << EOF >> $TEMPLATE_DIR/verticadb-operator-manager-role-cr.yaml
   - list
   - patch
   - update
+{{- end }}
 EOF
+
+# 21. Change change ClusterRoles/ClusterRoleBindings for the manager to be
+# Roles/RoleBindings if the operator is scoped to a single namespace.
+for f in $TEMPLATE_DIR/verticadb-operator-manager-clusterrolebinding-crb.yaml \
```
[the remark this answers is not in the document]
spilchen: There are three places:
- in helm-charts/verticadb-operator/templates/_helpers.tpl, we check to know if a ClusterRole/ClusterRoleBinding should really be a Role/RoleBinding. See the "vdb-op.roleKind" and "vdb-op.roleBindingKing" defines.
- in PodSecurityReconciler, with a namespace scoped operator we don't have permissions to read the annotations of a namespace anymore.
- in pkg/opcfg/config, we use the scope to know what namespace the operator needs to watch

roypaulin (approved):

PR #720, opened 2024-02-27
Resurface scrutinize tarball name

roypaulin:
This caches  the tarballName in our VSCR status after a successful scrutinize run. The generated tarball name follow a  specific format needed by `grasp`. To access the tarball name between reconcilers, we cache it in the scrutinize pod annotations.


roypaulin:
e2e expected to fail until the image is updated.

pkg/controllers/vscr/podpolling_reconciler.go
```diff
@@ -91,8 +92,13 @@ func (p *PodPollingReconciler) checkScrutinizeContainerStatus(ctx context.Contex
 	}
 	p.VRec.Eventf(p.Vscr, corev1.EventTypeNormal, events.VclusterOpsScrutinizeSucceeded,
 		"Successfully completed scrutinize run for the VerticaDB named '%s'", p.Vscr.Spec.VerticaDBName)
+	stat := &v1beta1.VerticaScrutinizeStatus{}
+	stat.PodName = p.Vscr.Status.PodName
+	stat.PodUID = p.Vscr.Status.PodUID
+	stat.TarballName = pod.Annotations[vmeta.ScrutinizeTarballName]
```
spilchen, line 98: Do we need the annotation? Can we just walk the pods arguments and find the tarball name that was passed in.

pkg/controllers/vscr/scrutinizepod_reconciler.go
```diff
@@ -144,6 +147,16 @@ func (s *ScrutinizeCmdArgs) buildScrutinizeCmdArgs() []string {
...
 		"--honor-user-input",
 	}
 }
+
+// generateScrutinizeID returns a string, with the format VerticaScrutinize.yyyymmddhhmmss,
+// that will be used as the name of the tarball generated by scrutinize
+func generateScrutinizeID() string {
+	const idPrefix = "VerticaScrutinize."
+	const timeFmt = "20060102150405" // using fixed reference time from pkg 'time'
+	idSuffix := time.Now().Format(timeFmt)
```
spilchen, line 160: I think `time.Now()` returns the local time. Should we use UTC? Is that was scrutinize does?
```suggestion
	idSuffix := time.Now().UTC().Format(timeFmt)
```

spilchen (commented):
Looks good. I had one suggestion that would prevent us from adding the tarballName parameter everywhere.

pkg/controllers/vscr/podpolling_reconciler.go
roypaulin: I thought about it. I just thought that the annotation was a cheap way to get the tarballName, but you are right looking up the arguments is simple.

pkg/controllers/vscr/scrutinizepod_reconciler.go
roypaulin: Yes this is the same function used in scrutinize.

roypaulin: This is just a copy of the function in vclusterops scrutinize.

tests/e2e-leg-6/collect-scrutinize/25-check-tarball.yaml
```diff
@@ -14,4 +14,6 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
-  - command: kubectl exec vertica-scrutinize-vdb-exist -n $NAMESPACE -c main -- sh -c "ls | grep -q VerticaScrutinize.*.tar"
\ No newline at end of file
+  - script: |-
+      tarball=$(kubectl -n $NAMESPACE get vscr vertica-scrutinize-vdb-exist -o json | jq '.status.tarballName' | grep -Eq '^VerticaScrutinize\.[0-9]{14}$')
+      kubectl exec vertica-scrutinize-vdb-exist -n $NAMESPACE -c main -- ls | grep -q $tarball\.tar
```
spilchen, line 19: I kind of expected that the file name we put in the status is the full file name. Do you think it's a bit odd that we have to append .tar to get the actual file name? It's not always convention that a tarball ends with .tar. If it's also gziped they end with .tar.gz.

pkg/controllers/vscr/scrutinizepod_reconciler.go
spilchen: I'm surprised by this. It could prevent ordering of these files if they were generated on separate machines.

tests/e2e-leg-6/collect-scrutinize/25-check-tarball.yaml
roypaulin: I did not want any dependencies on the extension. We can append .tar assuming we will stick with that in vcluster.

pkg/controllers/vscr/scrutinizepod_reconciler.go
spilchen: I spoke to Chris about this. He wasn't too worried either. No change needed here.

tests/e2e-leg-6/collect-scrutinize/25-check-tarball.yaml
spilchen: Yeah, maybe we just need to append '.tar' in `getTarballName`

spilchen (approved):

roypaulin:
I removed  `--honor-user-input` so this needs Cai's changes to pass.
Also I updated some files in the e2e tests related to scrutinize I will create a ticket to track them.

spilchen:
> I removed `--honor-user-input` so this needs Cai's changes to pass. 

This was in last nights build. So, we should be able to build the image and test that out. Did you have push access to vertica-k8s-private? If not, I can add one.

roypaulin:
> > I removed `--honor-user-input` so this needs Cai's changes to pass.
> 
> This was in last nights build. So, we should be able to build the image and test that out. Did you have push access to vertica-k8s-private? If not, I can add one.

I have push access, but can you add one just this time?

spilchen:
The new images should be up later tonight.

spilchen:
They are up now.

PR #722, opened 2024-02-28
Add experimental support to run the operator on arm64

spilchen:
The operator was initially designed exclusively for the amd64 platform. This modification aims to extend its compatibility to the arm64 platform. It's primarily intended for experimental purposes at this stage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -400,20 +400,24 @@ ifeq ($(shell $(KIND_CHECK)), 1)
 endif
 endif
 
-# PLATFORMS defines the target platforms for  the manager image be build to provide support to multiple
-# architectures. (i.e. make docker-buildx-operator OPERATOR_IMG=myregistry/mypoperator:0.0.1). To use this option you need to:
-# - able to use docker buildx . More info: https://docs.docker.com/build/buildx/
-# - have enable BuildKit, More info: https://docs.docker.com/develop/develop-images/build_enhancements/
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -400,20 +400,24 @@ ifeq ($(shell $(KIND_CHECK)), 1)
...
-# - able to use docker buildx . More info: https://docs.docker.com/build/buildx/
-# - have enable BuildKit, More info: https://docs.docker.com/develop/develop-images/build_enhancements/
-# - be able to push the image for your registry (i.e. if you do not inform a valid value via OPERATOR_IMG=<myregistry/image:<tag>> than the export will fail)
-# To properly provided solutions that supports more than one platform you should use this option.
-PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
-.PHONY: docker-buildx-operator
-docker-buildx-operator: test ## Build and push docker image for the manager for cross-platform support
+# PLATFORMS defines the target platforms that the image will be used for. Use
+# this with docker-build-crossplatform-* targets.
+PLATFORMS?=linux/arm64,linux/amd64
```
roypaulin, line 405: What exactly changed? other than the new target docker-build-crossplatform-operator. Did we already support running the operator on arm64?

spilchen: We always had the make target. We just never used it. What changed in this PR is that we'll use this make target to build the operator by default in the GitHub CI. We won't have any GitHub runners running on arm, so the arm aspect of the image won't be used. But it will make it easier to try out arm in the future (hopefully within the next couple of months)

roypaulin (approved):

PR #727, opened 2024-03-04
Fix e2e test for scrutinize

roypaulin:
This is to allow scrutinize to run successfully on the main branch.


roypaulin:
Let's use this PR to make sure scrutinize in successfully on the main branch with the latest image. I removed `--honor-user-input` which is not supported anymore.

spilchen (approved):

PR #729, opened 2024-03-05
Surface podName and state when running 'kubectl get vscr'

roypaulin:
Now when you have a VerticaScrutinize CR created, if you do kubectl get, you will get info about the scrutinize pod name and the state of the scrutinize run. The state is `Ready`, `PodCreated`, `ScrutinizeInProgress`, `ScrutinizeSucceeded` if successful and `NotReady:<reason>`, `PodCreationFailed`, `ScrutinizeFailed` in case of a failed run.

kubectl get vscr will look like this:
```
NAME                       STATE                 POD                        AGE
verticascrutinize-sample   ScrutinizeSucceeded   verticascrutinize-sample   84m
```


pkg/vscrstatus/status.go
```diff
@@ -92,3 +95,17 @@ func UpdateCondition(ctx context.Context, clnt client.Client, vscr *v1beta1.Vert
 
 	return update(ctx, clnt, vscr, refreshConditionInPlace)
 }
+
+// UpdateState will update the state field in vscr status
+func UpdateState(ctx context.Context, clnt client.Client, vscr *v1beta1.VerticaScrutinize, state string) error {
+	// refreshState will update the state field in vscr.  The update
+	// will be applied in-place.
+	refreshState := func(vscr *v1beta1.VerticaScrutinize) error {
```
spilchen, line 103: I noticed that none of the callback functions we pass to `update` return an error. They all return `nil`. Could we change the callback function signature to not return an error?

pkg/vscrstatus/status.go
```diff
@@ -70,6 +70,9 @@ func UpdateStatus(ctx context.Context, clnt client.Client, vscr *v1beta1.Vertica
 		if vscr.Status.TarballName != vscrChgStatus.TarballName {
 			vscr.Status.TarballName = vscrChgStatus.TarballName
 		}
+		if vscr.Status.State != vscrChgStatus.State {
```
spilchen, line 73: Any advantage in check the prior state? Would it be simple to just remove the if checks totally.

pkg/controllers/vscr/podpolling_reconciler.go
```diff
@@ -73,16 +73,21 @@ func (p *PodPollingReconciler) checkScrutinizeContainerStatus(ctx context.Contex
 	if cntStatus == nil {
 		return ctrl.Result{}, fmt.Errorf("could not find scrutinize container status")
 	}
+	stat := &v1beta1.VerticaScrutinizeStatus{}
+	stat.PodName = p.Vscr.Status.PodName
+	stat.PodUID = p.Vscr.Status.PodUID
```
spilchen, line 78: Can we just do a copy here? Then we don't need to enumerate each field.
```suggestion
	stat := p.Vscr.Status.DeepCopy()
```

api/v1beta1/verticascrutinize_types.go
```diff
@@ -103,6 +103,10 @@ type VerticaScrutinizeStatus struct {
 	// The name of the scrutinize tarball that was created.
 	TarballName string `json:"tarballName"`
 
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// Status message for scrutinize
+	State string `json:"state,omitempty"`
```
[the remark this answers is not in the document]
roypaulin: Yes, I forgot.

pkg/vscrstatus/status.go
roypaulin: Yes just a bad habit. I will remove the checks.

spilchen (approved):
Looks good

PR #731, opened 2024-03-06
Allow empty password for vcluster scrutinize

roypaulin:
When the db doesn't have a password, we append `password=` to scrutinize cmd.


pkg/builder/builder.go
```diff
@@ -1467,12 +1467,18 @@ func buildNMACommand() []string {
...
 	cmd := []string{
 		"/opt/vertica/bin/vcluster",
 		"scrutinize",
 	}
 	cmd = append(cmd, args...)
+	// if there is no password, we need to explicitly
+	// set the password flag with empty string as value,
+	// to still assume password as the authentication method
+	if vdb.Spec.PasswordSecret == "" {
+		cmd = append(cmd, "--password=")
```
spilchen, line 1480: I think we should move this to buildScrutinizeCmdArgs. It makes more sense to keep the arg building all in one place.

spilchen (approved):

PR #733, opened 2024-03-12
Fix  Staging Directory does not exist error

roypaulin:
We were getting `Staging Directory does not exist or have write permission:` because the https service could not access the staging dir during [HTTPSStageSystemTablesOp], `/tmp/scrutinize` was not mounted to the server container. 
This fixes that issue by simply mounting `/tmp/scrutinize` to the server container


spilchen (approved):
Looks good. I touched up the PR description a bit. Let me know if you agree with the changes.

roypaulin:
Looks good!

roypaulin:
I am going to open another PR today for the new env var.

PR #734, opened 2024-03-13
Make it easier to integrate custom containers with VerticaScrutinize

roypaulin:
Here are a few changes to make it easier to integrate custom init containers into the VerticaScrutinize CR:
- Include an environment variable, `SCRUTINIZE_TARBALL`, containing the full path of where the tarball resides.
- Mount the volume containing the tarball so that the containers have access to it.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -853,8 +853,8 @@ func makeScrutinizeInitContainer(vscr *v1beta1.VerticaScrutinize, vdb *vapi.Vert
 		Resources:    vscr.Spec.Resources,
 		Env:          buildCommonEnvVars(vdb),
 	}
-	cnt.Env = append(cnt.Env, buildNMATLSCertsEnvVars(vdb)...)
-	cnt.Env = append(cnt.Env, buildScrutinizeTarballEnvVar(tarballName))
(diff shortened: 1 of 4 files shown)


pkg/builder/builder.go
```diff
@@ -845,16 +854,17 @@ func makeScrutinizeInitContainer(vscr *v1beta1.VerticaScrutinize, vdb *vapi.Vert
 		Env:          buildCommonEnvVars(vdb),
 	}
 	cnt.Env = append(cnt.Env, buildNMATLSCertsEnvVars(vdb)...)
+	cnt.Env = append(cnt.Env, buildScrutinizeTarballEnvVar(tarballName))
```
spilchen, line 857: minor: could we rewrite this to combine into a single append?
```suggestion
	cnt.Env = append(cnt.Env, buildScrutinizeTarballEnvVar(tarballName),
	     buildNMATLSCertsEnvVars(vdb)...)
```

spilchen (approved):
look good, just had one minor comment

pkg/builder/builder.go
roypaulin: This syntax is not correct. How about:
```
cnt.Env = append(cnt.Env, append(buildNMATLSCertsEnvVars(vdb),
		buildScrutinizeTarballEnvVar(tarballName))...)
```

PR #735, opened 2024-03-14
Ensure proper service account for webhook rolebindings

spilchen:
Fixed an issue where deploying the Helm chart with a custom service account caused the operator to fail. This update ensures the correct service account is used for setting up the cluster role binding in the webhook config.

### DIFF
--- a/tests/e2e-leg-5/metrics-auth-proxy-cert/05-deploy-operator.yaml
+++ b/tests/e2e-leg-5/metrics-auth-proxy-cert/05-deploy-operator.yaml
@@ -15,4 +15,4 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
   - command: sh -c "cd ../../.. && make undeploy-operator || true"
-  - command: sh -c "cd ../../.. && make deploy-operator DEPLOY_WITH=helm NAMESPACE=$NAMESPACE HELM_OVERRIDES='--set prometheus.tlsSecret=custom-cert,prometheus.expose=EnableWithAuthProxy'"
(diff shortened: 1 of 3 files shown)


tests/e2e-leg-5/metrics-auth-proxy-cert/05-deploy-operator.yaml
```diff
@@ -15,4 +15,4 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
   - command: sh -c "cd ../../.. && make undeploy-operator || true"
-  - command: sh -c "cd ../../.. && make deploy-operator DEPLOY_WITH=helm NAMESPACE=$NAMESPACE HELM_OVERRIDES='--set prometheus.tlsSecret=custom-cert,prometheus.expose=EnableWithAuthProxy'"
+  - command: sh -c "cd ../../.. && make deploy-operator DEPLOY_WITH=helm NAMESPACE=$NAMESPACE HELM_OVERRIDES='--set prometheus.tlsSecret=custom-cert,prometheus.expose=EnableWithAuthProxy,serviceAccountNameOverride=special-sa'"
```
roypaulin, line 18: At some point in this test are we checking that `special-sa` exists?

spilchen: No, I don't check that. The operator will crash if the webhook ClusterRoleBinding is setup correctly. That was my thought process here. But I think checking specifically for the sa makes sense.

roypaulin (approved):

PR #739, opened 2024-03-20
Resolved Service Annotation Loop Issue in ROSA Environment

spilchen:
In ROSA (RedHat OpenShift on AWS), we noticed that when setting up a network LoadBalancer an annotation would automatically be added to the Service object. This caused the operator to start a reconcile loop. It would remove the annotation, only to have OpenShift add it back. So, the operator was in a continuous reconcile loop.

To fix this we now allow manual annotations be added to service objects. The operator will only ensure the annotations that it generates are the correct value. It will ignore any additional annotation that was added outside of the VerticaDB.

I am also cleaning up the PVC expansion events. This can cause quite a lot of noise about skipping expansion if we continuously are doing reconciles. The skip events have been changed to log entries instead.

### DIFF
--- a/pkg/controllers/vdb/obj_reconciler.go
+++ b/pkg/controllers/vdb/obj_reconciler.go
@@ -294,11 +294,11 @@ func (o *ObjReconciler) reconcileSvc(ctx context.Context, expSvc *corev1.Service
 		return o.createService(ctx, expSvc, svcName)
 	}
 
-	// We reconcile annotations to include both new and existing entries. If an
-	// annotation is removed from the spec.subclusters[].serviceAnnotations
-	// field, it cannot be automatically removed due to potential ambiguity in
-	// how it was initially added. Certain cloud platforms may employ webhooks
-	// to automatically append their annotations.
+	// Annotations are always additive. We never remove an annotation if it's
+	// not in expSvc. Since we don't know how an annotation was added, we can't
+	// guess if it should be removed. Platforms like OpenShift may add
+	// annotations via a webhook, so removing them could lead to them being
+	// added back automatically.
 	for k, v := range curSvc.Annotations {
 		if _, ok := expSvc.Annotations[k]; !ok {
 			expSvc.Annotations[k] = v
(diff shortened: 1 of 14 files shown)


pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -294,6 +294,17 @@ func (o *ObjReconciler) reconcileSvc(ctx context.Context, expSvc *corev1.Service
 		return o.createService(ctx, expSvc, svcName)
 	}
 
+	// We reconcile annotations to include both new and existing entries. If an
```
roypaulin, line 297: Any chance to make this comment less complex?

tests/e2e-leg-4/pvc-expansion/verify-pvc-change/volume-expansion-disabled/assert.yaml
```diff
@@ -11,12 +11,4 @@
...
-apiVersion: v1
-kind: Event
-reason: SkipPVCExpansion
-source:
-  component: verticadb-operator
-involvedObject:
-  apiVersion: vertica.com/v1
-  kind: VerticaDB
-  name: v-pvc-expansion
+# No event to check if expansion is skipped
```
roypaulin, line 14: Why do you still keep this file?

roypaulin (approved):
Looks good!

roypaulin:
Is this related to the crash from yesterday?

spilchen:
> Is this related to the crash from yesterday?

Yes, we were repeatedly doing reconcile iterations. Eventually, the K8s OOMKiller stepped in and killed the pod. We were allocating/freeing too much memory that the garbage collection couldn't keep up.

tests/e2e-leg-4/pvc-expansion/verify-pvc-change/volume-expansion-disabled/assert.yaml
spilchen: If we run with a PVC that allows for expansion, then the step can still be used to verify that.

PR #742, opened 2024-03-21
Change VerticaScrutinize to read password from file

roypaulin:
Previously, we set the password secret environment variables to allow scrutinize, when invoked via the VerticaScrutinize CRD, to read the password from the secret. Now, when the password secret is in Kubernetes, we mount it into the scrutinize init container and configure the scrutinize command to read the password from a file. This eliminates the need to create a role granting read access to the secret when the secret is in Kubernetes.


pkg/builder/builder.go
```diff
@@ -478,6 +479,16 @@ func buildScrutinizeVolumes(vscr *v1beta1.VerticaScrutinize, vdb *vapi.VerticaDB
 		secrets.IsK8sSecret(vdb.Spec.NMATLSSecret) {
 		vols = append(vols, buildNMACertsSecretVolume(vdb))
 	}
+	// we add a volume for the password when the password secret
+	// is on k8s
+	if vdb.Spec.PasswordSecret != "" &&
+		secrets.IsK8sSecret(vdb.Spec.PasswordSecret) {
+		projSources := []corev1.VolumeProjection{
+			// If these is a superuser password, include that in the projection
```
spilchen, line 487:
```suggestion
			// If there is a superuser password, include that in the projection
```

pkg/builder/builder.go
```diff
@@ -507,11 +518,16 @@ func buildPodInfoVolume(vdb *vapi.VerticaDB) corev1.Volume {
 	projSources := []corev1.VolumeProjection{
 		{DownwardAPI: buildDownwardAPIProjection()},
 		// If these is a superuser password, include that in the projection
```
spilchen, line 520:
```suggestion
		// If there is a superuser password, include that in the projection
```

pkg/builder/builder.go
```diff
@@ -862,12 +884,24 @@ func makeScrutinizeInitContainer(vscr *v1beta1.VerticaScrutinize, vdb *vapi.Vert
...
-			names.GenNamespacedName(vscr, vdb.Spec.PasswordSecret))...)
+		if secrets.IsK8sSecret(vdb.Spec.PasswordSecret) {
+			// we mount the password into the scrutinize init container
+			// only when the password secret in on k8s
+			cnt.VolumeMounts = append(cnt.VolumeMounts, corev1.VolumeMount{
+				Name:      scrutinizeDBpasswordMountName,
+				MountPath: paths.ScrutinizeDBpasswordPath,
+			})
+		} else {
+			// the password secret env vars are needed only when the secret
```
spilchen, line 896: We should make sure that QA verifies this part that we can still run scrutinize if the dbadmin password is stored in AWS or GSM.

pkg/builder/builder.go
```diff
@@ -478,6 +479,16 @@ func buildScrutinizeVolumes(vscr *v1beta1.VerticaScrutinize, vdb *vapi.VerticaDB
...
 	}
+	// we add a volume for the password when the password secret
+	// is on k8s
+	if vdb.Spec.PasswordSecret != "" &&
+		secrets.IsK8sSecret(vdb.Spec.PasswordSecret) {
+		projSources := []corev1.VolumeProjection{
+			// If these is a superuser password, include that in the projection
+			{Secret: buildPasswordProjection(vdb)},
+		}
+		vols = append(vols, buildVolumeFromProjections(projSources, scrutinizeDBpasswordMountName))
```
spilchen, line 490: Does the password need to be stored in a project volume? It would be simple if its just a regular volume.

pkg/paths/paths.go
```diff
@@ -40,6 +40,7 @@ const (
 	LogPath                   = "/opt/vertica/log"
 	ScrutinizeTmp             = "/tmp/scrutinize"
 	ScrutinizeLogFile         = "/tmp/scrutinize/vcluster.log"
+	ScrutinizeDBpasswordPath  = "/etc/password"
```
spilchen, line 43: minor:
```suggestion
	ScrutinizeDBPasswordPath  = "/etc/password"
```

pkg/builder/builder.go
```diff
@@ -1594,3 +1628,7 @@ func GetTarballName(cmd []string) string {
 	}
 	return ""
 }
+
+func GetScrutinizeDBPasswordFullPath() string {
```
spilchen, line 1632: I think this should be a const in the `paths` package.

spilchen (commented):
Looks good. I'm glad we are able to get this in for the 24.2.0 cutoff.

spilchen (approved):

PR #746, opened 2024-03-25
Operator now defaults endpoints for S3/GCS paths.

spilchen:
Previously, the operator required specifying an endpoint for S3 and GCS communal paths. This pull request modifies that behavior, allowing the endpoint to be left off. In such cases, the server will automatically select a suitable default. For S3 communal paths, the server defaults to https://s3.amazonaws.com, while for GCS endpoints, it defaults to https://storage.googleapis.com.

### DIFF
--- a/pkg/vdbconfig/config.go
+++ b/pkg/vdbconfig/config.go
@@ -206,10 +206,14 @@ func (g *ConfigParamsGenerator) setS3AuthParms(ctx context.Context) (ctrl.Result
 		}
 	}
 
+	// If the endpoint isn't provided, we let the server pick an appropriate
+	// default by omitting that config parm.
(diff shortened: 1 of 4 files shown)


pkg/vdbconfig/config.go
```diff
@@ -448,6 +460,10 @@ func (g *ConfigParamsGenerator) GetCommunalEndpoint() string {
 
 // getEnableHTTPS will return "1" if connecting to https otherwise return "0"
 func (g *ConfigParamsGenerator) GetEnableHTTPS() string {
+	// If no endpoint, then we let the server pick the endpoint and decide if HTTPS is enabled.
```
roypaulin, line 463: Maybe you can do the same in `GetCommunalEndpoint()`

PR #748, opened 2024-03-26
Assign subclusters to replica groups for replicated upgrade

spilchen:
During a replicated upgrade, we must categorize the subclusters into one of two replica groups. Initially, all subclusters participating in the upgrade will be placed in replica group A. Subsequently, new subclusters will be generated throughout the process and assigned to the second group, named replica group B. This change manages the assignment of subclusters at the beginning of the upgrade process. To facilitate this change, new status fields will be introduced to the VerticaDB, enabling the tracking of this state across reconcile iterations.


roypaulin (approved):
Looks good!

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -51,6 +55,14 @@ func MakeReplicatedUpgradeReconciler(vdbrecon *VerticaDBReconciler, log logr.Log
 	}
 }
 
+const (
+	// Entries into the status.upgradeState.replicas array for each replica.
+	// Replica A has all of the primary subclusters and some secondary
+	// subclusters. Replica B has only secondary subclusters.
+	replicaA = 0
```
roypaulin, line 62: Doesn't replicaGroup/replicaSet make more sense? My assumption is that replica=pod/node

spilchen: This is good feedback. I think I'll use replicaGroup. ReplicaSet is term used in k8s already, so may be confusing.

PR #752, opened 2024-04-01
Clear filters in VerticaRestorePointsQuery sample

chinhtranvan:
This PR eliminates filter options and labels from the VerticaRestorePointsQuery sample. The intention is to enhance the user experience on OpenShift when generating this CR from the webUI. OpenShift utilizes this sample to prepopulate fields in the webUI interface.


jizhuoyu:
@spilchen do I also need to remove labels in vrep sample?

spilchen (approved):

PR #754, opened 2024-04-02
Remove labels in `et` sample config

jizhuoyu:
This PR eliminates labels from the `EventTrigger` sample config. The intention is to enhance the user experience on OpenShift when generating this CR from the webUI. OpenShift utilizes this sample to prepopulate fields in the webUI interface.

### DIFF
--- a/config/samples/v1beta1_eventtrigger.yaml
+++ b/config/samples/v1beta1_eventtrigger.yaml
@@ -1,12 +1,6 @@
 apiVersion: vertica.com/v1beta1
 kind: EventTrigger
 metadata:
-  labels:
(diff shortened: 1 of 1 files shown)


config/samples/v1beta1_eventtrigger.yaml
```diff
@@ -1,12 +1,6 @@
 apiVersion: vertica.com/v1beta1
 kind: EventTrigger
 metadata:
-  labels:
```
roypaulin, line 4: Should we remove these labels from other samples?

jizhuoyu: There are no labels left in other sample configs

PR #755, opened 2024-04-02
Add webhook rules for replicated upgrade

spilchen:
This adds a few webhook rules for replicated upgrade. It's probably not a complete list, but we can always add more later when we find rules to check. This commit adds the following:
- ensures the subclusters involved in the upgrade are stable. They cannot be removed and the sizes can't change. 
- We do allow new subclusters to be added to one of the replicae groups. But they must be secondary subclusters. No new primaries.
- ensures that subclusters listed in the `.status.updateState.replicaGroups` aren't repeated.

### DIFF
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -140,6 +140,7 @@ func (v *VerticaDB) validateImmutableFields(old runtime.Object) field.ErrorList 
 	allErrs = v.checkImmutableS3ServerSideEncryption(oldObj, allErrs)
 	allErrs = v.checkImmutableDepotVolume(oldObj, allErrs)
 	allErrs = v.checkImmutablePodSecurityContext(oldObj, allErrs)
+	allErrs = v.checkImmutableSubclusterDuringUpgrade(oldObj, allErrs)
 	return allErrs
 }
 
@@ -176,6 +177,7 @@ func (v *VerticaDB) validateVerticaDBSpec() field.ErrorList {
 	allErrs = v.hasValidNMAResourceLimit(allErrs)
 	allErrs = v.hasValidCreateDBTimeout(allErrs)
 	allErrs = v.hasValidUpgradePolicy(allErrs)
+	allErrs = v.hasValidReplicaGroups(allErrs)
 	if len(allErrs) == 0 {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


api/v1/verticadb_webhook.go
```diff
@@ -1261,6 +1285,57 @@ func checkInt64PtrChange(prefix *field.Path, fieldName string,
...
 
+// checkImmutableSubclusterDuringUpgrade will ensure we don't scale, add or
+// remove subclusters during a replicated upgrade.
+func (v *VerticaDB) checkImmutableSubclusterDuringUpgrade(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {
+	// This entire check can be skipped if we aren't doing replicated upgrade.
+	if oldObj.Status.UpgradeState == nil || len(oldObj.Status.UpgradeState.ReplicaGroups) == 0 {
+		return allErrs
+	}
+
+	// The subclusters defined in status.upgradeState.replicaGroups must stay
```
roypaulin, line 1296: I am not sure I understand this comment.

api/v1/verticadb_webhook.go
```diff
@@ -1261,6 +1285,57 @@ func checkInt64PtrChange(prefix *field.Path, fieldName string,
...
+			if !newScFound {
+				err := field.Invalid(path,
+					v.Spec.Subclusters,
+					fmt.Sprintf("Cannot remove subcluster %q while database is being upgraded", scName))
+				allErrs = append(allErrs, err)
+				continue
+			}
+			// The operator can create new subclusters that are used for the
+			// upgrade process. These must be secondary subclusters.
+			if !oldScFound {
```
roypaulin, line 1317: I am confused by this oldScFound/newScFound checks. How can we check them both  together?

api/v1/verticadb_webhook.go
spilchen: The subclusters, whose names are stored in replicaGroups, must stay static throughout the upgrade. The size of the them cannot change. We cannot remove them. But, new subclusters may be added as this is something the upgrade process will do. So, we have to allow for that.

api/v1/verticadb_webhook.go
spilchen: oldScFound = true, newScFound = false => this can happen if a subcluster is attempted to be removed
oldScFound = false, newScFound = true => this can happen if a new subcluster is added. We allow this as long as the subcluster is a secondary.

roypaulin (approved):

PR #761, opened 2024-04-04
Fix e2e test stability

spilchen:
This is a continuation of issue #750. In summary, we're encountering intermittent test failures. Upon deleting the pod in admintools deployments, we observed that the Vertica process shows up in the `ps` output. A new pod should not have Vertica already running because the operator should be triggering that. Although this doesn't impact the operator's functionality, it disrupts end-to-end tests for upgrades, as they rely on the operator's sequential actions. 

I encountered another instance of this issue while investigating with the debug code introduced in #750. It's evident from the `ps` output that the Vertica process belongs to a previous instance of the pod. One clear indication is that the host IP provided to the Vertica process doesn't match the pod's IP. Therefore, I've implemented this check to address the test failure.

### DIFF
--- a/pkg/controllers/vdb/podfacts.go
+++ b/pkg/controllers/vdb/podfacts.go
@@ -200,8 +200,6 @@ type GatherState struct {
 	DBExists               bool            `json:"dbExists"`
 	VerticaPIDRunning      bool            `json:"verticaPIDRunning"`
 	VerticaProcess         string          `json:"verticaProcess"`
-	SpreadPIDRunning       bool            `json:"spreadPIDRunning"`
-	SpreadProcess          string          `json:"spreadProcess"`
 	UpNode                 bool            `json:"upNode"`
 	StartupComplete        bool            `json:"startupComplete"`
 	Compat21NodeName       string          `json:"compat21NodeName"`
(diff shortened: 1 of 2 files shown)


pkg/controllers/vdb/podfacts.go
```diff
@@ -512,20 +510,24 @@ func (p *PodFacts) genGatherScriptBase(vdb *vapi.VerticaDB, pf *PodFact) string
...
 }
 
-func (p *PodFacts) genGatherScriptForVerticaPIDCollection() string {
+func (p *PodFacts) genGatherScriptForVerticaPIDCollection(pf *PodFact) string {
 	// This is only be run when vertica process is running in the container we
 	// are doing the gather. And this is only true if we aren't running the nma
 	// in a separate container.
-	return dedent.Dedent(`
+	//
+	// The pgrep command is used to search for the Vertica process with the '-h
```
roypaulin, line 518: I did not know this could even happen.

pkg/controllers/vdb/podfacts.go
```diff
@@ -512,20 +510,24 @@ func (p *PodFacts) genGatherScriptBase(vdb *vapi.VerticaDB, pf *PodFact) string
...
+	// the '-h <podIP>' parameter ensures that we identify the Vertica process
+	// that started after the pod was created, adding an extra layer of
+	// fail-safe.
+	return dedent.Dedent(fmt.Sprintf(`
 		echo -n 'verticaPIDRunning: '
-		[[ $(pgrep -f "^.*vertica\s-D") ]] && echo true || echo false
+		[[ $(pgrep -f "/opt/vertica/bin/vertica.*-h %s") ]] && echo true || echo false
 		echo -n 'verticaProcess: '
 		pgrep -f "^.*vertica\s-D" -a | tail -1 || echo error
-		echo -n 'spreadPIDRunning: '
```
roypaulin, line 524: Why were we checking spread pid?

roypaulin (approved):
Looks good!

pkg/controllers/vdb/podfacts.go
spilchen: It was a prior round of debug code. I was curious to see if vertica and spread process both show up after a pod deletion.

PR #763, opened 2024-04-05
Add new leg for 24.3.0 tests

spilchen:
This adds a new leg, e2e-leg-9, to the CI. This is going to be used to test changes done in 24.3.0. One new change with this is that it will use a vertica license so that we can test scaling past 3 nodes. This will be required in order to test out the new replicated upgrade work.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -50,6 +50,7 @@ on:
         - vcluster leg 6
         - vcluster leg 7
         - leg 8
+        - vcluster leg 9
         - vcluster server upgrade
         - vcluster udx
       run_security_scan:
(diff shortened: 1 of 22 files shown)


.github/workflows/e2e.yml
```diff
@@ -50,6 +50,7 @@ on:
         - vcluster leg 6
         - vcluster leg 7
         - leg 8
```
jizhuoyu, line 52: I see leg 8 doesn't have a "vcluster" prefix, should we add that?

jizhuoyu:
LGTM, only have a minor comment

.github/workflows/e2e.yml
spilchen: I saw that too. But leg 8 is special in that it does both admintools and vcluster. It will start as admintools, do an upgrade and end up as vcluster. So, it's kind of both.

jizhuoyu: Agreed.

jizhuoyu (approved):

PR #764, opened 2024-04-08
`vrep` reconciler enforces minimum version of source and target `vdb`s

jizhuoyu:
`vrep` reconciler enforces the minimum source and target db versions to be at least 24.3.0


api/v1beta1/verticareplicator_types.go
```diff
@@ -22,12 +22,62 @@ import (
 
 // VerticaReplicatorSpec defines the desired state of VerticaReplicator
 type VerticaReplicatorSpec struct {
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced"
```
roypaulin, line 27:
```suggestion
```
This should not have the `advanced` marker as it is a required field.

api/v1beta1/verticareplicator_types.go
```diff
@@ -22,12 +22,62 @@ import (
...
+	// Information of the source Vertica database to replicate from
+	Source VerticaReplicatorDatabaseInfo `json:"source"`
+
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced"
```
roypaulin, line 33:
```suggestion
```
This should not have the `advanced` marker as it is a required field.

api/v1beta1/verticareplicator_types.go
```diff
@@ -22,12 +22,62 @@ import (
...
+	// Optional TLS configuration to use when connecting from the source database to the target database;
+	// it refers to an existing TLS config that already exists in the source
+	TLSConfig string `json:"tlsConfig,omitempty"`
+}
+
+// VerticaReplicatorDatabaseInfo defines the information related to either the source or target Vertica database
+// involved in a replication
+type VerticaReplicatorDatabaseInfo struct {
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
```
roypaulin, line 49:
```suggestion
```
You don't need to specify this twice. Same remark for other fields.

api/v1beta1/verticareplicator_types.go
```diff
@@ -22,12 +22,62 @@ import (
...
+	SandboxName string `json:"sandboxName,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:text"
+	// Optional username with which to connect to either source or target Vertica database;
+	// if omitted the superuser of the database is assumed
+	UserName string `json:"userName,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:text"
```
roypaulin, line 67:
```suggestion
	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:Secret"
```

pkg/controllers/vrep/vdbverify_reconciler.go
```diff
@@ -0,0 +1,91 @@
...
+	vinfSource, err := vdbSource.MakeVersionInfoCheck()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	vinfTarget, err := vdbTarget.MakeVersionInfoCheck()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if !vinfSource.IsEqualOrNewer(vapi.ReplicationSupportedMinVersion) {
```
roypaulin, line 68: I assume the 2 vdbs are not required to have the same version?

jizhuoyu: I don't know whether we need to enforce the source and target to have the same version. The JIRA ticket only mentions that we need to check whether both meet the minimum version requirement. I think we need more info on the version requirement for source and target `vdb`s here. @chinhtranvan do you have more info?

api/v1beta1/verticareplicator_types.go
jizhuoyu: should I just remove `xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced"`?

api/v1beta1/verticareplicator_types.go
```diff
@@ -22,12 +22,57 @@ import (
...
+	// if omitted the superuser of the database is assumed
+	UserName string `json:"userName,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:Secret"
+	// Optional name of the secret which contains password for the specified username with which to
+	// connect to either source or target Vertica database;
+	// if omitted or empty no password is provided;
+	// the secret is assumed to be a k8s secret;
+	// if it has the secret path reference (i.e. gsm://), it reads the secret
+	// from the external secret storage manager
```
spilchen, line 69:
```suggestion
        // The password secret for the given Username is stored in this field. If
	// this field and Username are omitted, the default is set to the superuser
	// password secret found in the VerticaDB. An empty value indicates the
	// absence of a password.
        //
	// The secret is assumed to be a Kubernetes (k8s) secret unless a secret
	// path reference is specified. In the latter case, the secret is retrieved
	// from an external secret storage manager.
```

api/v1beta1/verticareplicator_types.go
```diff
@@ -22,12 +22,57 @@ import (
...
+	PasswordSecret string `json:"passwordSecret,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:text"
+	// Optional name of the service that will be connected to in either source or target Vertica database;
+	// if omitted or empty the service object for the first primary subcluster is assumed
```
spilchen, line 74:
```suggestion
	// This field allows you to specify the name of the service object that will
	// be used to connect to the database. If you do not specify a name, the
	// service object for the first primary subcluster will be used.
```

pkg/controllers/vrep/vdbverify_reconciler.go
chinhtranvan: I don't think it's necessary to require the same version for both the source and target databases. The main idea is to upgrade all nodes in a database (or sandbox) first, and then replicate to another database (or sandbox). In the sandbox, it will be a separate cluster that no longer sees modifications from the main cluster. Therefore, the version may differ from the main cluster.

pkg/controllers/vrep/vdbverify_reconciler.go
```diff
@@ -0,0 +1,91 @@
...
+// Reconcile will verify the verticaDBs in the vrep CR source and target exist and
+// both vertica versions support the replication feature
+func (r *VdbVerifyReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	// no-op if the check has already been done once
+	isSet := r.Vrep.IsStatusConditionTrue(v1beta1.ReplicationReady)
+	if isSet {
+		return ctrl.Result{}, nil
+	}
+
+	vdbSource := &vapi.VerticaDB{}
```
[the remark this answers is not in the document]
jizhuoyu: do we check this only for source or for both `vdb`s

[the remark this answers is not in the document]
jizhuoyu: I assume yes as minimum version for target "v23.3.0" doesn't support `vclusterops`?

pkg/controllers/vrep/vdbverify_reconciler.go
```diff
@@ -0,0 +1,91 @@
...
+			"The target Vertica version '%s' doesn't support replication", vinfTarget.VdbVer)
+		err = vrepstatus.Update(ctx, r.VRec.Client, r.VRec.Log, r.Vrep,
+			[]*metav1.Condition{vapi.MakeCondition(v1beta1.ReplicationReady, metav1.ConditionFalse, "IncompatibleTargetDB")}, stateIncompatibleDB)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	return ctrl.Result{}, vrepstatus.Update(ctx, r.VRec.Client, r.VRec.Log, r.Vrep,
```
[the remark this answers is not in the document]
jizhuoyu: will add one

pkg/controllers/vrep/vdbverify_reconciler.go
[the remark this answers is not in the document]
spilchen: Yes, it is just the source we need to check. The operator doesn't talk to the target. That's only done from the source server.

pkg/events/event.go
```diff
@@ -54,6 +54,7 @@ const (
 	ShowRestorePointsStarted        = "ShowRestorePointsStarted"
 	ShowRestorePointsFailed         = "ShowRestorePointsFailed"
 	RestoreNotSupported             = "RestoreNotSupported"
+	ReplicationNotSupported         = "ReplicationNotSupported"
```
spilchen, line 57: I noticed that we have been splitting the constants out by controllers. Some of the vrpq constants are in here too. Could you move them to separate blocks?

spilchen (approved):
Looks good. Just one additional comment that should be easy to address.

roypaulin (approved):

pkg/controllers/vrep/vdbverify_reconciler.go
```diff
@@ -0,0 +1,120 @@
...
+		VRec: r,
+		Vrep: vrep,
+		Log:  log.WithName("VdbVerifyReconciler"),
+	}
+}
+
+// Reconcile will verify the verticaDBs in the vrep CR source and target exist and
+// both vertica versions support the replication feature
+func (r *VdbVerifyReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	// no-op if the check has already been done once
```
roypaulin, line 57:
```suggestion
	// no-op if the check has already been done once and was successful
```
Because the condition could be false. The question do we want to keep checking or abort the reconciliation all together.

jizhuoyu: So do we want to keep checking until the check succeeds? @spilchen @roypaulin

roypaulin: Now that I think about it, VerticaRepliactor is not something to use once and discard like scrutinize, so the current code makes sense.

jizhuoyu: I think the CR is intended for one-time use, similar to `vrpq`, and currently it's the verification part of it (checking vdb version and deployment type) that we want to keep retrying until we succeed, and once it succeeds the actual replication process is done exactly once without retrying. After discussing with @spilchen we think this might not be the ideal behavior, as the operator would keep retrying as long as the verification fails, and it's possible that it retries while the vdb is upgrading or right after a vdb is just created and not fully ready yet. During such moments the version annotation and deployment method annotation may already indicates the new state which would satisfy the `vrep` reconciler for verification purpose but the source and target vdbs are not fully ready yet for a replication. We could in theory wait for some status conditions of source and target vdbs that indicate full readiness for replication, but that adds complexity to the implementation and may be unworthwhile. The simpler solution is to treat the whole replication attempt, including the verification part of it, as a one-shot attempt, so that it's less confusing for users. This applies similarly to the `vrpq` reconciler, and there will be a dedicated JIRA ticket for this.

roypaulin (approved):

PR #765, opened 2024-04-08
Replicated upgrade: create secondaries based on primaries

spilchen:
Replicated upgrade needs to mimic the primaries as they will be sandboxed and eventually form the new cluster. This adds creation of secondaries for each of the primaries.

I also changed the replicated upgrade state a bit. I removed the status.upgradeState.replicaGroups field. Instead I am going to annotate individual subclusters through a new spec.subclusters[].annotations field.


pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -93,42 +101,192 @@ func (r *ReplicatedUpgradeReconciler) Reconcile(ctx context.Context, _ *ctrl.Req
...
+// need to add to the VerticaDB to mimic the primaries. The new subclusters will
+// be added directly to r.VDB.
+func (r *ReplicatedUpgradeReconciler) addNewSubclustersForPrimaries() (bool, error) {
+	oldImage, found := r.Manager.fetchOldImage()
+	if !found {
+		return false, errors.New("Could not find old image needed for new subclusters")
 	}
+	newSubclusters := []vapi.Subcluster{}
+	scMap := r.VDB.GenSubclusterMap()
+	for i := range r.VDB.Spec.Subclusters {
```
roypaulin, line 180: Anyway the code inside this for block can be turned into a function?

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -93,42 +101,192 @@ func (r *ReplicatedUpgradeReconciler) Reconcile(ctx context.Context, _ *ctrl.Req
...
+}
+
+// assignSubclustersToReplicaGroupACallback is a callback method to update the
+// VDB. It will assign each subcluster to replica group A by setting an
+// annotation.
+func (r *ReplicatedUpgradeReconciler) assignSubclustersToReplicaGroupACallback() (bool, error) {
+	annotatedAtLeastOnce := false
+	for inx := range r.VDB.Spec.Subclusters {
+		sc := &r.VDB.Spec.Subclusters[inx]
+		if val, found := sc.Annotations[vmeta.ReplicaGroupAnnotation]; !found ||
```
roypaulin, line 245: Shouldn't check if the annotations == nil first?

api/v1/helpers.go
```diff
@@ -421,7 +410,13 @@ func (v *VerticaDB) GetUpgradePolicyToUse() UpgradePolicyType {
...
 	// complete the implementation of this new policy.
 	if v.Spec.UpgradePolicy == ReplicatedUpgrade {
-		if v.HasSecondarySubclusters() && vinf.IsEqualOrNewer(ReplicatedUpgradeVersion) {
+		// Replicated upgrade requires that we scale out the cluster. See if
+		// there is evidence that we have already scaled past 3 nodes (CE
+		// license limit), or we have a license defined.
+		const ceLicenseLimit = 3
+		if vinf.IsEqualOrNewer(ReplicatedUpgradeVersion) &&
+			!v.IsKSafety0() &&
+			(v.getNumberOfNodes() > ceLicenseLimit || v.Spec.LicenseSecret != "") {
```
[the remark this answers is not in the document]
spilchen: It can happen if you install the license with other means. One example is a migrated database. You installed the license prior to moving to k8s. When you do migrate to k8s, you don't need to fill in the license secret since its already in the catalog.

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -93,42 +101,192 @@ func (r *ReplicatedUpgradeReconciler) Reconcile(ctx context.Context, _ *ctrl.Req
...
+	return rec.Reconcile(ctx, &ctrl.Request{})
+}
+
+// runAddNodesReconciler will run the reconciler to scale out any subclusters.
+func (r *ReplicatedUpgradeReconciler) runAddNodesReconciler(ctx context.Context) (ctrl.Result, error) {
+	rec := MakeDBAddNodeReconciler(r.VRec, r.Log, r.VDB, r.PFacts.PRunner, r.PFacts, r.Dispatcher)
+	return rec.Reconcile(ctx, &ctrl.Request{})
+}
+
+// generatePrimariesForReplicaGroupB will create new secondary subclusters for each of the
```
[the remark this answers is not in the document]
spilchen: Good point. Maybe `assignSubclusterToReplicaGroupB`. This will mirror the function `assignSubclusterToReplicaGroupA`.

pkg/controllers/vdb/replicatedupgrade_reconciler.go
spilchen: Yes, will do. Good observation.

pkg/controllers/vdb/replicatedupgrade_reconciler.go
spilchen: No. You can always do lookups into a nil map. You'll never find a key obviously, but its valid Go code to do.

api/v1/helpers.go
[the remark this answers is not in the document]
roypaulin: Yes that's correct, this makes sense! Thanks!

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -290,3 +257,44 @@ func (r *ReplicatedUpgradeReconciler) genNewSubclusterName(baseName string, scMa
 	}
 	return "", errors.New("failed to generate a unique subcluster name")
 }
+
+// duplicateSubcluster will return a new vapi.Subcluster that is based on
+// baseSc. This is used to mimic the primaries in replica group B.
+func (r *ReplicatedUpgradeReconciler) duplicateSubcluster(baseSc *vapi.Subcluster, newSCName, oldImage string) *vapi.Subcluster {
```
roypaulin, line 263: If this function is used only for replica group B can you give it a more specific name like `duplicateSubclusterInReplicaGroupB` or `duplicateSubclusterForReplicaGroupB`

roypaulin (approved):

PR #766, opened 2024-04-09
Add restart reconciler to sandbox controller

roypaulin:
we want to use restart reconciler in both the VerticaDB controller and the sandbox controller. This makes some changes to the restart reconciler in order to achieve that. The sandbox name (empty string if VerticaDB controller) is passed to the reconciler and it uses it to target only pods belonging to that sandbox( or to target only pods belonging to the main cluster if sandbox name is empty)

### DIFF
--- a/api/v1/version.go
+++ b/api/v1/version.go
@@ -75,6 +75,8 @@ const (
 	// Starting in v24.2.0, vcluster scrutinize command can read the
 	// database password from secret(k8s, aws, gsm)
 	ScrutinizeDBPasswdInSecretMinVersion = "v24.2.0"
+	// Starting in v24.3.0, sandboxing with vclusterops is supported
+	SandboxSupportedMinVersion = "v24.3.0"
 )
 
 // GetVerticaVersionStr returns the vertica version, in string form, that is stored
--- a/pkg/builder/labels_annotations.go
+++ b/pkg/builder/labels_annotations.go
@@ -92,8 +92,12 @@ func MakeLabelsForPodObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster) map[string
 }
 
 // MakeLabelsForStsObject constructs the labels that are common for all statefulsets.
-func MakeLabelsForStsObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster) map[string]string {
-	return makeLabelsForObject(vdb, sc, false)
+func MakeLabelsForStsObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster, sandbox string) map[string]string {
+	labels := makeLabelsForObject(vdb, sc, false)
+	if sandbox != "" {
+		labels[vmeta.SandboxNameLabel] = sandbox
+	}
+	return labels
 }
 
 // MakeLabelsForSvcObject will create the set of labels for use with service objects
--- a/pkg/controllers/sandbox/sandbox_controller.go
+++ b/pkg/controllers/sandbox/sandbox_controller.go
@@ -22,15 +22,22 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
 	v1 "github.com/vertica/vertica-kubernetes/api/v1"
+	"github.com/vertica/vertica-kubernetes/pkg/cmds"
 	"github.com/vertica/vertica-kubernetes/pkg/controllers"
+	vdbcontroller "github.com/vertica/vertica-kubernetes/pkg/controllers/vdb"
 	verrors "github.com/vertica/vertica-kubernetes/pkg/errors"
 	"github.com/vertica/vertica-kubernetes/pkg/events"
 	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
 	"github.com/vertica/vertica-kubernetes/pkg/names"
+	"github.com/vertica/vertica-kubernetes/pkg/vadmin"
+	"github.com/vertica/vertica-kubernetes/pkg/version"
 	"github.com/vertica/vertica-kubernetes/pkg/vk8s"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -38,11 +45,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 const (
 	verticaDBNameKey = "verticaDBName"
+	sandboxNameKey   = "sandboxName"
 )
 
 // SandboxConfigMapReconciler reconciles a ConfigMap for sandboxing
@@ -62,6 +73,11 @@ func (r *SandboxConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(
 			&corev1.ConfigMap{},
+			builder.WithPredicates(r.predicateFuncs(), predicate.ResourceVersionChangedPredicate{}),
+		).
+		Watches(
+			&source.Kind{Type: &appsv1.StatefulSet{}},
+			handler.EnqueueRequestsFromMapFunc(r.findObjectsForStatesulSet),
 			builder.WithPredicates(r.predicateFuncs(), predicate.ResourceVersionChangedPredicate{}),
 		).
 		WithOptions(controller.Options{MaxConcurrentReconciles: r.Concurrency}).
@@ -97,6 +113,16 @@ func (r *SandboxConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, err
 	}
 
+	sandboxName := configMap.Data[sandboxNameKey]
+	// we check that sandbox name in configmap.Data is the same as
+	// the one in the labels
+	if sandboxName != configMap.Labels[vmeta.SandboxNameLabel] {
+		err = fmt.Errorf("sandbox name %q in configmap data does not match sandbox label value %q",
+			sandboxName, configMap.Labels[vmeta.SandboxNameLabel])
+		log.Error(err, "configmap contains two different sandbox names")
+		return ctrl.Result{}, err
+	}
+
 	// Fetch the VDB found in the configmap
 	vdb := &v1.VerticaDB{}
 	nm := names.GenNamespacedName(configMap, configMap.Data[verticaDBNameKey])
@@ -104,10 +130,22 @@ func (r *SandboxConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if res, err = vk8s.FetchVDB(ctx, r, configMap, nm, vdb); verrors.IsReconcileAborted(res, err) {
 		return res, err
 	}
+	err = checkDeployment(configMap, vmeta.UseVClusterOps(vdb.Annotations))
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 	log = log.WithValues("verticadb", vdb.Name)
 
+	passwd, err := r.GetSuperuserPassword(ctx, log, vdb)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	prunner := cmds.MakeClusterPodRunner(log, r.Cfg, vdb.GetVerticaUser(), passwd)
+	pfacts := vdbcontroller.MakePodFacts(r, prunner, log)
+	dispatcher := vadmin.MakeVClusterOps(log, vdb, r.Client, passwd, r.EVRec, vadmin.SetupVClusterOps)
+
 	// Iterate over each actor
-	actors := r.constructActors(vdb, log)
+	actors := r.constructActors(vdb, log, prunner, &pfacts, dispatcher, sandboxName)
 	for _, act := range actors {
 		log.Info("starting actor", "name", fmt.Sprintf("%T", act))
 		res, err = act.Reconcile(ctx, &req)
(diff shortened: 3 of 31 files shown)


pkg/controllers/vdb/podfacts.go
```diff
@@ -366,6 +369,14 @@ func (p *PodFacts) collectPodByStsIndex(ctx context.Context, vdb *vapi.VerticaDB
 			return err
 		}
 		pf.hasNMASidecar = vk8s.HasNMAContainer(&pod.Spec)
+		// we get the sandbox name from the sts labels if the subcluster
+		// belongs to a sanbox. If the node is up, we will later retrieve
+		// the sandbox state from the catalog
+		val, ok := sts.Labels[vmeta.SandboxNameLabel]
+		if ok {
+			pf.sandbox = val
+			pf.isPrimary = true
```
roypaulin, line 378: This is temporary. We cannot rely on the subcluster object to get the type once the node belongs to a sandbox. How are we going to get the subcluster type of a sandboxed node when it is down? Using the sandboxes statuses in vdb? Adding a label to the configmap? From the doc, the 1st sc to be added to the sandbox is the primary and the other will be secondary so all we need is that primary sc and we can deduce the type of other sc. Unless there is something I am missing, still trying to understand sandbox.

pkg/controllers/vdb/podfacts.go
```diff
@@ -366,6 +369,14 @@ func (p *PodFacts) collectPodByStsIndex(ctx context.Context, vdb *vapi.VerticaDB
 			return err
 		}
 		pf.hasNMASidecar = vk8s.HasNMAContainer(&pod.Spec)
+		// we get the sandbox name from the sts labels if the subcluster
+		// belongs to a sanbox. If the node is up, we will later retrieve
```
spilchen, line 373:
```suggestion
		// belongs to a sandbox. If the node is up, we will later retrieve
```

pkg/controllers/vdb/podfacts.go
spilchen: Lets defer how we are going to figure this out. I think it's a reasonable assumption to make to start. Can we add a check somewhere that says you can only have a single subcluster in the sandbox? I would suggest we move this to a function with lots of comments too so that its clear it is a work in progress.

pkg/controllers/vdb/podfacts.go
spilchen, line 1: one change I expected in here was that we would filter for a sandbox quite early. Maybe even had a sandbox field to FindSubclusters. It doesn't make sense to collect podfacts for pods that aren't part of the sandbox.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -126,7 +128,7 @@ func (r *RestartReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctr
 	// admin commands to run.  Cluster operations only apply if the entire
 	// vertica cluster is managed by k8s.  We skip that if initPolicy is
 	// ScheduleOnly.
-	if r.PFacts.getUpNodeAndNotReadOnlyCount() == 0 &&
+	if r.PFacts.getUpNodeAndNotReadOnlyCount(r.SandboxName) == 0 &&
```
spilchen, line 131: I think we can avoid the filter for most of the podfacts APIs if we collected podfacts for only a single sandbox.

pkg/controllers/vdb/revivedb_reconciler.go
```diff
@@ -243,7 +243,7 @@ func (r *ReviveDBReconciler) getPodList() ([]*PodFact, bool) {
 // findPodToRunInit will return a PodFact of the pod that should run the init
 // command from
 func (r *ReviveDBReconciler) findPodToRunInit() (*PodFact, bool) {
-	return r.PFacts.findPodToRunAdminCmdOffline()
+	return r.PFacts.findPodToRunAdminCmdOffline("" /* sandbox name*/)
```
spilchen, line 246: I think this API call will change, but we should introduce a constant called `MainCluster`, which is an empty string.

pkg/controllers/sandbox/sandbox_controller.go
```diff
@@ -196,3 +236,54 @@ func (r *SandboxConfigMapReconciler) Eventf(obj runtime.Object, eventtype, reaso
...
+	return requests
+}
+
+// GetSuperuserPassword returns the superuser password if it has been provided
+func (r *SandboxConfigMapReconciler) GetSuperuserPassword(ctx context.Context, log logr.Logger, vdb *v1.VerticaDB) (string, error) {
+	return vk8s.GetSuperuserPassword(ctx, r.Client, log, r, vdb)
+}
+
+// checkDeployment checks if version supports sandboxing with vclusterops
+func checkDeployment(cm *corev1.ConfigMap, isVclusterOps bool) error {
```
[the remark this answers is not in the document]
roypaulin: Yes a vdbcheck kind of reconciler. I will do that.

pkg/controllers/vdb/podfacts.go
roypaulin: That makes sense but wouldn't there be a case later where we need to have all podfacts to make some decision.


Hum thinking about it I don't think that will the case for restartreconciler

spilchen: I can't think of any case either. It will make it easier to reason about things if the sandbox controls all aspects of the pod while they are in the sandbox.

pkg/iter/sc_finder.go
```diff
@@ -109,16 +109,17 @@ func (m *SubclusterFinder) FindPods(ctx context.Context, flags FindFlags) (*core
...
 
 	if flags&FindInVdb != 0 {
-		for i := range m.Vdb.Spec.Subclusters {
-			subclusters = append(subclusters, &m.Vdb.Spec.Subclusters[i])
-		}
+		subclusters = append(subclusters, m.getVdbSubclusters(sandbox)...)
 	}
 
-	if flags&FindNotInVdb != 0 || flags&FindExisting != 0 {
+	// a sanboxed subcluster can only be one of those in the vdb so we skip this part if
```
spilchen, line 120:
```suggestion
	// a sandboxed subcluster can only be one of those in the vdb so we skip this part if
```

pkg/iter/sc_finder.go
spilchen, line 1: I think `FindPods` needs to return pods for a sandbox too.

pkg/builder/labels_annotations.go
```diff
@@ -93,7 +93,12 @@ func MakeLabelsForPodObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster) map[string
 
 // MakeLabelsForStsObject constructs the labels that are common for all statefulsets.
 func MakeLabelsForStsObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster) map[string]string {
-	return makeLabelsForObject(vdb, sc, false)
+	labels := makeLabelsForObject(vdb, sc, false)
+	sandbox := vdb.GetSubclusterSandboxName(sc.Name)
+	if sandbox != vapi.MainCluster {
+		labels[vmeta.IsSandboxConfigLabel] = vmeta.IsSandboxConfigTrue
```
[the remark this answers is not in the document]
roypaulin: Because we also watch the sts to check if a sandbox node is down.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -365,9 +365,6 @@ func (r *RestartReconciler) doesDBHaveQuorum(clusterState map[string]string) boo
 	totalPrimaryCount := 0
 	upPrimaryCount := 0
 	for _, pod := range r.PFacts.Detail {
-		if !pod.isPrimary {
```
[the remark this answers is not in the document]
roypaulin: A mistake when resolving conflicts, I will put it back.

pkg/builder/labels_annotations.go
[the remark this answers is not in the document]
spilchen: I would prefer a differently named label then. It's too specific to the config map, so it seems odd that we set it for the statefulset too.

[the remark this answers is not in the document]
roypaulin: "is-sandbox-resource"? "managed-by-sandbox"? "is-sandbox"? "used-by-sandbox"?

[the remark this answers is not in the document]
roypaulin: I think I made a mistake we do need the sandbox name in the label but for the sts and pods not the configmap. It is helpful watch to trigger the reconciler for a configmap whose snadbox matches the name on the sts label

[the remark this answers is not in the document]
roypaulin: We should have `is-sandbox-config` in configmap and `vertica.com/sandbox` in pods and sts.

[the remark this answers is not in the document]
spilchen: Having the two makes sense. Why do we need the label in the pod?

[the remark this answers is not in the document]
roypaulin: Oh right we already know the subcluster the pod belongs to from  a sc label so we can use that to filter pods, we don't need to add an extra label.

pkg/controllers/sandbox/verifydeployment_reconciler.go
```diff
@@ -0,0 +1,79 @@
...
+func (s *VerifyDeploymentReconciler) checkDeployment() error {
+	var msg string
+	if !vmeta.UseVClusterOps(s.Vdb.Annotations) {
+		msg = fmt.Sprintf("the VerticaDB named %q has vclusterops disabled", s.ConfigMap.Data[verticaDBNameKey])
+		s.Log.Info(msg)
+		return errors.New(msg)
+	}
+	// we want the sandboxed cluster version that is why we get the version
+	// the sandbox configmap
+	vdbVer := s.ConfigMap.Annotations[vmeta.VersionAnnotation]
```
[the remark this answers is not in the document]
roypaulin: I will do. I used this because I assumed the configmap will content the current version running in the sandbox, given that version can be different from the one in the main server. I think I just overthought this.

pkg/iter/sc_finder.go
```diff
@@ -216,3 +222,81 @@ func getLabelsFromObject(obj runtime.Object) (map[string]string, bool) {
...
+// filterPods returns a pod list without main cluster pods if a non-empty sandbox named is
+// passed in, or pod list with only main cluster pods if no sandbox name is passed
+func (m *SubclusterFinder) filterPods(oldPods *corev1.PodList, sandbox string) *corev1.PodList {
+	newPods := []corev1.Pod{}
+	scMap := m.getSubclusterSandboxStatusMap()
+	for i := range oldPods.Items {
+		pod := oldPods.Items[i]
+		isSandbox := false
+		sbName := vapi.MainCluster
+		sc, isSCNode := pod.Labels[vmeta.SubclusterNameLabel]
```
spilchen, line 286: Are there cases where the pod won't have the subcluster name label?

pkg/iter/sc_finder.go
```diff
@@ -216,3 +222,81 @@ func getLabelsFromObject(obj runtime.Object) (map[string]string, bool) {
...
+		for _, scName := range sb.Subclusters {
+			subclusters = append(subclusters, m.Subclusters[scName])
+		}
+	}
+	return subclusters
+}
+
+// filterPods returns a pod list without main cluster pods if a non-empty sandbox named is
+// passed in, or pod list with only main cluster pods if no sandbox name is passed
+func (m *SubclusterFinder) filterPods(oldPods *corev1.PodList, sandbox string) *corev1.PodList {
```
spilchen, line 279:
```suggestion
func (m *SubclusterFinder) filterPodsBySandbox(oldPods *corev1.PodList, sandbox string) *corev1.PodList {
```

pkg/controllers/sandbox/verifydeployment_reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if !vinf.IsEqualOrNewer(v1.SandboxSupportedMinVersion) {
+		s.SRec.Eventf(s.Vdb, corev1.EventTypeWarning, events.SandboxNotSupported,
+			"The Vertica version %q does not support sandboxing with vclusterops",
+			vinf.VdbVer)
+		return ctrl.Result{Requeue: true}, nil
+	}
+	s.Log.Info(fmt.Sprintf("The VerticaDB named '%s' is configured for sandboxing with vclusterops", s.Vdb.Name))
```
spilchen, line 71: minor: we shouldn't use Spintf to format a log message. The string message should be constant and any additional data passed as key/data.
```suggestion
	s.Log.Info("The VerticaDB is configured for sandboxing with vclusterops", "vdbName", s.Vdb.Name)
```

api/v1/version.go
```diff
@@ -79,7 +79,7 @@ const (
 	// Starting in v24.2.0, vcluster scrutinize command can read the
 	// database password from secret(k8s, aws, gsm)
 	ScrutinizeDBPasswdInSecretMinVersion = "v24.2.0"
-	// Starting in v24.3.0, sandboxing a subcluster is supported
+	// Starting in v24.3.0, sandboxing a subcluster  is supported
```
spilchen, line 82: I think your earlier version of this comment was better.
```suggestion
	// Starting in v24.3.0, sandboxing a subcluster with the operator is supported
```

pkg/controllers/vscr/scrutinizepod_reconciler.go
```diff
@@ -92,7 +92,7 @@ func (s *ScrutinizePodReconciler) Reconcile(ctx context.Context, _ *ctrl.Request
 // to the vcluster scrutinize CLI
 func (s *ScrutinizePodReconciler) collectInfoFromVdb(ctx context.Context) (ctrl.Result, error) {
 	finder := iter.MakeSubclusterFinder(s.VRec.Client, s.Vdb)
-	pods, err := finder.FindPods(ctx, iter.FindExisting)
+	pods, err := finder.FindPods(ctx, iter.FindExisting, v1.MainCluster)
```
spilchen, line 95: I guess we can't ever collect scrutinize inside a sandbox. Should have a jira for this?

pkg/controllers/vdb/podfacts.go
spilchen, line 1: I think we'll want to move this out of this package into a separate one. It will make it easier for reuse in other controllers. I think @jizhuoyu needs to use this in the vertica replicator controller too. Something to do as a follow-on task.

spilchen (commented):
I think we are close. Just a few minor comments.

pkg/controllers/vscr/scrutinizepod_reconciler.go
roypaulin: You mean to collect scrutinize inside a sandbox? Yes we can I will open a jira for that.

pkg/controllers/vdb/podfacts.go
roypaulin: You are right, I did not think we would end up using it in multiple controllers. I will open a follow-on PR for this

What about the reconcilers shared by some controllers? It looks strange to see one controller depend on another.

pkg/iter/sc_finder.go
roypaulin: The initial list contains the pods owned by the operator, I know that's a stretch but I thought a pod owned by the operator does not mean it belongs/has belonged to a sc. If you think it can confuse I can remove it.

pkg/controllers/vdb/podfacts.go
spilchen: Yes, its possible we may want to move reconcilers out that are shared between controllers. This is something to keep an eye on as we develop the sandbox controller more.

pkg/controllers/vscr/scrutinizepod_reconciler.go
spilchen: Yes, exactly.

pkg/iter/sc_finder.go
spilchen: We should already omit pods that aren't part of a subcluster. In `buildObjList`, see the check for `hasSubclusterNameLabel`. We may be able to simplify this a bit if we assume all pods will have the label you are after.

spilchen (approved):

PR #767, opened 2024-04-09
Add webhook rules for sandboxes

cchen-vertica:
This PR adds below webhooks rules for sandboxes in CRD:
- cannot scale (up or down) any subcluster that is in a sandbox
- cannot remove a subcluster that is sandboxed. It must be unsandboxed first.
- cannot have multiple sandboxes with the same name
- cannot have the image of a sandbox be different than the main cluster before the sandbox has been setup
- cannot be used on versions older than 24.3.0
- cannot be used before the database has been initialized
- cannot have duplicate subclusters defined in a sandbox
- cannot have a subcluster defined in multiple sandboxes
- cannot have a non-existing subcluster defined in a sandbox

We could add more rules later for sandboxes when we have needs.

### DIFF
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -21,8 +21,10 @@ import (
 	"reflect"
 	"strings"
 
+	vutil "github.com/vertica/vcluster/vclusterops/util"
 	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
 	"github.com/vertica/vertica-kubernetes/pkg/paths"
+	vversion "github.com/vertica/vertica-kubernetes/pkg/version"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -141,6 +143,7 @@ func (v *VerticaDB) validateImmutableFields(old runtime.Object) field.ErrorList 
 	allErrs = v.checkImmutableDepotVolume(oldObj, allErrs)
 	allErrs = v.checkImmutablePodSecurityContext(oldObj, allErrs)
 	allErrs = v.checkImmutableSubclusterDuringUpgrade(oldObj, allErrs)
+	allErrs = v.checkImmutableSubclusterInSandbox(oldObj, allErrs)
 	return allErrs
 }
 
@@ -178,6 +181,7 @@ func (v *VerticaDB) validateVerticaDBSpec() field.ErrorList {
 	allErrs = v.hasValidCreateDBTimeout(allErrs)
 	allErrs = v.hasValidUpgradePolicy(allErrs)
 	allErrs = v.hasValidReplicaGroups(allErrs)
+	allErrs = v.validateSandboxes(allErrs)
 	if len(allErrs) == 0 {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


api/v1/verticadb_webhook.go
```diff
@@ -1021,6 +1025,108 @@ func (v *VerticaDB) hasValidProbeOverride(allErrs field.ErrorList, fieldPath *fi
 	return allErrs
 }
 
+// validateSandboxes validates if provided sandboxes info is correct
+func (v *VerticaDB) validateSandboxes(allErrs field.ErrorList) field.ErrorList {
```
roypaulin, line 1029: I think you should some checks in this function in their own functions

api/v1/verticadb_webhook.go
```diff
@@ -1029,6 +1135,16 @@ func (v *VerticaDB) isDBInitialized() bool {
 	return v.IsStatusConditionTrue(DBInitialized)
 }
 
+// isSandboxInitialized returns ture when the sandbox has been created in the database
+func (v *VerticaDB) isSandboxInitialized(targetSc string) bool {
```
roypaulin, line 1139: I know it is not a big deal
```suggestion
func (v *VerticaDB) isSandboxInitialized(targetSb string) bool {
```

api/v1/verticadb_webhook.go
```diff
@@ -1029,6 +1135,16 @@ func (v *VerticaDB) isDBInitialized() bool {
...
+	for _, sc := range v.Status.Sandboxes {
```
roypaulin, line 1140:
```suggestion
	for _, sb := range v.Status.Sandboxes {
```

api/v1/verticadb_webhook.go
```diff
@@ -1341,6 +1457,85 @@ func (v *VerticaDB) checkImmutableSubclusterDuringUpgrade(oldObj *VerticaDB, all
...
+	for _, sandbox := range oldSandboxes {
+		for _, sc := range sandbox.Subclusters {
+			scInSandbox[sc.Name] = struct{}{}
+		}
+	}
+	removedScs := vutil.MapKeyDiff(oldScMap, newScMap)
+	for _, sc := range removedScs {
+		if _, ok := scInSandbox[sc]; ok {
+			err := field.Invalid(path,
+				v.Spec.Subclusters,
```
spilchen, line 1529: general: this may generate a lot of output in the error because the message will print out the entire subcluster list. We may want to only include the offending subcluster. Something that we find in one of the subcluster maps should be fine.

api/v1/verticadb_webhook.go
```diff
@@ -1021,6 +1025,108 @@ func (v *VerticaDB) hasValidProbeOverride(allErrs field.ErrorList, fieldPath *fi
...
+		// check if we have duplicate sandboxes
+		if _, ok := seenSandbox[sandbox.Name]; ok {
+			err := field.Invalid(path,
+				v.Spec.Sandboxes,
+				fmt.Sprintf("sandbox %s already exists", sandbox.Name))
+			allErrs = append(allErrs, err)
+		}
+		seenSandbox[sandbox.Name] = struct{}{}
+		// check if sandbox image is different than main cluster's before sandbox has been setup
+		if sandbox.Image != mainClusterImage && !v.isSandboxInitialized(sandbox.Name) {
```
spilchen, line 1049: The `sandbox.Image` can be an empty string. That just means it will inherit the image from vdb.spec.image. So, we should allow that.

api/v1/verticadb_webhook.go
cchen-vertica: Do you mean each check should have its own function?

cchen-vertica:
Ready for second round review.

api/v1/verticadb_webhook.go
cchen-vertica: I'd like to save some spaces for validateVerticaDBSpec(). It will become too long if we have multiple check methods for one component.  And putting sandbox checks in one function will be faster since we looped over all sandboxes/subclusters only once for multiple checks.

spilchen (approved):

PR #768, opened 2024-04-11
Add stubs for replicated upgrade

spilchen:
This change adds stubs to the replicated upgrade reconciler. It attempts to have the functions present that are needed to do the upgrade. Due to dependencies, such as sandboxing and replication, comprehensive testing is currently limited to unit tests. While this logic might evolve during integration, it gives a framework for how the upgrade should function.

Additionally, I identified new requirements for state management. Two new annotations have been introduced to the VerticaDB: 
- vertica.com/replicated-upgrade-sandbox: to record the name of the sandbox created for the upgrade
- vertica.com/replicated-upgrade-replicator-name: to track the name of the VerticaReplicator CR utilized for replicating data to the sandbox.

### DIFF
--- a/pkg/controllers/vdb/replicatedupgrade_reconciler.go
+++ b/pkg/controllers/vdb/replicatedupgrade_reconciler.go
@@ -216,7 +216,7 @@ func (r *ReplicatedUpgradeReconciler) sandboxReplicaGroupB(ctx context.Context) 
 	return ctrl.Result{}, nil
 }
 
-// waitForSandboToFinish will block the upgrade until the sandbox controller has
+// waitForSandboxToFinish will block the upgrade until the sandbox controller has
 // sandboxed the subclusters that are part of replica group B.
 func (r *ReplicatedUpgradeReconciler) waitForSandboxToFinish(ctx context.Context) (ctrl.Result, error) {
 	sb := r.VDB.GetSandbox(r.sandboxName)
@@ -224,7 +224,7 @@ func (r *ReplicatedUpgradeReconciler) waitForSandboxToFinish(ctx context.Context
 		return ctrl.Result{}, fmt.Errorf("could not find sandbox %q", r.sandboxName)
 	}
 	// If the image in the sandbox matches the image in the spec. We have
-	// already attempted to upgrade the sandbox. So, this implies they have been
+	// already attempted to upgrade the sandbox. So, this implies it has been
 	// created already.
 	if sb.Image == r.VDB.Spec.Image {
 		return ctrl.Result{}, nil
@@ -256,7 +256,7 @@ func (r *ReplicatedUpgradeReconciler) upgradeSandbox(ctx context.Context) (ctrl.
 	}
 
 	// We can skip if the image in the sandbox matches the image in the vdb.
-	// This is the new version that we are upgrading too.
+	// This is the new version that we are upgrading to.
 	if sb.Image == r.VDB.Spec.Image {
 		return ctrl.Result{}, nil
 	}
@@ -341,7 +341,7 @@ func (r *ReplicatedUpgradeReconciler) startReplicationToReplicaGroupB(ctx contex
 func (r *ReplicatedUpgradeReconciler) waitForReplicateToReplicaGroupB(ctx context.Context) (ctrl.Result, error) {
 	vrepName := vmeta.GetReplicatedUpgradeReplicator(r.VDB.Annotations)
 	if vrepName == "" {
-		r.Log.Info("skipping wait for VerticaReplicator because name cannot be found")
+		r.Log.Info("skipping wait for VerticaReplicator because name cannot be found in vdb annotations")
 		return ctrl.Result{}, nil
 	}
 
@@ -358,7 +358,7 @@ func (r *ReplicatedUpgradeReconciler) waitForReplicateToReplicaGroupB(ctx contex
 			r.Log.Info("VerticaReplicator is not found. Skipping wait", "name", vrepName)
 			return ctrl.Result{}, nil
 		}
-		return ctrl.Result{}, fmt.Errorf("failed trying to fetche VerticaReplicator: %w", err)
+		return ctrl.Result{}, fmt.Errorf("failed trying to fetch VerticaReplicator: %w", err)
 	}
 
 	if !vrep.IsStatusConditionTrue(v1beta1.ReplicationComplete) {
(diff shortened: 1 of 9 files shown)


pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -246,12 +522,40 @@ func (r *ReplicatedUpgradeReconciler) genNewSubclusterName(baseName string, scMa
...
+// genNameWithUUID will return a unique name with a uuid suffix. The caller has
+// to provide a lookup function to verify the name generated isn't used. If the
+// lookupFunc returns true, that means the name is in use and another one will
+// be generated.
+func (r *ReplicatedUpgradeReconciler) genNameWithUUID(baseName string, lookupFunc func(nm string) bool) (string, error) {
 	// Add a uuid suffix.
 	const maxAttempts = 100
 	for i := 0; i < maxAttempts; i++ {
 		u := uuid.NewString()
 		nm := fmt.Sprintf("%s-%s", baseName, u[0:5])
```
roypaulin, line 557: There is a lot of effort on making sure we generate a unique sandbox name. Is that necessary? I see you are only using a subset of the uuid I guess that decrease the uniqueness. Is that for readability?

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -221,16 +452,61 @@ func (r *ReplicatedUpgradeReconciler) assignSubclustersToReplicaGroupACallback()
...
+	}
+	sandbox := vapi.Sandbox{
+		Name:  sandboxName,
+		Image: oldImage,
+	}
+	for _, nm := range scNames {
+		sandbox.Subclusters = append(sandbox.Subclusters, vapi.SubclusterName{Name: nm})
+	}
+	r.VDB.Annotations[vmeta.ReplicatedUpgradeSandboxAnnotation] = sandboxName
+	r.VDB.Spec.Sandboxes = append(r.VDB.Spec.Sandboxes, sandbox)
```
roypaulin, line 480: I guess this is vdb is updated elsewhere. Can you add a comment about why we do this here and when/where this will be used?

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -221,16 +452,61 @@ func (r *ReplicatedUpgradeReconciler) assignSubclustersToReplicaGroupACallback()
...
 func (r *ReplicatedUpgradeReconciler) countSubclustersForReplicaGroup(groupName string) int {
-	count := 0
+	scNames := r.getSubclustersForReplicaGroup(groupName)
+	return len(scNames)
+}
+
+// getSubclustersForReplicaGroup returns the names of the subclusters that are part of a replica group.
+func (r *ReplicatedUpgradeReconciler) getSubclustersForReplicaGroup(groupName string) []string {
+	scNames := []string{}
 	for i := range r.VDB.Spec.Subclusters {
```
roypaulin, line 504: Should this be a function in v1?

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -299,3 +603,37 @@ func (r *ReplicatedUpgradeReconciler) duplicateSubclusterForReplicaGroupB(
...
+	for i := range r.VDB.Status.Sandboxes {
+		sb := r.VDB.Status.Sandboxes[i]
+		if sb.Name != r.sandboxName {
+			continue
+		}
+		foundSandbox = true
+
+		// Sandbox was found. Do additional verification that the subclusters
+		// contained within it are correct.
+		if len(sb.Subclusters) != len(scNames) {
```
roypaulin, line 625: Can you add what can cause this?

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -165,7 +199,204 @@ func (r *ReplicatedUpgradeReconciler) assignSubclustersToReplicaGroupB(ctx conte
...
+
+// waitForSandboToFinish will block the upgrade until the sandbox controller has
+// sandboxed the subclusters that are part of replica group B.
+func (r *ReplicatedUpgradeReconciler) waitForSandboxToFinish(ctx context.Context) (ctrl.Result, error) {
+	sb := r.VDB.GetSandbox(r.sandboxName)
+	if sb == nil {
+		return ctrl.Result{}, fmt.Errorf("could not find sandbox %q", r.sandboxName)
+	}
+	// If the image in the sandbox matches the image in the spec. We have
+	// already attempted to upgrade the sandbox. So, this implies they have been
```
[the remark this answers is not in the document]
spilchen: I'll say "So, this implies it has been created already"

pkg/controllers/vdb/replicatedupgrade_reconciler.go
spilchen: This is used as a callback function `updateVDBWithRetry` to do the actual update.

pkg/controllers/vdb/replicatedupgrade_reconciler.go
spilchen: I'm not sure if we'll need this outside of replicated upgrade though. We can always move it later.

pkg/controllers/vdb/replicatedupgrade_reconciler.go
spilchen: We need to come up with a unique name for any subcluster or sandbox we create. The preference is for a human readable name. So, we try to create a sandbox named "replica-group-b". However, if this is already picked, then we will simply add a UUID suffice (of 5 characters) on the end. Similar to how k8s will generate an object name. So, it could end being something like "replica-group-b-abcef". There isn't a way to avoid generating a unique name. The database will not like it if we try to reuse a name that already exists.

pkg/controllers/vdb/replicatedupgrade_reconciler.go
spilchen: This is just defensive programming. This should always be the case. But its good practice to check in case of some programming error.

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -435,7 +436,8 @@ func (r *ReplicatedUpgradeReconciler) addNewSubclustersForPrimaries() (bool, err
 
 // assignSubclustersToReplicaGroupACallback is a callback method to update the
 // VDB. It will assign each subcluster to replica group A by setting an
-// annotation.
+// annotation. This is a callback function for updatedVDBWithRetry to prepare
+// the update for an update.
```
roypaulin, line 440: This does not sound right.

spilchen: You're right :)

roypaulin (approved):

PR #769, opened 2024-04-12
Use VClusterOps API to get node details in podfacts

cchen-vertica:
This adds a new fetcher in podfacts to fetch node details from the running database inside the pod. This new fetcher will call VClusterOps API which will send an HTTP request to the database. The new fetcher will be enabled when VclusterOps annotation is set, and Vertica server version is not older than v24.3.0. The new fetcher should be faster and more reliable than the old fetcher which will execute vsql within the pod.


pkg/controllers/vdb/podfacts.go
```diff
@@ -805,29 +744,45 @@ func (p *PodFacts) checkIsDBCreated(_ context.Context, vdb *vapi.VerticaDB, pf *
...
 
-// makeNodeInfoFetcher will create a NodeInfoFetcher object based on the feature flags set
-// and the server version
+// makeNodeInfoFetcher will create a node-info Fetcher object based on the vclusterOps annotation
+// and the server version. If we are using vclusterops and server version is not older than v24.3.0,
+// the operator will call vclusterOps API to collect the node info. Otherwise, the operator will
+// execute vsql inside the pod to collect the node info.
 func (p *PodFacts) makeNodeInfoFetcher(vdb *vapi.VerticaDB, pf *PodFact) catalog.Fetcher {
-	return catalog.MakeVSQL(vdb, p.PRunner, pf.name, pf.execContainerName)
+	const vclusterAPISupportedMinVersion = "v24.3.0"
```
roypaulin, line 752: Can you call this const `FetchNodeDetailsWithVclusterOpsMinVersion` and add it to v1/version.go?

pkg/controllers/vdb/obj_reconciler_test.go
```diff
@@ -46,7 +46,7 @@ var _ = Describe("obj_reconcile", func() {
 
 	runReconciler := func(vdb *vapi.VerticaDB, expResult ctrl.Result, mode ObjReconcileModeType) {
 		// Create any dependent objects for the CRD.
-		pfacts := MakePodFacts(vdbRec, &cmds.FakePodRunner{}, logger)
+		pfacts := MakePodFacts(vdbRec, &cmds.FakePodRunner{}, logger, "")
```
roypaulin, line 49:
```suggestion
		pfacts := MakePodFacts(vdbRec, &cmds.FakePodRunner{}, logger, vapi.MainCluster)
```
minor: use v1.MainCluster const in place of empty sandbox name

spilchen: I think this new empty string is for the dbadmin password. It might be clear if we still used a const for it. i.e. TestEmptyPassword

pkg/controllers/vdb/podfacts.go
```diff
@@ -852,23 +807,6 @@ func parseVerticaNodeName(stdout string) string {
 	return ""
```
spilchen, line 807: It's not directly related to your PR, but I saw it when reviewing your change, can you remove `parseVerticaNodeName`? I think we only use it for test. We don't have any non-test operator code that relies on it.

pkg/controllers/vdb/podfacts.go
```diff
@@ -805,29 +744,45 @@ func (p *PodFacts) checkIsDBCreated(_ context.Context, vdb *vapi.VerticaDB, pf *
...
-// and the server version
+// makeNodeInfoFetcher will create a node-info Fetcher object based on the vclusterOps annotation
+// and the server version. If we are using vclusterops and server version is not older than v24.3.0,
+// the operator will call vclusterOps API to collect the node info. Otherwise, the operator will
+// execute vsql inside the pod to collect the node info.
 func (p *PodFacts) makeNodeInfoFetcher(vdb *vapi.VerticaDB, pf *PodFact) catalog.Fetcher {
-	return catalog.MakeVSQL(vdb, p.PRunner, pf.name, pf.execContainerName)
+	const vclusterAPISupportedMinVersion = "v24.3.0"
+	vdbVer, ok := vdb.GetVerticaVersionStr()
+	if ok {
```
spilchen, line 754: It might be good to log an entry if we cannot get the version. Just so we know why we reverted back to the old vsql behaviour.

pkg/catalog/fetch_node_details_vcluster.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// parseVNodeDetails will parse node details returned by vclusterOps API
+func (nodeDetails *NodeDetails) parseVNodeDetails(vnodeDetails *vclusterops.NodeDetails) {
+	nodeDetails.Name = vnodeDetails.Name
+	nodeDetails.State = vnodeDetails.State
+	nodeDetails.SubclusterOid = strconv.FormatUint(vnodeDetails.SubclusterID, 10)
+	nodeDetails.ReadOnly = vnodeDetails.IsReadOnly
+	nodeDetails.SandboxName = vnodeDetails.SandboxName
+	nodeDetails.ShardSubscriptions = int(vnodeDetails.NumberShardSubscriptions)
```
spilchen, line 49: I believe there a bit of inconsistency with this one. The HTTPS endpoint will return the shard subscriptions, including the replica shard. Where as the old method that did vsql would could the number of subscriptions and explicitly avoided the replica shard in the WHERE clause. I wonder if we can just subtract the number by 1 if it's positive?

pkg/catalog/fetch_node_details_vcluster.go
```diff
@@ -47,6 +47,9 @@ func (nodeDetails *NodeDetails) parseVNodeDetails(vnodeDetails *vclusterops.Node
 	nodeDetails.ReadOnly = vnodeDetails.IsReadOnly
 	nodeDetails.SandboxName = vnodeDetails.SandboxName
 	nodeDetails.ShardSubscriptions = int(vnodeDetails.NumberShardSubscriptions)
+	if nodeDetails.ShardSubscriptions > 0 {
```
spilchen, line 50: I suggest we leave a comment so that we know why we are decrementing the subscription count.
```suggestion
	// The shard subscriptions we get from vcluster includes the replica shard.
	// We decrement that by one to account for that. We want to know when there
	// are 0 shard subscriptions in order to drive a shard rebalance.
	if nodeDetails.ShardSubscriptions > 0 {
```

spilchen (approved):
Looks good. I just had one additional suggestion about a comment we should have.

PR #771, opened 2024-04-15
Redirect clients to replica group B during replicated upgrade

spilchen:
In the replicated upgrade process, we must pause and redirect client connections from subclusters in replica group A to those in replica group B. This is the initial stage of the change. Currently, pause/redirect semantics are not supported, as they require new server modifications that have not yet been implemented. Therefore, we will perform an old-style drain to pause the process and maintain service labels correctly to point to subclusters in replica group B for redirection.

### DIFF
--- a/pkg/controllers/vdb/podfacts.go
+++ b/pkg/controllers/vdb/podfacts.go
@@ -240,6 +240,18 @@ func MakePodFactsForSandbox(vrec config.ReconcilerInterface, prunner cmds.PodRun
 	return pf
 }
 
+// Clone will make a new copy of the podfacts, but setup for the given sandbox name.
+func (p *PodFacts) Clone(sandbox string) PodFacts {
+	ret := *p
+	// Clear out fields we don't want to copy to the new copy.
+	ret.NeedCollection = true
+	ret.Detail = make(PodFactDetail)
+	// This function is intended to get a PodFacts for a different sandbox, so
+	// use the sandbox name that is passed in.
+	ret.SandboxName = sandbox
+	return ret
+}
+
 // Collect will gather up the for facts if a collection is needed
 // If the facts are already up to date, this function does nothing.
 func (p *PodFacts) Collect(ctx context.Context, vdb *vapi.VerticaDB) error {
--- a/pkg/controllers/vdb/podfacts.go
+++ b/pkg/controllers/vdb/podfacts.go
@@ -240,8 +240,8 @@ func MakePodFactsForSandbox(vrec config.ReconcilerInterface, prunner cmds.PodRun
 	return pf
 }
 
-// Clone will make a new copy of the podfacts, but setup for the given sandbox name.
-func (p *PodFacts) Clone(sandbox string) PodFacts {
+// Copy will make a new copy of the podfacts, but setup for the given sandbox name.
+func (p *PodFacts) Copy(sandbox string) PodFacts {
 	ret := *p
 	// Clear out fields we don't want to copy to the new copy.
 	ret.NeedCollection = true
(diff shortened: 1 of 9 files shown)


pkg/controllers/vdb/podfacts.go
```diff
@@ -240,6 +240,18 @@ func MakePodFactsForSandbox(vrec config.ReconcilerInterface, prunner cmds.PodRun
 	return pf
 }
 
+// Clone will make a new copy of the podfacts, but setup for the given sandbox name.
+func (p *PodFacts) Clone(sandbox string) PodFacts {
```
roypaulin, line 244: I would suggest calling this function `Copy` or `CopyForSandbox`. `Clone` gives the impression we are not going to change anything.

pkg/controllers/vdb/upgrade.go
```diff
@@ -500,3 +502,69 @@ func (i *UpgradeManager) fetchOldImage() (string, bool) {
...
 	return "", false
 }
+
+func (i *UpgradeManager) traceActorReconcile(actor controllers.ReconcileActor) {
+	i.Log.Info("starting actor for upgrade", "name", fmt.Sprintf("%T", actor))
+}
+
+// isSubclusterIdle will run a query to see the number of connections
+// that are active for a given subcluster.  It returns a requeue error if there
+// are active connections still.
```
roypaulin, line 512:
```suggestion
// are still active connections.
```

pkg/controllers/vdb/upgrade.go
```diff
@@ -500,3 +502,69 @@ func (i *UpgradeManager) fetchOldImage() (string, bool) {
...
+func (i *UpgradeManager) isSubclusterIdle(ctx context.Context, pfacts *PodFacts, scName string) (ctrl.Result, error) {
+	pf, ok := pfacts.findPodToRunVsql(true, scName)
+	if !ok {
+		i.Log.Info("No pod found to run vsql.  Skipping active connection check")
+		return ctrl.Result{}, nil
+	}
+
+	sql := fmt.Sprintf(
+		"select count(session_id) sessions"+
```
roypaulin, line 521: Any plans to use an endpoint to get the sessions?

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -158,23 +158,35 @@ func (r *ReplicatedUpgradeReconciler) assignSubclustersToReplicaGroupA(ctx conte
...
+func (r *ReplicatedUpgradeReconciler) runObjReconcilerForMainCluster(ctx context.Context) (ctrl.Result, error) {
+	rec := MakeObjReconciler(r.VRec, r.Log, r.VDB, r.PFacts[vapi.MainCluster], ObjReconcileModeAll)
+	r.Manager.traceActorReconcile(rec)
+	res, err := rec.Reconcile(ctx, &ctrl.Request{})
+	r.PFacts[vapi.MainCluster].Invalidate()
+	return res, err
 }
 
 // runAddSubclusterReconciler will run the reconciler to create any necessary subclusters
 func (r *ReplicatedUpgradeReconciler) runAddSubclusterReconciler(ctx context.Context) (ctrl.Result, error) {
```
roypaulin, line 173:
```suggestion
func (r *ReplicatedUpgradeReconciler) runAddSubclusterReconcilerForMainCluster(ctx context.Context) (ctrl.Result, error) {
```
To be consistent. Let's update the description too.

pkg/controllers/vdb/podfacts.go
spilchen: Good point. I'll use `Copy`

pkg/controllers/vdb/upgrade.go
spilchen: No plans at the moment. There is an endpoint, but I'm expecting this usage to be temporary. It is waiting on the server logic for pause/redirect that Fen is working on.

roypaulin (approved):

PR #772, opened 2024-04-16
Update upgrade reconciler to handle subclusters in sandbox

roypaulin:
This change will modify the upgrade reconcilers, both offline and online, to function in either the main cluster or a sandbox. Our immediate plan is to use the offline upgrade reconciler within the sandbox controller, although this will be done in a follow-on task.


pkg/controllers/vdb/upgrade.go
```diff
@@ -46,18 +46,20 @@ type UpgradeManager struct {
...
-	statusCondition string,
+	statusCondition, sandbox string,
 	isAllowedForUpgradePolicyFunc func(vdb *vapi.VerticaDB) bool) *UpgradeManager {
 	return &UpgradeManager{
 		VRec:                          vdbrecon,
 		Vdb:                           vdb,
 		Log:                           log,
 		Finder:                        iter.MakeSubclusterFinder(vdbrecon.Client, vdb),
 		StatusCondition:               statusCondition,
+		SandboxName:                   sandbox,
```
spilchen, line 62: I don't think setting sandbox for the entire upgrade manager makes sense. For replicated upgrade, it will span both the main cluster and a sandbox. So, it doesn't really make sense. I think each function in UpgradeManager that needs the sandbox should accept it as input.

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -233,7 +233,7 @@ func (o *ObjReconciler) checkForDeletedSubcluster(ctx context.Context) (ctrl.Res
 	finder := iter.MakeSubclusterFinder(o.VRec.Client, o.Vdb)
 
 	// Find any statefulsets that need to be deleted
-	stss, err := finder.FindStatefulSets(ctx, iter.FindNotInVdb)
+	stss, err := finder.FindStatefulSets(ctx, iter.FindNotInVdb, vapi.MainCluster)
```
spilchen, line 236: Rather than hard coding the main cluster, could we use the sandbox found in the podfacts?

pkg/controllers/vdb/offlineupgrade_reconciler.go
```diff
@@ -228,7 +230,7 @@ func (o *OfflineUpgradeReconciler) checkNMADeploymentChange(ctx context.Context)
 		return ctrl.Result{}, nil
 	}
 
-	stss, err := o.Finder.FindStatefulSets(ctx, iter.FindExisting)
+	stss, err := o.Finder.FindStatefulSets(ctx, iter.FindExisting, o.SandboxName)
```
spilchen, line 233: General comment, could we use the sandbox from the podfacts if its available in the reconciler?

pkg/iter/sc_finder.go
```diff
@@ -64,11 +64,18 @@ func MakeSubclusterFinder(cli client.Client, vdb *vapi.VerticaDB) SubclusterFind
...
+//nolint:dupl
+func (m *SubclusterFinder) FindStatefulSets(ctx context.Context, flags FindFlags, sandbox string) (*appsv1.StatefulSetList, error) {
 	sts := &appsv1.StatefulSetList{}
 	if err := m.buildObjList(ctx, sts, flags); err != nil {
 		return nil, err
 	}
+	// we can skip the filtering if all of the sts we are
+	// looking for are not for a subcluster in vdb
+	if flags&FindInVdb != 0 || flags&FindExisting != 0 {
+		sts = m.filterStatefulsetsBySandbox(sts, sandbox)
```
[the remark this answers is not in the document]
spilchen: It should be setup correctly. We have full control over it, so it would a programming error if not set.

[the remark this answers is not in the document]
roypaulin: Ok then let's set the label in both sts and pods.

pkg/iter/sc_finder.go
```diff
@@ -80,7 +80,7 @@ func (m *SubclusterFinder) FindStatefulSets(ctx context.Context, flags FindFlags
 // FindServices returns service objects that are in use for subclusters
 func (m *SubclusterFinder) FindServices(ctx context.Context, flags FindFlags) (*corev1.ServiceList, error) {
 	svcs := &corev1.ServiceList{}
-	if err := m.buildObjList(ctx, svcs, flags); err != nil {
+	if err := m.buildObjList(ctx, svcs, flags, vapi.MainCluster); err != nil {
```
spilchen, line 83: This is correct, just because we know how FindServices is called, but to be consistent with the other exported APIs, we should take the sandbox in as input parm to FindServices. That will keep this package general purpose.

pkg/iter/sc_finder.go
```diff
@@ -125,7 +120,7 @@ func (m *SubclusterFinder) FindSubclusters(ctx context.Context, flags FindFlags,
 	// a sandboxed subcluster can only be one of those in the vdb so we skip this part if
 	// we are looking for sandboxed subclusters
 	if isMainCluster && (flags&FindNotInVdb != 0 || flags&FindExisting != 0) {
-		missingSts, err := m.FindStatefulSets(ctx, flags & ^FindInVdb)
+		missingSts, err := m.FindStatefulSets(ctx, flags & ^FindInVdb, vapi.MainCluster)
```
spilchen, line 123: Does this still make sense to pass in main cluster here? Any harm in removing the `isMainCluster`, getting in here for a sandbox and passing the sandbox parm here. It would keep this logic simple. No need for the comment at line 120-121.

pkg/iter/sc_finder.go
```diff
@@ -194,6 +195,15 @@ func (m *SubclusterFinder) buildObjList(ctx context.Context, list client.ObjectL
 	return meta.SetList(list, rawObjs)
 }
 
+// shouldSkipBasedOnSandboxState returns true if the object whose labels
+// is passed does not belong to the given sandbox or main cluster
+func shouldSkipBasedOnSandboxState(l map[string]string, sandbox string) bool {
+	if sandbox == vapi.MainCluster {
+		return l[vmeta.SandboxNameLabel] != vapi.MainCluster
+	}
+	return l[vmeta.SandboxNameLabel] != sandbox
```
spilchen, line 204: Couldn't you just simply it this way?
```suggestion
	return l[vmeta.SandboxNameLabel] != sandbox
```

pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -56,25 +56,31 @@ type ReplicatedUpgradeReconciler struct {
...
+			replicatedUpgradeAllowed),
 		Dispatcher: dispatcher,
 	}
 }
 
 // Reconcile will automate the process of a replicated upgrade.
 func (r *ReplicatedUpgradeReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
-	if ok, err := r.Manager.IsUpgradeNeeded(ctx); !ok || err != nil {
+	sandbox := r.PFacts.getSandboxName()
+	if ok, err := r.Manager.IsUpgradeNeeded(ctx, sandbox); !ok || err != nil {
```
spilchen, line 72: When you merge up `r.PFacts` is a map now (sandbox name to podfacts). This only needs to look at the main cluster, so you probably can just pass in `vapi.MainCluster` here.
```suggestion
	if ok, err := r.Manager.IsUpgradeNeeded(ctx, vapi.MainCluster); !ok || err != nil {
```

pkg/iter/sc_finder.go
roypaulin: This was to avoid calling the client to fetch sts when we don't need, but I see what you mean, now that we have moved the filter in buildObjList let's just pass the sandbox name.

spilchen (approved):
Everything looks good. Thanks for doing the revisions. I noticed in the last run leg 9 failed related to replicated upgrade. I didn't look at it in too much detail. So, I'm not sure if its a flaky test or not. I'm approving assuming it's a flaky test or requires a small change to fix.

pkg/builder/labels_annotations.go
```diff
@@ -88,6 +88,12 @@ func makeLabelsForObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster, forPod bool)
 
 // MakeLabelsForPodObject constructs the labels that are common for all pods
 func MakeLabelsForPodObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster) map[string]string {
+	return makeLabelsForObject(vdb, sc, true)
```
roypaulin, line 91: I think the test was failing because when the replicaB is sandboxed, the sts podTemplateSpec is updated(sandbox name label added) which causes a pod replacement with a rollingUpdateStrategy. That's why `*-pri1-sb-2` is restarted(and will never be up again as restart reconciler is never called) and we get `upNodeCount mismatch expect 3 actual 2`. I don't think we really want to rolling update those pods, don't we?
So if we want to add sandbox label to the pods, maybe we should do so after the sandbox reconciler. What do you think?

spilchen: Yeah, makes sense. Since the pods have to be restarted by the sandbox controller, it seems like it should apply the label then.

PR #780, opened 2024-04-29
Decouple subcluster name from its statefulset name

spilchen:
This adds a new subcluster annotation to control the statefulset name. The new annotation is: `vertica.com/statefulset-name-override`. This is needed for replicated upgrade when the sandbox gets promoted to the main cluster. We will want to rename the sandbox subcluster to match the original names. However, the statefulset name always included the subcluster name. This new annotation will allow us to rename the subcluster while keeping the statefulset name constant. This ensurse we can do a rename without needing to rebuild the statefulset.

### DIFF
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -148,6 +148,7 @@ func (v *VerticaDB) validateImmutableFields(old runtime.Object) field.ErrorList 
 	allErrs = v.checkImmutablePodSecurityContext(oldObj, allErrs)
 	allErrs = v.checkImmutableSubclusterDuringUpgrade(oldObj, allErrs)
 	allErrs = v.checkImmutableSubclusterInSandbox(oldObj, allErrs)
+	allErrs = v.checkImmutableStsName(oldObj, allErrs)
 	return allErrs
 }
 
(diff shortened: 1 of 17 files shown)


api/v1/verticadb_webhook.go
```diff
@@ -1617,6 +1621,39 @@ func (v *VerticaDB) checkImmutableSubclusterInSandbox(oldObj *VerticaDB, allErrs
...
+		}
+		oldStsName := oldSc.GetStatefulSetName(oldObj)
+		newStsName := newSc.GetStatefulSetName(v)
+		if oldStsName != newStsName {
+			scInx := scIndexMap[scName]
+			path := field.NewPath("spec").Child("subclusters").Index(scInx).
+				Child("annotations").Key(vmeta.StsNameOverrideAnnotation)
+			err := field.Invalid(path,
+				newStsName,
+				fmt.Sprintf("Unable to rename the statefulset of subcluster %q to %q after creation of the subcluster.",
```
roypaulin, line 1649: Does `renaming the statefulset of subcluster %q after creation of the subcluster is not allowed` sound better? I think we should make it clear that we prohibit sts name change.

pkg/iter/sc_finder.go
```diff
@@ -127,7 +127,13 @@ func (m *SubclusterFinder) FindSubclusters(ctx context.Context, flags FindFlags,
 		// indication the subcluster is being removed.
 		for i := range missingSts.Items {
 			scName := missingSts.Items[i].Labels[vmeta.SubclusterNameLabel]
-			subclusters = append(subclusters, &vapi.Subcluster{Name: scName, Size: 0})
+			subclusters = append(subclusters, &vapi.Subcluster{
+				Name: scName,
+				Size: 0,
+				Annotations: map[string]string{
+					vmeta.StsNameOverrideAnnotation: missingSts.Items[i].ObjectMeta.Name,
```
roypaulin, line 134: Why do you need to set this when the subcluster is being removed?

spilchen: It's because the statefulset still exists and callers need to be able to generate its name to delete it. Consider the scenario where the VerticaDB has been updated by removing a subcluster but we haven't finished removing it in the Vertica catalog. We need to do the cleanup logic, which includes eventually deleting the statefulset itself.

roypaulin (approved):

PR #781, opened 2024-04-30
Give the vdb verification exactly once semantics for vrpq

chinhtranvan:
The VerticaRestorePointsQuery reconciler will keep retrying the query when a previous query has failed. We should check if QueryComplete is present instead of checking if QueryComplete is true to ensure the verification of exactly-once semantics.
The VRep reconciler change will be opened in another PR in vnext branch

### DIFF
--- a/api/v1beta1/helpers.go
+++ b/api/v1beta1/helpers.go
@@ -142,6 +142,10 @@ func (vrpq *VerticaRestorePointsQuery) IsStatusConditionFalse(statusCondition st
 	return meta.IsStatusConditionFalse(vrpq.Status.Conditions, statusCondition)
 }
 
+func (vrpq *VerticaRestorePointsQuery) IsStatusConditionPresent(statusCondition string) bool {
+	return meta.FindStatusCondition(vrpq.Status.Conditions, statusCondition) != nil
+}
+
 func MakeSampleVrpqName() types.NamespacedName {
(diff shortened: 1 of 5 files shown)


spilchen:
The e2e failures are likely something we inherit from the latest server image. I see them in my PR too.

chinhtranvan:
> The e2e failures are likely something we inherit from the latest server image. I see them in my PR too.



> The e2e failures are likely something we inherit from the latest server image. I see them in my PR too.

Yeah, I notice that the log rotation on my server (e2e-leg-6) failed with the latest server image. I will open a pull request (PR) to fix it.

spilchen:
I am following up with a server issue. In the mean time, to get the e2e tests clean, I rolled back the latest image to point to the day before. So, lets see if that resolves things.

chinhtranvan:
> I am following up with a server issue. In the mean time, to get the e2e tests clean, I rolled back the latest image to point to the day before. So, lets see if that resolves things.

Did your recent PR use the image from the day before? I noticed that it ran successfully for e2e leg 6

spilchen:
> > I am following up with a server issue. In the mean time, to get the e2e tests clean, I rolled back the latest image to point to the day before. So, lets see if that resolves things.
> 
> The image from the day before worked well; I noticed that it ran successfully for all e2e tests in your PR.

Yes, I resubmitted the failed runs after reverting the latest images.

PR #785, opened 2024-04-30
Add offline upgrade reconciler to sandbox controller

roypaulin:
This will allow us to upgrade the subclusters in a sandbox. There is a new image field in the VerticaDB for the sandbox. When that is changed, it will cause the sandbox controller to drive the upgrade. The vdb controller will check if a sandbox needs an upgrade and will wake up the sandbox controller by setting the vdb resource version in the configmap annotations. 
You can follow a sandbox upgrade progress through the upgradestate field in the sandbox status.


roypaulin:
I have partially tested this. Sandbox Offline upgrade cannot complete successfully for now because there is no way to find sandboxed pods. We use sandboxname label to filter pods but it is not set yet. You work on that needs to go in first.

pkg/controllers/vdb/upgrade.go
```diff
@@ -87,15 +90,22 @@ func (i *UpgradeManager) IsUpgradeNeeded(ctx context.Context, sandbox string) (b
...
 	// We first check if the status condition indicates the upgrade is in progress
-	isSet := i.Vdb.IsStatusConditionTrue(i.StatusCondition)
+	isSet := i.IsUpgradeStatusTrue(sbName)
 	if isSet {
 		i.ContinuingUpgrade = true
 	}
 	return isSet
 }
 
+func (i *UpgradeManager) IsUpgradeStatusTrue(sbName string) bool {
```
spilchen, line 102: Can we avoid exporting if we can?
```suggestion
func (i *UpgradeManager) isUpgradeStatusTrue(sbName string) bool {
```

api/v1/verticadb_types.go
```diff
@@ -844,6 +844,23 @@ type VerticaDBStatus struct {
...
+type SandboxUpgradeState struct {
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// UpgradeInProgress indicates if the sandbox is in the process
+	// of having its image change
+	UpgradeInProgress bool `json:"upgradeInProgress"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// OfflineUpgradeInProgress indicates if the sandbox is in the middle
+	// of an offline upgrade
+	OfflineUpgradeInProgress bool `json:"offlineUpgradeInProgress"`
```
spilchen, line 856: Can we get away with just having a single bool for upgrade? Just the `UpgradeInProgress` one. I don't think we need to distinguish the different types of upgrade. Especially since we will only support one to start with.

api/v1/verticadb_types.go
```diff
@@ -844,6 +844,23 @@ type VerticaDBStatus struct {
 	Sandboxes []SandboxStatus `json:"sandboxes,omitempty"`
 }
 
+type SandboxUpgradeState struct {
```
spilchen, line 847: Can we make all of the types in here optional? 
```
// +optional
```

pkg/vadmin/stop_db_vc.go
```diff
@@ -55,6 +55,8 @@ func (v *VClusterOps) genStopDBOptions(s *stopdb.Parms) vops.VStopDatabaseOption
 	opts.DBName = &v.VDB.Spec.DBName
 	opts.IsEon = v.VDB.IsEON()
 
+	*opts.Sandbox = s.Sandbox
```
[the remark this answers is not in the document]
roypaulin: This is how it works:
- sandbox="" && maincluster=false -> all subclusters
- sandbox="" && maincluster=true -> maincluster only
- sandbox="anything" && maincluster=false -> sandbox only
So yes I think we have to set that one too for the main cluster case.

pkg/controllers/vdb/sandboxupgrade_reconciler.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// SandboxUpgradeReconciler will handle the process when a sandbox
+// image changes
+type SandboxUpgradeReconciler struct {
+	VRec       *VerticaDBReconciler
+	Log        logr.Logger
+	Vdb        *vapi.VerticaDB // Vdb is the CRD we are acting on
+	PFacts     *PodFacts
+	Finder     iter.SubclusterFinder
```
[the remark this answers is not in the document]
roypaulin: My bad , forgot to remove them.

pkg/controllers/vdb/sandboxupgrade_reconciler.go
```diff
@@ -0,0 +1,103 @@
...
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+// reconcileSandboxImage will handle sandbox configmap update based on the sandbox image change
+func (s *SandboxUpgradeReconciler) reconcileSandboxImage(ctx context.Context, sbName string) error {
+	if !s.doesSandboxExist(sbName) {
+		return nil
```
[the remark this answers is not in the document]
roypaulin: Yes, probably sandbox reconciler. Let's requeue.

pkg/controllers/vdb/upgrade.go
```diff
@@ -554,7 +600,10 @@ func (i *UpgradeManager) getTargetImage(sandbox string) (string, error) {
 	}
 	// if the target cluster is a sandbox, the target image
 	// is the one set for that specific sandbox
-	return sb.Image, nil
+	if sb.Image != "" {
+		return sb.Image, nil
+	}
+	return i.Vdb.Spec.Image, nil
```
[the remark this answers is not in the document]
roypaulin: Yes you are right. We may want to internally set sandboxes[i].image like we do for nmaTLSSecret.

pkg/controllers/vdb/sandbox_trigger.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// triggerSandboxController will wake up the sandbox controller by setting
+// the vdb resource version in the sandbox configmap annotations
+func (s *SandboxTrigger) triggerSandboxController(ctx context.Context) (bool, error) {
+	if err := s.fetchConfigMap(ctx); err != nil {
+		return false, err
+	}
+	anns := map[string]string{
+		vmeta.VDBResourceVersion: s.vdb.ResourceVersion,
```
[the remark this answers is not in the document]
roypaulin: uuid sounds safer. You are right, we can't 100% guarantee the resource version will be different each time

roypaulin:
Ready for another round.

pkg/controllers/vdb/sandboxupgrade_reconciler.go
```diff
@@ -0,0 +1,100 @@
...
+	if ok, err := s.isSandboxUpgradeNeeded(ctx, sbName); !ok || err != nil {
+		return ctrl.Result{}, err
+	}
+	// Once we find out that a sandbox upgrade is needed, we need to wake up
+	// the sandbox controller to drive it. We will use a SandboxTrigger object
+	// that will update that sandbox's configmap watched by the sandbox controller
+	sbTrigger := MakeSandboxTrigger(s.VRec, s.Vdb, sbName)
+	triggered, err := sbTrigger.triggerSandboxController(ctx)
+	if triggered {
+		s.Log.Info("Sandbox ConfigMap updated. The sandbox controller will drive the upgrade", "Sandbox", sbName)
```
spilchen, line 79: It might be useful for debug purposes to log the uuid used for the trigger value.

spilchen (commented):
What happens if we pick a really old version for the sandbox? Say like we pick version 12.0.4. It doesn't sandboxing. Are we able to get out of this and go back to a good version? Or does the vdb verification in the sandbox controller prevent that?

roypaulin:
> What happens if we pick a really old version for the sandbox? Say like we pick version 12.0.4. It doesn't sandboxing. Are we able to get out of this and go back to a good version? Or does the vdb verification in the sandbox controller prevent that?

@spilchen Do you mean if we change the sandbox version to 12.0.4?

spilchen:
> > What happens if we pick a really old version for the sandbox? Say like we pick version 12.0.4. It doesn't sandboxing. Are we able to get out of this and go back to a good version? Or does the vdb verification in the sandbox controller prevent that?
> 
> @spilchen Do you mean if we change the sandbox version to 12.0.4?

Yes. If the user pick a really old image for the sandbox. Is there a way to get out of that scenario? Or will the vdb verifier in the sandbox controller prevent that.

roypaulin:
> > > What happens if we pick a really old version for the sandbox? Say like we pick version 12.0.4. It doesn't sandboxing. Are we able to get out of this and go back to a good version? Or does the vdb verification in the sandbox controller prevent that?
> > 
> > 
> > @spilchen Do you mean if we change the sandbox version to 12.0.4?
> 
> Yes. If the user pick a really old image for the sandbox. Is there a way to get out of that scenario? Or will the vdb verifier in the sandbox controller prevent that.

vdb verifier checks only the vdb version(annotations) so not really useful for sandbox upgrade and we don't know the  new image is bad until we create new pods running it. I think what is needed here is to update the `imageversion_reconciler`. We can't rely on vdb annotations to get the old sandbox version so I think we need to cache that in the configmap annotations. So during offlineupgrade, at `checkVersion` step, when we call `imageversion_reconciler`, we check upgrade path, deployment type, the new version and update the version in the configmap. What do you think?

spilchen:
The latest changes look good. I'll approve after I see the next set of changes that will set the version annotation in the sandbox ConfigMap.

roypaulin:
@spilchen Ready for another round.

pkg/controllers/vdb/imageversion_reconciler.go
```diff
@@ -108,6 +109,17 @@ func (v *ImageVersionReconciler) Reconcile(ctx context.Context, _ *ctrl.Request)
 	return v.verifyNMADeployment(vinf, pod)
 }
 
+// genSandboxOldVersionInfo will build and return the sandbox in the configmap
+// annotations
+func (v *ImageVersionReconciler) genSandboxOldVersionInfo(ctx context.Context) (*version.Info, error) {
+	sbTrigger := MakeSandboxTrigger(v.Rec, v.Vdb, v.PFacts.SandboxName, "" /*no uuid*/)
+	oldVersion, err := sbTrigger.getSandboxVersion(ctx)
```
spilchen, line 116: Using the sandbox trigger to fetch the version is an odd choice. I guess it has the convenience that it already knows how to fetch the configMap. Is there a better way to do that?

spilchen, line 116: Does this handle the case where the version annotation isn't present? I think to be consistent that should be tolerated.

roypaulin: How about renaming sandbox trigger?

roypaulin: Got you!

spilchen: Yes, that works too.

roypaulin:
Ready!

spilchen (approved):

PR #786, opened 2024-05-01
add test for `vrep` custom username / password and tls config

jizhuoyu:
test the following scenarios:
- replicate from main cluster of source db with default username (`dbadmin`) to a sandbox cluster of target db with custom username and no password
- replicate from main cluster of source db with default username (`dbadmin`) to a sandbox cluster of target db with custom username and password
- replicate from main cluster of source dbwith default username (`dbadmin`) to a sandbox cluster of target db with default username (`dbadmin`) and source TLS config


tests/e2e-leg-9/verify-vrep/setup-vdb-target/base/setup-vdb.yaml
```diff
@@ -32,10 +32,28 @@ spec:
     - name: sc2
       size: 1
       type: secondary
+    - name: sc3
```
spilchen, line 35: You may need to split up the assertion for the create db step. There are a lot of things it needs to do, and it only has a 10 minute timeout. You may need to split the things to check over multiple steps so that it can complete within time. Each step has its own 10 minute timeout.

PR #789, opened 2024-05-02
Vcluster API Refresh

chinhtranvan:


pkg/vadmin/fetch_node_state_vc.go
```diff
@@ -26,17 +26,23 @@ import (
...
 // FetchNodeState will determine if the given set of nodes are considered UP
 // or DOWN in our consensous state. It returns a map of vnode to its node state.
-func (v *VClusterOps) FetchNodeState(_ context.Context, opts ...fetchnodestate.Option) (map[string]string, ctrl.Result, error) {
+func (v *VClusterOps) FetchNodeState(ctx context.Context, opts ...fetchnodestate.Option) (map[string]string, ctrl.Result, error) {
 	v.setupForAPICall("FetchNodeState")
 	defer v.tearDownForAPICall()
 	v.Log.Info("Starting vcluster FetchNodeState")
 
+	// get the certs
+	certs, err := v.retrieveNMACerts(ctx)
```
chinhtranvan, line 35: @spilchen The new fetchNodeState API use both nma and http endpoint instead of only http endpoint  https://git.verticacorp.com/projects/VER/repos/server/browse/platform/vcluster/vclusterops/fetch_node_state.go#177-186, It cause a lot of failures for e2e tests. I just get certs for NMA.

However, we still got problem at this line https://github.com/vertica/vertica-kubernetes/blob/main/pkg/controllers/vdb/restart_reconciler.go#L390. It appears that if I provide NMA certs, they are only used for the initialPodIP. Previously behavior, we only use http for only inital node to get all nodes state. Now, with the recent change, we need to configure all NMA certificates for all pods to use NMA read catalog editor to retrieve information. Is there any existing function available to retrieve pods in the database?

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -385,9 +385,14 @@ func (r *RestartReconciler) doesDBHaveQuorum(clusterState map[string]string) boo
...
 // it could doing the initialization phase.
-func (r *RestartReconciler) fetchClusterNodeStatus(ctx context.Context) (map[string]string, ctrl.Result, error) {
+func (r *RestartReconciler) fetchClusterNodeStatus(ctx context.Context, downPodsFact []*PodFact) (map[string]string, ctrl.Result, error) {
+	downPods := make([]string, 0, len(downPodsFact))
+	for i := range downPodsFact {
+		downPods = append(downPods, downPodsFact[i].podIP)
+	}
 	opts := []fetchnodestate.Option{
 		fetchnodestate.WithInitiator(r.InitiatorPod, r.InitiatorPodIP),
+		fetchnodestate.WithHosts(downPods),
```
chinhtranvan, line 395: @spilchen The new fetchNodeState API use both nma and http endpoint instead of only http endpoint  https://git.verticacorp.com/projects/VER/repos/server/browse/platform/vcluster/vclusterops/fetch_node_state.go#177-186, It cause a lot of failures for e2e tests. I just provide certs for NMA to solve issue

Previously behavior, we only use http for only inital node and use http get all nodes state. Now, with the recent change, we need to configure all NMA certificates for all pods to use NMA read catalog editor endpoint to retrieve information. I just provide downPods to solve it, but let me know if we any existing function available to retrieve pods in the database?

spilchen: Shouldn't there be a change in `fetch_node_state_vc.go`? I understand that we need to pass in the certs (probably with a change to `fetch_node_state_vc.go`). But why do we have to pass in the down hosts?

spilchen: EDIT: now I see the `fetch_node_state_vc.go`. But I still don't understand why the down nodes are needed.

chinhtranvan: I got timeout error for down nodes at NMAReadVerticaVersionOp  step in fech_node_state
```
controllers.VerticaDB.FetchNodeState.NMAReadVerticaVersionOp	[NMAReadVerticaVersionOp] result from host 10.244.0.215 summary FAILURE, details: fail to send request v1/vertica/version on host 10.244.0.215, details Get "https://10.244.0.215:5554/v1/vertica/version": dial tcp 10.244.0.215:5554: connect: connection timed out	{"verticadb": "kuttl-test-lasting-parrot/v-restart", "reconcile-uuid": "8075ac96-4a1d-4afb-941d-3602af402426"}
verticadb-operator verticadb-operator-manager-7d4bb66d4-9hfj8 manager github.com/vertica/vcluster/vclusterops/vlog.(*Printer).PrintError
```
It seems we don't have certs in down nodes

spilchen: I think this needs to do back to Le. Down nodes are probably running with the NMA, but the IPs in the catalog are different, so the endpoint request is timing out.

spilchen: In the meantime, do we need the latest change from vclusterops? I think the fetch node state change was only in the last commit. I wonder if we can pull in the vclusterOps changes from before the last commit. We still need to address this issue, but at least we can make some progress.

chinhtranvan:
I close this PR and move it to https://github.com/vertica/vertica-kubernetes/pull/792

PR #790, opened 2024-05-02
Create sandbox config maps in sandbox reconciler

cchen-vertica:
This will create/update a sandbox config map whenever we create a new sandbox in sandbox reconciler. The config map is watched by sandbox controller for restarting/upgrading the nodes in a sandbox. The config map will be deleted if vdb is deleted or the sandbox is unsandboxed.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -1308,8 +1308,8 @@ func BuildSandboxConfigMap(nm types.NamespacedName, vdb *vapi.VerticaDB, sandbox
 		// the data should be immutable since dbName and sandboxName are fixed
 		Immutable: &immutable,
 		Data: map[string]string{
-			"verticaDBName": vdb.Spec.DBName,
-			"sandboxName":   sandbox,
+			vapi.VerticaDBNameKey: vdb.Spec.DBName,
+			vapi.SandboxNameKey:   sandbox,
 		},
 	}
 }
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -1290,6 +1290,30 @@ func BuildStsSpec(nm types.NamespacedName, vdb *vapi.VerticaDB, sc *vapi.Subclus
 	}
 }
 
+// BuildSandboxConfigMap builds a config map for sandbox controller
+func BuildSandboxConfigMap(nm types.NamespacedName, vdb *vapi.VerticaDB, sandbox string) *corev1.ConfigMap {
+	immutable := true
+	return &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ConfigMap",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            nm.Name,
+			Namespace:       nm.Namespace,
+			Labels:          MakeLabelsForSandboxConfigMap(vdb),
+			Annotations:     MakeAnnotationsForSandboxConfigMap(vdb),
+			OwnerReferences: []metav1.OwnerReference{vdb.GenerateOwnerReference()},
+		},
+		// the data should be immutable since dbName and sandboxName are fixed
+		Immutable: &immutable,
+		Data: map[string]string{
+			"verticaDBName": vdb.Spec.DBName,
+			"sandboxName":   sandbox,
+		},
+	}
+}
+
 // BuildScrutinizePod construct the spec for the scrutinize pod
 func BuildScrutinizePod(vscr *v1beta1.VerticaScrutinize, vdb *vapi.VerticaDB, args []string) *corev1.Pod {
 	return &corev1.Pod{
--- a/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
+++ b/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
@@ -201,17 +201,29 @@ func (s *SandboxSubclusterReconciler) updateSandboxConfigMapFields(curCM, newCM 
 	// exclude sandbox controller trigger ID from the annotations because
 	// vdb controller will set this in current config map, and the new
 	// config map cannot get it
-	triggerID, ok := curCM.Annotations[vmeta.SandboxControllerTriggerID]
-	if ok {
+	triggerID, hasTriggerID := curCM.Annotations[vmeta.SandboxControllerTriggerID]
+	if hasTriggerID {
 		delete(curCM.Annotations, vmeta.SandboxControllerTriggerID)
 	}
+	delete(newCM.Annotations, vmeta.SandboxControllerTriggerID)
+	// exclude version annotation because vdb controller can set a different
+	// vertica version annotation for a sandbox in current config map
+	version, hasVersion := curCM.Annotations[vmeta.VersionAnnotation]
+	if hasVersion {
+		delete(curCM.Annotations, vmeta.VersionAnnotation)
+	}
+	delete(newCM.Annotations, vmeta.VersionAnnotation)
 	if stringMapDiffer(curCM.ObjectMeta.Annotations, newCM.ObjectMeta.Annotations) {
 		updated = true
 		curCM.ObjectMeta.Annotations = newCM.ObjectMeta.Annotations
 	}
 	// add sandbox controller trigger ID back to the annotations
-	if ok {
+	if hasTriggerID {
 		curCM.Annotations[vmeta.SandboxControllerTriggerID] = triggerID
+	}
+	// add vertica version back to the annotations
+	if hasVersion {
+		curCM.Annotations[vmeta.VersionAnnotation] = version
 	}
 	if stringMapDiffer(curCM.ObjectMeta.Labels, newCM.ObjectMeta.Labels) {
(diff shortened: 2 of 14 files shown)


pkg/names/names.go
```diff
@@ -60,6 +60,11 @@ func GenStsName(vdb *vapi.VerticaDB, sc *vapi.Subcluster) types.NamespacedName {
 	return GenNamespacedName(vdb, vdb.Name+"-"+sc.GenCompatibleFQDN())
 }
 
+// GenSandboxConfigMapName returns the name of the sandbox config map
+func GenSandboxConfigMapName(vdb *vapi.VerticaDB, sandbox string) types.NamespacedName {
```
spilchen, line 64: Roy added a similar function in his PR (#785) called `GenConfigMapName`, although I prefer your name here because its clear that the configMap is for the sandbox. We should make sure we don't end up with two functions that do the same thing.

pkg/builder/builder.go
```diff
@@ -1290,6 +1290,30 @@ func BuildStsSpec(nm types.NamespacedName, vdb *vapi.VerticaDB, sc *vapi.Subclus
...
+			Namespace:       nm.Namespace,
+			Labels:          MakeLabelsForSandboxConfigMap(vdb),
+			Annotations:     MakeAnnotationsForSandboxConfigMap(vdb),
+			OwnerReferences: []metav1.OwnerReference{vdb.GenerateOwnerReference()},
+		},
+		// the data should be immutable since dbName and sandboxName are fixed
+		Immutable: &immutable,
+		Data: map[string]string{
+			"verticaDBName": vdb.Spec.DBName,
+			"sandboxName":   sandbox,
```
spilchen, line 1312: We should sync with Roy's PR (#785). He added `const` values for these key names.

pkg/builder/builder.go
```diff
@@ -1290,6 +1290,30 @@ func BuildStsSpec(nm types.NamespacedName, vdb *vapi.VerticaDB, sc *vapi.Subclus
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            nm.Name,
+			Namespace:       nm.Namespace,
+			Labels:          MakeLabelsForSandboxConfigMap(vdb),
+			Annotations:     MakeAnnotationsForSandboxConfigMap(vdb),
+			OwnerReferences: []metav1.OwnerReference{vdb.GenerateOwnerReference()},
+		},
+		// the data should be immutable since dbName and sandboxName are fixed
+		Immutable: &immutable,
```
spilchen, line 1309: Cool, I didn't even know you could do this with a ConfigMap.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -68,9 +72,51 @@ func (s *SandboxSubclusterReconciler) Reconcile(ctx context.Context, _ *ctrl.Req
...
 
+// reconcileSandboxStatus will update sandbox status for the subclusters that are already sandboxed
+func (s *SandboxSubclusterReconciler) reconcileSandboxStatus(ctx context.Context) error {
+	sbScMap := make(map[string][]string)
+	seenScs := make(map[string]any)
+	for _, v := range s.PFacts.Detail {
+		if _, ok := seenScs[v.subclusterName]; ok {
+			continue
+		}
+		if v.sandbox != vapi.MainCluster {
```
spilchen, line 96: I expected there would be a change to podfacts.go. It looks like we set v.sandbox initially in `collectPodByStsIndex` and we get it from the requested sandbox we are requesting. That doesn't seem right. I feel like we should pull it from a label in the statefulset. The comment seems to indicate that, but I'm not sure if that's exactly what's going on.

This is the code I am talking about:
```
		// we get the sandbox name from the sts labels if the subcluster
		// belongs to a sandbox. If the node is up, we will later retrieve
		// the sandbox state from the catalog
		pf.sandbox = p.SandboxName
```

`p.SandboxName` is just an input parameter. I wonder if it should be `sts.Labels[vmeta.SandboxNameLabel`.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -68,9 +72,51 @@ func (s *SandboxSubclusterReconciler) Reconcile(ctx context.Context, _ *ctrl.Req
 		return ctrl.Result{}, err
 	}
 
+	// reconcile sandbox status for the subclusters that are already sandboxed
+	if err := s.reconcileSandboxStatus(ctx); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// reconcile sandbox config maps for the existing sandboxes
+	if err := s.reconcileSandboxConfigMaps(ctx); err != nil {
```
spilchen, line 81: I wonder if there will be race conditions by creating the configMap before the actual sandbox command. Once the ConfigMap is created, the Sandbox reconciler can start to act on the pods. What is preventing them from attempting to restart the pods while we are in the middle of sandboxing them. I think this needs to happen after the sandbox command.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -104,21 +150,81 @@ func (s *SandboxSubclusterReconciler) sandboxSubclusters(ctx context.Context) (c
...
+	}
+	s.Log.Info("Found an existing sandbox config map with correct content, skip updating it", "Name", nm)
+	return nil
+}
+
+// updateSandboxConfigMapFields checks if we need to update the content of a config map,
+// if so, we will update the content of that config map and return true
+func (s *SandboxSubclusterReconciler) updateSandboxConfigMapFields(curCM, newCM *corev1.ConfigMap) bool {
+	updated := false
+	if stringMapDiffer(curCM.ObjectMeta.Annotations, newCM.ObjectMeta.Annotations) {
```
spilchen, line 200: Roy's PR (#785) is going to add an annotation that the vdb controller will use to wake up the sandbox controller. So, we may need to relax this. Or maybe compare everything except a specific annotation. fyi @roypaulin

pkg/names/names.go
cchen-vertica: I will wait for Roy's PR(#785) to be merged first, then I can update his function.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
cchen-vertica: I will change it to sts.Labels[vmeta.SandboxNameLabel] since it probably causes an error when the node is down. That time we cannot update it to the correct sandbox name from checkNodeDetails. Then it could make the node restart fail.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
cchen-vertica: Here we just create config maps for the sandboxes that have subclusters. This is for the case we sandbox a subcluster successfully but create a config map failed. Then the reconciler will fix the config map here. I have the code to create the configMap after the actual sandbox command as well.

spilchen: Ah ok, yes I see that. You can disregard this comment then.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
cchen-vertica: Will that annotation be modified by sandbox controller? In which case that annotation will be changed?

spilchen: The vdb controller will modify it to wake up the sandbox controller. It sets a new uuid anytime it needs the sandbox controller to do something. I don't think the sandbox controller will modify it at all.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -104,21 +151,92 @@ func (s *SandboxSubclusterReconciler) sandboxSubclusters(ctx context.Context) (c
...
+}
+
+// updateSandboxConfigMapFields checks if we need to update the content of a config map,
+// if so, we will update the content of that config map and return true
+func (s *SandboxSubclusterReconciler) updateSandboxConfigMapFields(curCM, newCM *corev1.ConfigMap) bool {
+	updated := false
+	// exclude sandbox controller trigger ID from the annotations because
+	// vdb controller will set this in current config map, and the new
+	// config map cannot get it
+	triggerID, ok := curCM.Annotations[vmeta.SandboxControllerTriggerID]
```
spilchen, line 204: We should also exclude the version (vmeta.VersionAnnotation). That can be set in the ConfigMap too.

cchen-vertica: Is the version value same as vdb's "vertica.com/version"? If so, the new config map will have the correct version annotation too, and we don't need to exclude the version.

spilchen: The values could differ. We will be able to upgrade the sandbox independently from the main cluster.

spilchen (approved):

PR #791, opened 2024-05-02
Use static labels in pods

spilchen:
During a replicated upgrade we will promote a sandbox to the main cluster, which cause it to drop its sandbox state, and we need to rename subclusters to match the original names. Prior to this change, the pods deployed by the operator had labels for the sandbox and subcluster name. And the labels were set in the statefulset pod template. Now, in order to change those, the k8s controller will actually do a rollout. This means it needs to restart each of the pods. But we require sandbox promotion and subcluster rename to happen without a pod restart. So, the change here to use different labeling to ensure we don't need to restart the pods. In some cases, if we have a pod we have to refer back to the statefulset in order to figure certain metadata. 

Since we are changing the statefulset selector values, the statefulset from old operator need to be recreated. We had done this in the past, so I repurposed the upgradeoperator120_reconciler.go for this purpose.

A summary of all of the labels that are set is as follows.

- All objects will have these labels set:
```
app.kubernetes.io/instance: <CR name>
app.kubernetes.io/managed-by: verticadb-operator
app.kubernetes.io/component: database
app.kubernetes.io/name: vertica
app.kubernetes.io/version: 2.2.0
vertica.com/database: <vertica db name>
```
- StatefulSets and Service objects will have the following labels:
```
vertica.com/subcluster-name: <subcluster-name>
vertica.com/subcluster-type: <primary|secondary>
vertica.com/sandbox: <sandbox name if applicable>
```
- Further Service objects will have this additional label:
```
vertica.com/svc-type: <external|headless>
```
- Pods will have the following labels:
```
vertica.com/subcluster-svc: <name of service object>
vertica.com/client-routing: true # Only set if not doing a drain or pending removal
vertica.com/subcluster-selector-name: <full name of the pod's statefulset>
```
- ConfigMaps used by the sandbox controller will have this label:
```
vertica.com/watched-by-sandbox-controller: true
```
- When the statefulset sets up the pod selector, it will use these labels:
```
app.kurbernetes.io/instance
vertica.com/subcluster-selector-name
```
- When the service objects sets up its pod selector, it will use these labels:
```
app.kurbernetes.io/instance
vertica.com/subcluster-selector-name
vertica.com/client-routing
```


pkg/controllers/vdb/dbremovesubcluster_reconciler.go
```diff
@@ -159,15 +159,15 @@ func (d *DBRemoveSubclusterReconciler) resetDefaultSubcluster(ctx context.Contex
 		scFinder := iter.MakeSubclusterFinder(d.VRec.Client, d.Vdb)
 		// We use the FindServices() API to get subclusters that already exist.
 		// We can only change the default subcluster to one of those.
-		svcs, err := scFinder.FindServices(ctx, iter.FindInVdb, vapi.MainCluster)
+		stss, err := scFinder.FindStatefulSets(ctx, iter.FindInVdb, vapi.MainCluster)
 		if err != nil {
 			return err
 		}
 		// If we don't find a service object we don't fail.  The attempt to
```
roypaulin, line 166: Should we update this comment now that we find sts instead of svc?

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -426,12 +424,17 @@ func (o *OnlineUpgradeReconciler) processSecondary(ctx context.Context, sts *app
...
+	if stsScType != scType {
+		return false
+	}
+
+	transientName, hasTransient := o.Vdb.GetTransientSubclusterName()
+	if !hasTransient {
+		return true
 	}
-	return sts.Labels[vmeta.SubclusterTypeLabel] == scType && !isTransient, nil
+	return sts.Labels[vmeta.SubclusterNameLabel] != transientName
```
roypaulin, line 437: I know it is been a while but why dowe need to return false for transient sc?

pkg/iter/sc_finder.go
```diff
@@ -207,12 +211,16 @@ func (m *SubclusterFinder) buildObjList(ctx context.Context, list client.ObjectL
 // shouldSkipBasedOnSandboxState returns true if the object whose labels
 // is passed does not belong to the given sandbox or main cluster
 func shouldSkipBasedOnSandboxState(l map[string]string, sandbox string) bool {
+	// Note, for a pod, the labels passed in will be from the statefulset. So,
```
roypaulin, line 214: Is it better to add this to the function's description?

pkg/meta/labels.go
```diff
@@ -16,13 +16,78 @@
...
-	SubclusterTransientLabel  = "vertica.com/subcluster-transient"
+	//
+	// The type of the subcluster. Values are can be either: primary or
+	// secondary.
+	SubclusterTypeLabel = "vertica.com/subcluster-type"
+	//
+	// This label is added to a statefulset to indicate the sandbox it belongs
+	// to. This will allow the operator to filter these objects if it is looking
+	// for objects from the main cluster or a sandbox.  Moreover, the sandbox
+	// controller will be watching statefulsets witGh this label and will trigger
```
roypaulin, line 79:
```suggestion
	// controller will be watching statefulsets with this label and will trigger
```

roypaulin (commented):
Looks good!

pkg/controllers/vdb/onlineupgrade_reconciler.go
spilchen: When online upgrade iterates over subclusters (i.e. to upgrade, add client routing, etc), we always want to skip transient. Those aren't ever upgraded. They are meant to only serve read requests during the upgrade.

pkg/iter/sc_finder.go
spilchen: yeah, that makes sense. There are 3 functions that have this comment. So, I'll change all 3.

pkg/meta/labels.go
```diff
@@ -34,38 +99,21 @@ const (
 	// to be removed by a pending node removal.
 	ClientRoutingLabel = "vertica.com/client-routing"
 	ClientRoutingVal   = "true"
+	//
+	// This is set in the pods, and is used by statefulset as a pod selector. This
+	// stays constant for the life of the statefulset, and is unique across
+	// statefulsets for other subclusters in the database. It may appear like it
+	// is derived from a subcluster name, but is does not get updated for
+	// subcluster rename. So, it shouldn't be treated as such.
+	SubclusterSelectorLabel = "vertica.com/subcluster-selector-name"
```
roypaulin, line 108: Does that mean the sts name no longer changes when the subcluster is renamed?

spilchen: Yes, this will stay static. This is tied to the statefulset name, which we'll keep static for the rename too.

roypaulin (approved):
I made a lot of changes for upgrade so I wonder if this can go in after my PR?

spilchen:
> I made a lot of changes for upgrade so I wonder if this can go in after my PR?

Yeah, no problem. I'll wait.

PR #793, opened 2024-05-03
Add status reconciler to sandbox controller

roypaulin:
The status reconciler in vdb controller was removing a subcluster's status when the subcluster is in a sandbox. Now it will keep unchanged the status of subclusters that do not belong to the passed in sandbox. To do that, it finds all the subclusters regardless of their cluster.


pkg/controllers/vdb/status_reconciler.go
```diff
@@ -73,25 +73,46 @@ func (s *StatusReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl
 
 // updateStatusFields will refresh the status fields in the vdb
 func (s *StatusReconciler) updateStatusFields(ctx context.Context) error {
-	// Use all subclusters, even ones that are scheduled for removal.  We keep
+	// Use all subclusters regardless of the sandbox they belong to,
+	// even ones that are scheduled for removal. We keep
 	// reporting status on the deleted ones until the statefulsets are gone.
 	finder := iter.MakeSubclusterFinder(s.Client, s.Vdb)
-	subclusters, err := finder.FindSubclusters(ctx, iter.FindAll, vapi.MainCluster)
+	subclusters, err := finder.FindSubclusters(ctx, iter.FindAllAcrossSandboxes, s.PFacts.GetSandboxName())
```
spilchen, line 80: I think we should move this call within the `refreshStatus` closure too.

roypaulin:
> Are there any e2e test changes we can do now that we can rely on the status fields across sandboxes?

I can update the sandbox test case

spilchen (approved):

spilchen:
@roypaulin a follow-on PR for this status change is that we want `.status.subclusters[].detail[].addedToDB` to control if a pod has a database. This is needed by @cchen-vertica to properly start a node after unsandboxing. We can discuss more in today's scrum.

PR #795, opened 2024-05-06
Add sandbox option in vrep

chinhtranvan:
This PR incorporates the latest changes from the vCluster API to update the sandbox option in vrep


spilchen:
Can I assume that without this change, replication from a source sandbox would not be possible? Even though the operator provides the source IP for a sandbox, vclusterOps couldn't utilize the source IP without recognizing that it originated from a sandbox.

spilchen:
I have no issues with this change, assuming my understanding is correct. You should coordinate with @jizhuoyu as you both are changing the same test (see #786)

chinhtranvan:
> Can I assume that without this change, replication from a source sandbox would not be possible? Even though the operator provides the source IP for a sandbox, vclusterOps couldn't utilize the source IP without recognizing that it originated from a sandbox.

Yes, your assumption is correct, vclusterOps needs to identify that the source IP belongs to a sandbox in order to function properly

PR #798, opened 2024-05-07
Update sandbox label

chinhtranvan:
A nondeterministic failure occurs when a label is updated based on the sandbox spec. We need to ensure that the label is updated only after the status is set by the sandbox subcluster


chinhtranvan:
All tests have passed! After this PR is merged, I will proceed to merge my changes for the replicator reconciler.

spilchen (approved):

PR #799, opened 2024-05-07
Add support for multiple subclusters in a sandbox

spilchen:
This adds support for multiple subclusters in a single sandbox. In order to denote which subclusters are primary in the sandbox, a new subcluster type is added called 'sandboxprimary'. New webhook rules have been added to account for the new subcluster type.

The new type will get added to the first subcluster in a sandbox. We now add subclusters to a sandbox in the order they are defined in the vdb spec. This ensures that we pick a deterministic subcluster as the first one in a sandbox.

When adding a subcluster to an existing sandbox, we need to pass in a host from the main cluster and the sandbox too. This reason this is needed is that we need to replicate the new sandbox state in both the main cluster and in the sandbox. So, the vclusterOps API will update both clusters.

### DIFF
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -104,8 +104,6 @@ func (v *VerticaDB) ValidateCreate() error {
 	verticadblog.Info("validate create", "name", v.Name, "GroupVersion", GroupVersion)
 
 	allErrs := v.validateVerticaDBSpec()
-	// temporary check added here to make testing easier
-	allErrs = v.validateSandboxSize(allErrs)
 	if allErrs == nil {
 		return nil
 	}
@@ -117,8 +115,6 @@ func (v *VerticaDB) ValidateUpdate(old runtime.Object) error {
 	verticadblog.Info("validate update", "name", v.Name, "GroupVersion", GroupVersion)
 
 	allErrs := append(v.validateImmutableFields(old), v.validateVerticaDBSpec()...)
-	// temporary check added here to make testing easier
-	allErrs = v.validateSandboxSize(allErrs)
 	if allErrs == nil {
 		return nil
 	}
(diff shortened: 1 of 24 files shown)


api/v1/verticadb_webhook.go
```diff
@@ -1170,12 +1164,23 @@ func (v *VerticaDB) validateSubclustersInSandboxes(allErrs field.ErrorList) fiel
...
+	primarySubclustersInSandboxes := map[string]int{}
 	for sc, i := range seenScWithSbIndex {
 		if scInfo, ok := scMap[sc]; !ok {
 			err := field.Invalid(path.Index(i),
 				sandboxes[i],
 				fmt.Sprintf("subcluster %s does not exist", sc))
 			allErrs = append(allErrs, err)
+		} else if scInfo.IsSandboxPrimary() {
+			primarySubclustersInSandboxes[sandboxes[i].Name]++
+			if primarySubclustersInSandboxes[sandboxes[i].Name] > 1 {
```
cchen-vertica, line 1178: I guess this check could be relaxed by calling promote_subcluster endpoint in the future?

api/v1/verticadb_webhook.go
```diff
@@ -1618,6 +1609,43 @@ func (v *VerticaDB) checkImmutableSubclusterInSandbox(oldObj *VerticaDB, allErrs
 		}
 	}
 
+	newScInSandbox := v.GenSubclusterSandboxMap()
+	oldSbIndexMap := v.GenSandboxIndexMap()
+	oldSbMap := v.GenSandboxMap()
+
+	// This loop ensures a couple of things:
```
cchen-vertica, line 1616: Can demote/promote_subcluster endpoint eliminate this limitation?

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -153,27 +158,86 @@ func (s *SandboxSubclusterReconciler) sandboxSubclusters(ctx context.Context) (c
...
+		if err := pfs.Collect(ctx, s.Vdb); err != nil {
+			return nil, ctrl.Result{}, err
+		}
+		// If this is the first pod in the sandbox, then only an API from the
+		// main cluster is needed.
+		if len(pfs.Detail) == 0 {
+			return []string{s.InitiatorIPs[vapi.MainCluster]}, ctrl.Result{}, nil
+		}
+		pf, found := pfs.findFirstPodSorted(func(v *PodFact) bool {
+			return v.upNode
```
cchen-vertica, line 232: Use primary up node should be better 
```suggestion
			return v.upNode && v.IsSandboxPrimary()
```

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -153,27 +158,86 @@ func (s *SandboxSubclusterReconciler) sandboxSubclusters(ctx context.Context) (c
...
+			// The first subcluster in a sandbox turns into a primary. Set
+			// state in the vdb to indicate that.
+			if j == 0 {
+				_, err := vk8s.UpdateVDBWithRetry(ctx, s.VRec, s.Vdb, func() (bool, error) {
+					scMap := s.Vdb.GenSubclusterMap()
+					vdbSc, found := scMap[sc]
+					if !found {
+						return false, fmt.Errorf("subcluster %q missing in vdb %q", sc, s.Vdb.Name)
+					}
+					vdbSc.Type = vapi.SandboxPrimarySubcluster
```
cchen-vertica, line 186: Why do we need to update this? Isn't this set by the user in the spec? I think we just need to sandbox the subcluster with "sandboxprimary" first.

api/v1/verticadb_webhook.go
```diff
@@ -149,6 +145,44 @@ func (v *VerticaDB) validateImmutableFields(old runtime.Object) field.ErrorList
...
+	}
+	// Go through new object to see that existing subclusters have a valid type change.
+	for i := range v.Spec.Subclusters {
+		sc := v.Spec.Subclusters[i]
+		oldType, ok := nameToTypeMap[sc.Name]
+		// Can skip new subclusters
+		if !ok {
+			continue
+		}
+		if (oldType == PrimarySubcluster || oldType == TransientSubcluster) && oldType != sc.Type {
```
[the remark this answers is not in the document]
spilchen: Yes, if users want more control over the node types in the sandbox. I'll open a follow-on jira for this.

api/v1/verticadb_webhook.go
spilchen: Yes, that's correct. We could revisit this. VER-94031 is opened to address this.

api/v1/verticadb_webhook.go
spilchen: Yeah, it's possible this could be relaxed.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
spilchen: Yeah, I see your point. We should be using a primary. I think you mean `v.upNode && v.IsPrimary()` though?

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
spilchen: It makes it easier to sandbox if there is one less thing the user needs to update. Plus, if we needed the user to change the type of the subcluster they want as the primary, we would need to ensure we sandbox that subcluster first. I think it's easier if we assume the first subcluster listed in the sandbox becomes the primary.

cchen-vertica: I think we can improve this in VER-94031

cchen-vertica (approved):

PR #802, opened 2024-05-08
Set VerticaDB name in ConfigMap

spilchen:
The ConfigMap should have the name of the VerticaDB, not the database name. The sandbox controller uses this to figure out the correct VerticaDB to fetch.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -1308,7 +1308,7 @@ func BuildSandboxConfigMap(nm types.NamespacedName, vdb *vapi.VerticaDB, sandbox
 		// the data should be immutable since dbName and sandboxName are fixed
 		Immutable: &immutable,
 		Data: map[string]string{
-			vapi.VerticaDBNameKey: vdb.Spec.DBName,
+			vapi.VerticaDBNameKey: vdb.Name,
 			vapi.SandboxNameKey:   sandbox,
 		},
 	}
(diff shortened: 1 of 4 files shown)


pkg/builder/builder.go
```diff
@@ -1308,7 +1308,7 @@ func BuildSandboxConfigMap(nm types.NamespacedName, vdb *vapi.VerticaDB, sandbox
 		// the data should be immutable since dbName and sandboxName are fixed
 		Immutable: &immutable,
 		Data: map[string]string{
-			vapi.VerticaDBNameKey: vdb.Spec.DBName,
+			vapi.VerticaDBNameKey: vdb.Name,
```
roypaulin, line 1311: If the configmap name is <vdbname>-<sandbox> we don't need to pass the namespaced name

roypaulin (approved):
Looks good!

pkg/builder/builder.go
spilchen: The VerticaDB and ConfigMap are assumed to be in the same namespace.

cchen-vertica (approved):

pkg/builder/builder.go
roypaulin: I meant e configmap name is <vdbname>-<sandbox> we don't need to pass the namespaced name at 1294.

PR #803, opened 2024-05-08
Add unsandbox reconciler to vdb and sandbox controllers

cchen-vertica:
This PR added two unsandbox reconcilers: one in vdb controller and the other in sandbox controller. The one in vdb controller will update sandbox config map for triggering sandbox controller. The one in sandbox controller will call vclusterOps to unsandbox the subclusters, update vdb sandbox status, and delete the config map. After that, the restart reconciler in vdb controller should restart the nodes that have been unsandboxed.


pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,244 @@
...
+	// reconcile sandbox status for the subclusters that are already unsandboxed
+	if err := r.reconcileSandboxStatus(ctx); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// reconcile the sandbox config map if it expires
+	err, deleted := r.reconcileSandboxConfigMap(ctx)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
roypaulin, line 85: I think these two should be in  a separate reconciller so that this one can only be about unsandbox. So here ypu only unsandbox and update status and in a separate reconciler you check the status.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// Reconcile will unsandbox subclusters, remove the sandbox config map, and update sandbox status of vdb
+func (r *UnsandboxSubclusterReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	// no-op for ScheduleOnly init policy or enterprise db
+	if r.Vdb.Spec.InitPolicy == vapi.CommunalInitPolicyScheduleOnly || !r.Vdb.IsEON() {
+		return ctrl.Result{}, nil
+	}
+
+	// collect pod facts for main cluster
```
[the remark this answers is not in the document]
cchen-vertica: I made a copy of the sandbox podFacts in the constructor so here it collected main cluster's pod facts.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,244 @@
...
+		if sb.Name == r.ConfigMap.Data[vapi.SandboxNameKey] {
+			foundSbInStatus = true
+			break
+		}
+	}
+	// if we cannot find the sandbox in status, it means the sandbox has been unsandboxed
+	// or does not exist
+	if !foundSbInStatus {
+		cmName := r.ConfigMap.Name
+		err := r.Client.Delete(ctx, r.ConfigMap)
```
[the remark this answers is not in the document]
cchen-vertica: I think unsandbox controller deleting the map will be faster and intuitive. After unsandbox controller unsandbox the subclusters, it knows if it can delete the config map. We don't need to wait for a loop in vdb controller. Before vdb controller runs into unsandbox_reconciler, the config map could be picked up again in sandbox controller, that will have some cost.

pkg/controllers/vdb/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,139 @@
...
+	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
+	"github.com/vertica/vertica-kubernetes/pkg/names"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// UnsandboxSubclusterReconciler will update sandbox config maps for letting
+// sandbox controller to unsandbox the subclusters
+type UnsandboxSubclusterReconciler struct {
```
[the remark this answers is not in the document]
cchen-vertica: When sandbox status is changed by sandbox controller, the current reconciler(restart_reconciler) should be able to handle putting sandboxed subclusters in the main cluster by restarting the nodes.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
cchen-vertica: You mean a separate reconciler running before unsandboxsubcluster_reoncilier? That could work.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+	ConfigMap   *corev1.ConfigMap
+	InitiatorIP string // The IP of the pod that we run vclusterOps from
+}
+
+func MakeUnsandboxSubclusterReconciler(r *SandboxConfigMapReconciler, vdb *vapi.VerticaDB, log logr.Logger,
+	cli client.Client, pfacts *vdbcontroller.PodFacts, dispatcher vadmin.Dispatcher, configMap *corev1.ConfigMap) controllers.ReconcileActor {
+	pfactsForMainCluster := pfacts.Copy(vapi.MainCluster)
+	return &UnsandboxSubclusterReconciler{
+		SRec:       r,
+		Log:        log.WithName("SandboxControllerUnsandboxSubclusterReconciler"),
```
spilchen, line 54: The sandbox controller should already be a prefix.
```suggestion
		Log:        log.WithName("UnsandboxSubclusterReconciler"),
```

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// reconcileSandboxStatus will update sandbox status for the subclusters that are already unsandboxed
+func (r *UnsandboxSubclusterReconciler) reconcileSandboxStatus(ctx context.Context) error {
+	scSbInStatus := make(map[string]string)
+	for _, sb := range r.Vdb.Status.Sandboxes {
+		for _, sc := range sb.Subclusters {
+			scSbInStatus[sc] = sb.Name
+		}
+	}
```
spilchen, line 112: Roy was adding a helper to api/v1 to do this in #793 (`GenSubclusterSandboxStatusMap`). Perhaps, we can duplicate the helper in this PR -- second person in will be forced to resolve the conflict.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+	sbScMap := r.PFacts.FindUnsandboxedSubclustersStillInSandboxStatus(scSbInStatus)
+	for sb, scs := range sbScMap {
```
spilchen, line 114: Shouldn't this just take care of its own sandbox? The one defined in the ConfigMap. We shouldn't be doing this for all sandboxes. Otherwise, we could have multiple threads all doing the same thing.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+			"Failed to unsandbox subcluster %q", scName)
+		return err
+	}
+	r.SRec.Eventf(r.Vdb, corev1.EventTypeNormal, events.UnsandboxSubclusterSucceeded,
+		"Successfully unsandboxed subcluster %q", scName)
+	return nil
+}
+
+// updateSandboxStatus will update sandbox status in vdb
+func (r *UnsandboxSubclusterReconciler) updateSandboxStatus(ctx context.Context, sbName string, scNames []string) error {
```
spilchen, line 251: I think `scNames` are the names of the subclusters that are being removed from the sandbox. Perhaps a better name to denote this:
```suggestion
func (r *UnsandboxSubclusterReconciler) updateSandboxStatus(ctx context.Context, sbName string, unsandboxedScNames []string) error {
```

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+	updateStatus := func(vdbChg *vapi.VerticaDB) error {
+		// update the sandbox's subclusters in sandbox status
+		for i := len(vdbChg.Status.Sandboxes) - 1; i >= 0; i-- {
+			if vdbChg.Status.Sandboxes[i].Name == sbName {
+				vdbChg.Status.Sandboxes[i].Subclusters = vutil.SliceDiff(vdbChg.Status.Sandboxes[i].Subclusters, scNames)
+			}
+			if len(vdbChg.Status.Sandboxes[i].Subclusters) == 0 {
+				vdbChg.Status.Sandboxes = append(vdbChg.Status.Sandboxes[:i], vdbChg.Status.Sandboxes[i+1:]...)
+			}
```
spilchen, line 260: We are only updating a single sandbox in this function. Should we have an early out?
```suggestion
			if vdbChg.Status.Sandboxes[i].Name != sbName {
				continue
			}
			vdbChg.Status.Sandboxes[i].Subclusters = vutil.SliceDiff(vdbChg.Status.Sandboxes[i].Subclusters, scNames)
			if len(vdbChg.Status.Sandboxes[i].Subclusters) == 0 {
				vdbChg.Status.Sandboxes = append(vdbChg.Status.Sandboxes[:i], vdbChg.Status.Sandboxes[i+1:]...)
			}
			break
```

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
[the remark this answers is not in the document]
spilchen: I like that we delete the ConfigMap in here too. It's a clear distinction that once the ConfigMap is created, the sandbox controller is reponsible for its deletion. I noticed in the vdb controller we try to delete configMap's too. Is is necessary to delete it from there? Should we just have a single spot that does the deletion. That will be easier to reason about.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+	if deleted {
+		return ctrl.Result{}, nil
+	}
+
+	// only execute unsandbox op when unsandbox trigger id and sandbox name are set
+	if r.ConfigMap.Annotations[vmeta.SandboxControllerUnsandboxTriggerID] != "" &&
+		r.ConfigMap.Data[vapi.SandboxNameKey] != "" {
+		// we ignore the config map that contains wrong data,
+		// this kind of config map could be old or created by others
+		if r.ConfigMap.Data[vapi.VerticaDBNameKey] != r.Vdb.Name {
```
spilchen, line 95: Is this a legit error? The sandbox controller will fetch the vdb based on what's in the ConfigMap. So, this invariant should always hold.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+
+// Reconcile will unsandbox subclusters, remove the sandbox config map, and update sandbox status of vdb
+func (r *UnsandboxSubclusterReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	// no-op for ScheduleOnly init policy or enterprise db
+	if r.Vdb.Spec.InitPolicy == vapi.CommunalInitPolicyScheduleOnly || !r.Vdb.IsEON() {
+		return ctrl.Result{}, nil
+	}
+
+	// collect pod facts for main cluster
+	if err := r.PFacts.Collect(ctx, r.Vdb); err != nil {
```
spilchen, line 71: Can we move the collection to spot where we need to know its needed? For instance, if we come in here and there is no unsandboxing to do, we should not collect pod facts.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+	} else {
+		r.Log.Info("Requeue because there are no UP nodes in main cluster to execute unsandbox operation")
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	err := r.executeUnsandboxCommand(ctx)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	r.PFacts.Invalidate()
```
spilchen, line 175: `r.PFacts` is just a local copy since it collects for the main cluster. We should invalidate the pfacts that was passed into the contructor. As that is the one that has done collection for the sandbox.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,267 @@
...
+func (r *UnsandboxSubclusterReconciler) executeUnsandboxCommand(ctx context.Context) error {
+	unsandboxSbScMap := r.Vdb.GenSandboxSubclusterMapForUnsandbox()
+	sbName := r.ConfigMap.Data[vapi.SandboxNameKey]
+	scs, found := unsandboxSbScMap[sbName]
+	if !found {
+		r.Log.Info("Ignore the config map because the sandbox inside it does not need to be unsandboxed")
+		return nil
+	}
+	succeedScs := []string{}
+	for _, sc := range scs {
```
spilchen, line 191: Does the order of the subclusters matter? I would think that we need to unsandbox the one denoted as the primary last.

pkg/controllers/sandbox/sandbox_controller.go
```diff
@@ -150,11 +150,12 @@ func (r *SandboxConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 // Order matters in that some actors depend on the successeful execution of
 // earlier ones.
-func (r *SandboxConfigMapReconciler) constructActors(vdb *v1.VerticaDB, log logr.Logger,
-	prunner *cmds.ClusterPodRunner, pfacts *vdbcontroller.PodFacts, dispatcher vadmin.Dispatcher) []controllers.ReconcileActor {
+func (r *SandboxConfigMapReconciler) constructActors(vdb *v1.VerticaDB, log logr.Logger, prunner *cmds.ClusterPodRunner,
+	pfacts *vdbcontroller.PodFacts, dispatcher vadmin.Dispatcher, configMap *corev1.ConfigMap) []controllers.ReconcileActor {
 	// The actors that will be applied, in sequence, to reconcile a sandbox configmap.
 	return []controllers.ReconcileActor{
 		MakeVerifyDeploymentReconciler(r, vdb, log),
+		MakeUnsandboxSubclusterReconciler(r, vdb, log, r.Client, pfacts, dispatcher, configMap),
```
spilchen, line 158: I assume we don't need pods in the sandbox to be UP prior to unsandboxing. It's just an UP node from the main cluster, right?

pkg/controllers/vdb/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+	for sb := range unsandboxSbScMap {
+		nm := names.GenSandboxConfigMapName(r.Vdb, sb)
+		cm := &corev1.ConfigMap{}
+		err := r.Client.Get(ctx, nm, cm)
+		if err != nil {
+			r.Log.Error(err, "failed to retrieve sandbox config map", "configMapName", nm.Name)
+			return err
+		}
+		cm.Annotations[vmeta.SandboxControllerUnsandboxTriggerID] = uuid.NewString()
+		err = r.Client.Update(ctx, cm)
```
spilchen, line 82: Can we use the `SandboxConfigMapManager` for this? We do something similar to this to wake up the controller for upgrade.

pkg/controllers/vdb/sandbox_configmap.go
```diff
@@ -60,7 +60,7 @@ func (s *SandboxConfigMapManager) triggerSandboxController(ctx context.Context)
 		// This will ensure that we always set a new value
 		// for this annotation which will wake up the sandbox
 		// controller
-		vmeta.SandboxControllerTriggerID: triggerID,
+		vmeta.SandboxControllerUpgradeTriggerID: triggerID,
```
spilchen, line 63: Perhaps make this function more general so we can use it to trigger for upgrade as well as for unsandbox?

tests/e2e-leg-9/sandbox/50-assert.yaml
spilchen, line 1: We cannot assert that the subcluster sec1 came up after unsandboxing, right? I think it was because the vdb controller will attempt add node. We were going to use the restart option with unsandbox to work around that temporarily.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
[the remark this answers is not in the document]
cchen-vertica: Sure, I will remove the delete-config-map code in vdb controller.

pkg/controllers/vdb/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+	existingSbs := make(map[string]any)
+	for _, sb := range r.Vdb.Status.Sandboxes {
+		existingSbs[sb.Name] = struct{}{}
+	}
+	for i := range cmList.Items {
+		cm := &cmList.Items[i]
+		// only process the config maps for unsandbox operation
+		if cm.Annotations[vmeta.SandboxControllerUnsandboxTriggerID] != "" &&
+			cm.Data[vapi.SandboxNameKey] != "" {
+			if cm.Data[vapi.VerticaDBNameKey] != r.Vdb.Name {
```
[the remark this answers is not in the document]
cchen-vertica: Good to know we have a limit for listing objects. I will remove the whole function here. The config map deletion will be handled only by sandbox controller.

pkg/controllers/vdb/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+func (r *UnsandboxSubclusterReconciler) deleteExpiredSandboxConfigMaps(ctx context.Context) error {
+	cmList := &corev1.ConfigMapList{}
+	namespace := r.Vdb.GetNamespace()
+	err := r.Client.List(ctx, cmList, client.InNamespace(namespace))
+	if err != nil {
+		r.Log.Error(err, "failed to retrieve config map list")
+		return err
+	}
+	existingSbs := make(map[string]any)
+	for _, sb := range r.Vdb.Status.Sandboxes {
```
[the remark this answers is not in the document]
cchen-vertica: This entire function will be deleted, so nothing to do with this comment.

pkg/controllers/vdb/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+			if cm.Data[vapi.VerticaDBNameKey] != r.Vdb.Name {
+				r.Log.Info("Vdb name in sandbox config map doesn't match current vdb's, skip processing this config map",
+					"configMapName", cm.Name, "vdbNameInConfigMap", cm.Data[vapi.VerticaDBNameKey], "currentVdbName", r.Vdb.Name)
+				continue
+			}
+			sb := cm.Data[vapi.SandboxNameKey]
+			_, exists := existingSbs[sb]
+			// delete sandbox config maps for the non-existing sandboxes
+			if !exists {
+				cmName := cm.Name
```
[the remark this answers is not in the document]
cchen-vertica: This entire function will be deleted, so nothing to do with this comment.

pkg/controllers/vdb/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+
+// updateSandboxConfigMaps will add a trigger ID to sandbox config maps for triggering sandbox controller
+func (r *UnsandboxSubclusterReconciler) updateSandboxConfigMaps(ctx context.Context) error {
+	unsandboxSbScMap := r.Vdb.GenSandboxSubclusterMapForUnsandbox()
+	for sb := range unsandboxSbScMap {
+		nm := names.GenSandboxConfigMapName(r.Vdb, sb)
+		cm := &corev1.ConfigMap{}
+		err := r.Client.Get(ctx, nm, cm)
+		if err != nil {
+			r.Log.Error(err, "failed to retrieve sandbox config map", "configMapName", nm.Name)
```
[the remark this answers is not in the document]
cchen-vertica: In this case, I think we should sandbox first, then unsandbox. I cannot see a big advantage of switching the order. In the case someone simply renames the sandbox name in the vdb, the unsandbox only updated the config map, the sandbox will still need to wait until the next reconcile loop. It's the same as we sandbox first, then unsandbox.

pkg/controllers/vdb/podfacts.go
```diff
@@ -1184,3 +1184,32 @@ func checkIfNodeUpCmd(podIP string) string {
...
+	initiator, ok := p.findFirstPodSorted(func(v *PodFact) bool {
+		return v.sandbox == vapi.MainCluster && v.isPrimary && v.upNode
+	})
+	return initiator.podIP, ok
+}
+
+// FindUnsandboxedSubclustersStillInSandboxStatus returns a sandbox-subclustesrs map
+// that contains subclustesrs which has been unsandboxed but hasn't been removed
+// from sandbox status of VDB
+func (p *PodFacts) FindUnsandboxedSubclustersStillInSandboxStatus(scSbInVdbStatus map[string]string) map[string][]string {
```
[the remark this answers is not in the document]
cchen-vertica: This is for the case we successfully unsandboxed a subcluster but failed to update the status. This time the CR is not accurate.

go.sum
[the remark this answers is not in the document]
cchen-vertica: Right. After vclusterOps change is merged to github, I can update this.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
cchen-vertica: You are correct. Will remove this check.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
cchen-vertica: We need the podfacts to correct the sandbox status. The function right below it needs the pod facts.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
cchen-vertica: We only made a copy of pfacts that was passed into the constructor. Is there a need to invalidate that one? I think I can simply not invalidate r.PFacts here since unsandbox reconciler will collect podFacts every time it runs.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
cchen-vertica: The order doesn't matter. The unsandbox is done through main cluster nodes. If we unsandbox the primary subcluster first, the secondary subcluster can still be alive and unsandboxed. I tested this case using vcluster cli.

pkg/controllers/vdb/unsandboxsubcluster_reconciler.go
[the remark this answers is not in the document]
spilchen: But if you always sandbox first, won't we try to add the same subcluster to a second sandbox?

pkg/controllers/vdb/podfacts.go
```diff
@@ -1184,3 +1184,32 @@ func checkIfNodeUpCmd(podIP string) string {
...
 }
+
+// FindFirstPrimaryUpPodIP returns the ip of first pod that
+// has a primary up Vertica node, and a boolean that indicates
+// if we found such a pod
+func (p *PodFacts) FindFirstPrimaryUpPodIP() (string, bool) {
+	initiator, ok := p.findFirstPodSorted(func(v *PodFact) bool {
+		return v.sandbox == vapi.MainCluster && v.isPrimary && v.upNode
+	})
+	return initiator.podIP, ok
```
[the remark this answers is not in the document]
spilchen: But if initiator is nil, how are we able to follow the pointer to get `podIP`? It would be no different than this, which segfaulted in Go playground:
```
var initiator *PodFact = nil
fmt.Printf("%s\n", initiator.podIP)
```

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
spilchen: yeah, good point. I think the ConfigMap is currently only single threaded. But it may be multi-threaded in the future.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
spilchen: I guess they go into read-only mode then.

tests/e2e-leg-9/sandbox/50-assert.yaml
cchen-vertica: Unfortunately, even when we restart unsandboxed pods from vlcusterops, it cannot make the pod up. The node inside the pod is up, but the pod shows "CrashLoopBackOff"

cchen-vertica:
Create a separate Jira(VER-94063) to verify if we can move a subcluster between sandboxes. I think it can work with current design. If we have sc1 in sandbox1, and sc2 in sandbox2. If we switch the subclusters in the sandboxes, sc1 and sc2 will be recorded in config maps by vdb unsandbox reconciler, then sandbox controller will unsandbox them. The vdb will be updated after unsandbox operation which will trigger another round of vdb reconciliation.  In the new round, vdb restart reconciler and vdb sandbox reconciler should be able to sandbox sc1 and sc2 correctly.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -0,0 +1,275 @@
...
+	return ctrl.Result{}, nil
+}
+
+// reconcileSandboxStatus will update vdb for the subclusters that are already unsandboxed
+func (r *UnsandboxSubclusterReconciler) reconcileSandboxInfoInVdb(ctx context.Context) error {
+	scSbInStatus := r.Vdb.GenSubclusterSandboxStatusMap()
+	sbScMap := r.PFacts.FindUnsandboxedSubclustersStillInSandboxStatus(scSbInStatus)
+	for sb, scs := range sbScMap {
+		if sb == r.ConfigMap.Data[vapi.SandboxNameKey] {
+			err := r.updateSandboxInfoInVdb(ctx, sb, scs)
```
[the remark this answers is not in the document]
spilchen: The info message is nice. But I don't think we should mask out the error.

spilchen (commented):
I did another pass. I only had one comment left about the change to not return an error from reconcileSandboxInfoInVdb.

spilchen (approved):
Looks good. I'm approving even though you still need to up go.mod|sum after Chinh pushes his vcluster update. That should be a trivial change.

chinhtranvan:
@cchen-vertica you can pull vnext to get latest vclusterOps. I already merged it into vnext

PR #812, opened 2024-05-13
Decouple sandbox image from vdb.spec image

roypaulin:
When setting a sandbox, if the image is omitted, the operator will internally assume spec.image as the sandbox's image. The issue is that changing spec.image could trigger sandbox upgrade too. Now if the image is omitted, there is a new reconciler that will explicitly set the sandbox image in vdb to spec.image value.


spilchen (commented):
Would a webhook work? There is a function that is called to fill in default values (`func (v *VerticaDB) Default()`). That is called for each update to the vdb.

spilchen (approved):

PR #813, opened 2024-05-13
Applied new option --hosts-in-sandbox of vclusterOps start_db on the operator

cchen-vertica:
When restarting a pod in a database with a sandbox, it could fail when the IP of the pod changes. This is a bug in old vclusterOps code. We were trying to read node IPs from the catalog, but sandbox node IP change cannot be reflected in the main cluster catalog, and vice versa. This PR applied the new vclusterOps version and passed a new option to start_db to fix this issue. With this option, we will read main cluster node IPs from main cluster catalog and read sandbox node IPs from sandbox catalog.


pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -480,9 +480,20 @@ func (r *RestartReconciler) restartCluster(ctx context.Context, downPods []*PodF
 			startdb.WithConfigurationParams(r.ConfigurationParams.GetMap()),
 		)
 	}
+	hostsInOneSandbox := true
+	firstPodSandbox := ""
+	if len(downPods) > 0 && downPods[0].sandbox != vapi.MainCluster {
+		firstPodSandbox = downPods[0].sandbox
+	} else {
+		hostsInOneSandbox = false
+	}
```
roypaulin, line 489: `downPods` here should be assumed non-empty. All checks should be done before calling this function. It will also simplify the code in this block.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -480,9 +480,20 @@ func (r *RestartReconciler) restartCluster(ctx context.Context, downPods []*PodF
...
 	for i := range downPods {
+		if firstPodSandbox != downPods[i].sandbox {
+			hostsInOneSandbox = false
```
roypaulin, line 492: I think it is enough to check the sandbox name.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -480,9 +480,20 @@ func (r *RestartReconciler) restartCluster(ctx context.Context, downPods []*PodF
 			startdb.WithConfigurationParams(r.ConfigurationParams.GetMap()),
 		)
 	}
+	hostsInOneSandbox := true
```
roypaulin, line 483: Maybe `hostsInSandbox`? We only startdb on one cluster at a time.

spilchen (approved):
I wonder if this is related to the problem I saw in the upgrade demo. I'm curious to try my demo scenario again once this is in. I don't have any comments. I'm approving as it looks like Roy's comments should be minor changes anyway.

roypaulin (approved):

cchen-vertica:
I removed the e2e test change in this PR since those changes failed unsandbox tests. I will merge this PR first because it could improve the stability of the restart reconciler. Will add those e2e test changes back in future PR when vclusterOps bug(VER-94191) is fixed.

PR #814, opened 2024-05-13
Use subcluster status to update podfacts

roypaulin:
After unsandbox, the catalog dir is deleted. The issue we check its existence to set dbExists in the pod facts. With this change, dbExists is true if the catalog dir exists or if addedToDB field in the subcluster status detail for that specific pod is true.


pkg/controllers/vdb/podfacts.go
```diff
@@ -731,6 +737,7 @@ func (p *PodFacts) getEnvValueFromPod(cnt *corev1.Container, envName string) (st
 // If a db is found, we will set the vertica node name.
 func (p *PodFacts) checkIsDBCreated(_ context.Context, vdb *vapi.VerticaDB, pf *PodFact, gs *GatherState) error {
```
spilchen, line 738: Rather than a separate state, could we just overwrite whatever value is set for `dbExists`? Whatever is stored in the status field will supersede whatever we found in the gather script.
```
func (p *PodFacts) checkIsDBCreated(_ context.Context, vdb *vapi.VerticaDB, pf *PodFact, gs *GatherState) error {
	pf.dbExists = false

	// The gather state is empty if the pod isn't running.
	if pf.isPodRunning {
		pf.dbExists = gs.DBExists
		pf.vnodeName = gs.VNodeName
	}

	// Overwrite dbExists and vnodeName from state found in the vdb.status. We
	// cannot always trust on disk. When a pod is unsandboxed, the catalog is
	// removed for instance.
	scs, ok := vdb.FindSubclusterStatus(pf.subclusterName)
	if ok {
		// Set the db exists indicator first based on the count in the status
		// field.  We continue to check the path as we do that to figure out the
		// vnode.
		pf.dbExists = scs.AddedToDBCount > pf.podIndex
		// Inherit the vnode name if present
		if int(pf.podIndex) < len(scs.Detail) {
			pf.vnodeName = scs.Detail[pf.podIndex].VNodeName
			pf.dbExistsInStatus = scs.Detail[pf.podIndex].AddedToDB
		}
	}
	return nil
}
```

roypaulin: @spilchen Even if I update the status in `removenode_reconciler`,  `status_reconciler` will override those values so it will still be a problem in `obj_reconciler`. Moreover, pf.dbExists will no longer change as  it will getting its value from scs.Detail[pf.podIndex].AddedToDB which in turn will get its value from pf.dbExists, which is not good.

I think we still want to rely to some extent on the catalog dir too.

spilchen: Why would status_reconciler override the value? Do we invalidate the podfacts cache after removing the node?

roypaulin: Yes. But my point is if I update the subcluster status in removenode_reconciler(setting addedToDB for removed nodes to false), status reconciler will override that with the values from podfacts which will be true and later `obj_reconciler` will not be able to know that the pods have already been removed and will requeue.

spilchen: It sounds like status reconciler is using stale pod facts. The steps should have been:
- remove node succesfully called
- update addedToDB to false in vdb.status
- invalidate podfacts
- collect podfacts, should report dbExists as false
- update vdb.status using latest podfacts, keep addedToDB as false

roypaulin: I see your point and that works. Isn't it risky to modify subcluster status in multiple places?

spilchen: Actually, won't status_reconciler simply delete the entries for the subcluster detail based on the new size of the subcluster? We always remove the last node first.

spilchen (approved):

PR #815, opened 2024-05-16
Fix for many subclusters sandboxed at the same time

spilchen:
This update addresses an issue that occurs when multiple subclusters are added to a new sandbox during the same reconciliation iteration. The problem arose after sandboxing the first subcluster; the second subcluster failed to sandbox.

The failure was due to the need to provide both a host from the main cluster and a host from the sandbox to the vcluster API. To obtain the sandbox host, the system relies on the `vdb.status.sandbox` being updated. Since this update wasn't happening, the process of finding the sandbox host by checking pod facts didn't locate any sandboxed pods.

The solution is to update `vdb.status.sandboxes` as each subcluster is sandboxed. This update allows subsequent sandboxes to find the correct pod facts, ensuring the process completes successfully for all subclusters.

### DIFF
--- a/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
+++ b/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
@@ -229,7 +229,7 @@ func (s *SandboxSubclusterReconciler) findInitiatorIPs(ctx context.Context, sand
 		if err := pfs.Collect(ctx, s.Vdb); err != nil {
 			return nil, ctrl.Result{}, err
 		}
-		// If this is the first pod in the sandbox, then only an API from the
+		// If this is the first pod in the sandbox, then only an ip from the
 		// main cluster is needed.
 		if len(pfs.Detail) == 0 {
 			ips := []string{s.InitiatorIPs[vapi.MainCluster]}
--- a/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
+++ b/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
@@ -17,7 +17,6 @@ package vdb
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"reflect"
 
@@ -159,7 +158,6 @@ func (s *SandboxSubclusterReconciler) sandboxSubclusters(ctx context.Context) (c
 // executeSandboxCommand will call sandbox API in vclusterOps, create/update sandbox config maps,
 // and update sandbox status in vdb
 func (s *SandboxSubclusterReconciler) executeSandboxCommand(ctx context.Context, scSbMap map[string]string) (ctrl.Result, error) {
-	succeedSbScMap := make(map[string][]string)
 	seenSandboxes := make(map[string]any)
 
 	// We can simply loop over the scSbMap and sandbox each subcluster. However,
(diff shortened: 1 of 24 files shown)


tests/e2e-leg-9/verify-vrep/09-create-data-users-tlsconfig.yaml
```diff
@@ -61,7 +61,7 @@ data:
       echo "Assertion failed: expected 99, got $result"
       exit 1
     fi
-    result=$(kubectl exec $POD_NAME -i -c server -- bash -c "vsql -U dbadmin -w superuser -tAc \"SELECT COUNT(*) FROM tables;\"")
+    result=$(kubectl exec $POD_NAME -i -c server -- bash -c "vsql -U dbadmin -w superuser -tAc \"SELECT COUNT(*) FROM tables WHERE table_name = 'test_table';\"")
```
jizhuoyu, line 64: For these `WHERE table_name = 'test_table'` do we just add them as a general precaution or we're aware of some changes / future changes that will make `tables` not empty at the start of the test?

spilchen: No my CI runs failed without them. They were picking up the internal tables that have been recently added.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -232,7 +232,9 @@ func (s *SandboxSubclusterReconciler) findInitiatorIPs(ctx context.Context, sand
 		// If this is the first pod in the sandbox, then only an API from the
```
roypaulin, line 232:
```suggestion
		// If this is the first pod in the sandbox, then only an ip from the
```
?

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -378,7 +382,8 @@ func (s *SandboxSubclusterReconciler) sandboxSubcluster(ctx context.Context, sub
 	return ctrl.Result{}, nil
 }
 
-// updateSandboxStatus will update sandbox status in vdb
+// updateSandboxStatus will update sandbox status in vdb. This is a bulk update
+// and can handle multiple subclusters at once.
 func (s *SandboxSubclusterReconciler) updateSandboxStatus(ctx context.Context, originalSbScMap map[string][]string) error {
```
roypaulin, line 387: Looks like this is no longer needed.

tests/e2e-leg-9/verify-vrep/09-create-data-users-tlsconfig.yaml
jizhuoyu: Yeah I remember seeing some emails reminding devs of this change

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
jizhuoyu: I see this is still used in `reconcileSandboxStatus()`

tests/e2e-leg-9/verify-vrep/09-create-data-users-tlsconfig.yaml
chinhtranvan: Then, is this the new change from the server side? It is not related to the change in sandboxsubcluster_reconciler, right? @jizhuoyu, do you know the title of the email

spilchen: I think timing is involved. Last nights CI run was fine. But when I pulled last nights image down on my machine and ran the test I saw the issue.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
spilchen: Yes, we still need the bulk update.

roypaulin (approved):

tests/e2e-leg-9/verify-vrep/09-create-data-users-tlsconfig.yaml
jizhuoyu: I'm wondering whether this might be potentially related to Alex's email "Potential SF impact: internal tables have been merged"

jizhuoyu (approved):

PR #816, opened 2024-05-16
Add new test for sandbox upgrade

roypaulin:
This adds tests for sandbox upgrade:
- `sandbox-upgrade-happy-path`: runs a successful sandbox upgrade
- `sandbox-upgrade-bad-image`: starts an upgrade with a bad image, abort it and get it back to a good state

### DIFF
--- a/pkg/meta/annotations.go
+++ b/pkg/meta/annotations.go
@@ -20,7 +20,6 @@ import (
 	"strconv"
 	"strings"
 
-	vutil "github.com/vertica/vcluster/vclusterops/util"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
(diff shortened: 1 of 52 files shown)


spilchen:
Is `sandbox-upgrade-happy-path` needed? It seems like a subset of `sandbox-upgrade-bad-image` since that one finishes with a valid upgrade.

roypaulin:
> Is `sandbox-upgrade-happy-path` needed? It seems like a subset of `sandbox-upgrade-bad-image` since that one finishes with a valid upgrade.

I think both are needed. `sandbox-upgrade-happy-path` assumes the image is correct while `sandbox-upgrade-bad-image` checks that we can recover from a bad upgrade.

roypaulin:
`online-upgrade-kill-transient` is failing quite a lot. I will investigate.

spilchen (approved):

PR #817, opened 2024-05-17
Added a vertica image reconciler in sandbox controller

cchen-vertica:
This PR added a vertica image reconciler in sandbox controller. It could sync the vertica image across the main cluster. After we unsandboxed a subcluster, the subcluster could have different vertica image than the other subclusters in main cluster. This new reconciler will recreate the statefulset of that subcluster for fixing its vertica image.

### DIFF
--- a/pkg/controllers/sandbox/sandbox_controller.go
+++ b/pkg/controllers/sandbox/sandbox_controller.go
@@ -158,6 +158,8 @@ func (r *SandboxConfigMapReconciler) constructActors(vdb *v1.VerticaDB, log logr
 		MakeVerifyDeploymentReconciler(r, vdb, log),
 		// Move the subclusters from a sandbox to the main cluster
 		MakeUnsandboxSubclusterReconciler(r, vdb, log, r.Client, pfacts, dispatcher, configMap),
+		// Update the vertica image for unsandboxed subclusters
+		MakeVerticaImageReconciler(r, vdb, log, pfacts),
 		// Update the vdb status for the sandbox nodes/pods
 		vdbcontroller.MakeStatusReconciler(r.Client, r.Scheme, log, vdb, pfacts),
 		// Upgrade the sandbox using the offline method
--- a/pkg/controllers/sandbox/verticaimage_reconciler.go
+++ b/pkg/controllers/sandbox/verticaimage_reconciler.go
@@ -1,0 +1,111 @@
+/*
+ (c) Copyright [2021-2024] Open Text.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ You may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package sandbox
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	vapi "github.com/vertica/vertica-kubernetes/api/v1"
+	"github.com/vertica/vertica-kubernetes/pkg/builder"
+	"github.com/vertica/vertica-kubernetes/pkg/controllers"
+	vdbcontroller "github.com/vertica/vertica-kubernetes/pkg/controllers/vdb"
+	"github.com/vertica/vertica-kubernetes/pkg/names"
+	appsv1 "k8s.io/api/apps/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// VerticaImageReconciler will recreate the secondary subclusters' StatefulSets
+// in main cluster if those subclusters have a different vertica image than
+// the one in primary subclusters
+type VerticaImageReconciler struct {
+	SRec   *SandboxConfigMapReconciler
+	Vdb    *vapi.VerticaDB
+	Log    logr.Logger
+	PFacts *vdbcontroller.PodFacts
+}
+
+func MakeVerticaImageReconciler(r *SandboxConfigMapReconciler, vdb *vapi.VerticaDB,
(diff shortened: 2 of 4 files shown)


cchen-vertica:
I don't know how to test this in e2e test. We need to do an online upgrade for a sandbox. Is that doable in e2e test?

pkg/controllers/sandbox/sandbox_controller.go
```diff
@@ -158,6 +158,8 @@ func (r *SandboxConfigMapReconciler) constructActors(vdb *v1.VerticaDB, log logr
 		MakeVerifyDeploymentReconciler(r, vdb, log),
 		// Move the subclusters from a sandbox to the main cluster
 		MakeUnsandboxSubclusterReconciler(r, vdb, log, r.Client, pfacts, dispatcher, configMap),
+		// Update the vertica image for unsandboxed subclusters
+		MakeVerticaImageReconciler(r, vdb, log, pfacts),
```
spilchen, line 162: There may be problems if we do this after deleting the ConfigMap. Once the ConfigMap is deleted, the controller that is responsible for the pods has transferred back to the vdb controller. Is it possible to decouple the ConfigMap deletion for unsandbox into a separate reconciler that is called _after_ this reconciler? Or another approach is to work this reconciler into the vdb controller.

roypaulin: I prefer this reconciler in vdb controller. After unsandbox, the status are updated so the subcluster is no longer tied to the sandbox and should be handle by vdb controller. Instead of having `pfactsForMainCluster` ant not use the sandbox pfacts, you can simply implement this in the vdb controller.

pkg/controllers/vdb/podfacts.go
```diff
@@ -1204,3 +1204,43 @@ func (p *PodFacts) FindUnsandboxedSubclustersStillInSandboxStatus(scSbInVdbStatu
...
+		if v.isPrimary {
+			mcImage = v.image
+			break
+		}
+	}
+	// find secondary subclusters that has a different image
+	seenScs := make(map[string]any)
+	for _, v := range p.Detail {
+		// we expect the pfacts only contains the main cluster pods so we ignore the sandbox pods
+		if v.sandbox != vapi.MainCluster {
```
roypaulin, line 1228: You do not need this as a `PodFacts` object collects the pod facts for a specific sandbox or the main cluster and that info is in `p.SandboxName`

pkg/controllers/vdb/podfacts.go
```diff
@@ -1204,3 +1204,43 @@ func (p *PodFacts) FindUnsandboxedSubclustersStillInSandboxStatus(scSbInVdbStatu
 	}
 	return sbScMap
 }
+
+// FindSecondarySubclustersWithDifferentImage will scan the secondary subclusters in main cluster
+// and return the secondary subclusters that have different vertica image than main cluster
+func (p *PodFacts) FindSecondarySubclustersWithDifferentImage() []string {
```
roypaulin, line 1210: Can this be a general purpose function? A function that gets a an image an return a list of secondary subclusters not running that image. And the caller can pass in vdb.spec.Image.

pkg/controllers/vdb/podfacts.go
```diff
@@ -1204,3 +1204,43 @@ func (p *PodFacts) FindUnsandboxedSubclustersStillInSandboxStatus(scSbInVdbStatu
...
+		}
+		seenScs[v.subclusterName] = struct{}{}
+	}
+	return scs
+}
+
+// SetImage can set the image of a pod, it should only be used in
+// unit test of verticaimage_reconciler
+func (p *PodFact) SetImage(image string) {
+	p.image = image
```
roypaulin, line 1245: Why can't we just directly set this field?

pkg/controllers/vdb/podfacts.go
cchen-vertica: I think the pod facts have the most accurate image value. It's possible vdb.spec.Image is different than the one in a primary subcluster when the main cluster is under upgrade.

pkg/controllers/vdb/podfacts.go
cchen-vertica: Cause PodFact is in a different package, and its fields(e.g. image) aren't exposed to the outside.

pkg/controllers/vdb/podfacts.go
cchen-vertica: We don't know which podFacts the caller will use, if a future caller uses podFacts in a sandbox to call this function, that caller will get an empty slice. That is not likely to happen, but having one more check doesn't hurt.

pkg/controllers/vdb/podfacts.go
cchen-vertica: After I moved the image reconciler to vdb controller, this function isn't needed.

pkg/controllers/vdb/podfacts.go
roypaulin: Ok I see. Then update the function description to explain why we need to do this(unsandbox). Also looks like this is only useful for main cluster, then just do an early check at the beginning of the function, something like:
```
if p.GetSandboxName() != vapi.MainCluster {
       return []string{}
}
```

pkg/controllers/vdb/podfacts.go
roypaulin: I see. Then that also means we should not be using vdb.spec.image to set the image in the statefulset's spec like it is done in that reconciler.

pkg/controllers/sandbox/verticaimage_reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+		if err != nil {
+			r.Log.Error(err, "failed to recreate statefulset", "Name", nm)
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+// recreateSts will drop then create the statefulset
+func (r *VerticaImageReconciler) recreateSts(ctx context.Context, curSts, expSts *appsv1.StatefulSet) error {
```
roypaulin, line 97: I see there is a lot of code duplicated from `obj_reconciler`. Can you check if you can reuse `obj_reconciler`(with some changes) or at least avoid code dupliacation.

PR #818, opened 2024-05-17
Fixed polling timeout issue of start_db and start_node

cchen-vertica:
In k8s, the node state polling timeout is controlled by the annotation "vertica.com/restart-timeout". If it's not set, the polling timeout is set to 0 which will fail vclusterOps start_db and start_node. Then we will have many cluster/node restart failure events. This PR simply used vclusterOps default polling timeout if the annotation "vertica.com/restart-timeout" is not set.

### DIFF
--- a/pkg/vadmin/restart_node_vc.go
+++ b/pkg/vadmin/restart_node_vc.go
@@ -62,7 +62,10 @@ func (v *VClusterOps) genStartNodeOptions(s *restartnode.Parms, certs *HTTPSCert
 	opts.UserName = su
 	opts.Password = &v.Password
 	opts.Nodes = s.RestartHosts
-	opts.StatePollingTimeout = v.VDB.GetRestartTimeout()
+	vdbTimeout := v.VDB.GetRestartTimeout()
+	if vdbTimeout != 0 {
+		opts.StatePollingTimeout = vdbTimeout
+	}
 	if v.VDB.IsNMASideCarDeploymentEnabled() {
 		opts.StartUpConf = paths.StartupConfFile
 	}
--- a/pkg/vadmin/start_db_vc.go
+++ b/pkg/vadmin/start_db_vc.go
@@ -91,7 +91,10 @@ func (v *VClusterOps) genStartDBOptions(s *startdb.Parms, certs *HTTPSCerts) (vo
 	opts.Password = &v.Password
 
 	// timeout option
-	opts.StatePollingTimeout = v.VDB.GetRestartTimeout()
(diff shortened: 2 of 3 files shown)


pkg/vadmin/restart_node_vc.go
```diff
@@ -62,7 +62,10 @@ func (v *VClusterOps) genStartNodeOptions(s *restartnode.Parms, certs *HTTPSCert
 	opts.UserName = su
 	opts.Password = &v.Password
 	opts.Nodes = s.RestartHosts
-	opts.StatePollingTimeout = v.VDB.GetRestartTimeout()
+	vdbTimeout := v.VDB.GetRestartTimeout()
```
spilchen, line 65: Is this related to a specific change in vcluster? Or does this issue exist in 24.2.0. If it's an issue with 24.2.0, you may want to put the change into main first and pick it up in vnext via a merge.

chinhtranvan: * For restart_node, I think the reason comes from my change: https://git.verticacorp.com/projects/VER/repos/server/pull-requests/20821/diff#platform/vcluster/vclusterops/start_node.go?f=149. I moved the default value to func (options *VStartNodesOptions) setDefaultValues(). Thank you for finding this error.

* For start_db, I think it exists in 24.2.0.

spilchen: Because it is start_db, it probably should be in main and a changie entry added. The changie should only say you are fixing start_db. start_node is fine in prior releases.

roypaulin:
I was this in `annotations.go`
```
// GetRestartTimeout returns the timeout to use for restart node or start db. If
// 0 is returned, this means to use the default.
func GetRestartTimeout(annotations map[string]string) int {
	return lookupIntAnnotation(annotations, RestartTimeoutAnnotation, 0 /* default value */)
}
```

Should we also set the default value here to 20min? Because admintools too gets the timeout value from this annotation

pkg/vadmin/restart_node_vc.go
cchen-vertica: start_db is fine in main. The change to make start_db fail is made last month in this PR https://git.verticacorp.com/projects/VER/repos/server/pull-requests/20685/diff#platform/vcluster/vclusterops/start_db.go?f=111

We don't need to put the change into main.

roypaulin: I also suspect this is why the upgrade test fails in my PR

roypaulin (approved):
I am approving assuming start_db is fine in 24.2.0.
This will also fix my current PR.

PR #820, opened 2024-05-23
Fixed bugs in sandbox and unsandbox

cchen-vertica:
This PR applies the latest vclusterOps and passed more node information to vclusterOps for fixing the bugs in sandbox and unsandbox. The bugs can be reproduced when we kill a pod in a subcluster before we sandbox it and kill a pod in a sandbox before we unsandbox it. The issue is cause by the stale in the catalog. The vcluster re-ip only can modify the nodes' IPs on either main cluster or the sandbox. The new vclusterOps will do one more re-ip before sandboxing/unsandboxing.


tests/e2e-leg-9/sandbox-basic/47-kill-pods-in-sec3.yaml
roypaulin, line 1: I though all the nodes in the subcluster must be up before sandboxing. Do you expect them to be up before step 50?

tests/e2e-leg-9/sandbox-basic/57-kill-pods-in-sec3.yaml
roypaulin, line 1: Why do you need to kill the pods before unsandboxing?

tests/e2e-leg-9/sandbox-basic/47-kill-pods-in-sec3.yaml
cchen-vertica: Yes, I expect them to be up in step 50. There is one issue I had: if I assert upNodeCount in step 50, it can work in my local env but it cannot work in github CI since the upNodeCount field is missing in the CI.

cchen-vertica: I will add upNodeCount in step 50 and try it again.

tests/e2e-leg-9/sandbox-basic/57-kill-pods-in-sec3.yaml
cchen-vertica: This is a failure case this PR needs to address. When we kill a pod in the sandbox, and then do unsandboxing, the main cluster will have stale IPs so unsandboxing will fail. This step is for testing this case.

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -225,11 +232,38 @@ func (r *UnsandboxSubclusterReconciler) processConfigMap(ctx context.Context) er
...
+	// are the latest ones in the database, and vclusterOps will compare them with the ones in catalog
+	// of main cluster. If vclusterOps find catalog of main cluster has stale node addresses,
+	// it will use the correct addresses in this map to do a re-ip before unsandboxing.
+	nodeNameAddressMap := r.OriginalPFacts.FindNodeNameAndAddressInSubcluster(scName)
+
+	// remove startup.json in pod since vcluster unsandbox needs to poll node down.
+	// With that json file, the container will restart vertica automatically and fail
+	// vcluster unsandbox
+	podNames := r.OriginalPFacts.FindPodNamesInSubcluster(scName)
+	rmCmd := []string{"bash", "-c", fmt.Sprintf("rm -rf %s", paths.StartupConfFile)}
```
roypaulin, line 249: As you suggested I am going to move this to `shutdown_subcluster`.

roypaulin (approved):

PR #822, opened 2024-05-28
print a more useful error message if sandbox is supplied but empty of nodes

fenic-fawkes:


pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -225,6 +225,13 @@ func (r *ReplicationReconciler) collectPodFacts(ctx context.Context) (err error)
 // choose the source host and target host
 // (first host where db is up in the specified cluster)
 func (r *ReplicationReconciler) determineSourceAndTargetHosts() (err error) {
+	if len(r.SourcePFacts.Detail) == 0 && r.TargetPFacts.SandboxName != "" {
```
jizhuoyu, line 228: Typo? Should be `r.SourcePFacts.SandboxName`?

jizhuoyu, line 228: Do we prefer checking pod facts or checking status fields of `vdb`? Or either is OK? @cchen-vertica

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -225,6 +225,13 @@ func (r *ReplicationReconciler) collectPodFacts(ctx context.Context) (err error)
...
+		return fmt.Errorf("source sandbox '%s' does not exist or has no nodes assigned to it", r.SourcePFacts.SandboxName)
+	}
+	if len(r.TargetPFacts.Detail) == 0 && r.TargetPFacts.SandboxName != "" {
+		return fmt.Errorf("target sandbox '%s' does not exist or has no nodes assigned to it", r.TargetPFacts.SandboxName)
```
roypaulin, line 232: Can you move this to its own function and call that function after `collectPodFacts()`? And also add unit test.

pkg/controllers/vrep/replication_reconciler.go
roypaulin, line 228:
```suggestion
	if len(r.SourcePFacts.Detail) == 0 && r.TargetPFacts.SandboxName != vapi.MainCluster {
```
minor: a const we use and that is equal to empty string

roypaulin: Checking pod facts is enough

roypaulin (approved):

PR #825, opened 2024-05-31
Fixed intermittent failure in online-upgrade-kill-transient test

cchen-vertica:
This PR applied new vclusterOps library to fix intermittent failure in online-upgrade-kill-transient test. That intermittent failure is caused by IP inconsistency between the transient node and the catalog. In the new vclusterOps library, we won't check the version of the transient node in start_node, and we will do re-ip before removing the transient node.

### DIFF
--- a/pkg/vadmin/remove_sc_vc.go
+++ b/pkg/vadmin/remove_sc_vc.go
@@ -70,6 +70,8 @@ func (v *VClusterOps) genRemoveSubclusterOptions(s *removesc.Parms, certs *HTTPS
 	opts.DataPrefix = v.VDB.Spec.Local.DataPath
 	opts.IsEon = v.VDB.IsEON()
 	opts.ForceDelete = true
+	opts.PrimaryUpHost = s.InitiatorIP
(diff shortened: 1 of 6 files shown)


pkg/vadmin/remove_sc_vc.go
```diff
@@ -70,6 +70,8 @@ func (v *VClusterOps) genRemoveSubclusterOptions(s *removesc.Parms, certs *HTTPS
 	opts.DataPrefix = v.VDB.Spec.Local.DataPath
 	opts.IsEon = v.VDB.IsEON()
 	opts.ForceDelete = true
+	opts.PrimaryUpHost = s.InitiatorIP
```
roypaulin, line 73: Why weren't we using this before?

cchen-vertica: This is a new one I added to vclusterOps two weeks ago. Before, remove_subcluster doesn't do re_ip. Now I use this one and NodeNameAddressMap to do re_ip in remove_subcluster.

chinhtranvan (approved):
Looks good to me!

roypaulin (approved):

PR #826, opened 2024-06-03
parse serviceName in VerticaReplicator

fenic-fawkes:
allow users to provide a service name to use for source and target databases. if unspecified we use the first service assigned to the database with a cluster ip, if supplied we validate that it's assigned to the appropriate main/sandbox cluster


pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -238,33 +236,67 @@ func (r *ReplicationReconciler) checkSandboxExists() error {
...
-		return
+
+	if r.Vrep.Spec.Target.ServiceName != "" {
+		r.SourceInfo.IP, err = r.getServiceIP(ctx, r.Vrep.Spec.Target.ServiceName, r.Vrep.Spec.Target.VerticaDB)
+		if err != nil {
+			return err
+		}
 	} else {
-		r.TargetInfo.IP = upPodIP
+		// assume target must not be read-only, no subcluster constraints
```
roypaulin, line 282: If service name is empty you should pick the first primary service.

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -238,33 +236,67 @@ func (r *ReplicationReconciler) checkSandboxExists() error {
 	return nil
 }
 
+func (r *ReplicationReconciler) getServiceIP(ctx context.Context, svcName, dBname string) (ip string, err error) {
```
roypaulin, line 239: A simpler solution would be to just check if the service name belongs to one of the subclusters. If that's the case you fetch it and get the ip. You don't even need to check the labels.

pkg/controllers/vrep/replication_reconciler_test.go
roypaulin, line 1: We should also add a test case for that.

roypaulin:
We should also update e2e test. Let's discuss about this at the standup.

pkg/controllers/vrep/replication_reconciler_test.go
fenic-fawkes: are there examples of tests i could base one off of?

roypaulin: There is a e2e test `verify-vrep`. You probably need to update that one. @jizhuoyu can help.

fenic-fawkes: ok. i mostly got tests to work. i'm having trouble getting the test to generate a service for the sandbox with a cluster ip, but i'll keep working on it

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -238,33 +239,67 @@ func (r *ReplicationReconciler) checkSandboxExists() error {
...
+	}
+	err = fmt.Errorf("could not find service %s in %s", svcName, clusterName)
+	if svcName == "" {
+		err = fmt.Errorf("could not find a service in %s with an assigned ClusterIP", clusterName)
+	}
+	return "", err
+}
+
 // choose the source host and target host
 // (first host where db is up in the specified cluster)
```
jizhuoyu, line 281: could we update the comment to reflect the new logic related to service name?

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -238,33 +239,67 @@ func (r *ReplicationReconciler) checkSandboxExists() error {
 	return nil
 }
 
+func (r *ReplicationReconciler) getServiceHostname(ctx context.Context, vdb *vapi.VerticaDB, sandboxName, svcName string) (ip string, err error) {
```
jizhuoyu, line 242: could we update the comment for this function to explain the behavior, especially when `svcName` is empty string?

jizhuoyu:
> _ No description provided. _

Could we update the PR description please?

pkg/controllers/vrep/replication_reconciler_test.go
fenic-fawkes: oh, i'm a goomba. the services are being created fine but the statefulset isn't.
i could use help creating a sandbox in the tests, it doesn't look like there's an example in that file

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -238,33 +239,74 @@ func (r *ReplicationReconciler) checkSandboxExists() error {
...
+func (r *ReplicationReconciler) getServiceHostname(ctx context.Context, vdb *vapi.VerticaDB,
+	sandboxName, svcName string) (ip string, err error) {
+	finder := iter.MakeSubclusterFinder(r.VRec.Client, vdb)
+	var stss *appsv1.StatefulSetList
+	stss, err = finder.FindStatefulSets(ctx, iter.FindInVdb, sandboxName)
+	if err != nil {
+		return "", err
+	}
+
+	for i := range stss.Items {
```
roypaulin, line 255: An easier way would be:
- If svcName is non-empty: to iterate over vdb.spec.subclusters, get the servicename(`names.GenExtSvcName(vdb, sc)`). check if it is equal svcName, check if the sc sandboxName is equal to sandboxName.
- If svcName is empty: iterate over vdb.spec.subclusters, pick the 1st primary subcluster's service. main cluster primaries have `type = primary` while sandbox primaries have `type = sandboxPrimary`.

To know if a sc belongs to a sandbox you need to check the status. There are already functions like `vdb.GenSubclusterSandboxStatusMap()` or `vdb.GetSandboxStatus()` that can help.

fenic-fawkes:
needs one small vcluster change that's under review but then this should be good to merge

pkg/vadmin/replication_start_vc.go
```diff
@@ -68,11 +68,11 @@ func (v *VClusterOps) genReplicateDBOptions(s *replicationstart.Parms, certs *HT
 	} else {
 		opts.TargetPassword = &s.TargetPassword
 	}
-	opts.TargetHosts = append(opts.TargetHosts, s.TargetIP)
+	opts.TargetHosts = append(opts.TargetHosts, s.TargetHostname)
 	opts.SourceTLSConfig = s.SourceTLSConfig
 	opts.IsEon = v.VDB.IsEON()
 
-	opts.IPv6 = net.IsIPv6(s.SourceIP)
+	opts.IPv6 = net.IsIPv6(s.SourceHostname)
```
roypaulin, line 75: This only works with ips. Maybe we should just use the service clusterIP

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -238,33 +237,65 @@ func (r *ReplicationReconciler) checkSandboxExists() error {
...
+		clusterName += " (sandbox " + sandboxName + ")"
+	}
+
+	if svcName != "" {
+		inputName := types.NamespacedName{Namespace: vdb.Namespace, Name: svcName}
+		for i := range vdb.Spec.Subclusters {
+			sc := &vdb.Spec.Subclusters[i]
+			scSvcName := names.GenExtSvcName(vdb, sc)
+			if inputName == scSvcName && sandboxName == vdb.GetSubclusterSandboxName(sc.Name) {
+				return fmt.Sprintf("%s.%s.svc.cluster.local", scSvcName.Name, scSvcName.Namespace), nil
```
roypaulin, line 257: For precaution try to retrieve the service.

cchen-vertica:
I've synced the vclusterOps to the Github repository. You can update your go.mod to refer to the latest commit in vcluster repo.

pkg/controllers/vrep/replication_reconciler.go
fenic-fawkes: you mean the way that i did previously?

pkg/vadmin/replication_start_vc.go
fenic-fawkes: my issue is that if we use IPs then it could break tls. if the tls config being used for on the source side is validating hostnames it could validate the service name but not the ip address. in general we should prefer hostnames when tls might be involved. should i just remove this check?

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -238,33 +237,92 @@ func (r *ReplicationReconciler) checkSandboxExists() error {
...
+	if svcName != "" {
+		return scSvcName.Name == svcName, r.validateServiceIP(ctx, scSvcName)
+	}
+
+	if vdb.GetSubclusterSandboxName(sc.Name) != sandboxName {
+		return false, nil
+	}
+
+	return sc.IsPrimary(), r.validateServiceIP(ctx, scSvcName)
+}
```
roypaulin, line 270:
```suggestion
func (r *ReplicationReconciler) shouldPickService(ctx context.Context, vdb *vapi.VerticaDB, sc *vapi.Subcluster,
	sandboxName, svcName string) should bool {
	if vdb.GetSubclusterSandboxName(sc.Name) != sandboxName {
		return false
	}
	if svcName != "" {
		return sc.GetServiceName() == svcName
	}

	return sc.IsPrimary(), r.validateServiceIP(ctx, scSvcName)
}
```

roypaulin (commented):
Very close. Sorry if what I say is often confusing. Let me know If I misunderstood what you are trying to do.

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -238,33 +237,92 @@ func (r *ReplicationReconciler) checkSandboxExists() error {
...
+	sandboxName, svcName string) (should bool, err error) {
+	scSvcName := types.NamespacedName{Namespace: vdb.Namespace, Name: sc.ServiceName}
+	if scSvcName.Name == "" {
+		scSvcName = names.GenExtSvcName(vdb, sc)
+	}
+	if svcName != "" {
+		return scSvcName.Name == svcName, r.validateServiceIP(ctx, scSvcName)
+	}
+
+	if vdb.GetSubclusterSandboxName(sc.Name) != sandboxName {
```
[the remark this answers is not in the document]
fenic-fawkes: the reason i did it this way is so that specifying a service is enough to direct traffic to the sandbox. if you feel strongly that we shouldn't allow that behavior i'll reverse this

pkg/controllers/vrep/replication_reconciler.go
fenic-fawkes: the issue is that `sc.GetServiceName()` only returns the subcluster part of the service name if `sc.ServiceName` is empty. `k get service` will list service names as `<db-name>-<sc-name>`, so we should be comparing against the full service name, no? maybe this is just an artifact of the tests, but the `names.GenExtSvcName` call was required to get them to pass

fenic-fawkes: discussed with roy, i understand the desired behavior now. we should be comparing against the non-vdb portion of the service name

roypaulin:
Almost done! Resolve the conflicts and run the tests again.

PR #827, opened 2024-06-04
Comply with PSDL

roypaulin:
This addresses security issues.


cchen-vertica:
Looks good. I will approve this after you pass the e2e-leg-5 tests. 

It seems you tried to fix all the critical and high vulnerabilities. You can just fix the critical one if other fixes cause the failure in the test.

cchen-vertica (approved):

PR #834, opened 2024-06-15
Closed user sessions before db replication starts in db upgrade

cchen-vertica:
This PR used vsql to close all active user sessions before database replication starts in the replicated upgrade. This can speed up the upgrade process. This is a temporary change. We will improve this by pausing all user sessions soon.

### DIFF
--- a/pkg/controllers/vdb/upgrade.go
+++ b/pkg/controllers/vdb/upgrade.go
@@ -642,6 +642,26 @@ func (i *UpgradeManager) isSubclusterIdle(ctx context.Context, pfacts *PodFacts,
 	return res, nil
 }
 
+// closeAllSessions will run a query to close all active user sessions.
+func (i *UpgradeManager) closeAllSessions(ctx context.Context, pfacts *PodFacts) error {
+	pf, ok := pfacts.findFirstPodSorted(func(v *PodFact) bool {
+		return v.isPrimary && v.upNode
+	})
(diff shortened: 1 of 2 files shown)


pkg/controllers/vdb/upgrade.go
```diff
@@ -642,6 +642,26 @@ func (i *UpgradeManager) isSubclusterIdle(ctx context.Context, pfacts *PodFacts,
...
+func (i *UpgradeManager) closeAllSessions(ctx context.Context, pfacts *PodFacts) error {
+	pf, ok := pfacts.findFirstPodSorted(func(v *PodFact) bool {
+		return v.isPrimary && v.upNode
+	})
+	if !ok {
+		i.Log.Info("No pod found to run vsql. Skipping close all sessions")
+		return nil
+	}
+
+	sql := "select close_all_sessions();"
```
roypaulin, line 655: I am not an expert. Is it possible to close sessions on a specific subcluster?

cchen-vertica: There isn't an easy way to do that. We need to find all sessions using a query, then close the sessions one by one. And there isn't value in this. At this point, no one can connect to the sandbox since we haven't redirected the service yet. Close all sessions is ok, and we will replace this strategy soon with session draining.

roypaulin (approved):

PR #835, opened 2024-06-17
Ignore includeUIDInPath annotation when reviving db

roypaulin:
When `vertica.com/include-uid-in-path` is set to `true`, the operator will include the VerticaDB's UID in the path. The issue is that when reviving a db, we expect the path to an existing communal storage so that path should not be altered to avoid failure during revivedb. This PR adds a webhook rule that prevents setting the annotation to `true` when initPolicy is `Revive`


roypaulin:
I agree, adding a webhook is better to inform the user.

roypaulin:
I agree, adding a webhook rule is better.

PR #837, opened 2024-06-18
VER-93564-podfact-package

HaoYang0000:
This PR:
1. Move the podfact package out of vdb and its test file
2. Keeps the original podfact attribute private, adding Getters and Setters for other packages to assign or edit value(Otherwise we need to expose the attribute of podfact)
3. Adjust import and attribute reference in all packages and tests


HaoYang0000:
Duplicate as PR: https://github.com/vertica/vertica-kubernetes/pull/837

PR #838, opened 2024-06-18
support subcluster name "default_subcluster" in replicated upgrade

chinhtranvan:
This PR changes underscore to hyphen for the service name in the operator.


pkg/controllers/vdb/replicatedupgrade_reconciler.go
```diff
@@ -672,6 +672,8 @@ func (r *ReplicatedUpgradeReconciler) addNewSubclusters() (bool, error) {
 		if found {
 			continue
 		}
+		// replace underscore to hypen
+		sc.GenCompatibleFQDN()
```
cchen-vertica, line 676: This will change the subcluster names. In current behavior, if we create a subcluster named "default_subcluster"(not in replicated upgrade), that subcluster will still have the same name "default_subcluster". You shouldn't change the subcluster name. Instead, change the service name.

chinhtranvan: We don't have an issue with the service name. In https://github.com/vertica/vertica-kubernetes/blob/vnext/api/v1/verticadb_webhook.go#L518, sc.GetServiceName() already replaced the underscore with a hyphen. We got an error after changing the subcluster name in the statefulset at line https://github.com/vertica/vertica-kubernetes/blob/vnext/api/v1/verticadb_webhook.go#L505

```
"error": "failed trying to update VDB with new subclusters: admission webhook \"vverticadb.v1.kb.io\" denied the request: VerticaDB.vertica.com \"v-rep-target\" is invalid: spec.subclusters[5].name: Invalid value: \"v-rep-target-sec_1-sb\": subcluster name is not valid, change it so that 1. its length is greater than 0 and smaller than 254, 2. it matches regex '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'"}
```

We just need to replace underscore to hypen for the statefulset name to fix the issue

cchen-vertica (approved):

PR #841, opened 2024-06-21
Change Replicated upgrade policy to Online

roypaulin:
The new upgrade policy will be `Online` instead of `Replicated`. What was formerly called `Online` is now `ReadOnlyOnline` upgrade


cchen-vertica:
Even when online-upgrade(replicated upgrade) is robust, I think we still need to support read-only online-upgrade because it's much faster (no sandbox needed) and cheaper (no extra pods needed). If this is the case, read-only online-upgrade will be another policy type.

roypaulin:
> Even when online-upgrade(replicated upgrade) is robust, I think we still need to support read-only online-upgrade because it's much faster (no sandbox needed) and cheaper (no extra pods needed). If this is the case, read-only online-upgrade will be another policy type.

That's probably better. We need to document that change.

api/v1/verticadb_types.go
```diff
@@ -128,17 +128,17 @@ type VerticaDBSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:Auto","urn:alm:descriptor:com.tectonic.ui:select:Online","urn:alm:descriptor:com.tectonic.ui:select:Offline"}
 	// +kubebuilder:default:=Auto
 	// This setting defines how the upgrade process will be managed. The
-	// available values are Offline, Online, Replicated, and Auto.
+	// available values are Offline, Online, and Auto.
```
cchen-vertica, line 131: Add ReadOnlyOnline here

api/v1/verticadb_types.go
```diff
@@ -416,12 +416,14 @@ const (
 	// leaving the secondary subclusters in read-only mode.  When the primary
 	// subcluster comes back up, we restart/remove all of the secondary
 	// subclusters to take them out of read-only mode.
-	OnlineUpgrade UpgradePolicyType = "Online"
-	// Like online upgrade, however it allows for writes. This is done by
+	// This is only used internally by the operator to differentiate between
```
cchen-vertica, line 419: ReadOnlyOnlineUpgrade shouldn't be used internally.

config/manifests/bases/verticadb-operator.clusterserviceversion.yaml
```diff
@@ -227,6 +227,14 @@ spec:
         path: template.isPrimary
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:booleanSwitch
+      - description: A sandbox primary subcluster is a secondary subcluster that was
```
cchen-vertica, line 230: Thanks for adding this.

tests/e2e-leg-9/new-online-upgrade-sanity/co-assert.yaml
```diff
@@ -0,0 +1,77 @@
+# (c) Copyright [2021-2024] Open Text.
```
cchen-vertica, line 1: What is this for? When will it be verified?

tests/e2e-leg-9/new-online-upgrade-sanity/co-wait-for-sc-promotion.yaml
```diff
@@ -0,0 +1 @@
+# Intentionally empty to give this step a name in kuttl
```
cchen-vertica, line 1: Ditto

cchen-vertica:
Looks good overall. I only have some minor comments.

api/v1/verticadb_types.go
roypaulin: That's an old comment. Not relevant anymore, will update it

tests/e2e-leg-9/new-online-upgrade-sanity/co-assert.yaml
roypaulin: Should have ever been committed. Will remove it.

cchen-vertica (approved):

PR #843, opened 2024-06-25
Remove sandbox configmap after promotion

roypaulin:
After promoting sandbox to main, we need to delete sandbox config map to avoid any kind of conflicts as the sandbox does not exist anymore.

### DIFF
--- a/tests/e2e-leg-9/replicated-upgrade-sanity/40-errors.yaml
+++ b/tests/e2e-leg-9/replicated-upgrade-sanity/40-errors.yaml
@@ -25,3 +25,12 @@ apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   name: v-base-upgrade-sec1
+---
+apiVersion: v1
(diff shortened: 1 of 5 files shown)


cchen-vertica (approved):
Looks good to me. Once all the e2e tests pass, you can merge it.

tests/e2e-leg-9/replicated-upgrade-sanity/40-errors.yaml
```diff
@@ -24,4 +24,13 @@ metadata:
...
+  name: v-base-upgrade-sec1
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: v-base-upgrade-replica-group-b
+data:
+  sandboxName: replica-group-b
+  verticaDBName: v-base-upgrade
+immutable: true
```
chinhtranvan, line 36: I know I'm late  but what does immutable: true mean?

PR #853, opened 2024-07-09
set DisableNonReplicatableQueries before sandboxing subcluster during online upgrade

fenic-fawkes:
mostly based on the work in https://github.com/vertica/vertica-kubernetes/pull/831 plus fixing merge conflicts, etc


pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -293,6 +320,58 @@ func (r *OnlineUpgradeReconciler) assignSubclustersToReplicaGroupB(ctx context.C
...
+// postDisableNonReplicatableQueriesMsg will update the status message to indicate that
+// we are going to disable non-replicatble queries by setting config parameter DisableNonReplicatableQueries.
+func (r *OnlineUpgradeReconciler) postDisableNonReplicatableQueriesMsg(ctx context.Context) (ctrl.Result, error) {
+	return r.postNextStatusMsg(ctx, disableNonReplicatableQueriesMsgInx)
+}
+
+// setConfigParamDisableNonReplicatableQueries will run the reconciler to set the config parameter
+// DisableNonReplicatableQueries to true ("1") at database level within a given cluster
+func (r *OnlineUpgradeReconciler) setConfigParamDisableNonReplicatableQueries(ctx context.Context) (ctrl.Result, error) {
+	return r.setConfigParamDisableNonReplicatableQueriesImpl(ctx, ConfigParamDisableNonReplicatableQueriesValueTrue, vapi.MainCluster)
```
roypaulin, line 352: We should do this only if the original value is not "1".

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -293,6 +320,58 @@ func (r *OnlineUpgradeReconciler) assignSubclustersToReplicaGroupB(ctx context.C
...
+}
+
+// restoreConfigParamDisableNonReplicatableQueries will run the reconciler to restore the old value of config parameter
+// DisableNonReplicatableQueries before upgrade within a given cluster
+func (r *OnlineUpgradeReconciler) restoreConfigParamDisableNonReplicatableQueries(ctx context.Context) (ctrl.Result, error) {
+	return r.setConfigParamDisableNonReplicatableQueriesImpl(ctx, r.originalConfigParamDisableNonReplicatableQueriesValue, r.sandboxName)
```
roypaulin, line 358: We should do this only if old value != new value

pkg/controllers/vdb/getconfigparameter_reconciler.go
roypaulin, line 1: I don't think this needs to be a reconciler. It can be a simple struct with the function to perform `get` and `set`.
Let's talk about this.

scripts/guess-server-upgrade-base-image.sh
```diff
@@ -73,7 +73,7 @@ then
         print_vertica_k8s_img_with_tag $PRIVATE_REPO $PRIVATE_IMAGE "1f759615f0f723080b398edcf096a0bc8bc03aef-minimal"
     else
         # For all newer versions, we pick a random 24.3.0 image.
-        print_vertica_k8s_img_with_tag $PRIVATE_REPO $PRIVATE_IMAGE "20240526-minimal"
+        print_vertica_k8s_img_with_tag $PRIVATE_REPO $PRIVATE_IMAGE "20240701-minimal"
```
roypaulin, line 76: Why did you change the base image?t
The version in `docker-push-extra-vertica` in Makefile must always be greater than the base image.

fenic-fawkes: the server changes mean that we need an image that has the http endpoint change, which in this case means the image built on 24/7/12

pkg/catalog/config_parameter_vcluster.go
roypaulin, line 1: I suggest you move this to controllers/vdb and you can just use `dispatcher` from `onlineUpgradeReconciler` to call vcluster apis. These 2 functions are used only by online upgrade and in case we need them in a separate package we can just move them to their own package.

scripts/guess-server-upgrade-base-image.sh
```diff
@@ -73,7 +73,7 @@ then
         print_vertica_k8s_img_with_tag $PRIVATE_REPO $PRIVATE_IMAGE "1f759615f0f723080b398edcf096a0bc8bc03aef-minimal"
     else
         # For all newer versions, we pick a random 24.3.0 image.
-        print_vertica_k8s_img_with_tag $PRIVATE_REPO $PRIVATE_IMAGE "20240701-minimal"
+        print_vertica_k8s_img_with_tag $PRIVATE_REPO $PRIVATE_IMAGE "20240712-minimal"
```
roypaulin, line 76: You should change the version in tests/e2e-leg-9/multi-online-upgrade/30-initiate-upgrade.yaml and in `docker-push-extra-image`(in Makefile) to something newer than this version.

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -329,15 +328,17 @@ func (r *OnlineUpgradeReconciler) postQueryOriginalConfigParamDisableNonReplicat
...
-	)
-	r.Manager.traceActorReconcile(actor)
-	return actor.Reconcile(ctx, &ctrl.Request{})
+	initiator, ok := pf.findFirstUpPod(false /*not allow read-only*/, "" /*arbitrary subcluster*/)
+	if !ok {
+		r.Log.Info("No Up nodes found. Requeue reconciliation.")
+		return ctrl.Result{Requeue: true}, nil
+	}
+	vc := catalog.MakeVCluster(r.VDB, pf.VerticaSUPassword, initiator.podIP, r.Log, r.VRec.Client, r.VRec.EVRec)
+	r.originalConfigParamDisableNonReplicatableQueriesValue, err = vc.GetConfigurationParameter(ConfigParamDisableNonReplicatableQueries,
```
roypaulin, line 339: Does that mean that the min server version that supports online upgrade is 24.4?
@cchen-vertica what should we do? The next release is 24.3 and it does not support `non replicatable` as @fenic-fawkes added a change to the server recently.

pkg/controllers/vdb/onlineupgrade_reconciler.go
roypaulin: What about this?

fenic-fawkes: removed this for now. since this parameter is upgrade-only, it shouldn't ever be set by users

cchen-vertica:
You don't need to modify the added test step(36). It looks good. 

Once you fixed the conflicts and passed the test, I will approve it.

pkg/controllers/vdb/upgrade.go
```diff
@@ -719,3 +720,27 @@ func (i *UpgradeManager) getSubclusterNameFromSts(ctx context.Context, stsName s
...
+func (i *UpgradeManager) checkAllSubscriptionsActive(ctx context.Context, pfacts *PodFacts) (ctrl.Result, error) {
+	pf, ok := pfacts.findFirstPodSorted(func(v *PodFact) bool {
+		return v.isPrimary && v.upNode
+	})
+	if !ok {
+		i.Log.Info("No pod found to run vsql. Requeueing")
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	cmd := []string{"-tAc", "SELECT count(*) FROM node_subscriptions WHERE subscription_state not in ('ACTIVE', 'INACTIVE');"}
```
cchen-vertica, line 733: Can the subscriptions be "INACTIVE"? Why not using "subscription_state != 'ACTIVE'"?

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -340,6 +374,74 @@ func (r *OnlineUpgradeReconciler) assignSubclustersToReplicaGroupB(ctx context.C
...
+// DisableNonReplicatableQueries at database level within main cluster
+func (r *OnlineUpgradeReconciler) queryOriginalConfigParamDisableNonReplicatableQueries(ctx context.Context) (res ctrl.Result, err error) {
+	pf := r.PFacts[vapi.MainCluster]
+	initiator, ok := pf.findFirstUpPod(false /*not allow read-only*/, "" /*arbitrary subcluster*/)
+	if !ok {
+		r.Log.Info("No Up nodes found. Requeue reconciliation.")
+		return ctrl.Result{Requeue: true}, nil
+	}
+	vc := catalog.MakeVCluster(r.VDB, pf.VerticaSUPassword, initiator.podIP, r.Log, r.VRec.Client, r.VRec.EVRec)
+	r.originalConfigParamDisableNonReplicatableQueriesValue, err = vc.GetConfigurationParameter(ConfigParamDisableNonReplicatableQueries,
```
roypaulin, line 394: You should check if this is already set so we don't call `vc.GetConfigurationParameter()` everytime

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -340,6 +374,74 @@ func (r *OnlineUpgradeReconciler) assignSubclustersToReplicaGroupB(ctx context.C
...
+// postDisableNonReplicatableQueriesMsg updates the status message to indicate that
+// we are going to disable non-replicatable queries by setting config parameter DisableNonReplicatableQueries.
+func (r *OnlineUpgradeReconciler) postDisableNonReplicatableQueriesMsg(ctx context.Context) (ctrl.Result, error) {
+	return r.postNextStatusMsg(ctx, disableNonReplicatableQueriesMsgInx)
+}
+
+// setConfigParamDisableNonReplicatableQueries sets the config parameter
+// DisableNonReplicatableQueries to true ("1") at database level within a given cluster
+func (r *OnlineUpgradeReconciler) setConfigParamDisableNonReplicatableQueries(ctx context.Context) (ctrl.Result, error) {
+	if r.originalConfigParamDisableNonReplicatableQueriesValue != "1" {
```
roypaulin, line 408: Add a new inx(`setConfigParamInx`) value to `OnlineUpgradeStepInxAnnotation` for this step and check that `OnlineUpgradeStepInxAnnotation` value is not greater than that inx before you continue in this function.
See how `OnlineUpgradeStepInxAnnotation` is generally used.

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -88,13 +102,14 @@ const (
...
-	Dispatcher  vadmin.Dispatcher
-	sandboxName string // name of the sandbox created for replica group B
+	VRec                                                  *VerticaDBReconciler
+	Log                                                   logr.Logger
+	VDB                                                   *vapi.VerticaDB
+	PFacts                                                map[string]*PodFacts // We have podfacts for main cluster and replica sandbox
+	Manager                                               UpgradeManager
+	Dispatcher                                            vadmin.Dispatcher
+	sandboxName                                           string // name of the sandbox created for replica group B
+	originalConfigParamDisableNonReplicatableQueriesValue string
```
roypaulin, line 112: Aren't we going to set back the original value when upgrade is done?

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -340,6 +374,74 @@ func (r *OnlineUpgradeReconciler) assignSubclustersToReplicaGroupB(ctx context.C
...
+// we are going to query the original value of config parameter DisableNonReplicatableQueries.
+func (r *OnlineUpgradeReconciler) postQueryOriginalConfigParamDisableNonReplicatableQueriesMsg(
+	ctx context.Context) (ctrl.Result, error) {
+	return r.postNextStatusMsg(ctx, queryOriginalConfigParamDisableNonReplicatableQueriesMsgInx)
+}
+
+// queryOriginalConfigParamDisableNonReplicatableQueries gets value of the config parameter
+// DisableNonReplicatableQueries at database level within main cluster
+func (r *OnlineUpgradeReconciler) queryOriginalConfigParamDisableNonReplicatableQueries(ctx context.Context) (res ctrl.Result, err error) {
+	pf := r.PFacts[vapi.MainCluster]
```
roypaulin, line 387: You should check if `OnlineUpgradeStepInxAnnotation` value is greater than `setConfigParamInx` before going further in this func.

roypaulin (commented):
Merge the main into your branch to get the latest changes.

pkg/controllers/vdb/upgrade.go
fenic-fawkes: running a test, yes. subscriptions can be inactive and stay that way. for our use case we don't really care, this check is enough to make sure that sandbox can be run successfully

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -368,6 +405,84 @@ func (r *OnlineUpgradeReconciler) assignSubclustersToReplicaGroupB(ctx context.C
...
+// postClearConfigParamDisableNonReplicatableQueriesMsg updates the status message to indicate that
+// we are going to clear the config parameter DisableNonReplicatableQueries.
+func (r *OnlineUpgradeReconciler) postClearConfigParamDisableNonReplicatableQueriesMsg(ctx context.Context) (ctrl.Result, error) {
+	return r.postNextStatusMsg(ctx, clearDisableNonReplicatableQueriesMsgInx)
+}
+
+// clearConfigParamDisableNonReplicatableQueries clears the config parameter
+// DisableNonReplicatableQueries from the sandbox
+func (r *OnlineUpgradeReconciler) clearConfigParamDisableNonReplicatableQueries(ctx context.Context) (ctrl.Result, error) {
+	if vmeta.GetOnlineUpgradeStepInx(r.VDB.Annotations) >= upgradeSandboxInx {
```
roypaulin, line 461: It is better to add a new index: 
```suggestion
	if vmeta.GetOnlineUpgradeStepInx(r.VDB.Annotations) > clearConfigParamInx {
```

pkg/events/event.go
```diff
@@ -18,92 +18,98 @@ package events
...
+	DepotResized                       = "DepotResized"
+	MgmtFailed                         = "MgmtFailed"
+	MgmtFailedDiskFull                 = "MgmtFailedDiskfull"
+	LowLocalDataAvailSpace             = "LowLocalDataAvailSpace"
+	WrongImage                         = "WrongImage"
+	MonolithicContainerNotSupported    = "MonolithicContainerNotSupported"
+	InstallPackagesStarted             = "InstallPackagesStarted"
+	InstallPackagesFailed              = "InstallPackagesFailed"
+	InstallPackagesFinished            = "InstallPackagesFinished"
+	SetConfigurationParameterStarted   = "SetConfigurationParameterStarted"
```
cchen-vertica, line 104: These config param events are not used in other places. Can we remove them?

pkg/controllers/vdb/upgrade.go
cchen-vertica: My understanding is: if subscriptions on the nodes of a subcluster are inactive, the subcluster doesn't have coverage to do the sandbox. What do I miss?

tests/e2e-leg-9/new-online-upgrade-bad-image/32-assert.yaml
```diff
@@ -21,20 +21,12 @@ status:
...
-  apiVersion: vertica.com/v1
-  kind: VerticaDB
-  name: v-upgrade-bad-image
----
 apiVersion: vertica.com/v1
 kind: VerticaDB
 metadata:
   name: v-upgrade-bad-image
+  annotations:
+    vertica.com/online-upgrade-step-index: "4"
```
roypaulin, line 29: 4 means sandboxing is progress.

pkg/controllers/vdb/upgrade.go
fenic-fawkes: i'm not sure. all i know is that running rebalance_shards didn't change the subscription_state and sandboxing went fine

cchen-vertica: subscription_state probably is "PENDING" after add_node(before rebalance_shards) so this check will be fine. But the safer way is to verify all subscriptions are "ACTIVE".

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -68,7 +79,10 @@ var onlineUpgradeStatusMsgs = []string{
 const (
 	startOnlineUpgradeStatusMsgInx = iota
 	createNewSubclustersStatusMsgInx
+	queryOriginalConfigParamDisableNonReplicatableQueriesMsgInx
+	disableNonReplicatableQueriesMsgInx
 	sandboxSubclustersMsgInx
+	clearDisableNonReplicatableQueriesMsgInx
```
cchen-vertica, line 85: Remove this index

cchen-vertica (approved):

PR #861, opened 2024-07-23
Handle existing sandbox before online upgrade

roypaulin:
If online upgrade is done when there is already a sandbox, unsandboxing that sandbox after upgrade will fail. To prevent that, online upgrade will not be allowed if a sandbox exists. The operator will pick readonlyonlineupgrade


roypaulin:
@cchen-vertica I couldn't test online upgrade with existing subcluster in my own PC(I can't set up a k8s cluster). Can you try it with my branch? Then if it does not work I will add a webhook in this PR.

cchen-vertica:
> @cchen-vertica I couldn't test online upgrade with existing subcluster in my own PC(I can't set up a k8s cluster). Can you try it with my branch? Then if it does not work I will add a webhook in this PR.

I tested it. It cannot work and give me the error "cannot unsandbox a regular subcluster [sc1]".

roypaulin:
Instead of adding a webhook, I preferred being consistent with how we have been treating this kind of case. The operator will just switch to another policy if onlineupgrade cannot be done.

PR #886, opened 2024-07-31
Create restore point before replication

roypaulin:
Before starting replication, we back up the db. This will allow users to recover in case online upgrade fails.

Moreover an e2e test was added to make sure we con recover from a wrong image.

### DIFF
--- a/pkg/controllers/vdb/onlineupgrade_reconciler.go
+++ b/pkg/controllers/vdb/onlineupgrade_reconciler.go
@@ -57,6 +57,7 @@ var onlineUpgradeStatusMsgs = []string{
 	"Preparing replication",
 	"Back up database before replication",
 	"Replicate new data from main cluster to sandbox",
+	"Back up database after replication",
 	"Redirect connections to sandbox",
 	"Promote sandbox to main cluster",
 	"Remove original main cluster",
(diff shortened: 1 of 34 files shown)


pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -52,6 +54,8 @@ var onlineUpgradeStatusMsgs = []string{
 	"Promote secondaries whose base subcluster is primary",
 	"Upgrade sandbox to new version",
 	"Pause connections to main cluster",
+	"Preparing replication",
+	"Back up database before replication",
 	"Replicate new data from main cluster to sandbox",
 	"Redirect connections to sandbox",
```
cchen-vertica, line 60: Need to add "Back up database after replication" above this line

cchen-vertica (commented):
We need to treat the same version as a wrong image version. Probably need to change the version check accordingly.

roypaulin:
> We need to treat the same version as a wrong image version. Probably need to change the version check accordingly.

I will address this on a separate PR. The change is not that simple as we now have to take the hotfix into consideration, which means update some structs, the version parsing logic and tests.

cchen-vertica (approved):

PR #890, opened 2024-08-05
E2e push operator

HaoYang0000:
This PR add an action to check if docker image exists on remote internal repo(github private repo), if not we will push to it.


.github/workflows/build-images.yml
```diff
@@ -363,6 +363,15 @@ jobs:
         else
           make docker-build-operator
         fi
+        # Check if image exists in private repo, if not exist, push to repo
```
cchen-vertica, line 366: Have e2e tests run at this step? Before push image to private repo, we need to wait for all e2e tests to be done successfully.

cchen-vertica:
The target branch should be main, not vnext2. In your PR, could you do me a favor to remove vnext from e2e.yml since vnext branch has been removed?

.github/workflows/build-images.yml
HaoYang0000: No, this is build step for the e2e test. I thought for github one we just want to push if the image is not there. For the private repo(Opentext one) we want to check for e2e test result. Is that correct, or we want to check for both

HaoYang0000:
> The target branch should be main, not vnext2. In your PR, could you do me a favor to remove vnext from e2e.yml since vnext branch has been removed?

Done, removed in https://github.com/vertica/vertica-kubernetes/pull/890/commits/47b732b9df1d6fc1d81b4ad0c39e49175c8422b0

HaoYang0000:
Fix the CLA issue, create a new PR: https://github.com/vertica/vertica-kubernetes/pull/928

PR #900, opened 2024-08-19
poll for subcluster up in un/sandbox

fenic-fawkes:
fixes an issue where if pods are removed during sandboxing they will never be brought back up

### DIFF
--- a/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
+++ b/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
@@ -29,6 +29,7 @@ import (
 	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
 	"github.com/vertica/vertica-kubernetes/pkg/names"
 	"github.com/vertica/vertica-kubernetes/pkg/vadmin"
+	"github.com/vertica/vertica-kubernetes/pkg/vadmin/opts/pollscstate"
 	"github.com/vertica/vertica-kubernetes/pkg/vadmin/opts/sandboxsc"
 	"github.com/vertica/vertica-kubernetes/pkg/vdbstatus"
 	"github.com/vertica/vertica-kubernetes/pkg/vk8s"
@@ -203,6 +204,11 @@ func (s *SandboxSubclusterReconciler) executeSandboxCommand(ctx context.Context,
 			err = s.addSandboxedSubclusterToStatus(ctx, sc, sb)
 			if err != nil {
 				return ctrl.Result{}, err
+			}
+
+			res, err = s.waitForSandboxScUp(ctx, sb, sc)
+			if err != nil {
+				return res, err
 			}
 		}
 	}
@@ -446,4 +452,16 @@ func (s *SandboxSubclusterReconciler) addSandboxedSubclusterToStatus(ctx context
 		return nil
 	}
 	return vdbstatus.Update(ctx, s.Client, s.Vdb, updateStatus)
+}
+
+func (s *SandboxSubclusterReconciler) waitForSandboxScUp(ctx context.Context, sb, sc string) (ctrl.Result, error) {
+	initiatorIPs, res, err := s.findInitiatorIPs(ctx, sb)
+	if verrors.IsReconcileAborted(res, err) {
+		return res, err
+	}
+	err = s.Dispatcher.PollSubclusterState(ctx,
+		pollscstate.WithInitiators(initiatorIPs),
+		pollscstate.WithSubcluster(sc),
+	)
+	return ctrl.Result{Requeue: err != nil}, err
 }
--- a/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
+++ b/pkg/controllers/vdb/sandboxsubcluster_reconciler.go
@@ -173,9 +173,9 @@ func (s *SandboxSubclusterReconciler) executeSandboxCommand(ctx context.Context,
 			sb, found := scSbMap[sc]
 			if !found {
 				// assume it is already sandboxed
-				if i == 0 {
+				if j == 0 {
 					// we need the sandbox primary sc to be up so we can tell it about the sc being added
-					if res, err := s.waitForSandboxScUp(ctx, sb, sc); err != nil {
+					if res, err := s.waitForSandboxScUp(ctx, vdbSb.Name, sc); err != nil {
 						return res, err
 					}
 				}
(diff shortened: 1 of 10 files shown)


pkg/vadmin/opts/pollscstate/pollscstate.go
```diff
@@ -0,0 +1,43 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package pollscstate
+
+// Parms holds all of the options for FetchNodeState API call.
```
cchen-vertica, line 18: Change "FetchNodeState" to "PollSubclusterState"

pkg/vadmin/poll_sc_state_vc.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// required options
+	opts.DBName = v.VDB.Spec.DBName
+	opts.IsEon = v.VDB.IsEON()
+	opts.RawHosts = append(opts.RawHosts, s.InitiatorIPs...)
+	v.Log.Info("Setup poll sc state options", "hosts", opts.RawHosts)
+	opts.IPv6 = net.IsIPv6(s.InitiatorIPs[0])
+
+	opts.SCName = s.Subcluster
+
```
cchen-vertica, line 62: The new interface should accept a timeout option, and we can pass a value from vdb(in a new annotation) to that option like 'RestartTimeoutAnnotation'.

pkg/vadmin/poll_sc_state_vc.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// PollSubclusterState waits for a subcluster to come up
+func (v *VClusterOps) PollSubclusterState(ctx context.Context, opts ...pollscstate.Option) (err error) {
+	v.setupForAPICall("PollSubclusterState")
+	defer v.tearDownForAPICall()
+	v.Log.Info("Starting vcluster PollSubclusterState")
+
+	// get the certs
+	certs, err := v.retrieveNMACerts(ctx)
```
cchen-vertica, line 34: You don't need NMA certs for poll node states API.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -204,6 +205,11 @@ func (s *SandboxSubclusterReconciler) executeSandboxCommand(ctx context.Context,
 			if err != nil {
 				return ctrl.Result{}, err
 			}
+
+			res, err = s.waitForSandboxScUp(ctx, sb, sc)
+			if err != nil {
+				return res, err
```
cchen-vertica, line 211: We should just log the error and not end this method so other subclusters can be added to the sandbox without a reiteration.

fenic-fawkes: if this is the first subcluster to be sandboxed we can't continue until it's up, since we need to tell both main and sandbox about new subclusters being added.

pkg/vadmin/poll_sc_state_vc.go
fenic-fawkes: it currently uses the environment variable NODE_STATE_POLLING_TIMEOUT, you're saying we should add another way to configure the timeout?

cchen-vertica: Yes, that environment variable cannot be set by K8s users; it's only for VCluster users.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
cchen-vertica: If we poll subclsuter up failed, we will never come here again since the subcluster is correct in vdb.Status so next retry will skip that subcluster. We need to call this function in an earlier stage.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -206,6 +212,7 @@ func (s *SandboxSubclusterReconciler) executeSandboxCommand(ctx context.Context,
 				return ctrl.Result{}, err
 			}
 
+			s.Log.Info("waiting for newly sandboxed subcluster to be up", "sandbox", sb, "subcluster", sc)
 			res, err = s.waitForSandboxScUp(ctx, sb, sc)
 			if err != nil {
 				return res, err
```
cchen-vertica, line 218: Since we polled primary subcluster above, we don't need to return error here.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -173,6 +173,12 @@ func (s *SandboxSubclusterReconciler) executeSandboxCommand(ctx context.Context,
 			sb, found := scSbMap[sc]
 			if !found {
 				// assume it is already sandboxed
+				if j == 0 {
+					// we need the sandbox primary sc to be up so we can tell it about the sc being added
+					if res, err := s.waitForSandboxScUp(ctx, vdbSb.Name, sc); err != nil {
+						return res, err
```
cchen-vertica, line 179: Rather than returning an error, we should return a "ctrl.Result{Requeue: true}, nil" since error process is expensive in K8s.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -455,11 +462,20 @@ func (s *SandboxSubclusterReconciler) addSandboxedSubclusterToStatus(ctx context
 }
 
 func (s *SandboxSubclusterReconciler) waitForSandboxScUp(ctx context.Context, sb, sc string) (ctrl.Result, error) {
-	initiatorIPs, res, err := s.findInitiatorIPs(ctx, sb)
-	if verrors.IsReconcileAborted(res, err) {
-		return res, err
+	pfs := s.PFacts.Copy(sb)
+	if err := pfs.Collect(ctx, s.Vdb); err != nil {
```
cchen-vertica, line 466: Collect pod facts is expensive. We should collect it once for all subclusters in the sandbox. We should move it outside this function or change the argument `sc string` to `scs []string`. And move the line 216 below line 220.

cchen-vertica (approved):

PR #902, opened 2024-08-23
Install more packages for e2e tests in dockerfile

cchen-vertica:
This PR adds a new building argument to dockerfile of Vertica images. Devops will build the new daily Vertica images using the new dockerfile. Then our e2e tests don't need to access Vertica VPN any more.


cchen-vertica:
There will be a follow-up PR to update e2e tests after Devops team start using the new arg in their pipeline.

fenic-fawkes (approved):

PR #905, opened 2024-08-26
Processed hotfix number in vertica version

cchen-vertica:
This is an improvement of Vertica version process. Sometimes we support a feature from a specific hotfix of a version. For the low hotfix number, we don't support the feature. This PR allows us to process and compare hotfix numbers in the version.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -22,7 +22,7 @@ import (
 )
 
 type Components struct {
-	VdbMajor, VdbMinor, VdbPatch int
+	VdbMajor, VdbMinor, VdbPatch, VdbHotfix int
 }
 
 type Info struct {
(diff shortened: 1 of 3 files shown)


pkg/version/version.go
```diff
@@ -75,14 +75,25 @@ func MakeInfoFromStrCheck(curVer string) (*Info, error) {
...
 
+// HasEqualOrNewerHotfix checks if both versions have the same major, minor, patch numbers, and
+// if hotfix number in source version is equal or newer than the one in target version
+func (i *Info) HasEqualOrNewerHotfix(inVer string) bool {
+	comp, ok := parseVersion(inVer)
+	if !ok {
+		panic(fmt.Sprintf("could not parse input version: %s", inVer))
+	}
+	res := i.compareVersion(comp)
+	return res == compareEqual && i.VdbHotfix >= comp.VdbHotfix
```
fenic-fawkes, line 94: why not allow a newer major/minor? looks like this is only used by isOnlineUpgradeSupported which should accept 24.3-2 and >= 24.4, right?

fenic-fawkes (approved):

PR #907, opened 2024-08-28
Tighten the check for expired sandbox config map

cchen-vertica:
There is an ND error during the online upgrade. After a successful sandboxing, the config map is deleted by the sandbox controller which is not desired. There is a delay for sandbox controller to get the latest sandbox status. I made the expired sandbox check more strict in this PR. You can download the error logs from Jira VER-96105.

### DIFF
--- a/pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
+++ b/pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
@@ -134,7 +134,7 @@ func (r *UnsandboxSubclusterReconciler) reconcileSandboxConfigMap(ctx context.Co
 	cmName := r.ConfigMap.Name
 	sb := r.Vdb.GetSandboxStatus(sbName)
 	// if the sandbox doesn't have any subclusters, we delete the config map
-	if r.OriginalPFacts.IfNoPodsInSandbox(sbName) && (sb == nil || len(sb.Subclusters) == 0) {
+	if r.OriginalPFacts.IsSandboxEmpty(sbName) && (sb == nil || len(sb.Subclusters) == 0) {
 		err := r.Client.Delete(ctx, r.ConfigMap)
 		if err != nil {
 			r.Log.Error(err, "failed to delete expired sandbox config map", "configMapName", cmName)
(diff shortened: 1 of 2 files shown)


fenic-fawkes (approved):

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
```diff
@@ -126,11 +126,15 @@ func (r *UnsandboxSubclusterReconciler) reconcileSandboxInfoInVdb(ctx context.Co
 // reconcileSandboxConfigMap will update/delete sandbox config map if it expires, this function will return
 // an error and a boolean to indicate if sandbox config map is deleted
 func (r *UnsandboxSubclusterReconciler) reconcileSandboxConfigMap(ctx context.Context) (error, bool) {
+	if err := r.OriginalPFacts.Collect(ctx, r.Vdb); err != nil {
```
roypaulin, line 129: Maybe this is a sign that reconcileConfigMap should be in a separate reconciler.

pkg/controllers/vdb/podfacts.go
```diff
@@ -1290,3 +1290,11 @@ func (p *PodFacts) quorumCheckForRestartCluster(restartOnly bool) bool {
 	})
 	return restartablePrimaryNodeCount >= (primaryNodeCount+1)/2
 }
+
+// IfNoPodsInSandbox returns true if we cannot find any pods in the target sandbox
+func (p *PodFacts) IfNoPodsInSandbox(sandbox string) bool {
```
roypaulin, line 1295:
```suggestion
func (p *PodFacts) AnyPodsInSandbox(sandbox string) bool {
```
Or `IsSandboxEmpty()`

pkg/controllers/sandbox/unsandboxsubcluster_reconciler.go
cchen-vertica: I do have it in vdbcontroller.sandboxsubcluster_reconciler. But I want to keep one in sandbox controller too. The reconcilers inside sandbox controller will all have this issue. In the future, we can improve the method we used to udpate vdb: after we update vdb, we should poll/assign the vdb changes to the VDB of reconciler.

roypaulin (approved):

PR #910, opened 2024-08-29
Adds support for Multi-Arch image builds

narphu:
This PR updates the Make targets and Dockerfiles to allow passing a TARGET_ARCH argument that allows creating a Multi-architecture image, specifically intended towards ARM64 support. This does not yet create or test the multi-arch images. That will be handled in a following PR once we verify the Dev builds. This creates one image (each with a list of manifest)e and the container runtime picks the one architecture that it matches at runtime.

We leverage the buildx support by docker (something that we already use for the operator), where passing the TARGET_ARCH into the Dockerfile can be used to generate different layers based on the architecture. 
https://docs.docker.com/build/building/multi-platform/, https://forums.docker.com/t/build-multi-arch-images-with-different-commands-per-architecture-in-docker-file/134795
Note that for the operator cross platform builds, we can afford to do a cross compilation based build, but for vertica-k8s we would need to use emulation based build because we already have the vertica RPM that we want to install.

Some highlights are:

- The Makefile is updated such that it can read both x86 and aarch64 vertica RPMs and pass it on to the Dockerfile.
- Update the Dockerfile to check what RPMs are passed and accordingly make conditional installation of the respective architecture RPM in the image layer.
- This also updates some artifact download links to support future multi-arch builds and e2e tests: OPERATOR_SDK, CHANGIE_VERSION, ISTIO and OPM.
- We should eventually have one target for operator build as well when we are ready to use multiarch builds as default. Added a comment as a reminder for us to revisit, as changing that now might break other devops pipelines.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -359,6 +359,7 @@ jobs:
         # pushing it up to an external repository.
         if [ $GITHUB_EVENT_NAME != 'pull_request' ]
         then
+          # Will be removed once we merge this and make the multiarch builds as the default
           make docker-build-crossplatform-operator
         else
           make docker-build-operator
--- a/Makefile
+++ b/Makefile
@@ -161,6 +161,13 @@ E2E_TEST_DIRS?=tests/e2e-leg-1
 # Additional arguments to pass to 'kubectl kuttl'
 E2E_ADDITIONAL_ARGS?=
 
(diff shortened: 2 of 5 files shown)


narphu:
Declining this in favor of https://github.com/vertica/vertica-kubernetes/pull/911 as the tests need the Dockerhub Username and Password defined in the source branch that is not from a fork.

PR #911, opened 2024-08-30
Adds support for Multi-Arch image builds

narphu:
This PR updates the Make targets and Dockerfiles to allow passing a TARGET_ARCH argument that allows creating a Multi-architecture image, specifically intended towards ARM64 support. This does not yet create or test the multi-arch images. That will be handled in a following PR once we verify the Dev builds. This creates one image (each with a list of manifest)e and the container runtime picks the one architecture that it matches at runtime.

We leverage the buildx support by docker (something that we already use for the operator), where passing the TARGET_ARCH into the Dockerfile can be used to generate different layers based on the architecture.
https://docs.docker.com/build/building/multi-platform/, https://forums.docker.com/t/build-multi-arch-images-with-different-commands-per-architecture-in-docker-file/134795
Note that for the operator cross platform builds, we can afford to do a cross compilation based build, but for vertica-k8s we would need to use emulation based build because we already have the vertica RPM that we want to install.

Some highlights are:

- The Makefile is updated such that it can read both x86 and aarch64 vertica RPMs and pass it on to the Dockerfile.
- Update the Dockerfile to check what RPMs are passed and accordingly make conditional installation of the respective architecture RPM in the image layer.
- This also updates some artifact download links to support future multi-arch builds and e2e tests: OPERATOR_SDK, CHANGIE_VERSION, ISTIO and OPM.
- We should eventually have one target for operator build as well when we are ready to use multiarch builds as default. Added a comment as a reminder for us to revisit, as changing that now might break other devops pipelines.


docker-vertica-v2/Dockerfile
```diff
@@ -16,26 +16,35 @@ ARG BASE_OS_VERSION
...
+# Update is needed to be confident that we're picking up
+# fixed libraries.
+# Using --nobest to make it easier yum install to work. This stage isn't used
+# for the final image, so any package is good enough. We just need to install
+# the vertica rpm and copy that over.
+
+# It can be challenging to identify the exact architecture that the build file is targetting when doing a multi-platform build
+# The pain is described in https://github.com/BretFisher/multi-platform-docker-build quite well
+# We only need to do this because we have separate vertica RPMs for ARM and x86. 
+# There is no need for us to differentiate between the vriants of ARM. So we consider those as equivalent for our RPM install.
```
roypaulin, line 42:
```suggestion
# There is no need for us to differentiate between the variants of ARM. So we consider those as equivalent for our RPM install.
```

Makefile
```diff
@@ -161,6 +161,13 @@ E2E_TEST_DIRS?=tests/e2e-leg-1
 # Additional arguments to pass to 'kubectl kuttl'
 E2E_ADDITIONAL_ARGS?=
 
+# Target Architecture that the docker image can run on
+# By Default: linux/amd64,linux/arm64
+# If you wish to build the image targeting other platforms you can use the --platform flag: https://docs.docker.com/build/building/multi-platform/
+# (i.e. docker buildx build --platform=linux/amd64,linux/arm64). However, you must enable docker buildKit for it.
+# More info: https://docs.docker.com/develop/develop-images/build_enhancements/
+TARGET_ARCH?=linux/amd64
```
roypaulin, line 169: Is it eventually going to be `linux/arm64,linux/amd64`?

docker-vertica-v2/Dockerfile
```diff
@@ -44,10 +53,17 @@ RUN set -x \
...
-  && yum localinstall -y /tmp/${VERTICA_RPM} \
+  && ls /tmp/ \
+  && if [[ "$TARGETARCH" == "arm64" ]] ; then \
+      uname -m && \
+      yum localinstall -y /tmp/${VERTICA_ARM64_RPM} ; \
+    elif [[ "$TARGETARCH" == "amd64" ]] ; then \
+      yum localinstall -y /tmp/${VERTICA_X86_RPM} ; \
+    else \
+      exit 1; \
+    fi \
```
roypaulin, line 64: At the end there will be one image compatible with multiple architectures, right? How is the build process going to handle the fact that we have 2 rpms?

docker-vertica-v2/Dockerfile
```diff
@@ -44,10 +53,17 @@ RUN set -x \
   openssl \
   && /usr/sbin/groupadd -r verticadba \
   && /usr/sbin/useradd -r -m -s /bin/bash -g verticadba dbadmin \
-  && yum localinstall -y /tmp/${VERTICA_RPM} \
+  && ls /tmp/ \
+  && if [[ "$TARGETARCH" == "arm64" ]] ; then \
```
roypaulin, line 57: Where is `TARGETARCH` getting its value.

.github/workflows/build-images.yml
roypaulin, line 1: You probably also need to update the "download rpm" step for the rpm name to match with the one in dockerfile

Makefile
narphu: Yes, I will use this branch to create a Dev image build and make sure the E2E tests pass. Then we can change this to linux/arm64,linux/amd64 as default.

docker-vertica-v2/Dockerfile
narphu: Correct. We will need to download both x86 and aarch64 rpms to the packages directory. We can do the same in Jenkins as well for daily, private and public image builds.
The build is agnostic really, if we set TARGETARCH to `linux/arm64,linux/amd64`, it would be expected that both the RPMs are available in the packages/. Otherwise the docker build will fail. If we provide any one of them, it should still pass and build a single arch image, as long as the matching RPM is present.
Although, I expect that moving forward our default behavior will be multi-arch only.

.github/workflows/build-images.yml
narphu: Yes, thanks for pointer.

roypaulin (approved):

docker-vertica-v2/Makefile
```diff
@@ -1,14 +1,18 @@
...
 BASE_OS_VERSION?=9
 FOR_GITHUB_CI?=false
 VERTICA_IMG?=vertica-k8s
 MINIMAL_VERTICA_IMG?=
-VERTICA_VERSION?=$(shell rpm --nosignature -qp --queryformat '%{VERSION}-%{RELEASE}' packages/$(VERTICA_RPM))
+TARGET_ARCH?=linux/amd64
 VERTICA_ADDITIONAL_DOCKER_BUILD_OPTIONS?=
 
+VERTICA_X86_VERSION?=$(shell rpm --nosignature -qp --queryformat '%{VERSION}-%{RELEASE}' packages/$(VERTICA_X86_RPM))
+VERTICA_ARM_VERSION?=$(shell rpm --nosignature -qp --queryformat '%{VERSION}-%{RELEASE}' packages/$(VERTICA_ARM64_RPM))
```
cchen-vertica, line 14: These two variables are not used in docker-build-vertica. We should assign the value of them to VERTICA_VERSION.

docker-vlogger/Dockerfile
```diff
@@ -16,6 +16,7 @@
 
 ARG ALPINE_VERSION
 ARG BASE_IMG
+ARG TARGETARCH
```
cchen-vertica, line 19: This isn't used in current dockerfile, and I cannot find where we passed it in.

docker-vertica-v2/Dockerfile
```diff
@@ -16,26 +16,35 @@ ARG BASE_OS_VERSION
...
+# for the final image, so any package is good enough. We just need to install
+# the vertica rpm and copy that over.
+
+# It can be challenging to identify the exact architecture that the build file is targetting when doing a multi-platform build
+# The pain is described in https://github.com/BretFisher/multi-platform-docker-build quite well
+# We only need to do this because we have separate vertica RPMs for ARM and x86. 
+# There is no need for us to differentiate between the vriants of ARM. So we consider those as equivalent for our RPM install.
+
+# Don't move the ARG from here. The ARG is only read by the build system if its declared after the SHELL
+ARG TARGETARCH
```
cchen-vertica, line 45: How does the dockerfile receive the value of "TARGETARCH"? I didn't see it on build-arg list of the makefile below.

Makefile
narphu: Old Image size: 473.46 MB compressed
New Image size: 474.9 MB for linux/amd64 and 354.77 MB for linux/arm64 compressed

docker-vertica-v2/Makefile
narphu: Right, I had initally planned of using this to catch version mismatch if someone does not use the same RPM version. Do you think that is useful and a valid concern?

docker-vlogger/Dockerfile
narphu: This is an implicit Dockerfile arg that the Build system automatically recognizes: https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope

docker-vertica-v2/Dockerfile
narphu: It is a predefined arg that are set automatically. We do set the target platform with the --platform flag on docker build, docker pulls out the architecture component of TARGETPLATFORM (that is set by the --platform).

docker-vlogger/Dockerfile
cchen-vertica: You don't used it in this file, do we probably can remove it in this file?

docker-vertica-v2/Makefile
cchen-vertica: Is this for the case "TARGETARCH == linux/arm64,linux/amd64"? Yes, we need to verify both of them have the same version. And we need to label the version to the image.

cchen-vertica:
You can rerun the failure tests, and when all tests pass, you are good to merge this PR.

PR #912, opened 2024-09-02
config kind cluster to increase maximum log file size and event-ttl

qindotguan:
A larger operator log size and longer events duration would be useful for users who want to debug issues on a local Kind cluster. This PR will increase the maximum log size from 10Mi to 100Mi and the duration of events from 1 hour to 24 hours.

# Patch in kubelet to increase maximum log file size
On a newly deployed Kubernetes cluster, the pod's log is located under the host /var/log/pods/<pods-name> directory with a maximum log file size of 10Mi (openshift cluster is 50Mi). 
As our verticadb-operator pod's log grows fast as time passes by, especially with online upgrade operations, it's recommended to configure the kubelet argument containerLogMaxSize to increase the maximum log file size to 100Mi.

# Patch in apiserver to increase events ttl duration
By default, the k8s cluster clears the events after 1 hour and is not persistent on any disk. The direct way is to add extra args --event-ttl for apiserver to increase the events time-to-live (TTL) duration.

### DIFF
--- a/scripts/kind.sh
+++ b/scripts/kind.sh
@@ -29,6 +29,8 @@ KIND=$REPO_DIR/bin/kind
 REG_NAME='kind-registry'
 REG_PORT='5000'
 HANDLE_REGISTRY=1
+LOG_MAX_SIZE='100Mi'
+EVENT_TTL='24h'
 
 while getopts "ut:k:i:ap:xr:m:" opt
 do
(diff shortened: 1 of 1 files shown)


roypaulin:
What events? events like `CreateDBStart`?

qindotguan:
Yes. The goal is to view those events in the past 24 hours or longer.

roypaulin:
And why do we need this for? e2e tests? Can you add the reason to the description?

cchen-vertica (approved):

roypaulin (approved):

PR #916, opened 2024-09-06
Remove workarounds for online upgrade

fenic-fawkes:
the server can now reshard even with DisableNonReplicatableQueries set
sandboxing now properly clears DisableNonReplicatableQueries during sandbox setup, plus cat editor ignores it

### DIFF
--- a/pkg/controllers/vdb/onlineupgrade_reconciler.go
+++ b/pkg/controllers/vdb/onlineupgrade_reconciler.go
@@ -108,7 +108,6 @@ const (
 	addSubclustersInx
 	addNodeInx
 	rebalanceShardsInx
-	waitForActiveSubsInx
 	setConfigParamInx
 	sandboxInx
 	clearConfigParamInx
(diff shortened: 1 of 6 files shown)


fenic-fawkes:
is there a way to clean up space for the test runner? tests are failing all over the place due to lack of space, it looks like

roypaulin:
Yes there are only 14GB of disk space available.There is a lot of way to free some space in the runner like: splitting heavy jobs into smaller jobs, cleaning up docker images/containers the second we don't need them any more, cleaning up unnecessary dependencies.

I think we have already hit the 14GB limit so we need to update the tests to consume the least resource possible. But whatever we do, we are still going to hit this limit again but let's postpone it as much as possible.

I will open a ticket to improve the tests to consume less, in the meantime, rerunning the failed tests should work.

cchen-vertica:
> Yes there are only 14GB of disk space available.There is a lot of way to free some space in the runner like: splitting heavy jobs into smaller jobs, cleaning up docker images/containers the second we don't need them any more, cleaning up unnecessary dependencies.
> 
> I think we have already hit the 14GB limit so we need to update the tests to consume the least resource possible. But whatever we do, we are still going to hit this limit again but let's postpone it as much as possible.
> 
> I will open a ticket to improve the tests to consume less, in the meantime, rerunning the failed tests should work.

I planned to create an AWS runner for K8s e2e tests in the next PI. Then we could debug the e2e failures more easily: we can directly run the failure test on that AWS runner and check the logs in real time.  But Github runner is free, let's see how far you can go for the improvement.

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -188,9 +182,6 @@ func (r *OnlineUpgradeReconciler) Reconcile(ctx context.Context, _ *ctrl.Request
 		// replica group B.
 		r.postSandboxSubclustersMsg,
 		r.sandboxReplicaGroupB,
```
cchen-vertica, line 184: Could we uncomment the lines 197 and 198:
// r.postPrepareReplicationMsg,
// r.prepareReplication

Also remove //nolint:unused for those functions.

cchen-vertica (approved):

PR #920, opened 2024-09-11
Ver 90388 save restore point before upgrade

HaoYang0000:


pkg/controllers/vdb/offlineupgrade_reconciler.go
```diff
@@ -107,6 +107,8 @@ func (o *OfflineUpgradeReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
 
 	// Functions to perform when the image changes.  Order matters.
 	funcs := []func(context.Context) (ctrl.Result, error){
+		// Save restore point before start upgrade
+		o.saveRestorePoint,
```
HaoYang0000, line 111: @roypaulin For offline and read-only upgrade, we want the job "o.saveRestorePoint" before "startUpgrade", right? or other orders

pkg/controllers/vdb/offlineupgrade_reconciler.go
```diff
@@ -151,6 +153,15 @@ func (o *OfflineUpgradeReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
 	return ctrl.Result{}, o.Manager.logUpgradeSucceeded(sandbox)
 }
 
+func (o *OfflineUpgradeReconciler) saveRestorePoint(ctx context.Context) (ctrl.Result, error) {
+	actor := MakeSaveRestorePointReconciler(o.Rec, o.Vdb, o.Log, o.PFacts, o.Dispatcher, o.Rec.GetClient())
```
HaoYang0000, line 157: @roypaulin is this the proper way to call the other Reconciler and its method? 
MakeSaveRestorePointReconciler takes a VerticaDBReconciler which does not exist here, should I also create a VerticaDBReconciler here and pass it down to this make method?

pkg/controllers/vdb/upgrade.go
```diff
@@ -192,6 +192,15 @@ func (i *UpgradeManager) finishUpgrade(ctx context.Context, sbName string) (ctrl
 	return ctrl.Result{}, nil
 }
 
+// saveRestorePoint handles condition status and event recording for start of an upgrade
+func (i *UpgradeManager) saveRestorePoint(ctx context.Context) (ctrl.Result, error) {
```
HaoYang0000, line 196: @roypaulin Follow up question as https://github.com/vertica/vertica-kubernetes/pull/920/files#r1765057129, should we maintain the saveRestore logic for upgrade in one place. So in the offline/readonly/online reconciler we can call o.Manager.saveRestorePoint()

PR #921, opened 2024-09-11
Split leg-8 and leg-9 for improving stability of e2e tests

cchen-vertica:
leg-8 will use multiple Vertica images to do the test which requires a large disk size. It failed the e2e tests intermittently due to lower disk space. This PR split it into two legs.

leg-9 contains both sandbox tests and online-upgrade tests which cost a lot of time. Moreover, the operator log is long and hard to read. This PR keeps online-upgrade tests in leg-9 and extracted sandbox tests to leg-10.


.github/workflows/e2e.yml
```diff
@@ -465,6 +487,28 @@ jobs:
...
+          dockerhub-username: ${{ secrets.DOCKERHUB_USERNAME }}
+          dockerhub-token: ${{ secrets.DOCKERHUB_TOKEN }}
+          vlogger-image: ${{ needs.build.outputs.vlogger-image }}
+          operator-image: ${{ needs.build.outputs.operator-image }}
+          vertica-image: ${{ needs.build.outputs.minimal-vertica-image }}
+          vertica-deployment-method: vclusterops
+          communal-storage-type: s3
+          minimum-vertica-image: '24.3.0'
+          # Include the vertica license so we can test scaling past 3 nodes.
+          vertica-license: ${{ secrets.VERTICA_LICENSE }}
```
roypaulin, line 510: Must add `need-base-vertica-image: 'true'` after this line.

cchen-vertica: Yeah, forgot that we did upgrades in sandbox tests.

cchen-vertica:
> Maybe we should also isolate `revivedb-scaling-restore ` from leg-6. It always fails due to insufficient space. It can either be in a separate workflow or job

When I checked recent daily runs, `revivedb-scaling-restore` all passed. I will extract this test to a new leg and see if that helps.

BTW, I noticed that after I split e2e-leg-8, the readonly online upgrade is still not stable. Let's see if AWS runner can fix all those issues in next sprint.

roypaulin (approved):

PR #924, opened 2024-09-15
Recreated statefulSet after unsandboxing if Vertica image changes

cchen-vertica:
This PR added a Vertica image reconciler in vdb controller to recreate statefulSet of unsandboxed subclusters if those subclusters have a different Vertica image than primary subclusters. This can ensure the subclusters successfully join back to the main cluster and decrease the time for downgrading.


pkg/controllers/vdb/unsandboximageversion_reconciler.go
```diff
@@ -0,0 +1,118 @@
...
+		curSts := &appsv1.StatefulSet{}
+		expSts := builder.BuildStsSpec(nm, r.Vdb, scInVdb)
+		err := vk8s.SetVerticaImage(expSts.Spec.Template.Spec.Containers, priScImage, r.Vdb.IsNMASideCarDeploymentEnabled())
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		err = r.VRec.Client.Get(ctx, nm, curSts)
+		if err != nil && kerrors.IsNotFound(err) {
+			r.Log.Info("Creating statefulset", "Name", nm, "Size", expSts.Spec.Replicas, "Image", priScImage)
+			err = createSts(ctx, r.VRec, expSts, r.Vdb)
```
roypaulin, line 102: `obj_reconciler` should handle reconciliation of all vdb-owned resources (sts, service,etc...), let's not move that logic from there. 
Just update the image in the sts and delete the pods, they will get recreated from the stss with the right image

cchen-vertica:
The reason why we see isPrimary=false in primary pods is: the last step in online upgrade is renaming subclusters, but renaming subclusters only updated vdb, not the sts. In the next iteration, we get podfacts from both vdb and sts, so we get a wrong podfacts. After that, obj reconciler fixed sts, so nothing bad happens.

pkg/controllers/vdb/upgrade.go
```diff
@@ -132,12 +132,26 @@ func (i *UpgradeManager) isVDBImageDifferent(ctx context.Context, sandbox string
...
 		cntImage, err := vk8s.GetServerImage(sts.Spec.Template.Spec.Containers)
 		if err != nil {
 			return false, err
 		}
+
+		// If the current cluster has different images across its subclusters, it means
+		// that some subclusters have just been unsandboxed. In this case, we do not want
+		// the upgrade reconciler to be triggered. Instead, the unsandbox-image-version reconciler
+		// should first restore the image version for those subclusters.
+		if clusterImage != "" && clusterImage != cntImage {
```
roypaulin, line 147: Is it possible to test the case where you upgrade the main cluster in the middle of unsandboxing or just after?

cchen-vertica: I remember you added a webhook to prevent online upgrade if there is an existing sandbox so we cannot upgrade the main cluster in the middle of unsandboxing. I added one test for upgrading after unsandboxing.

Makefile
```diff
@@ -105,7 +105,7 @@ export LEG9
 # What alpine image does the vlogger image use
 VLOGGER_BASE_IMG?=alpine
 # What version of alpine does the vlogger image use
-VLOGGER_ALPINE_VERSION?=3.19
+VLOGGER_ALPINE_VERSION?=3.18
```
roypaulin, line 108: What was this about?

roypaulin (approved):

Makefile
cchen-vertica: If you see the error log of the previous commit, e2e tests cannot pull "alpine:3.19" since 3.19 doesn't support AMD. I will switch it back to 3.19 when alpine fixed that error.

PR #925, opened 2024-09-16
Installed strip in build image

cchen-vertica:
This PR simply installed "strip" command in the build image for stripping /opt/vertica. It should restore the size of Vertica minimal images.

### DIFF
--- a/docker-vertica-v2/Dockerfile
+++ b/docker-vertica-v2/Dockerfile
@@ -42,6 +42,7 @@ RUN set -x \
   glibc-langpack-en \
   iproute \
   openssl \
+  binutils \
   && /usr/sbin/groupadd -r verticadba \
   && /usr/sbin/useradd -r -m -s /bin/bash -g verticadba dbadmin \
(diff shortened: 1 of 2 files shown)


roypaulin (approved):

fenic-fawkes:
should we also add `set -e` at the top of `docker-vertica/packages/cleanup.sh` to make sure something like this doesn't happen again?

roypaulin:
> should we also add `set -e` at the top of `docker-vertica-v2/packages/cleanup.sh` to make sure something like this doesn't happen again?

Yes let's do it now that we do not control what is inside the base image.

cchen-vertica:
"set -e" will block most of strip commands since if stripping one library fails, all others cannot be stripped.  Many binaries and libraries cannot be stripped so we shouldn't use "set -e".

A failure example in "strip /opt/vertica/lib/*.so*":
strip:/opt/vertica/lib/stS2m6vi[.gnu.build.attributes.unlikely]: error: failed to locate merged notes: Invalid operation

Fixing all strip commands is time-consuming, so I just removed "set -e".

PR #928, opened 2024-09-19
add check to push image

HaoYang0000:
This PR add an action to check if docker image exists on remote internal repo(github private repo), if not we will push to it during the image build step.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -364,7 +364,15 @@ jobs:
         else
           make docker-build-operator
         fi
-
+        # Check if image exists in private repo, if not exist, push to repo
+        docker manifest inspect ${OPERATOR_IMG} > /dev/null 2>&1
+        EXIT_CODE=$?
+        if [ $EXIT_CODE -eq 0 ]; then
+          echo "$image exists"
+        else 
+          echo "$image does not exist, making a push"
+          make docker-push-operator
+        fi
     - name: Save the image for consumption by dependent jobs (PRs only)
       if: ${{ github.event_name == 'pull_request' }}
       run: |
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -349,7 +349,7 @@ jobs:
       uses: ./.github/actions/setup-go
       if: ${{ inputs.operator_image == '' }}
 
-    - name: Build and optionally push operator image
+    - name: Build operator image
       if: ${{ inputs.operator_image == '' }}
       run: |
         export OPERATOR_IMG=${{ steps.operator_image.outputs.value }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-images.yml
```diff
@@ -364,7 +364,19 @@ jobs:
         else
           make docker-build-operator
         fi
-
+    - name: Optionally push operator
+      if: ${{ github.event_name != 'pull_request' && inputs.operator_image == '' }}
+      run: |
+        # Check if image exists in private repo, if not exist, push to repo
+        export OPERATOR_IMG=${{ steps.operator_image.outputs.value }}
```
roypaulin, line 371: But what will be the value of `OPERATOR_IMG`? You need to make sure the name we picked is unique and contains the said repo.
And what repo is it?

.github/workflows/build-images.yml
```diff
@@ -364,7 +364,19 @@ jobs:
         else
           make docker-build-operator
         fi
-
+    - name: Optionally push operator
+      if: ${{ github.event_name != 'pull_request' && inputs.operator_image == '' }}
```
cchen-vertica, line 368: This case has been done in line 363. For non-pr e2e tests, we already push the image to Github repo. Just move your image-exists check mechanism to that line.

HaoYang0000: I moved the mechanism in https://github.com/vertica/vertica-kubernetes/pull/928/commits/1b99dc69afddfa8cdec57cab72316a1f9d5ddc3a
For the image check mechanism, we need to do the docker login first, to grant access to check the manifest. I was thinking doing that in the make file, but the credentials are stored in the git env, so keeping in the yml for now. I am open to other approaches though.

.github/workflows/build-images.yml
HaoYang0000: This part does not apply anymore, moved the logic to the existing step. Ref: https://github.com/vertica/vertica-kubernetes/pull/928#discussion_r1767095110

roypaulin (commented):
Please create a ticket for this as it has doc impact

.github/workflows/build-images.yml
```diff
@@ -360,7 +360,14 @@ jobs:
...
         if [ $GITHUB_EVENT_NAME != 'pull_request' ]
         then
-          make docker-build-crossplatform-operator
+          docker manifest inspect ${OPERATOR_IMG} > /dev/null 2>&1
+          EXIT_CODE=$?
+          if [ $EXIT_CODE -eq 0 ]; then
+            echo "$image exists"
+          else 
+            echo "$image does not exist, making a push"
+            make docker-build-crossplatform-operator
```
roypaulin, line 369: So we are not going to push to github repo for non-pr cases?

HaoYang0000: By existing logic, yes. If we need to change that I can do it in this PR as well.

cchen-vertica (approved):

PR #936, opened 2024-09-26
fix run condition

HaoYang0000:
That last PR added cmd docker manifest inspect IMAGE into non pr cases. However the pipeline will detect the exit code and stop the following tasks. Wrapped the cmd into a variable result, so the pipeline will ignore the exit code.
Also fixed a typo $.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -360,12 +360,11 @@ jobs:
         # pushing it up to an external repository.
         if [ $GITHUB_EVENT_NAME != 'pull_request' ]
         then
-          docker manifest inspect ${OPERATOR_IMG} > /dev/null 2>&1
-          EXIT_CODE=$?
+          EXIT_CODE="$(docker manifest inspect ${OPERATOR_IMG} > /dev/null ; echo $?)"
           if [ $EXIT_CODE -eq 0 ]; then
-            echo "$image exists"
+            echo "image exists"
           else 
-            echo "$image does not exist, making a push"
+            echo "image does not exist, making a push"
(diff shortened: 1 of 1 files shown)


HaoYang0000:
pipeline result for Non-pr run: https://github.com/vertica/vertica-kubernetes/actions/runs/11052203196

.github/workflows/build-images.yml
```diff
@@ -360,12 +360,11 @@ jobs:
         # pushing it up to an external repository.
         if [ $GITHUB_EVENT_NAME != 'pull_request' ]
         then
-          docker manifest inspect ${OPERATOR_IMG} > /dev/null 2>&1
-          EXIT_CODE=$?
+          EXIT_CODE="$(docker manifest inspect ${OPERATOR_IMG} > /dev/null ; echo $?)"
```
cchen-vertica, line 363: Why do we remove "2>&1"?

HaoYang0000: Mainly for debugging purposes, if we have 2>&1, then the error.output will not be printed on the log. We simply want to dump the success standard output from. 
With existing setup: > /dev/null
If image exist, then its success, standard.output will be redirect to  /dev/null, no output is ok.
If image doesn't exist, we will still see logs in error.output. Such as "manifest unknown", or other types like unauth, etc. So we know the true reason why the command returns the return code 1.

With the old setup: > /dev/null 2>&1
All output (standard.output, error.output) will be dumped, its hard for debugging.

Both setup works, because we just need the return code from last command. But use only "/dev/null" provides more useful info to us.

PR #939, opened 2024-09-30
add pause timeout

fenic-fawkes:
also makes online upgrade timeout configurable


pkg/controllers/vdb/upgrade.go
```diff
@@ -680,33 +680,81 @@ func (i *UpgradeManager) closeAllSessions(ctx context.Context, pfacts *PodFacts)
 	return nil
 }
 
+const userHasDbadminRole = "(all_roles like 'dbadmin' " +
```
cchen-vertica, line 683: If dbadmin is assigned to a role and this role is assinged to a user, we cannot see dbadmin in the user's all_roles colume. Can we use "has_role('dbadmin')"?

pkg/controllers/vdb/upgrade.go
```diff
@@ -680,33 +680,81 @@ func (i *UpgradeManager) closeAllSessions(ctx context.Context, pfacts *PodFacts)
...
+	errs := []error{}
+	for _, id := range strings.Split(strings.TrimSuffix(sessionIds, "\n"), "\n") {
+		if id == "" {
+			continue
+		}
+		killCmd := []string{"-tAc", fmt.Sprintf("select close_session('%s')", id)}
+		_, stderr, err = pfacts.PRunner.ExecVSQL(ctx, pf.name, names.ServerContainer, killCmd...)
+		if err != nil {
+			i.Log.Error(err, "failed to kill session", "session_id", id, "stderr", stderr)
+			errs = append(errs, err)
```
cchen-vertica, line 749: Define errs as an error, then use "errs = errors.Join(errs, err) to collect all errors.

pkg/controllers/vdb/upgrade.go
fenic-fawkes: we could... if we want to gather users and then run a sql statement for each user. `has_role` is a metafunction that can only go in a select statement on it's own. you can't even do `select has_role(user_name, 'pseudosuperuser') from users`. pseudosuperuser is a very powerful role. i can't imagine very many users would assign it to a role rather than just to a user directly. only a handful of users should ever have access to it.

cchen-vertica (approved):

pkg/controllers/vdb/upgrade.go
cchen-vertica: What if dbadmin role is assigned to a user?

cchen-vertica:
Left one comment about the dbadmin role. If you think that isn't a problem, you can merge this PR.

pkg/controllers/vdb/upgrade.go
fenic-fawkes: the pause only skips checking for pseudosuperuser, it was my mistake to check dbadmin.

PR #940, opened 2024-10-01
Support multiple local pathes in revive_db

cchen-vertica:
This PR supports reviving a database with multiple data paths, temp paths, user paths, and depot paths.

### DIFF
--- a/pkg/reviveplanner/analyze.go
+++ b/pkg/reviveplanner/analyze.go
@@ -17,12 +17,16 @@ package reviveplanner
 
 import (
 	"fmt"
+	"path/filepath"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 
 	"github.com/go-logr/logr"
 	vapi "github.com/vertica/vertica-kubernetes/api/v1"
+	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
+	"golang.org/x/exp/maps"
 )
 
 type Planner struct {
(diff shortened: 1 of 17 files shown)


roypaulin (commented):
There are some parts I don't understand. Can we have a walkthrough tomorrow so I can ask my questions live

pkg/reviveplanner/analyze.go
```diff
@@ -84,41 +85,89 @@ func (p *Planner) ApplyChanges(vdb *vapi.VerticaDB) (updated bool, err error) {
...
+	updated, err = p.updateLocalPathsInVdb(vdb, extraPaths)
+	updated = updated || updatedShardCount
 	if err != nil {
 		return updated, err
 	}
-	if catPath != vdb.Spec.Local.GetCatalogPath() {
-		p.logPathChange("catalog", vdb.Spec.Local.GetCatalogPath(), catPath)
-		vdb.Spec.Local.CatalogPath = catPath
+
+	otherPaths := p.Parser.GetOtherPaths()
```
roypaulin, line 99: What are `otherPaths`?

roypaulin (commented):
I think add a knob to switch between this and the old behavior(default). We do not want other to be blocked in case this causes unexpected behavior. This is more like a workaround and as such we should be able to control when to apply it, once on k8s they won't need it.
On k8s, users are less likely to create extra storage locations for insufficient space reasons as they can just expand the persistent volume

pkg/reviveplanner/analyze.go
cchen-vertica: TEMP and USER paths.

cchen-vertica:
> I think add a knob to switch between this and the old behavior(default). We do not want other to be blocked in case this causes unexpected behavior. This is more like a workaround and as such we should be able to control when to apply it, once on k8s they won't need it. On k8s, users are less likely to create extra storage locations for insufficient space reasons as they can just expand the persistent volume

The knob doesn't matter. On k8s, they still can create multiple paths through vsql. If they only have one path, the new code will have the same behavior as before. For non-k8s customers, we don't want them to fail the revive_db and ask us for the workaround. We should keep the knob on so they can revive successfully.

PR #942, opened 2024-10-01
Show a restore point info in the vdb status

roypaulin:
This fetches a restore point info through the show_restore_points api, and saves it to the status so that the user can easily find it.


pkg/meta/annotations.go
```diff
@@ -300,6 +300,11 @@ const (
 	OnlineUpgradeArchiveBeforeReplicationAnnotation = "vertica.com/online-upgrade-archive-before-replication"
 
 	SaveRestorePointAnnotation = "vertica.com/save-restore-point-on-upgrade"
+	// Allow users to skip the creation of a VerticaRestorePointsQuery object after saving
+	// a restore point. When this is false, after saving a restore point, the operator
+	// will create a VerticaRestorePointsQuery in order to expose the restore point info
+	// the user.
+	SkipVRPQCreationAnnotation = "vertica.com/skip-restore-points-query-creation"
```
cchen-vertica, line 307: Why do we need this annotation? We should always create the VRep CR to collect the restore point's info for the user.

cchen-vertica (commented):
After creating the vrep, could we collect the restore point's id and index and store it in vdb or somewhere? Then we remove the vrep so the user doesn't know we did a vrep creation. 

Or we just call vclusterOps API to get the restore point's info rather than creating a vrep CR.

roypaulin:
> After creating the vrep, could we collect the restore point's id and index and store it in vdb or somewhere? Then we remove the vrep so the user doesn't know we did a vrep creation.
> 
> Or we just call vclusterOps API to get the restore point's info rather than creating a vrep CR.

A few reasons why I did it this way:
- I don't want show_restore_points failure to fail the whole save_restore_point.
- I don't want to slow down save_restore_point while there is a separate controller dedicated to show_restore_points.*
- Duplicate information. There is already a CR that can contain the info we want, and we will be duplicating that info over to the VDB CR.
- A user can create multiple restore_points which mean we would need to add an array to the vdb status to contain the info about all of them

roypaulin:
I went a slightly different direction. I added a reconciler for show_restore_point. It will detect if show_restore_point is needed from the status and will update the status.

PR #945, opened 2024-10-03
fix online upgrade pause timeout logic

fenic-fawkes:
kill all sessions during online upgrade when hitting pause timeout when the vertica version is too old to tell us which sessions are paused

### DIFF
--- a/api/v1/version.go
+++ b/api/v1/version.go
@@ -178,7 +178,7 @@ func (v *VerticaDB) IsUpgradePathSupported(newAnnotations map[string]string) (ok
 	return
 }
 
-// isOnlineUpgradeSupported returns true if the version in the Vdb is is equal or newer than
+// isOnlineUpgradeSupported returns true if the version in the Vdb is equal or newer than
 // 24.3.0-2.
 func (v *VerticaDB) isOnlineUpgradeSupported(vinf *version.Info) bool {
 	return vinf.IsEqualOrNewerWithHotfix(OnlineUpgradeVersion)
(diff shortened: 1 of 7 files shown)


api/v1/version.go
```diff
@@ -176,11 +178,20 @@ func (v *VerticaDB) IsUpgradePathSupported(newAnnotations map[string]string) (ok
 	return
 }
 
-// isOnlineUpgradeSupported returns true if the version in the Vdb is for a version that supports online upgrade
-// 24.3.0-4 and 24.4.0-1 are the oldest version that support online upgrade, anything newer is fine
+// isOnlineUpgradeSupported returns true if the version in the Vdb is is equal or newer than
```
roypaulin, line 181:
```suggestion
// isOnlineUpgradeSupported returns true if the version in the Vdb is equal or newer than
```

api/v1/version.go
```diff
@@ -176,11 +178,20 @@ func (v *VerticaDB) IsUpgradePathSupported(newAnnotations map[string]string) (ok
...
+func (v *VerticaDB) IsPausedSessionsSupported() bool {
+	vinf, ok := v.MakeVersionInfo()
+	if !ok {
+		return false
+	}
+	if vinf.IsEqualOrNewer(MinPauseSessionsVersion243) {
+		return vinf.IsEqualOrNewerWithHotfix(MinPauseSessionsVersion243)
 	}
-	return vinf.IsEqualOrNewerWithHotfix(OnlineUpgradeVersion243)
+	return vinf.IsEqualOrNewerWithHotfix(MinPauseSessionsVersion243)
```
roypaulin, line 196: You need to update this function. You never use `MinPauseSessionsVersion244`.
Is it easier to just exclude versions  that do not support paused sessions(lower than v24.3.0-4 and equal to v24.4.0-0)?

cchen-vertica (approved):

PR #952, opened 2024-10-08
Prevent skipping some tests

roypaulin:
The reason why normal ci tests were oddly more stable than daily build tests was because we were skipping a lot of them.
We rely on the vertica-version label in the image to skip some tests in case the vertica version is lower than the min required version by the test. The issue is that label is empty for `latest-test-master` which leads to all the tests using that image to be skipped. We need to set that label properly for that image too but in the meantime, we are adding this work-around to not skip some tests.

### DIFF
--- a/scripts/is-image.sh
+++ b/scripts/is-image.sh
@@ -60,6 +60,14 @@ then
 fi
 
 logAndRunCommand docker pull $IMG
+# Normally we would rely on the vertica-version label of the image but it is empty
+# for the latest-test-master image. Until we figure out why, we are adding
+# this work around to not skip some tests.
+if [[ "$IMG" == "docker.io/opentext/vertica-k8s-private:latest-test-master" ]]
+then
+    logInfo "$IMG is newer"
+    exit 1
+fi
 IMG_VER=$(determine_image_version $IMG)
 logInfo "Image $IMG has version $IMG_VER"
 logInfo "Checking if $IMG_VER is $COMPARE_TYPE than $COMPARE_VERSION"
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
-	"time"
 
 	vutil "github.com/vertica/vcluster/vclusterops/util"
 	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
@@ -28,7 +27,6 @@ import (
 	vversion "github.com/vertica/vertica-kubernetes/pkg/version"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -95,19 +93,6 @@ func (v *VerticaDB) Default() {
 	}
 	if v.Spec.TemporarySubclusterRouting != nil {
 		v.Spec.TemporarySubclusterRouting.Template.Type = SecondarySubcluster
-	}
-	// Set required status conditions fields if they are unset
-	for i := range v.Status.Conditions {
-		cond := &v.Status.Conditions[i]
-		if cond.LastTransitionTime.IsZero() {
-			cond.LastTransitionTime = metav1.NewTime(time.Now())
-		}
-		if cond.Reason == "" {
-			cond.Reason = DefaultReason
-		}
-		if cond.Message == "" {
-			cond.Message = DefaultMsg
-		}
 	}
 	v.setDefaultServiceName()
 	v.setDefaultSandboxImages()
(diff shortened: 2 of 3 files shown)


roypaulin:
For a while there are quite a few tests that were skipped including save-restore-point, that's why my solution looked like it worked when it does not. 
We might see more failures in our normal ci tests.

scripts/is-image.sh
```diff
@@ -60,6 +60,14 @@ then
 fi
 
 logAndRunCommand docker pull $IMG
+# Normally we would rely on the vertica-version label of the image but it is empty
+# for the latest-test-master image. Until we figure out why, we are adding
+# this work around to not skip some tests.
+if [[ "$IMG" == "docker.io/opentext/vertica-k8s-private:latest-test-master" ]]
```
cchen-vertica, line 66: what's the version of docker.io/opentext/vertica-k8s-private:latest-test-master? If it's larger than 24.2, we should fix this script.

roypaulin: It is the same as `latest-master` or `latest-minimal-master` which is the daily version. The thing is that the vertica-version label is not set, and we rely on it. We need to address that but until then we can use this workaround.
The issue is not the script but the image.

cchen-vertica (approved):

scripts/is-image.sh
cchen-vertica: I can see the vertica-version in latest-minimal-master so it's our script's issue.
[
    {
        "Id": "sha256:5b70a076596ceff33316611bd179d1d8d48f5cf95364451b92904b5187f16278",
        "RepoTags": [
            "opentext/vertica-k8s-private:latest-minimal-master"
        ],
        "RepoDigests": [
            "opentext/vertica-k8s-private@sha256:ae91071f044ed091fabec7c8a1581fd36ebd276c735f1f34e7c98accfb480495"
        ],
        "Parent": "",
        "Comment": "buildkit.dockerfile.v0",
        "Created": "2024-10-09T03:24:59.960983375-04:00",
        "DockerVersion": "",
        "Author": "",
        "Config": {
            "Hostname": "",
            "Domainname": "",
            "User": "daemon",
            "AttachStdin": false,
            "AttachStdout": false,
            "AttachStderr": false,
            "ExposedPorts": {
                "5433/tcp": {},
                "8443/tcp": {}
            },
            "Tty": false,
            "OpenStdin": false,
            "StdinOnce": false,
            "Env": [
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/vertica/bin:/opt/vertica/sbin",
                "JAVA_HOME=/usr",
                "VERTICA_STARTUP_LOG_DUPLICATE=/proc/1/fd/1"
            ],
            "Cmd": null,
            "Image": "",
            "Volumes": null,
            "WorkingDir": "/",
            "Entrypoint": null,
            "OnBuild": null,
            "Labels": {
                "image-name": "vertica-k8s",
                "maintainer": "K8s Team",
                "minimal": "yes",
                "org.opencontainers.image.description": "Runs the Vertica server that is optimized for use with the VerticaDB operator",
                "org.opencontainers.image.documentation": "https://www.vertica.com/docs/latest/HTML/Content/Authoring/Containers/ContainerizedVertica.htm",
                "org.opencontainers.image.source": "https://github.com/vertica/vertica-kubernetes/tree/main/docker-vertica-v2",
                "org.opencontainers.image.title": "Vertica Server",
                "org.opencontainers.image.url": "https://github.com/vertica/vertica-kubernetes/",
                "os-family": "rockylinux",
                "os-version": "9.202407.2494951-vertica",
                "vertica-deployment-method": "vclusterops",
                "vertica-version": "25.1.0-20241009"
            }
        },
        "Architecture": "amd64",
        "Os": "linux",
        "Size": 906770720,
        "GraphDriver": {
            "Data": {
                "LowerDir": "/var/lib/docker/overlay2/c3353034cbaeb3be168464f03f4251feb14ed2242130963decd58ab88ca1533b/diff:/var/lib/docker/overlay2/8cb2cc2a184d3d4df549c68ff84046cd55393e8f3c3329fbc8404c83fa8e114d/diff",
                "MergedDir": "/var/lib/docker/overlay2/be6539d6e7a256fea8d7e9213e4f1fe1b6c709be4abd05e97778667b2b32906b/merged",
                "UpperDir": "/var/lib/docker/overlay2/be6539d6e7a256fea8d7e9213e4f1fe1b6c709be4abd05e97778667b2b32906b/diff",
                "WorkDir": "/var/lib/docker/overlay2/be6539d6e7a256fea8d7e9213e4f1fe1b6c709be4abd05e97778667b2b32906b/work"
            },
            "Name": "overlay2"
        },
        "RootFS": {
            "Type": "layers",
            "Layers": [
                "sha256:781baa185249ea4ee5cef59a0a672831d1e98373414c4261ecabed0e43292da9",
                "sha256:76dffd59e1a9e13f9768fd20cf3d8bee7d857af47268315c5525cab4c208ced5",
                "sha256:b5edce646a7d636ca3a8b542d3a243fab2e1c692844f2eae4955d94f1f6aaa50"
            ]
        },
        "Metadata": {
            "LastTagTime": "0001-01-01T00:00:00Z"
        }
    }
]

roypaulin: The script simply retrieves the version which is an empty string for `latest-test-master`.

roypaulin: 2 places to look at: the jenkinsfile and the `docker-vertica-v2/Makefile` as those are the 2 places we set the version.

cchen-vertica: Remove the condition "ifeq ($(FOR_GITHUB_CI), false)" in docker-vertica-v2/Makefile should fix that error.

roypaulin: Did not work. I will merge this so we can reenable the other tests. We can properly fix this on a separate PR

PR #953, opened 2024-10-09
fix value in integration test

HaoYang0000:
The value is not setup correctly so we see error in the integration test:
 kubectl patch vdb vertica-db --type='json' -p='[{
> "op": "add",
> "path": "/status/conditions/-",
> "value": {"type": "SaveRestorePointNeeded", "status": "True"}
> }]' --subresource='status'
The VerticaDB "vertica-db" is invalid:
* conditions[3].lastTransitionTime: Required value
* conditions[3].message: Required value
* conditions[3].reason: Required value

We need to add value for the missing attribute.


HaoYang0000:
Duplicate with https://github.com/vertica/vertica-kubernetes/pull/952, closing

PR #955, opened 2024-10-09
Add support for stop sandbox

roypaulin:
This is the first step to support stop sandbox for realdb. Normally, if a pod is down the operator will try to restart it. However, for the realdb use case we want to be able to stop a standbox and to prevent the operator from restarting it until we decide otherwise.
This adds the logic to stop_db a sandbox when `spec.sandboxes[].shutdown` is true, a new reconciler has been added to the vdb controller to wake up the sandbox controller. As long as that field is true, the operator will ignore the sandbox's pods. As soon as it is set to false, the operator will restart the sandbox.

Scaling down the sts and adding webhook rules will be done on a follow-up ticket.


pkg/controllers/sandbox/restartsandbox_reconciler.go
```diff
@@ -0,0 +1,114 @@
...
+	return &RestartSandboxReconciler{
+		SRec:   r,
+		Log:    log.WithName("RestartSandboxReconciler"),
+		Vdb:    vdb,
+		PFacts: pfacts,
+	}
+}
+
+func (r *RestartSandboxReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	if res, err := r.updateSandboxShutdownState(ctx); verrors.IsReconcileAborted(res, err) {
```
cchen-vertica, line 53: Add r.PFacts.Collect() here since the previous reconciler could invalidate the pod facts.

pkg/controllers/sandbox/restartsandbox_reconciler.go
```diff
@@ -0,0 +1,114 @@
...
+	return err
+}
+
+func (r *RestartSandboxReconciler) updateSubclustersShutdownStateCallback() (bool, error) {
+	sb := r.Vdb.GetSandbox(r.PFacts.SandboxName)
+	if sb == nil {
+		return false, errors.New("sandbox not found")
+	}
+	needUpdate := false
+	scMap := r.Vdb.GenSubclusterMap()
```
cchen-vertica, line 105: We should only set shutdown status in sandbox.subclusters, not set in spec.subclusters. Then the user cannot set shutdown status for the subclusters that aren't sandboxed.

pkg/controllers/sandbox/sandbox_controller.go
```diff
@@ -164,8 +164,12 @@ func (r *SandboxConfigMapReconciler) constructActors(vdb *v1.VerticaDB, log logr
 		vdbcontroller.MakeOfflineUpgradeReconciler(r, log, vdb, prunner, pfacts, dispatcher),
 		// Add annotations/labels to each pod about the host running them
 		vdbcontroller.MakeAnnotateAndLabelPodReconciler(r, log, vdb, pfacts),
+		// Stop Vertica in the sandbox if the shutdown state is true
+		vdbcontroller.MakeStopDBReconciler(r, vdb, prunner, pfacts, dispatcher),
 		// Restart any down pods
 		vdbcontroller.MakeRestartReconciler(r, log, vdb, prunner, pfacts, true, dispatcher),
+		// Update the sandbox and its subclusters' shutdown state
+		MakeRestartSandboxReconciler(r, vdb, pfacts, log),
```
cchen-vertica, line 172: Move this above MakeStopDBReconciler so we can set the shutdown state and then stop the sandbox in one iteration.

pkg/controllers/vdb/podfacts.go
```diff
@@ -1318,3 +1370,12 @@ func (p *PodFacts) FindSecondarySubclustersWithDifferentImage() (scs []string, p
...
 	return scs, priScImage
 }
+
+func isSubclusterShutdown(vdb *vapi.VerticaDB, sc *vapi.Subcluster) bool {
+	sbName := vdb.GetSubclusterSandboxName(sc.Name)
+	sb := vdb.GetSandbox(sbName)
+	if sb != nil {
+		return sb.Shutdown
+	}
+	return sc.Shutdown
```
cchen-vertica, line 1380: If sb == nil, we should return false since the subcluster isn't in a sandbox.

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -126,9 +126,14 @@ func (r *RestartReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctr
 	// admin commands to run.  Cluster operations only apply if the entire
 	// vertica cluster is managed by k8s.  We skip that if initPolicy is
 	// ScheduleOnly.
-	if r.PFacts.getUpNodeAndNotReadOnlyCount() == 0 &&
+	if r.PFacts.GetUpNodeAndNotReadOnlyCount() == 0 &&
 		r.Vdb.Spec.InitPolicy != vapi.CommunalInitPolicyScheduleOnly {
-		return r.reconcileCluster(ctx)
+		// Abort cluster restart if at least one pod must not be
+		// restarted.
+		if r.PFacts.GetShutdownCount() == 0 {
```
cchen-vertica, line 133: Can we add one more check here? if r.PFacts.GetShutdownCount() == 0 && r.PFacts.GetSandboxName() == vapi.MainCluster

pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -126,9 +126,14 @@ func (r *RestartReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctr
...
+			return r.reconcileCluster(ctx)
+		}
+		return ctrl.Result{}, nil
```
cchen-vertica, line 136: What if GetShutdownCount() < clusterSize? Do we need to support stop part of the sandbox?

pkg/controllers/vdb/sandboxshutdown_reconciler.go
```diff
@@ -0,0 +1,87 @@
...
+	vdb *vapi.VerticaDB) controllers.ReconcileActor {
+	return &SandboxShutdownReconciler{
+		VRec: vdbrecon,
+		Log:  log.WithName("SandboxShutdownReconciler"),
+		Vdb:  vdb,
+	}
+}
+
+func (s *SandboxShutdownReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	// no-op as there is no sandbox
```
cchen-vertica, line 49: Collect pfacts here.

pkg/controllers/vdb/podfacts.go
roypaulin: Later we are going to support `stop subcluster` and it won't be limited to sandboxes. So it will be possible for a main cluster's subcluster to be "shutdown".

pkg/controllers/vdb/restart_reconciler.go
roypaulin: Later we will support `stop subcluster` on any sandbox including the main cluster, that is why I did not add an extra check for the sandbox.

pkg/controllers/vdb/restart_reconciler.go
roypaulin: This path is for `startdb`. We don't allow `startdb` if there is at least one subcluster with `shutdown` set to true. Keep in mind we will support `stop_subcluster` so it will be possible to stop only specific subcluster(s), although realdb does not have that use case.

pkg/controllers/vdb/sandboxshutdown_reconciler.go
roypaulin: This reconciler does not rely on pfacts. It checks only `(spec/status).sandboxes[].shutdown`.

pkg/controllers/sandbox/restartsandbox_reconciler.go
roypaulin: I collect pfacts at line 76.

pkg/controllers/sandbox/sandbox_controller.go
roypaulin: This achieves the same result. Here is the workflow:
`spec.sandboxes[].shutdown`
- false -> true: stopdb_reconciler stops the sandbox and updates `status.sandboxes[].shutdown`, no-op for restartsandbox_reconciler
- true -> false: restart_reconciler will restarts the sandbox and then restartsandbox_reconciler will update `status.sandboxes[].shutdown`.
So This updates `status.sandboxes[].shutdown` only when `spec.sandboxes[].shutdown` goes from true to false.

pkg/controllers/sandbox/restartsandbox_reconciler.go
roypaulin: We will support `stop_subcluster` regardless of sandboxes.

pkg/controllers/vdb/stopdb_reconciler.go
```diff
@@ -120,3 +129,47 @@ func (s *StopDBReconciler) runATCmd(ctx context.Context, initiatorName types.Nam
...
+		return err
+	}
+	// here mark all subclusters in the sandbox for shutdown
+	_, err = vk8s.UpdateVDBWithRetry(ctx, s.VRec, s.Vdb, s.markSandboxSubclustersForShutDown)
+	return err
+}
+
+// markSandboxSubclustersForShutDown marks the subclusters so that the operator
+// does not restart them.
+func (s *StopDBReconciler) markSandboxSubclustersForShutDown() (bool, error) {
```
cchen-vertica, line 160: The spec.subclusters.Shutdown should be set no matter whether we stopped the sandbox or not. This should be executed in another reconciler. Here we should update status.subclusters.Shutdown.

api/v1/verticadb_types.go
```diff
@@ -888,6 +901,12 @@ type SandboxStatus struct {
 	// +optional
 	// State of the current running upgrade in the sandbox
 	UpgradeState SandboxUpgradeState `json:"upgradeState,omitempty"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// +optional
+	// State of the sandbox. true means the sandbox is currently down
+	// and must not be restarted.
+	Shutdown bool `json:"shutdown"`
```
cchen-vertica, line 909: This should be removed since you already have one in SubclusterStatus. If a sandbox is stopped, we should update all subclusters' status inside it.

api/v1/verticadb_types.go
```diff
@@ -952,6 +971,12 @@ type SubclusterStatus struct {
 
 	// +operator-sdk:csv:customresourcedefinitions:type=status
 	Detail []VerticaDBPodStatus `json:"detail"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// +optional
+	// State of the subcluster. true means the subcluster is currently down
+	// and must not be restarted.
+	Shutdown bool `json:"shutdown,omitempty"`
```
cchen-vertica, line 979: I cannot see this field is updated in this PR.

api/v1beta1/verticadb_types.go
```diff
@@ -969,6 +982,12 @@ type SandboxStatus struct {
 	// +optional
 	// State of the current running upgrade in the sandbox
 	UpgradeState SandboxUpgradeState `json:"upgradeState"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// +optional
+	// State of the sandbox. true means the sandbox is currently down
+	// and must not be restarted.
+	Shutdown bool `json:"shutdown"`
```
cchen-vertica, line 990: Ditto

pkg/controllers/sandbox/restartsandbox_reconciler.go
```diff
@@ -0,0 +1,114 @@
...
+	for i := range sb.Subclusters {
```
cchen-vertica, line 106: We shouldn't update the spec. Instead, we should update status.subclusters.

cchen-vertica (commented):
If we will support stop subcluster in main cluster, we probably need to support stop subcluster in the sandbox as well since sandbox is a special database. Then we can have a consistent way to process the Shutdown field: the user can set Shutdown field in spec.subclusters, and then operator will update Shutdown field in status.subclusters. For stopping a sandbox, we provide an easy way: set spec.sandbox.Shutdown to be true. Then operator will update spec.subclusters.Shutdown to be true.

Besides this, do we support unsandboxing a sandbox that is stopped? Should we error it out in webhook or we internally restart the sandbox first, then unsandbox it.

pkg/controllers/vdb/sandbox_configmap.go
```diff
@@ -74,6 +75,8 @@ func (s *SandboxConfigMapManager) triggerSandboxController(ctx context.Context,
 		anns[vmeta.SandboxControllerUpgradeTriggerID] = triggerID
 	case Unsandbox:
 		anns[vmeta.SandboxControllerUnsandboxTriggerID] = triggerID
+	case Shutdown:
+		anns[vmeta.SandboxControllerShutdownTriggerID] = triggerID
```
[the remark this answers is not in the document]
roypaulin: It is not a requirement to always check an added id in the sandbox controller(we don't for sandbox_upgrade). The fact is we need to wake up the sandbox controller for shutdown/restart and this is the way we do it. The (stopdb/restart) reconciler will know what to do based on the podfacts and vdb spec/status.

api/v1/verticadb_types.go
roypaulin: This is still useful:
- From the realdb script you can just check this field, instead of all the sandbox subclusters status, to know if the sandbox  shutdown state is being enforced.
- Updating `shutdown` in podfacts is done earlier. As soon as the user sets `spec.sandboxes[].shutdown` to true all the nodes in that subcluster should inherit that state and must not be restarted by the operator. This makes the operator more reactive to this change as it does not have to wait for the subcluster(s) update.

Simply put it does not hurt to have this.

pkg/controllers/vdb/stopdb_reconciler.go
roypaulin: Here is how I see things. Only 2 entities can update `spec.subclusters[].shutdown`:
- The user, only for stop_sc. They will not be allowed to change that field for a subcluster that belongs to a sandbox with shutdown set to true.
- The operator, after stop_sandbox.

For both cases, there will be a separate reconciler(used for stop_sc) that will update `status.subclusters[].shutdown`. That reconciler will be agnostic to whether we are doing stop_sandbox or stop_sc, and must rely only on `spec.subclusters[].shutdown` and podfacts.
status field must never be updated before the corresponding spec field.

roypaulin:
> Besides this, do we support unsandboxing a sandbox that is stopped? Should we error it out in webhook or we internally restart the sandbox first, then unsandbox it.

No, we are not going to support it. We should restrict what the operator can do when `shutdown` is true. We will enforce it with  a set of webhook rules.

api/v1/verticadb_types.go
roypaulin: Because it is not needed so far. This PR is only for stopping/restarting a sandbox

pkg/controllers/vdb/restart_reconciler.go
roypaulin: It is. This path is for start_db. We do not allow start_db if there is at least one node that belongs to a subcluster with shutdown true. So, the moment we find one, we exit. All subclusters must have shutdown false in order to do start_db. I will add a log msg.

pkg/controllers/vdb/podfacts.go
roypaulin: Because we want the pfacts to reflect the reality as soon as possible. The moment the user set `spec.sandboxes[].shutdown` to true, we no longer allow the operator to restart those pods. So if a reconciler collects the pfacts before the operator updates the subclusters, the will see those pods with the correct shutdown state. The sandbox owns its subcluster.
I know about the limitation( if you stop_sc in a sandbox you will get false although subcluster shutdown state is true. This PR is for stop_sandbox so I was not thinking too much about `stop_sc`).  Here is the change I can make:
```
func (p *PodFacts) isSubclusterShutdown(vdb *vapi.VerticaDB, sc *vapi.Subcluster) bool {
        if p.SandboxName == vapi.MainCluster {
                return sc.Shutdown
        }
	sb := vdb.GetSandbox(p.SandboxName)
	if sb != nil {
                 if sb.Shutdown {
                         return true
                 }
	}
	return sc.Shutdown
}
```
In case of `stop_sandbox` getting the info from the sandbox is faster.

pkg/controllers/sandbox/restartsandbox_reconciler.go
roypaulin: Either we update spec, or we update both. We must not update status before spec. status update is driven by spec. You change spec, that pushes the operator to do something and then update the status to say that thing was successfully done.
In this case you change `spec.Subcluster[].shutdown` to false, that pushes the operator to look if that subcluster is still down(and restart it if so) , and update `status.Subcluster[].shutdown` as a proof the subcluster is now up.

roypaulin:
@cchen-vertica Given that `status.sandboxes[].shutdown` is confusing too you, I will remove it. After stop_db on sandbox, I will update `spec.subclusters[].shutdown` to true(. I will add a separate reconciler(stopsubcluster_reconciler) that will iterate over the subclusters and if spec shutdown is true will stop the subcluster(if at least one node is up or the subcluster is not in a sandbox with shutdown == true) and will update `status.subclusters[].shutdown` to true.
When `spec.subclusters[].shutdown` is set back to false, the operator will startdb the sandbox and update both spec and status shutdown.
Do you find this reasonable?

roypaulin:
I am keeping `spec.sandboxes[].shutdown`. It provides a straightforward and easy way to trigger sandbox shutdown. Moreover the primary use case realdb and it is easier to update a single field so we don't have to look up the subclusters.

pkg/controllers/vdb/shutdownspec_reconciler.go
```diff
@@ -0,0 +1,94 @@
...
+				// differentiate this case from when the user is explicitly
+				// changing the subcluster's shutdown field.
+				sc.Annotations[vmeta.ShutdownDrivenBySandbox] = "true"
+			} else {
+				// If the shutdown/restart is not controlled by the sandbox,
+				// we skip to the next subcluster.
+				if !vmeta.GetShutdownDrivenBySandbox(sc.Annotations) {
+					continue
+				}
+				delete(sc.Annotations, vmeta.ShutdownDrivenBySandbox)
```
cchen-vertica, line 87: We should add "continue" after this line. When sb.Shutdown == false, we don't want to execute line 89 since single subcluster can be shut down.

cchen-vertica (commented):
All looks good, only a minor comment needs to be addressed.

pkg/controllers/vdb/shutdownspec_reconciler.go
roypaulin: Line 84 takes care of that. The "shutdown-driven-by-sandbox" is set only when shutdown is done through spec.sandboxes[]. shutdown(total sandbox shutdown). This reconciler will not touch subclusters whose shutdown is controlled by the user(user explicitly changed the subcluster's shutdown field).
There will be webhook rules to make sure users cannot change a  sandbox subcluster's shutdown field if the shutdown is controlled by the sandbox(total sandbox shutdown). This feature can keep users clusters down if they are not careful that's why I want to control its use.

cchen-vertica: That makes sense with webhook rules.

cchen-vertica (approved):

PR #957, opened 2024-10-14
trivy scan setup backup url for downloading db artifacts

HaoYang0000:
We have seen trivy scan failure multiple times due to network issues. To make it more robust, maybe add retry to those steps, so we don't have to re-trigger it manually every time. 
First option, using https://github.com/Wandalen/wretry.action. This one doesn't apply good enough in docker run, get docker os issue when trying

Second solution, add a retry step after the build: https://stackoverflow.com/questions/71574593/how-to-automatically-retry-github-action-jobs-on-failure. 

Third option, wrap the trivy script and then call with run script, that way we can use other github actions like: https://github.com/marketplace/actions/retry-step This requires more efforts and need to download the trivy binary to the repo and run as script.

Fourth option, use backup URL to download if failed, as suggested in: https://github.com/aquasecurity/trivy/discussions/7668#discussioncomment-10884984

Let's give the last option a shoot, and see if that makes the run more stable.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -120,15 +120,18 @@ jobs:
     # We never run the sarif scanner in PRs because PRs don't have permission
     # to upload the results to github.
     - name: Run the Trivy vulnerability scanner (sarif)
-      uses: aquasecurity/trivy-action@0.24.0
+      uses: Wandalen/wretry.action@master
       if: ${{ inputs.run_security_scan == 'all' && github.event_name != 'pull_request' }}
       with:
-        image-ref: ${{ steps.full_vertica_image.outputs.value }}
-        ignore-unfixed: true
-        security-checks: vuln
-        timeout: '20m0s'
-        format: 'sarif'
-        output: 'trivy-results-vertica-image.sarif'
+        action: aquasecurity/trivy-action@0.24.0
+        retry_condition: steps._this.outputs.code == 0
+        with:
+          image-ref: ${{ steps.full_vertica_image.outputs.value }}
+          ignore-unfixed: true
+          security-checks: vuln
+          timeout: '20m0s'
+          format: 'sarif'
+          output: 'trivy-results-vertica-image.sarif'
 
     - name: Upload Trivy scan results to GitHub Security tab
       uses: github/codeql-action/upload-sarif@v3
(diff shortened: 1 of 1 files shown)


HaoYang0000:
We can see that the fallback url was caught in one of the test run: https://github.com/vertica/vertica-kubernetes/actions/runs/11343558108/job/31546362845

2024-10-15T09:50:49Z	ERROR	[vulndb] Failed to download artifact	repo="ghcr.io/aquasecurity/trivy-db:2" err="oci download error: failed to fetch the layer: GET https://ghcr.io/v2/aquasecurity/trivy-db/blobs/sha256:8dcb18f131bc5d8d17159f7682ebd49b2ea9c784edbb03aa87fe92927ff1d851: TOOMANYREQUESTS: retry-after: 719.785µs, allowed: 44000/minute"
2024-10-15T09:50:49Z	INFO	[vulndb] Trying to download artifact from other repository...
2024-10-15T09:50:49Z	INFO	[vulndb] Downloading artifact...	repo="public.ecr.aws/aquasecurity/trivy-db:2"

.github/workflows/build-images.yml
```diff
@@ -146,6 +149,9 @@ jobs:
         timeout: '20m0s'
         format: 'table'
         output: 'trivy-results-vertica-image.out'
+      env:
+        TRIVY_DB_REPOSITORY: ghcr.io/aquasecurity/trivy-db,public.ecr.aws/aquasecurity/trivy-db
```
cchen-vertica, line 153: Could we define the env vars at job level then we can easily change the repos in future?

roypaulin:
What's the trivy version in the fallback url?

.github/workflows/build-images.yml
HaoYang0000: Sure, updated in https://github.com/vertica/vertica-kubernetes/pull/957/commits/cbdbd45965fc99d402b47c73742878051c3448ea

HaoYang0000:
> What's the trivy version in the fallback url?

Those URL are for trivy to download security knowledges to its running process database, I don't think the trivy version is provided in the URL. 
The problem here is that the default ghcr.io domain can't take too many requests and sometimes it failed to handle the HTTP request. Then the fallback URL can be a backup server for trivy to download required data. 

But for this feature to work(fallback URL), it required to use github action trivy-action >=0.27, which is trivy v0.56.1.

cchen-vertica (approved):

PR #959, opened 2024-10-14
Remove sandbox and subclusters from vdb at the same time

roypaulin:
This fixes the issue that happens when you remove a sandbox and all its subclusters at the same time from the vdb, the pods will be down but will never get deleted.
Now, as soon as unsandbox is done the operator will find the statefulset make it rejoin the main cluster where it will be removed.

### DIFF
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -1605,7 +1605,15 @@ func (v *VerticaDB) checkImmutableSubclusterInSandbox(oldObj *VerticaDB, allErrs
 	newScMap := v.GenSubclusterMap()
 	path := field.NewPath("spec").Child("subclusters")
 
-	// check if the sizes of subclusters in a sandbox get changed
+	allErrs = v.checkSubclusterSizeInSandbox(allErrs, persistScsWithSbIndex, newScIndexMap, oldScMap,
+		newScMap, path)
+	allErrs = v.checkSandboxSubclustersRemoved(allErrs, oldObj, oldScIndexMap, oldScMap, newScMap, path)
+	return v.checkSandboxPrimary(allErrs, oldObj, oldScIndexMap, oldScMap, path)
+}
+
+// checkSubclusterSizeInSandbox checks if the sizes of subclusters in a sandbox get changed
+func (v *VerticaDB) checkSubclusterSizeInSandbox(allErrs field.ErrorList, persistScsWithSbIndex, newScIndexMap map[string]int,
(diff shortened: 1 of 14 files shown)


pkg/iter/sc_finder_test.go
```diff
@@ -326,6 +326,29 @@ var _ = Describe("sc_finder", func() {
...
+		vdb.Status.Sandboxes = []vapi.SandboxStatus{
+			{Name: sbName, Subclusters: scNames[:2]},
+		}
+		test.CreatePods(ctx, k8sClient, vdb, test.AllPodsRunning)
+		defer test.DeletePods(ctx, k8sClient, vdb)
+
+		vdb.Spec.Subclusters = vdb.Spec.Subclusters[1:]
+		vdb.Spec.Subclusters = vdb.Spec.Subclusters[:2]
+		scNames = append(scNames[0:1], scNames[3])
+		verifySubclusters(ctx, vdb, scNames, []int32{0, 0}, sbName, FindNotInVdbAcrossSandboxes)
```
cchen-vertica, line 350: Why the size is set to {0, 0}?

api/v1/verticadb_webhook.go
```diff
@@ -1625,31 +1633,45 @@ func (v *VerticaDB) checkImmutableSubclusterInSandbox(oldObj *VerticaDB, allErrs
...
 		}
 	}
+	return allErrs
+}
 
-	// find subclusters that are sandboxed in old vdb but removed in new vdb
+// checkSandboxSubclustersRemoved checks subclusters that are sandboxed in old vdb but removed in new vdb
+func (v *VerticaDB) checkSandboxSubclustersRemoved(allErrs field.ErrorList, oldObj *VerticaDB, oldScIndexMap map[string]int,
+	oldScMap, newScMap map[string]*Subcluster, path *field.Path) field.ErrorList {
 	oldScInSandbox := oldObj.GenSubclusterSandboxMap()
```
cchen-vertica, line 1642: We should modify this to: newScInSandbox := v.GenSubclusterSandboxMap(). Then in line 1647, verify the removed sc isn't in newScInSandbox. The check in line 1650(if v.GetSandbox(sb) == nil) is not accurate since the sc could be in a new sandbox of new vdb.

pkg/iter/sc_finder_test.go
roypaulin: Because when the operator constructs subclusters no longer in vdb from their statefulsets, it gives them size 0 as they will soon be removed.

cchen-vertica (approved):
In your next PR: you can remove the timeout you added in the e2e tests.

PR #961, opened 2024-10-17
online upgrade with package installation and pending pods

qindotguan:
Fixes two issues during the online upgrade process:
1. Try to restart the pending nodes and report a warning event to remind the user to check the DB configuration.
2. To print a hint that package installation failure may be due to the lack of memory resources.

An e2e test was added to cover the test with package installation and a pending pod.


pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -290,6 +292,30 @@ func (r *OnlineUpgradeReconciler) loadUpgradeState(ctx context.Context) (ctrl.Re
...
+// checkUpNodes will requeue the upgrade process if not all the main cluster nodes are up.
+func (r *OnlineUpgradeReconciler) checkUpNodes(ctx context.Context) (ctrl.Result, error) {
+	// We skip this if we have already added the new subclusters
+	if vmeta.GetOnlineUpgradeStepInx(r.VDB.Annotations) > addSubclustersInx {
+		return ctrl.Result{}, nil
+	}
+
+	mainPFacts := r.PFacts[vapi.MainCluster]
+	// All nodes in the main cluster must be up before running online upgrade
+	if mainPFacts.getUpNodeCount() != len(mainPFacts.Detail) {
```
roypaulin, line 304: We want all the pods to be running before starting online upgrade. They don't all have to be up.

qindotguan: It seems not working after changed to `countRunningAndInstalled()`. When the pending nodes just became running (not up yet), the upgrade process continued to duplicate the main cluster but returned VAddnode error. That running node and the new added nodes never came up.

roypaulin: In that case, don't even check the number of upnode. Just call the restart reconciler, if any node is down it will start it, if no node is down it will do nothing, if any pod is not running, it will requeue.

pkg/vadmin/install_packages_at.go
```diff
@@ -34,7 +35,8 @@ func (a *Admintools) InstallPackages(ctx context.Context, opts ...installpackage
 
 	status := genInstallPackageStatus(stdout)
 	if err != nil && len(status.Packages) == 0 {
-		_, logErr := a.logFailure("install_package", events.InstallPackagesFailed, stdout, err)
+		pkgErr := errors.New(err.Error() + "This may due to lack of memory resources.")
```
HaoYang0000, line 38: Do we have a way to distinguish the cause of the error? Simply adding a string here can be misleading for future debugging.

pkg/controllers/vdb/onlineupgrade_reconciler.go
qindotguan: Thanks for the instruction. It is updated accordingly.

pkg/vadmin/install_packages_at.go
qindotguan: Thanks. Now it only logs a message if 0 package was installed.

pkg/controllers/vdb/upgrade.go
```diff
@@ -845,8 +845,17 @@ func (i *UpgradeManager) routeClientTraffic(ctx context.Context, pfacts *PodFact
 func (i *UpgradeManager) logEventIfRequestedUpgradeIsDifferent(actualUpgrade vapi.UpgradePolicyType) {
 	if !i.ContinuingUpgrade && i.Vdb.Spec.UpgradePolicy != actualUpgrade && i.Vdb.Spec.UpgradePolicy != vapi.AutoUpgrade {
 		actualUpgradeAsText := strings.ToLower(string(actualUpgrade))
+
 		i.Rec.Eventf(i.Vdb, corev1.EventTypeNormal, events.IncompatibleUpgradeRequested,
 			"Requested upgrade is incompatible with the Vertica deployment. Falling back to %s upgrade.", actualUpgradeAsText)
+
+		if i.Vdb.Spec.UpgradePolicy == "Online" {
```
roypaulin, line 852: Let's move this before the event recorder.

roypaulin:
The main now has the lint fix.

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -165,6 +165,8 @@ func (r *OnlineUpgradeReconciler) Reconcile(ctx context.Context, _ *ctrl.Request
 
 	// Functions to perform when the image changes.  Order matters.
 	funcs := []func(context.Context) (ctrl.Result, error){
+		// Requeue if not all nodes are running
+		r.requeuePodsNotRunning,
```
roypaulin, line 169: Add a status msg in `onlineUpgradeStatusMsgs` for this step.

tests/e2e-leg-9/online-upgrade-with-package-install/02-assert.yaml
roypaulin, line 1: This new test does too many things already validated by other tests. I think what you want to check is that we do not pass the first function in online upgrade if some pods are pending so make some pod pending, start online upgrade, wait a little and check the upgradeStatus to verify that we are still stuck in that function, make the pending pods run and now wait until we create the new subclusters and the test can end there.

pkg/controllers/vdb/onlineupgrade_reconciler.go
qindotguan: Done.

tests/e2e-leg-9/online-upgrade-with-package-install/02-assert.yaml
qindotguan: Good suggestion. I removed the duplicated checks and renamed the test to online-upgrade-pods-pending.

pkg/controllers/vdb/onlineupgrade_reconciler.go
roypaulin: Let's put them after postStartOnlineUpgradeMsg

tests/e2e-leg-9/online-upgrade-pods-pending/26-assert.yaml
```diff
@@ -0,0 +1,20 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: v-base-upgrade-sec2
+status:
+  availableReplicas: 0
+  replicas: 1
```
roypaulin, line 20: It does not necessarily mean the pods are pending. You need to validate the pod itself:
```suggestion
apiVersion: v1
kind: Pod
metadata:
  name: v-base-upgrade-sec2-0
status:
 phase: pending
```

tests/e2e-leg-9/online-upgrade-pods-pending/29-set-pod-running.yaml
roypaulin, line 1: Add an assert file where you check if the pod is running

tests/e2e-leg-9/online-upgrade-pods-pending/27-initiate-upgrade.yaml
roypaulin, line 1: Add an assert that checks the upgradeStatus:

```
apiVersion: v1
kind: Pod
metadata:
  name: v-base-upgrade-sec2-0
status:
 phase: pending
---
kind: VerticaDB
metadata:
  name: v-base-upgrade
status:
  upgradeStatus: "Requeue as pods not running"
```

tests/e2e-leg-9/online-upgrade-pods-pending/26-assert.yaml
qindotguan: Thanks. This is helpful.

tests/e2e-leg-9/online-upgrade-pods-pending/27-initiate-upgrade.yaml
qindotguan: Updated according to the comments. Please help to review again.

tests/e2e-leg-9/online-upgrade-pods-pending/setup-vdb/base/setup-vdb.yaml
```diff
@@ -24,7 +24,7 @@ metadata:
 spec:
   image: kustomize-vertica-image
   dbName: repUP
-  initPolicy: Create
+  initPolicy: CreateSkipPackageInstall
```
roypaulin, line 27: Why?

qindotguan: As the test online-upgrade-pods-pending stops at the steps of wait-for-scs-mimic, the package installation (at the step of a new Vertica version is installed) is not covered in the test. So we can either
1. use CreateSkipPackageInstall to skip 
2. add the upgrade steps back to cover the package installation

roypaulin (approved):

PR #962, opened 2024-10-17
stabilize e2e test case sandbox-basic

LiboYu2:
Increased the timeout for steps that frequently fail. 
Add some latency to stabilize the cluster before moving to the unsandboxing step.
I ran the test cases locally for 5 times in a roll and they all passed.


kuttl-test.yaml
```diff
@@ -13,6 +13,7 @@
 
 apiVersion: kuttl.dev/v1beta1
 kind: TestSuite
+kindNodeCache: true
```
LiboYu2, line 16: This allows the downloaded images to be cached on the node. This will speed up test case execution.

tests/e2e-leg-10/sandbox-basic/59-wait-for-steady-state.yaml
```diff
@@ -15,3 +15,4 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
   - command: bash -c "../../../scripts/wait-for-verticadb-steady-state.sh -n verticadb-operator -t 360 $NAMESPACE"
+  - command: sleep 120
```
LiboYu2, line 18: Add this latency to stabilize the cluster before step 60 starts.

roypaulin: You don't need to add this sleep. It will take the test longer to complete. **If** this usually fails due to insufficient timeout, then just increase the timeout in the script call above.

LiboYu2: This is different from timeout. The timeout is the maximum wait time for the whole test step to finish. This sleep call gives the cluster some time to stabilize its state before the next step starts. It makes the test run longer but it makes the test pass.

roypaulin: The script on line 17 waits for the operator to be steady(gives the cluster some time to stabilize its state before the next step) meaning there is no error and nothing going on. There is no benefit in adding another wait after. If this step fails and more time is needed, you should increase the time passed as argument to the script.

LiboYu2: This is from that wait-for-verticadb-steady-state.sh

`timeout $TIMEOUT bash -c -- "while ! $LOG_CMD | \
    grep $WEBHOOK_FILTER | \
    grep $DEPRECATION_FILTER | \
    grep $VDB_FILTER | \
    tail -1 | grep --quiet '\"result\": {\"Requeue\":false,\"RequeueAfter\":0}, \"err\": null'; do sleep 1; done" &
pid=$!
wait $pid
`

This is from "man timeout":
`timeout - run a command with a time limit`

If the script runs longer than the $TIMEOUT, an error will be reported. 
If the script finishes within the $TIMEOUT, the next step starts right away.
What I want to achieve is to add a latency between the two steps to make sure the latter
step is not impacted by the previous step. Increasing the timeout will not achieve that.

cchen-vertica:
As discussed, we should use an environment variable for all extended timeout values (900), and try to print all the events on vdb.

LiboYu2:
> > As discussed, we should use an environment variable for all extended timeout values (900), and try to print all the events on vdb.
> 
> What do you mean by "printing all the events on vdb"

Cai's point is to let vdb receive all the events.

LiboYu2:
> As discussed, we should use an environment variable for all extended timeout values (900), and try to print all the events on vdb.

Totally forgot about that:)  I just tried it out. It turns out environmental variables can only be used in commands.  https://kuttl.dev/docs/testing/steps.html#running-commands

We cannot use the environmental variables as a configuration parameter. Here is the error I got after using environmental variables:    harness.go:397: loading /home/lyu/test-repos/vertica-kubernetes/tests/e2e-leg-10/sandbox-basic/60-assert.yaml: error converting unstructured object TestAssert:/ (/home/lyu/test-repos/vertica-kubernetes/tests/e2e-leg-10/sandbox-basic/60-assert.yaml): error converting TestAssert:/ from unstructured error: unrecognized type: int

tests/e2e-leg-10/sandbox-basic/setup-vdb/base/setup-vdb.yaml
```diff
@@ -23,7 +23,7 @@ spec:
   initPolicy: CreateSkipPackageInstall
   communal: {}
   local:
-    requestSize: 250Mi
+    requestSize: 270Mi
```
cchen-vertica, line 26: Will 20Mi make any difference? We shouldn't make this change.

tests/e2e-leg-10/sandbox-basic/60-unsandbox-sec3.yaml
```diff
@@ -10,7 +10,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+apiVersion: kuttl.dev/v1beta1
```
cchen-vertica, line 13: The changes in this file should be removed since you already extended the timeout in 60-assert.yaml.

cmd/operator/main.go
```diff
@@ -239,7 +240,7 @@ func main() {
 	if opcfg.GetLoggingFilePath() != "" {
 		log.Printf("Now logging in file %s", opcfg.GetLoggingFilePath())
 	}
-
+	var multibroadcaster = record.NewBroadcasterWithCorrelatorOptions(record.CorrelatorOptions{BurstSize: 100})
```
roypaulin, line 243: Let's just make this configurable through an env var. Look at examples in opcfg/config.go5(DEPLOY_WITH? WEBHOOKS_ENABLED). See how and where they are changed and do the same for the burstsize. By default it will be 25(k8s default), if a value lower than 25 is passed we still pick 25.

tests/e2e-leg-10/sandbox-basic/setup-vdb/base/setup-vdb.yaml
LiboYu2: I noticed the vdb received low disk space event when I ran the test locally. When free disk space is less than 10MB, that event will be sent.

cmd/operator/main.go
LiboYu2: Made the change. Please take a look.

roypaulin: @LiboYu2,  I do not see the change.

LiboYu2: Forgot to push it. My bad. Now you should be able to see it.

tests/kustomize-defaults.cfg
```diff
@@ -111,6 +111,11 @@ USE_SERVER_MOUNT_PATCH=
 # expansion.  Set this value to 1 for those environments.
 ALLOW_VOLUME_EXPANSION=
 
+# this will set up the threshold used by the spam filer 
+# default threshold is 25, which is too low to run the test cases
+# some test cases rely on event verification
+BROADCASTER_BURST_SIZE=100
```
roypaulin, line 117: It is not needed here.

config/manager/operator-envs
```diff
@@ -14,3 +14,4 @@ CONCURRENCY_VERTICARESTOREPOINTSQUERY
 CONCURRENCY_VERTICASCRUTINIZE
 CONCURRENCY_SANDBOXCONFIGMAP
 CONCURRENCY_VERTICAREPLICATOR
+BROADCASTER_BURST_SIZE
```
roypaulin, line 17: Sorry I forgot to tell you. You also need to add this to `template-helm-chart.sh`. Also need to add a new parameter to values.yaml. Look at `WEBHOOKS_ENABLED`(line 227 in template-helm-chart.sh)  as an example and do the same for `BROADCASTER_BURST_SIZE`

LiboYu2: No worries. I added a new commit to address that. You can take a look.  I built locally and deployed to my local kind cluster. The burst size is 100. No external environmental variable was used.

pkg/opcfg/config.go
```diff
@@ -34,6 +34,16 @@ func GetIsWebhookEnabled() bool {
 	return lookupBoolEnvVar("WEBHOOKS_ENABLED", envMustExist)
 }
 
+// GetBroadcasterBurstSize returns the customizable burst size for broadcaster.
+func GetBroadcasterBurstSize() int {
+	burstSize := lookupIntEnvVar("BROADCASTER_BURST_SIZE", envCanNotExist)
+	if burstSize < 25 {
```
HaoYang0000, line 40: Minor: We can make a constant DEFAULT_BURST_SIZE = 25 for maintaining the int, and we don't need an else block here, we can do:
`if burstSize >= DEFAULT_BURST_SIZE {
    return burstSize
}
return DEFAULT_BURST_SIZE 
`

tests/e2e-leg-10/sandbox-basic/60-assert.yaml
```diff
@@ -11,6 +11,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 900
```
HaoYang0000, line 16: I remember the default timeout for all job is 1000, any reason we want to reduce the time out here?

roypaulin: The default is 600s in kuttl-test.yaml

helm-charts/verticadb-operator/values.yaml
```diff
@@ -86,6 +86,10 @@ webhook:
   # can be used to skip that and still deploy the operator.
   enable: true
 
+  # this will increase the default threshold used by spam filter in controller runtime from 25 to 100
+  # this is important for running e2e test as some test cases require verification of events
+  burstSize: 100
```
roypaulin, line 91: Move this under controllers.

pkg/opcfg/config.go
LiboYu2: made the change.

helm-charts/verticadb-operator/values.yaml
```diff
@@ -86,6 +86,10 @@ webhook:
   # can be used to skip that and still deploy the operator.
   enable: true
 
+  # this will increase the default threshold used by spam filter in controller runtime from 25 to 100
+  # this is important for running e2e test as some test cases require verification of events
```
roypaulin, line 90:
```suggestion
  # Use this if you want to provide a custom burst size for event recording in the operator.
  # Increasing this allows the controllers to record more events in a short period.
```

helm-charts/verticadb-operator/values.yaml
LiboYu2: Moved.

tests/e2e-leg-10/sandbox-basic/setup-vdb/base/setup-vdb.yaml
cchen-vertica: Then give it more like 500Mi.

cchen-vertica:
The unit tests failed. Some go-lint errors exist. You need to fix those errors.

tests/e2e-leg-10/sandbox-basic/setup-vdb/base/setup-vdb.yaml
LiboYu2: 500Mi is great for ci server. But when we run the test locally, it may be a challenge if the free disk space is not large enough. When parallel is set to 2, there will 20 pods at most at the same time. How about 300Mi?

roypaulin (approved):

PR #965, opened 2024-10-22
increase hugeparam default size

HaoYang0000:
The new linter version breaks unit test because of the hugeParam size

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,11 +22,6 @@ linters-settings:
       - octalLiteral
       - whyNoLint
       - wrapperFunc
-    settings:
-      hugeParam:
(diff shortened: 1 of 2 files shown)


roypaulin:
Let's instead use pointers in verticadb_conversion.go. Like in Cai's PR. Can you update it here?

HaoYang0000:
> Let's instead use pointers in verticadb_conversion.go. Like in Cai's PR. Can you update it here?

Yeah, updated in https://github.com/vertica/vertica-kubernetes/pull/965/commits/aca9bb9784b1c1ad960e3de06c6b25e25785b89f

roypaulin (approved):
Looks good! I will merge it once all the tests pass.

PR #968, opened 2024-10-23
Delete a subcluster pods after it is shut down

roypaulin:
After a subcluster is "shut down"(shutdown set to true), its pods will stay down and cannot get restarted by the operator. We go a step further and delete the pods by scaling the statefulset to zero. This is useful for realdb as we do not want to keep down pods around and the EKS instances will also be shut down.
When shutdown is set back to false, the statefulset will be scale up to its original size(subcluster size in spec).


pkg/controllers/sandbox/scalestatefulset_reconciler.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+			newSize := sc.GetStsSize(s.Vdb)
+			if oldSize == &newSize {
+				return nil
+			}
+			sts.Spec.Replicas = &newSize
+			return s.VRec.GetClient().Update(ctx, sts)
+		})
+		if err != nil {
+			return ctrl.Result{}, err
```
HaoYang0000, line 78: Should we log the error here, or the reconciler has a layer to handle those errors(logging)?

pkg/controllers/vdb/status_reconciler.go
```diff
@@ -196,6 +196,14 @@ func (s *StatusReconciler) calculateSubclusterStatus(ctx context.Context, sc *va
 		if !ok {
 			continue
 		}
+		stsSize := sc.GetStsSize(s.Vdb)
+		if stsSize == 0 && stsSize != sc.Size {
+			s.setSubclusterStatusWhenShutdown(podIndex, curStat)
```
HaoYang0000, line 201: If we want to skip the value assignment, why do we need to change the value to false on UpNode, installed, AddedToDB?

roypaulin: In case they were true before statefulset's scale down. After scale down those fields will always be false because the pods have been deleted.

pkg/controllers/sandbox/scalestatefulset_reconciler.go
roypaulin: Yes, the sandbox controller, which is the top layer caller, will log the error.

pkg/controllers/sandbox/scalestatefulset_reconciler.go
```diff
@@ -0,0 +1,82 @@
...
+			if err != nil {
+				return err
+			}
+			oldSize := sts.Spec.Replicas
+			sc := scMap[sts.Labels[vmeta.SubclusterNameLabel]]
+			if sc == nil {
+				return nil
+			}
+			newSize := sc.GetStsSize(s.Vdb)
+			if oldSize == &newSize {
```
cchen-vertica, line 71: Why do we compare the addressses here? We should do "if *oldSize == newSize".

pkg/controllers/sandbox/scalestatefulset_reconciler.go
```diff
@@ -0,0 +1,82 @@
...
+				return nil
+			}
+			sts.Spec.Replicas = &newSize
+			return s.VRec.GetClient().Update(ctx, sts)
```
cchen-vertica, line 75: Should we recreate the statefulset here? Otherwise, we will shutdown the pods one by one.

api/v1/helpers.go
```diff
@@ -970,6 +985,22 @@ func (v *VerticaDB) GetSubclusterSandboxName(scName string) string {
 	return MainCluster
 }
 
+func (v *VerticaDB) GenShutdownSubclusterMap() map[string]*Subcluster {
```
cchen-vertica, line 988: This function isn't used in this PR?

pkg/controllers/sandbox/scalestatefulset_reconciler.go
roypaulin: I don't think it is a problem; this is actually the recommended way to scale down stateful app. This is how we have always done scale down. The default grace period is zero anyway so each pod will get deleted quickly. But if the user has set a custom grace period, we still want to respect it. We usually recreate the sts when we need the pods back running asap.

pkg/controllers/sandbox/scalestatefulset_reconciler.go
```diff
@@ -0,0 +1,82 @@
...
+		err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			nm := names.GenNamespacedName(s.Vdb, sts.Name)
+			err := s.VRec.GetClient().Get(ctx, nm, sts)
+			if err != nil {
+				return err
+			}
+			oldSize := sts.Spec.Replicas
+			sc := scMap[sts.Labels[vmeta.SubclusterNameLabel]]
+			if sc == nil {
+				return nil
```
LiboYu2, line 68: Do we need to raise an error here? Same question for line 72.

roypaulin: I will add at line 68. We don't need at 72 as it is just a no-op

cchen-vertica (approved):

PR #971, opened 2024-10-28
webhook rules for sandbox shutdown

LiboYu2:
Here are the rules that are implemented by this change.

- When creating a vdb, all subclusters and sandboxes(if any) must not have `shutdown` set to true.
- If a sandbox has `shutdown`(spec) true and any of its subclusters has the annotation "vertica.com/shutdown-driven-by-sandbox" set to true in the old vdb, we should prevent the user from updating that subcluster's shutdown field.
- You cannot unsandbox a subcluster with `shutdown` true(spec and/or status) or part of a sandbox with `shutdown` true
- You cannot change spec.sandboxes[].image if the sandbox has `shutdown` true or has any subcluster with `shutdown` true(spec and/or status)
- You cannot terminate a sandbox if it has has `shutdown` true. terminate means removing the sandbox in spec.sandboxes[] and its subclusters in spec.subclusters[] at the same time.

### DIFF
--- a/api/v1/helpers.go
+++ b/api/v1/helpers.go
@@ -226,6 +226,16 @@ func (v *VerticaDB) GenStatusSubclusterMap() map[string]*SubclusterStatus {
 	return statusSclusterMap
 }
 
+// GenStatusSClusterIndexMap will organize all of the subclusters into a map so we
+// can quickly find its index in the status.subclusters[] array.
+func (v *VerticaDB) GenStatusSClusterIndexMap() map[string]int {
+	m := make(map[string]int)
+	for i := range v.Status.Subclusters {
+		m[v.Status.Subclusters[i].Name] = i
+	}
+	return m
+}
+
 // GenSandboxSubclusterMapForUnsandbox will compare sandbox status and spec
 // for finding subclusters that need to be unsandboxed, this function returns a map
 // with sandbox name as the key and its subclusters (need to be unsandboxed) as the value
--- a/api/v1/helpers.go
+++ b/api/v1/helpers.go
@@ -216,7 +216,7 @@ func (v *VerticaDB) GenStatusSandboxMap() map[string]*SandboxStatus {
 	return statusSboxMap
 }
 
-// GenStatusSandboxMap() returns a map from status. The key is subcluster name and value is the subcluster pointer
+// GenStatusSubclusterMap() returns a map from status. The key is subcluster name and value is the subcluster pointer
 func (v *VerticaDB) GenStatusSubclusterMap() map[string]*SubclusterStatus {
 	statusSclusterMap := make(map[string]*SubclusterStatus)
 	for i := range v.Status.Subclusters {
--- a/api/v1/helpers.go
+++ b/api/v1/helpers.go
@@ -206,6 +206,26 @@ func (v *VerticaDB) GenSubclusterSandboxStatusMap() map[string]string {
 	return scSbMap
 }
 
+// GenStatusSandboxMap() returns a map from status. The key is sandbox name and value is the sandbox pointer
+func (v *VerticaDB) GenStatusSandboxMap() map[string]*SandboxStatus {
+	statusSboxMap := make(map[string]*SandboxStatus)
+	for i := range v.Status.Sandboxes {
+		sBox := &v.Status.Sandboxes[i]
+		statusSboxMap[sBox.Name] = sBox
+	}
+	return statusSboxMap
+}
+
+// GenStatusSandboxMap() returns a map from status. The key is subcluster name and value is the subcluster pointer
+func (v *VerticaDB) GenStatusSubclusterMap() map[string]*SubclusterStatus {
+	statusSclusterMap := make(map[string]*SubclusterStatus)
+	for i := range v.Status.Subclusters {
+		sCluster := &v.Status.Subclusters[i]
+		statusSclusterMap[sCluster.Name] = sCluster
+	}
+	return statusSclusterMap
+}
+
 // GenSandboxSubclusterMapForUnsandbox will compare sandbox status and spec
 // for finding subclusters that need to be unsandboxed, this function returns a map
 // with sandbox name as the key and its subclusters (need to be unsandboxed) as the value
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -1788,6 +1788,7 @@ func (v *VerticaDB) checkUnsandboxShutdownConditions(oldObj *VerticaDB, allErrs 
 	statusSClusterIndexMap := v.GenStatusSClusterIndexMap()
 	subclustersToUnsandbox := v.findSclustersToUnsandbox(oldObj)
 	subclusterIndexMap := v.GenSubclusterIndexMap()
+	sandboxIndexMap := v.GenSandboxIndexMap()
 	sandboxesWithError := map[string]bool{}
 	for oldSubcluster, oldSandbox := range subclustersToUnsandbox {
 		oldSubclusterIndex, found := statusSClusterIndexMap[oldSubcluster.Name]
@@ -1802,6 +1803,7 @@ func (v *VerticaDB) checkUnsandboxShutdownConditions(oldObj *VerticaDB, allErrs 
 			continue
 		}
 		if oldSandbox.Shutdown {
+			i := sandboxIndexMap[oldSandbox.Name]
 			_, found := sandboxesWithError[oldSandbox.Name]
 			if !found { // this is to avoid duplicate error messages
 				sandboxesWithError[oldSandbox.Name] = true
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -1599,17 +1599,17 @@ func (v *VerticaDB) findPersistScsInSandbox(oldObj *VerticaDB) map[string]int {
 }
 
 // findPersistSandboxes returns a slice of sandbox names that are in both new and old vdb.
-func (v *VerticaDB) findPersistSandboxes(oldObj *VerticaDB) []string {
+func (v *VerticaDB) findPersistSandboxes(oldObj *VerticaDB) map[string]bool {
 	oldSandboxes := oldObj.Spec.Sandboxes
 	newSandboxes := v.Spec.Sandboxes
 	oldSandboxMap := make(map[string]int)
 	for _, oldSandbox := range oldSandboxes {
 		oldSandboxMap[oldSandbox.Name] = 1
 	}
-	persistSandboxes := make([]string, 0)
+	persistSandboxes := make(map[string]bool)
 	for _, newSandbox := range newSandboxes {
 		if _, ok := oldSandboxMap[newSandbox.Name]; ok {
-			persistSandboxes = append(persistSandboxes, newSandbox.Name)
+			persistSandboxes[newSandbox.Name] = true
 		}
 	}
 	return persistSandboxes
@@ -1757,7 +1757,7 @@ func (v *VerticaDB) checkNewSBoxOrSClusterShutdownUnset(allErrs field.ErrorList)
 	return allErrs
 }
 
-// findSclustersToUnsandboxreturn a map whose key is the pointer of a subcluster that is to be unsandboxed.
+// findSclustersToUnsandbox return a map whose key is the pointer of a subcluster that is to be unsandboxed.
 // The value is the pointer to a sandbox where the subcluster lives in
 func (v *VerticaDB) findSclustersToUnsandbox(old runtime.Object) map[*Subcluster]*Sandbox {
 	oldObj := old.(*VerticaDB)
@@ -1768,13 +1768,16 @@ func (v *VerticaDB) findSclustersToUnsandbox(old runtime.Object) map[*Subcluster
 	newSubclusterMap := v.GenSubclusterMap()
 	sclusterSboxMap := make(map[*Subcluster]*Sandbox)
 	for oldSubclusterName, oldSandboxName := range oldSubclusterInSandbox {
-		_, oldSubclusterInNewSandboxes := newSuclusterInSandbox[oldSubclusterName]
+		newSandboxName, oldSubclusterInNewSandboxes := newSuclusterInSandbox[oldSubclusterName]
 		_, oldSubclusterInPersist := newSubclusterMap[oldSubclusterName]
+		oldSandbox := oldSandboxMap[oldSandboxName]
 		// for unsandboxing, check shutdown field of the subcluster and sandbox
-		if !oldSubclusterInNewSandboxes && oldSubclusterInPersist {
-			oldSubcluster := oldSubclusterMap[oldSubclusterName]
-			oldSandbox := oldSandboxMap[oldSandboxName]
-			sclusterSboxMap[oldSubcluster] = oldSandbox
+		if oldSubclusterInPersist {
+			if !oldSubclusterInNewSandboxes || (oldSubclusterInNewSandboxes && oldSandbox.Name != newSandboxName) {
+				//either subcluster is not in any sbox or is moved to a different sbox in new vdb
+				oldSubcluster := oldSubclusterMap[oldSubclusterName]
+				sclusterSboxMap[oldSubcluster] = oldSandbox
+			}
 		}
 	}
 	return sclusterSboxMap
@@ -1787,9 +1790,10 @@ func (v *VerticaDB) checkUnsandboxShutdownConditions(old runtime.Object, allErrs
 	subclustersToUnsandbox := v.findSclustersToUnsandbox(old)
 	subclusterIndexMap := v.GenSubclusterIndexMap()
 	sandboxIndexMap := v.GenSandboxIndexMap()
+	unsandboxIndex := -1
 	for oldSubcluster, oldSandbox := range subclustersToUnsandbox {
 		oldSubclusterIndex, found := statusSClusterIndexMap[oldSubcluster.Name]
-		if oldSubcluster.Shutdown || found && v.Status.Subclusters[oldSubclusterIndex].Shutdown {
+		if oldSubcluster.Shutdown || (found && v.Status.Subclusters[oldSubclusterIndex].Shutdown) {
 			i := subclusterIndexMap[oldSubcluster.Name]
 			p := field.NewPath("spec").Child("subclusters").Index(i)
 			err := field.Invalid(p,
@@ -1801,13 +1805,16 @@ func (v *VerticaDB) checkUnsandboxShutdownConditions(old runtime.Object, allErrs
 		}
 		if oldSandbox.Shutdown {
 			i := sandboxIndexMap[oldSandbox.Name]
-			p := field.NewPath("spec").Child("sandboxes").Index(i)
-			err := field.Invalid(p,
-				oldSubcluster.Name,
-				fmt.Sprintf("cannot unsandbox subcluster %q in sandbox %q that has Shutdown field set to true",
-					oldSubcluster.Name, oldSandbox.Name))
-			allErrs = append(allErrs, err)
-			continue
+			if i != unsandboxIndex { // this is to avoid duplicate error messages
+				unsandboxIndex = i
+				p := field.NewPath("spec").Child("sandboxes").Index(i)
+				err := field.Invalid(p,
+					oldSubcluster.Name,
+					fmt.Sprintf("cannot unsandbox subcluster %q in sandbox %q that has Shutdown field set to true",
+						oldSubcluster.Name, oldSandbox.Name))
+				allErrs = append(allErrs, err)
+				continue
+			}
 		}
 	}
 	return allErrs
@@ -1823,7 +1830,7 @@ func (v *VerticaDB) checkAnnotatedSubclustersInShutdownSandbox(old runtime.Objec
 	newSubclusterIndexMap := v.GenSubclusterIndexMap()
 	newSubclusterMap := v.GenSubclusterMap()
 	oldSubclusterMap := oldObj.GenSubclusterMap()
-	for _, sandboxName := range persistSandboxes {
+	for sandboxName := range persistSandboxes {
 		newSandbox := newSandboxMap[sandboxName]
 		oldSandbox := oldSandboxMap[sandboxName]
 		if newSandbox.Shutdown {
@@ -1856,7 +1863,7 @@ func (v *VerticaDB) checkShutdownSandboxImage(old runtime.Object, allErrs field.
 	newSandboxIndexMap := v.GenSandboxIndexMap()
 	newSubclusterMap := v.GenSubclusterMap()
 	statusSClusterIndexMap := v.GenStatusSClusterIndexMap()
-	for _, sandboxName := range persistSandboxes {
+	for sandboxName := range persistSandboxes {
 		oldSandbox := oldSandboxMap[sandboxName]
 		newSandbox := newSandboxMap[sandboxName]
 		if oldSandbox.Image != newSandbox.Image {
@@ -1906,6 +1913,7 @@ func (v *VerticaDB) checkTerminatingSandboxes(old runtime.Object, allErrs field.
 }
 
 // checkSboxForShutdown checks if sbox or its scluster has shutdown field set to true in spec/status
+// It returns a slice of strings. Each string is an error message
 func (v *VerticaDB) checkSboxForShutdown(newSandbox *Sandbox, newSClusterMap map[string]*Subcluster, statusSClusterIndexMap map[string]int) []string {
 	errMsgs := []string{}
 	if newSandbox.Shutdown {
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -1764,19 +1764,15 @@ func (v *VerticaDB) findSclustersToUnsandbox(oldObj *VerticaDB) map[*Subcluster]
 	newSuclusterInSandbox := v.GenSubclusterSandboxMap()
 	oldSubclusterMap := oldObj.GenSubclusterMap()
 	oldSandboxMap := oldObj.GenSandboxMap()
-	newSubclusterMap := v.GenSubclusterMap()
 	sclusterSboxMap := make(map[*Subcluster]*Sandbox)
 	for oldSubclusterName, oldSandboxName := range oldSubclusterInSandbox {
 		newSandboxName, oldSubclusterInNewSandboxes := newSuclusterInSandbox[oldSubclusterName]
-		_, oldSubclusterInPersist := newSubclusterMap[oldSubclusterName]
 		oldSandbox := oldSandboxMap[oldSandboxName]
 		// for unsandboxing, check shutdown field of the subcluster and sandbox
-		if oldSubclusterInPersist {
-			if !oldSubclusterInNewSandboxes || (oldSubclusterInNewSandboxes && oldSandbox.Name != newSandboxName) {
-				// either subcluster is not in any sbox or is moved to a different sbox in new vdb
-				oldSubcluster := oldSubclusterMap[oldSubclusterName]
-				sclusterSboxMap[oldSubcluster] = oldSandbox
-			}
+		if !oldSubclusterInNewSandboxes || (oldSubclusterInNewSandboxes && oldSandbox.Name != newSandboxName) {
+			// either subcluster is not in any sbox or is moved to a different sbox in new vdb
+			oldSubcluster := oldSubclusterMap[oldSubclusterName]
+			sclusterSboxMap[oldSubcluster] = oldSandbox
 		}
 	}
 	return sclusterSboxMap
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -1744,7 +1744,7 @@ func (v *VerticaDB) validateNewSBoxOrSClusterShutdownUnset(allErrs field.ErrorLi
 		}
 	}
 	for i := range v.Spec.Subclusters {
-		newSCluster := v.Spec.Subclusters[i]
+		newSCluster := &v.Spec.Subclusters[i]
 		if _, foundInStatus := statusSClusterMap[newSCluster.Name]; !foundInStatus && newSCluster.Shutdown {
 			p := field.NewPath("spec").Child("subclusters").Index(i)
 			err := field.Invalid(p,
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -1816,8 +1816,8 @@ func (v *VerticaDB) checkUnsandboxShutdownConditions(oldObj *VerticaDB, allErrs 
 	return allErrs
 }
 
-// checkAnnotatedSubclustersInShutdownSandbox ensures: when a sandbox is to be shutdown, if a subcluster in it has annotation
-// vertica.com/shutdown-driven-by-sandbox set to true, the subcluster's shutdown field should not be updated
+// checkSubclustersInShutdownSandbox ensures: when a sandbox is marked to be shutdown,
+// the subcluster's shutdown field should not to set to false by a user
 func (v *VerticaDB) checkSubclustersInShutdownSandbox(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {
 	persistSandboxes := v.findPersistSandboxes(oldObj)
 	newSandboxMap := v.GenSandboxMap()
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -108,9 +108,6 @@ func (v *VerticaDB) ValidateCreate() error {
 	allErrs := v.validateVerticaDBSpec()
 	v.hasNoShutdownSubclusters(allErrs)
 	v.hasNoShutdownSandboxes(allErrs)
-	if allErrs == nil {
-		return nil
-	}
 	return apierrors.NewInvalid(schema.GroupKind{Group: Group, Kind: VerticaDBKind}, v.Name, allErrs)
 }
 
@@ -119,13 +116,10 @@ func (v *VerticaDB) ValidateUpdate(old runtime.Object) error {
 	verticadblog.Info("validate update", "name", v.Name, "GroupVersion", GroupVersion)
 
 	allErrs := append(v.validateImmutableFields(old), v.validateVerticaDBSpec()...)
-	v.validateSandboxImage(old, allErrs)
+	v.validateShutdownSandboxImage(old, allErrs)
 	v.validateTerminatingSandboxes(old, allErrs)
 	v.validateAnnotatedSubclustersInShutdownSandbox(old, allErrs)
-	v.checkUnsandboxShutdownConditions(old, allErrs)
-	if allErrs == nil {
-		return nil
-	}
+	v.validateUnsandboxShutdownConditions(old, allErrs)
 	return apierrors.NewInvalid(schema.GroupKind{Group: Group, Kind: VerticaDBKind}, v.Name, allErrs)
 }
 
@@ -155,7 +149,6 @@ func (v *VerticaDB) validateImmutableFields(old runtime.Object) field.ErrorList 
 	allErrs = v.checkImmutableStsName(oldObj, allErrs)
 	allErrs = v.checkValidSubclusterTypeTransition(oldObj, allErrs)
 	allErrs = v.checkSandboxesDuringUpgrade(oldObj, allErrs)
-
 	return allErrs
 }
 
@@ -269,6 +262,7 @@ func (v *VerticaDB) hasValidSubclusterTypes(allErrs field.ErrorList) field.Error
 	return allErrs
 }
 
+// check if a vdb to be created has any subcluster whose shutdown field is true
 func (v *VerticaDB) hasNoShutdownSubclusters(allErrs field.ErrorList) field.ErrorList {
 	for i := range v.Spec.Subclusters {
 		subcluster := &v.Spec.Subclusters[i]
@@ -284,6 +278,7 @@ func (v *VerticaDB) hasNoShutdownSubclusters(allErrs field.ErrorList) field.Erro
 	return allErrs
 }
 
+// check if a vdb to be created has any sandbox whose shutdown field is true
 func (v *VerticaDB) hasNoShutdownSandboxes(allErrs field.ErrorList) field.ErrorList {
 	sandboxes := v.Spec.Sandboxes
 	path := field.NewPath("spec").Child("sandboxes")
@@ -1633,13 +1628,13 @@ func (v *VerticaDB) findPersistScsInSandbox(oldObj *VerticaDB) map[string]int {
 func (v *VerticaDB) findPersistSandboxes(oldObj *VerticaDB) []string {
 	oldSandboxes := oldObj.Spec.Sandboxes
 	newSandboxes := v.Spec.Sandboxes
-	sandboxMap := make(map[string]int)
+	oldSandboxMap := make(map[string]int)
 	for _, oldSandbox := range oldSandboxes {
-		sandboxMap[oldSandbox.Name] = 1
+		oldSandboxMap[oldSandbox.Name] = 1
 	}
 	persistSandboxes := make([]string, 0)
 	for _, newSandbox := range newSandboxes {
-		if _, ok := sandboxMap[newSandbox.Name]; ok {
+		if _, ok := oldSandboxMap[newSandbox.Name]; ok {
 			persistSandboxes = append(persistSandboxes, newSandbox.Name)
 		}
 	}
@@ -1759,21 +1754,18 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
 	return allErrs
 }
 
-// checkUnsandboxShutdownConditions ensures we will not unsandbox a subcluster that is to be shut down, or a subcluster
-// in a sandbox that is to be shut down
-func (v *VerticaDB) checkUnsandboxShutdownConditions(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
+// validateUnsandboxShutdownConditions ensures we will not unsandbox a subcluster that has shutdown set to true, or a subcluster
+// in a sandbox where the shutdown field for the sandbox is set to true
+func (v *VerticaDB) validateUnsandboxShutdownConditions(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
 	oldObj := old.(*VerticaDB)
 	oldSubclusterInSandbox := oldObj.GenSubclusterSandboxMap()
 	newSuclusterInSandbox := v.GenSubclusterSandboxMap()
 	oldSubclusterMap := oldObj.GenSubclusterMap()
 	oldSandboxMap := oldObj.GenSandboxMap()
-	newSubclusterIndexMap := v.GenSubclusterIndexMap()
 	oldSubclusterIndexMap := oldObj.GenSubclusterIndexMap()
 
 	for oldSubclusterName, oldSandboxName := range oldSubclusterInSandbox {
 		_, oldSubclusterInNewSandboxes := newSuclusterInSandbox[oldSubclusterName]
-		i := newSubclusterIndexMap[oldSubclusterName]
-		// an error will be reported in checkImmutableSubclusterInSandbox() when oldSubclusterName is not found
 
 		// for unsandboxing, check shutdown field of the subcluster and sandbox
 		if !oldSubclusterInNewSandboxes {
@@ -1782,18 +1774,17 @@ func (v *VerticaDB) checkUnsandboxShutdownConditions(old runtime.Object, allErrs
 			oldSubclusterIndex := oldSubclusterIndexMap[oldSubclusterName]
 			if oldSubcluster.Shutdown || v.Status.Subclusters[oldSubclusterIndex].Shutdown {
 				p := field.NewPath("spec").Child("subclusters")
-				err := field.Invalid(p.Index(i),
+				err := field.Invalid(p,
 					oldSubclusterName,
-					fmt.Sprintf("cannot unsandbox subcluster %q that is to be shut down",
+					fmt.Sprintf("cannot unsandbox subcluster %q that has Shutdown field set to true",
 						oldSubclusterName))
 				allErrs = append(allErrs, err)
 			}
 			if oldSandbox.Shutdown {
-				i := newSubclusterIndexMap[oldSubclusterName]
-				p := field.NewPath("spec").Child("subclusters")
-				err := field.Invalid(p.Index(i),
+				p := field.NewPath("spec").Child("sandboxes")
+				err := field.Invalid(p,
 					oldSubclusterName,
-					fmt.Sprintf("cannot unsandbox subcluster %q in sandbox %q that is to be shut down",
+					fmt.Sprintf("cannot unsandbox subcluster %q in sandbox %q that has Shutdown field set to true",
 						oldSubclusterName, oldSandboxName))
 				allErrs = append(allErrs, err)
 			}
@@ -1825,7 +1816,7 @@ func (v *VerticaDB) validateAnnotatedSubclustersInShutdownSandbox(old runtime.Ob
 						p := field.NewPath("spec").Child("subclusters")
 						err := field.Invalid(p.Index(index),
 							subclusterName.Name,
-							fmt.Sprintf("cannot change shutdown field for subcluster %q that has annotation \"vertica.com/shutdown-driven-by-sandbox\" set to true",
+							fmt.Sprintf("cannot change Shutdown field for subcluster %q that has annotation \"vertica.com/shutdown-driven-by-sandbox\" set to true",
 								subclusterName.Name))
 						allErrs = append(allErrs, err)
 					}
@@ -1836,7 +1827,7 @@ func (v *VerticaDB) validateAnnotatedSubclustersInShutdownSandbox(old runtime.Ob
 	return allErrs
 }
 
-func (v *VerticaDB) validateSandboxImage(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
+func (v *VerticaDB) validateShutdownSandboxImage(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
 	oldObj := old.(*VerticaDB)
 	persistSandboxes := v.findPersistSandboxes(oldObj)
 	oldSandboxMap := oldObj.GenSandboxMap()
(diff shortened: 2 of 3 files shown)


api/v1/verticadb_webhook.go
```diff
@@ -105,7 +106,9 @@ func (v *VerticaDB) ValidateCreate() error {
 	verticadblog.Info("validate create", "name", v.Name, "GroupVersion", GroupVersion)
 
 	allErrs := v.validateVerticaDBSpec()
-	if allErrs == nil {
+	allErrs = v.hasNoShutdownSubclusters(allErrs)
+	allErrs = v.hasNoShutdownSandboxes(allErrs)
```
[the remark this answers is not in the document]
LiboYu2: I believe the original 1st rule is still a valid one. The two cover different scenarios. Should we add what you proposed as a new rule?

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1760,162 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+	if sandbox.Shutdown {
+		errMsgs = append(errMsgs, fmt.Sprintf("Shutdown field of sandbox %q is set to true", sandbox.Name))
+	}
+	for _, subcluster := range sandbox.Subclusters {
+		if subclusterMap[subcluster.Name].Shutdown {
+			errMsgs = append(errMsgs, fmt.Sprintf("Shutdown field of subcluster %q is set to true", subcluster.Name))
+		}
+	}
+	return errMsgs
+}
```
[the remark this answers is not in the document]
LiboYu2: "make sure the subcluster exists in status"   
I got your point. When a user tries to change the sandbox image, the subcluster is already shutdown at this time. This is different from my original understanding. I will make a change.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1760,162 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
 	return allErrs
 }
 
+// validateUnsandboxShutdownConditions ensures we will not unsandbox a subcluster that has shutdown set to true, or a subcluster
+// in a sandbox where the shutdown field for the sandbox is set to true
+func (v *VerticaDB) validateUnsandboxShutdownConditions(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
```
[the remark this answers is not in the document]
LiboYu2: "check if the subcluster shutdown(spec/status)"
for the condition shutdown(spec/status), is it an OR or AND? Currently I implemented it as an OR.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1760,162 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+						allErrs = append(allErrs, err)
+					}
+				}
+			}
+		}
+	}
+	return allErrs
+}
+
+func (v *VerticaDB) validateShutdownSandboxImage(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
```
cchen-vertica, line 1839: Add a comment for this function.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1760,162 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+					fmt.Sprintf("cannot change the image for sandbox %q because %q",
+						sandboxName, strings.Join(errMsgs, ",")))
+				allErrs = append(allErrs, err)
+			}
+		}
+	}
+	return allErrs
+}
+
+func (v *VerticaDB) validateTerminatingSandboxes(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
```
cchen-vertica, line 1868: Add a comment for this function.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1760,162 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+					fmt.Sprintf("cannot terminate sandbox %q that has Shutdown field set to true",
+						sandboxToBeRemoved))
+				allErrs = append(allErrs, err)
+			} // else looks good
+		} // else is expected
+	}
+	return allErrs
+}
+
+func (v *VerticaDB) checkSboxForShutdownInSpecAndStatus(sandbox *Sandbox, subclusterMap map[string]*Subcluster, subclusterIndexMap map[string]int, errMsgs []string) []string {
```
cchen-vertica, line 1897: Add a comment for this function.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1760,162 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+	errMsgs = v.checkSboxForShutdownInSpec(sandbox, subclusterMap, errMsgs)
+	for _, subcluster := range sandbox.Subclusters {
+		if v.Status.Subclusters[subclusterIndexMap[subcluster.Name]].Shutdown {
+			errMsgs = append(errMsgs, fmt.Sprintf("Shutdown status of subcluster %q is true", subcluster.Name))
+		}
+	}
+	return errMsgs
+}
+
+func (v *VerticaDB) checkSboxForShutdownInSpec(sandbox *Sandbox, subclusterMap map[string]*Subcluster, errMsgs []string) []string {
```
cchen-vertica, line 1907: Add a comment for this function.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1760,162 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+		if oldSandbox.Image != newSandbox.Image {
+			newSandboxIndex := newSandboxIndexMap[sandboxName]
+			errMsgs := make([]string, 0)
+			errMsgs = v.checkSboxForShutdownInSpecAndStatus(newSandbox, newSubclusterMap, oldSubclusterIndexMap, errMsgs)
+			if len(errMsgs) != 0 {
+				p := field.NewPath("spec").Child("sandboxes").Index(newSandboxIndex)
+				err := field.Invalid(p.Child("image"),
+					sandboxName,
+					fmt.Sprintf("cannot change the image for sandbox %q because %q",
+						sandboxName, strings.Join(errMsgs, ",")))
```
cchen-vertica, line 1860: errMsgs should only get the names of the subclusters that are shut down. Then here we can say "cannot change the image for sandbox %q because its subclusters %v are already shut down. "

api/v1/verticadb_webhook.go
cchen-vertica, line 1868: We can simplify this rule to "we cannot remove any subclusters that are shut down". Preventing removing the sandbox should already be covered in validateUnsandboxShutdownConditions.  

We should remove this rule in future since a stopped subcluster/sandbox should be able to removed.

api/v1/verticadb_webhook.go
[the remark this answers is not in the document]
LiboYu2: Got it. It is different from your original comment: "for each subcluster find the shutdown value in both spec **and** status(make sure the subcluster exists in status)."

api/v1/verticadb_webhook.go
cchen-vertica, line 1839: We should have one more rule: if the sandbox is shut down or partial shut down, we cannot add or remove any subclusters in it.

api/v1/verticadb_webhook.go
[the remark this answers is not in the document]
cchen-vertica: We need to add one more limitation: we cannot add one or multiple subclusters to a sandbox if the sandbox is shutdown or partially shutdown. I didn't test it, but I think we cannot restart a sandboxed subcluster if that sandbox changed its group members.

api/v1/verticadb_webhook.go
```diff
@@ -116,7 +119,11 @@ func (v *VerticaDB) ValidateUpdate(old runtime.Object) error {
 	verticadblog.Info("validate update", "name", v.Name, "GroupVersion", GroupVersion)
 
 	allErrs := append(v.validateImmutableFields(old), v.validateVerticaDBSpec()...)
-	if allErrs == nil {
+	allErrs = v.validateShutdownSandboxImage(old, allErrs)
```
cchen-vertica, line 122: The new functions should be put in validateImmutableFields().

api/v1/verticadb_webhook.go
LiboYu2: The remove scenario is covered by unsandbox. We can discuss with Roy about the add scenario.

api/v1/verticadb_webhook.go
LiboYu2: moved. Three unrelated test cases were broken and fixed to pass the newly added tests.

api/v1/verticadb_webhook.go
LiboYu2: FYI:
I did some manual testing. When I tried to add a new subcluster to a shutdown sandbox. Webhook did not complain. The spec was updated but nothing happened. It seems that is not supported and we may need a new rule.

roypaulin: Yes, and this is because we want to keep things simple, the users must first create the subcluster/sandbox before they can shut it down.

LiboYu2: The proposed new rule can be found in another webhook JIRA VER-97254

api/v1/verticadb_webhook_test.go
```diff
@@ -1298,6 +1299,202 @@ var _ = Describe("verticadb_webhook", func() {
...
+			}
+			newVdb := oldVdb.DeepCopy()
+			newVdb.Spec.Subclusters[2].Shutdown = false
+			Ω(newVdb.validateAnnotatedSubclustersInShutdownSandbox(oldVdb, field.ErrorList{})).Should(HaveLen(1))
+			newVdb.Spec.Subclusters[2].Shutdown = true
+			Ω(newVdb.validateAnnotatedSubclustersInShutdownSandbox(oldVdb, field.ErrorList{})).Should(HaveLen(0))
+			newVdb.Spec.Subclusters[1].Shutdown = false
+			Ω(newVdb.validateAnnotatedSubclustersInShutdownSandbox(oldVdb, field.ErrorList{})).Should(HaveLen(1))
+			newVdb.Spec.Subclusters[1].Shutdown = true
+			Ω(newVdb.validateAnnotatedSubclustersInShutdownSandbox(oldVdb, field.ErrorList{})).Should(HaveLen(0))
```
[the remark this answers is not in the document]
LiboYu2: new cases added

api/v1/verticadb_webhook.go
LiboYu2: This piece of code is already refactored and strings.join(0 is removed to address Roy's concern that there are too many err messages.

api/v1/verticadb_webhook.go
LiboYu2: This is from the JIRA requirement: "terminate means removing the sandbox in spec.sandboxes[] and its subclusters in spec.subclusters[] at the same time."
There is another rule that does not allow unsandboxing a subcluster that is shut down. The two scenarios require different handling logics.

api/v1/verticadb_webhook.go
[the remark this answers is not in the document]
LiboYu2: The proposed new rule can be found in another webhook JIRA VER-97254

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,204 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+	}
+	return errMsgs
+} */
+
+func (v *VerticaDB) checkSboxForShutdown(sandbox *Sandbox, newSClusterMap map[string]*Subcluster, oldSClusterIndexMap map[string]int) string {
+	if sandbox.Shutdown {
+		return fmt.Sprintf("Shutdown field of sandbox %q is set to true", sandbox.Name)
+	}
+	for _, subcluster := range sandbox.Subclusters {
+		i, foundInOldSBox := oldSClusterIndexMap[subcluster.Name]
```
roypaulin, line 1921: Why are you getting the index from old subcluster to use it in new? 
Maybe add a function to get a map[string]int to get the subcluster index in status. Most of the time the index is spec is the same in status but there are some exceptions we do not want to deal with. Anyway, I think the index should be from new vdb. We don't care if the subcluster was in old vdb sandbox

roypaulin: Let me know if you have objections.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,204 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+		}
+	}
+	for i := range v.Spec.Subclusters {
+		newSCluster := v.Spec.Subclusters[i]
+		if _, foundInStatus := statusSClusterMap[newSCluster.Name]; !foundInStatus && newSCluster.Shutdown {
+			p := field.NewPath("spec").Child("subclusters")
+			err := field.Invalid(p.Index(i),
+				newSCluster.Name,
+				fmt.Sprintf("cannot add subcluster %q that has Shutdown field set to true",
+					newSCluster.Name))
```
roypaulin, line 1753:
```suggestion
			p := field.NewPath("spec").Child("subclusters").Index(i).Shutdown
			err := field.Invalid(p,
				newSCluster.Shutdown,
				fmt.Sprintf("shutdown must be false when adding subcluster %q",
					newSCluster.Name))
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,204 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+	statusSClusterMap := v.GenStatusSubclusterMap()
+	statusSBoxMap := v.GenStatusSandboxMap()
+	for i := range v.Spec.Sandboxes {
+		newSBox := v.Spec.Sandboxes[i]
+		if _, foundInStatus := statusSBoxMap[newSBox.Name]; !foundInStatus && newSBox.Shutdown {
+			p := field.NewPath("spec").Child("sandboxes")
+			err := field.Invalid(p.Index(i),
+				newSBox.Name,
+				fmt.Sprintf("cannot add sandbox %q that has Shutdown field set to true",
+					newSBox.Name))
```
roypaulin, line 1742:
```suggestion
			p := field.NewPath("spec").Child("sandboxes").Index(i).Shutdown
			err := field.Invalid(p,
				newSBox.Shutdown,
				fmt.Sprintf("shutdown must be false when adding sandbox %q",
					newSBox.Name))
```

api/v1/verticadb_webhook.go
LiboYu2: That is a good point. I will make the change.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+			} // else looks good
+		} // else is expected
+	}
+	return allErrs
+}
+
+// checkSboxForShutdown checks if sbox or its scluster has shutdown field set to true in spec/status
+func (v *VerticaDB) checkSboxForShutdown(newSandbox *Sandbox, newSClusterMap map[string]*Subcluster, statusSClusterIndexMap map[string]int) string {
+	if newSandbox.Shutdown {
+		return fmt.Sprintf("Shutdown field of sandbox %q is set to true", newSandbox.Name)
```
roypaulin, line 1907:
```suggestion
		errMsgs = append(errMsgs, fmt.Sprintf("Shutdown field of sandbox %q is set to true", newSandbox.Name))
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+	}
+
+	for _, subcluster := range newSandbox.Subclusters {
+		i, foundInStatus := statusSClusterIndexMap[subcluster.Name]
+		if foundInStatus && (newSClusterMap[subcluster.Name].Shutdown || v.Status.Subclusters[i].Shutdown) {
+			return fmt.Sprintf("Subcluster %q is marked for shut down or has already been shut down", subcluster.Name)
```
roypaulin, line 1913:
```suggestion
			errMsgs = append(errMsgs, fmt.Sprintf("Subcluster %q is marked for shut down or has already been shut down", subcluster.Name))
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+		}
+	}
+	return ""
```
roypaulin, line 1916:
```suggestion
	return errMsgs
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+						sandboxToBeRemoved))
+				allErrs = append(allErrs, err)
+			} // else looks good
+		} // else is expected
+	}
+	return allErrs
+}
+
+// checkSboxForShutdown checks if sbox or its scluster has shutdown field set to true in spec/status
+func (v *VerticaDB) checkSboxForShutdown(newSandbox *Sandbox, newSClusterMap map[string]*Subcluster, statusSClusterIndexMap map[string]int) string {
```
roypaulin, line 1905:
```suggestion
func (v *VerticaDB) checkSboxForShutdown(newSandbox *Sandbox, newSClusterMap map[string]*Subcluster, statusSClusterIndexMap map[string]int) []string {
       errMsgs := []string{}
```

We still want to return all the error msgs. We just did not want to return one batch for spec and another for status.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+	for _, sandboxName := range persistSandboxes {
+		oldSandbox := oldSandboxMap[sandboxName]
+		newSandbox := newSandboxMap[sandboxName]
+		if oldSandbox.Image != newSandbox.Image {
+			newSandboxIndex := newSandboxIndexMap[sandboxName]
+			errMsg := v.checkSboxForShutdown(newSandbox, newSubclusterMap, statusSClusterIndexMap)
+			if errMsg != "" {
+				p := field.NewPath("spec").Child("sandboxes").Index(newSandboxIndex)
+				err := field.Invalid(p.Child("image"),
+					sandboxName,
```
roypaulin, line 1864:
```suggestion
					newSandbox.Image,
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+		}
+	}
+	return sclusterSboxMap
+}
+
+// validateUnsandboxShutdownConditions ensures we will not unsandbox a subcluster that has shutdown set to true, or a subcluster
+// in a sandbox where the shutdown field for the sandbox is set to true
+func (v *VerticaDB) validateUnsandboxShutdownConditions(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
+	statusSClusterIndexMap := v.GenStatusSClusterIndexMap()
+	sclustersToUnsandbox := v.findSclustersToUnsandbox(old)
```
roypaulin, line 1787:
```suggestion
	subclustersToUnsandbox := v.findSclustersToUnsandbox(old)
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+			allErrs = append(allErrs, err)
+		}
+	}
+	return allErrs
+}
+
+// return a map whose key is the pointer of a subcluster that is to be unsandboxed. The value is the pointer to a sandbox
+// where the subcluster lives in
+func (v *VerticaDB) findSclustersToUnsandbox(old runtime.Object) map[*Subcluster]*Sandbox {
+	oldObj := old.(*VerticaDB)
```
roypaulin, line 1763: Why not just past this as argument?

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+	return sclusterSboxMap
+}
+
+// validateUnsandboxShutdownConditions ensures we will not unsandbox a subcluster that has shutdown set to true, or a subcluster
+// in a sandbox where the shutdown field for the sandbox is set to true
+func (v *VerticaDB) validateUnsandboxShutdownConditions(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
+	statusSClusterIndexMap := v.GenStatusSClusterIndexMap()
+	sclustersToUnsandbox := v.findSclustersToUnsandbox(old)
+	for oldSubcluster, oldSandbox := range sclustersToUnsandbox {
+		oldSubclusterIndex := statusSClusterIndexMap[oldSubcluster.Name]
```
roypaulin, line 1789: Always make sure the subcluster is in the status. If it is not(very unlikely), skip it.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+		if oldSubcluster.Shutdown || v.Status.Subclusters[oldSubclusterIndex].Shutdown {
+			p := field.NewPath("spec").Child("subclusters")
```
roypaulin, line 1791:
```suggestion
			p := field.NewPath("spec").Child("subclusters").Index(...)
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+			err := field.Invalid(p,
+				oldSubcluster.Name,
```
roypaulin, line 1793:
```suggestion
			err := field.Invalid(p.Child("shutdown"),
				oldSubcluster.Shutdown,
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+		}
+		if oldSandbox.Shutdown {
+			p := field.NewPath("spec").Child("sandboxes")
+			err := field.Invalid(p,
+				oldSubcluster.Name,
+				fmt.Sprintf("cannot unsandbox subcluster %q in sandbox %q that has Shutdown field set to true",
+					oldSubcluster.Name, oldSandbox.Name))
+			allErrs = append(allErrs, err)
+			continue
+		}
```
roypaulin, line 1807:
```suggestion
			allErrs = append(allErrs, err)
		}
		if oldSandbox.Shutdown {
			p := field.NewPath("spec").Child("sandboxes").Index(...)
			err := field.Invalid(p.Child("shutdown"),
				oldSandbox.Name,
				fmt.Sprintf("cannot unsandbox subcluster %q in sandbox %q that has shutdown field set to true",
					oldSubcluster.Name, oldSandbox.Name))
			allErrs = append(allErrs, err)
		}
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+		if newSandbox.Shutdown {
+			for _, subclusterName := range oldSandbox.Subclusters {
+				oldSubcluster := oldSubclusterMap[subclusterName.Name]
+				if vmeta.GetShutdownDrivenBySandbox(oldSubcluster.Annotations) {
+					newSubcluster, oldSclusterPersist := newSubclusterMap[subclusterName.Name]
+					if oldSclusterPersist && oldSubcluster.Shutdown != newSubcluster.Shutdown {
+						index := newSubclusterIndexMap[subclusterName.Name]
+						p := field.NewPath("spec").Child("subclusters")
+						err := field.Invalid(p.Index(index),
+							subclusterName.Name,
```
roypaulin, line 1834:
```suggestion
						err := field.Invalid(p.Index(index).Child("shutdown"),
							newSubcluster.Shutdown,
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+							fmt.Sprintf("cannot change Shutdown field for subcluster %q that has annotation \"vertica.com/shutdown-driven-by-sandbox\" set to true",
```
roypaulin, line 1835:
```suggestion
							fmt.Sprintf("cannot change shutdown field for subcluster %q whose shutdown is driven by sandbox %q",
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
 
+// ensures a new subcluster or sandbox to be added to a vdb has Shutdown field set to false
+func (v *VerticaDB) validateNewSBoxOrSClusterShutdownUnset(allErrs field.ErrorList) field.ErrorList {
+	statusSClusterMap := v.GenStatusSubclusterMap()
+	statusSBoxMap := v.GenStatusSandboxMap()
+	for i := range v.Spec.Sandboxes {
+		newSBox := v.Spec.Sandboxes[i]
+		if _, foundInStatus := statusSBoxMap[newSBox.Name]; !foundInStatus && newSBox.Shutdown {
+			p := field.NewPath("spec").Child("sandboxes").Index(i)
+			err := field.Invalid(p,
```
roypaulin, line 1739:
```suggestion
			err := field.Invalid(p.Child("shutdown"),
```

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,194 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+				fmt.Sprintf("shutdown must be false when adding sandbox %q",
+					newSBox.Name))
+			allErrs = append(allErrs, err)
+		}
+	}
+	for i := range v.Spec.Subclusters {
+		newSCluster := &v.Spec.Subclusters[i]
+		if _, foundInStatus := statusSClusterMap[newSCluster.Name]; !foundInStatus && newSCluster.Shutdown {
+			p := field.NewPath("spec").Child("subclusters").Index(i)
+			err := field.Invalid(p,
```
roypaulin, line 1750:
```suggestion
			err := field.Invalid(p.child("shutdown"),
```

roypaulin (commented):
We are getting close.

api/v1/verticadb_webhook.go
LiboYu2: checkSboxForShutdown() is used within a loop inside  validateShutdownSandboxImage()
 We use the returned string return value to check if there is an error. If string is empty, that means no error. Otherwise, there is an error.

api/v1/verticadb_webhook.go
LiboYu2: same comment as the above one.

api/v1/verticadb_webhook.go
LiboYu2: same comment as the above one. When it returns empty string, it means no error.

api/v1/verticadb_webhook.go
LiboYu2: Now I know what you want to achieve. I withdraw my previous comments and will make the change.

api/v1/verticadb_webhook.go
roypaulin: Yes, but it can happen that the sandbox has shutdown true and also all its subclusters have shutdown true. In that case we want to report all the error msgs so the user is aware.

api/v1/verticadb_webhook.go
LiboYu2: Then the casting will be done in the caller of this function. It makes no difference.

api/v1/verticadb_webhook.go
```diff
@@ -1902,18 +1906,19 @@ func (v *VerticaDB) validateTerminatingSandboxes(old runtime.Object, allErrs fie
 }
 
 // checkSboxForShutdown checks if sbox or its scluster has shutdown field set to true in spec/status
-func (v *VerticaDB) checkSboxForShutdown(newSandbox *Sandbox, newSClusterMap map[string]*Subcluster, statusSClusterIndexMap map[string]int) string {
+func (v *VerticaDB) checkSboxForShutdown(newSandbox *Sandbox, newSClusterMap map[string]*Subcluster, statusSClusterIndexMap map[string]int) []string {
+	errMsgs := []string{}
 	if newSandbox.Shutdown {
-		return fmt.Sprintf("Shutdown field of sandbox %q is set to true", newSandbox.Name)
+		return append(errMsgs, fmt.Sprintf("Shutdown field of sandbox %q is set to true", newSandbox.Name))
```
roypaulin, line 1912: Do not return.

api/v1/verticadb_webhook.go
```diff
@@ -1902,18 +1906,19 @@ func (v *VerticaDB) validateTerminatingSandboxes(old runtime.Object, allErrs fie
...
 	}
 
 	for _, subcluster := range newSandbox.Subclusters {
 		i, foundInStatus := statusSClusterIndexMap[subcluster.Name]
 		if foundInStatus && (newSClusterMap[subcluster.Name].Shutdown || v.Status.Subclusters[i].Shutdown) {
-			return fmt.Sprintf("Subcluster %q is marked for shut down or has already been shut down", subcluster.Name)
+			return append(errMsgs, fmt.Sprintf("Subcluster %q is marked for shut down or has already been shut down", subcluster.Name))
```
roypaulin, line 1918: Do not return.
Also to make the error msg shorter, we can first append the subcluster names and then add the msg. The err msg will look like: ......Subclusters sc1, sc2, sc3 are marked for shut down or have already been shut down.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,199 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+		numOfSclustersInSbox := len(oldSandboxMap[sandboxToBeRemoved].Subclusters)
+		sclustersToTerminate := numOfSclustersInSbox
+		for _, subcluster := range oldSandboxMap[sandboxToBeRemoved].Subclusters {
+			if _, ok := newSubclusterMap[subcluster.Name]; ok {
+				sclustersToTerminate--
+			}
+		}
+		// sandboxToBeRemoved and all its subclusters are not found in new spec
+		if sclustersToTerminate == numOfSclustersInSbox {
+			if oldSandboxMap[sandboxToBeRemoved].Shutdown {
```
roypaulin, line 1895: This check should be at the beginning of the loop so you can skip to the next sandbox if the current sandbox has shutdown true.

api/v1/verticadb_webhook.go
```diff
@@ -148,6 +150,11 @@ func (v *VerticaDB) validateImmutableFields(old runtime.Object) field.ErrorList
 	allErrs = v.checkImmutableStsName(oldObj, allErrs)
 	allErrs = v.checkValidSubclusterTypeTransition(oldObj, allErrs)
 	allErrs = v.checkSandboxesDuringUpgrade(oldObj, allErrs)
+	allErrs = v.validateShutdownSandboxImage(old, allErrs)
+	allErrs = v.validateTerminatingSandboxes(old, allErrs)
+	allErrs = v.validateUnsandboxShutdownConditions(old, allErrs)
+	allErrs = v.validateAnnotatedSubclustersInShutdownSandbox(old, allErrs)
+	allErrs = v.validateNewSBoxOrSClusterShutdownUnset(allErrs)
```
roypaulin, line 157: Let's keep the same function structure and call these `check...` like `checkShutdownSandboxImage`

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,199 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
 	return allErrs
 }
 
+// ensures a new subcluster or sandbox to be added to a vdb has Shutdown field set to false
```
roypaulin, line 1731:
```suggestion
// validateNewSBoxOrSClusterShutdownUnset ensures a new subcluster or sandbox to be added to a vdb has Shutdown field set to false
```
Follow the same forma for other functions

roypaulin (commented):
This is the last round. After my comments are addressed, I will approve

api/v1/verticadb_webhook.go
LiboYu2: This if statement compares the number of subclusters to be removed with the number of subclusters in the sandbox.
If they equal, that means the sandbox will be terminated.
There is another for loop on the top of this if statement. It is used to calculate "**sclustersToTerminate**" used in the if statement.

api/v1/helpers.go
```diff
@@ -206,6 +206,36 @@ func (v *VerticaDB) GenSubclusterSandboxStatusMap() map[string]string {
...
+func (v *VerticaDB) GenStatusSandboxMap() map[string]*SandboxStatus {
+	statusSboxMap := make(map[string]*SandboxStatus)
+	for i := range v.Status.Sandboxes {
+		sBox := &v.Status.Sandboxes[i]
+		statusSboxMap[sBox.Name] = sBox
+	}
+	return statusSboxMap
+}
+
+// GenStatusSandboxMap() returns a map from status. The key is subcluster name and value is the subcluster pointer
```
cchen-vertica, line 219: Minor: update GenStatusSandboxMap to GenStatusSubclusterMap

api/v1/verticadb_webhook.go
```diff
@@ -1591,6 +1598,23 @@ func (v *VerticaDB) findPersistScsInSandbox(oldObj *VerticaDB) map[string]int {
 	return persistScsWithSbIndex
 }
 
+// findPersistSandboxes returns a slice of sandbox names that are in both new and old vdb.
+func (v *VerticaDB) findPersistSandboxes(oldObj *VerticaDB) []string {
+	oldSandboxes := oldObj.Spec.Sandboxes
+	newSandboxes := v.Spec.Sandboxes
+	oldSandboxMap := make(map[string]int)
```
cchen-vertica, line 1605: Minor: this map should be a map[string]any or map[string]bool.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,202 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+				newSCluster.Shutdown,
+				fmt.Sprintf("shutdown must be false when adding subcluster %q",
+					newSCluster.Name))
+			allErrs = append(allErrs, err)
+		}
+	}
+	return allErrs
+}
+
+// findSclustersToUnsandboxreturn a map whose key is the pointer of a subcluster that is to be unsandboxed.
```
cchen-vertica, line 1760: add space in front of "return"

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,202 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+func (v *VerticaDB) findSclustersToUnsandbox(old runtime.Object) map[*Subcluster]*Sandbox {
+	oldObj := old.(*VerticaDB)
+	oldSubclusterInSandbox := oldObj.GenSubclusterSandboxMap()
+	newSuclusterInSandbox := v.GenSubclusterSandboxMap()
+	oldSubclusterMap := oldObj.GenSubclusterMap()
+	oldSandboxMap := oldObj.GenSandboxMap()
+	newSubclusterMap := v.GenSubclusterMap()
+	sclusterSboxMap := make(map[*Subcluster]*Sandbox)
+	for oldSubclusterName, oldSandboxName := range oldSubclusterInSandbox {
+		_, oldSubclusterInNewSandboxes := newSuclusterInSandbox[oldSubclusterName]
```
cchen-vertica, line 1771: You need to check if the old sandbox name and new sandbox name are the same. If we move a subcluster from a sandbox to another in new vdb, that subcluster is still in new vdb, but it will be unsandboxed.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,202 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+// checkUnsandboxShutdownConditions ensures we will not unsandbox a subcluster that has shutdown set to true, or a subcluster
+// in a sandbox where the shutdown field for the sandbox is set to true
+func (v *VerticaDB) checkUnsandboxShutdownConditions(old runtime.Object, allErrs field.ErrorList) field.ErrorList {
+	statusSClusterIndexMap := v.GenStatusSClusterIndexMap()
+	subclustersToUnsandbox := v.findSclustersToUnsandbox(old)
+	subclusterIndexMap := v.GenSubclusterIndexMap()
+	sandboxIndexMap := v.GenSandboxIndexMap()
+	for oldSubcluster, oldSandbox := range subclustersToUnsandbox {
+		oldSubclusterIndex, found := statusSClusterIndexMap[oldSubcluster.Name]
+		if oldSubcluster.Shutdown || found && v.Status.Subclusters[oldSubclusterIndex].Shutdown {
```
cchen-vertica, line 1792: Add parenthesis for readability.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,202 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+			i := subclusterIndexMap[oldSubcluster.Name]
+			p := field.NewPath("spec").Child("subclusters").Index(i)
+			err := field.Invalid(p,
+				oldSubcluster.Shutdown,
+				fmt.Sprintf("cannot unsandbox subcluster %q that has Shutdown field set to true",
+					oldSubcluster.Name))
+			allErrs = append(allErrs, err)
+			continue
+		}
+		if oldSandbox.Shutdown {
```
cchen-vertica, line 1802: For one sandbox, we probably reach here multiple times so we probably return duplicate errors.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,202 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+	newSandboxMap := v.GenSandboxMap()
+	oldSandboxMap := oldObj.GenSandboxMap()
+	newSubclusterIndexMap := v.GenSubclusterIndexMap()
+	newSubclusterMap := v.GenSubclusterMap()
+	oldSubclusterMap := oldObj.GenSubclusterMap()
+	for _, sandboxName := range persistSandboxes {
+		newSandbox := newSandboxMap[sandboxName]
+		oldSandbox := oldSandboxMap[sandboxName]
+		if newSandbox.Shutdown {
+			for _, subclusterName := range oldSandbox.Subclusters {
```
cchen-vertica, line 1830: Why are we checking old sandbox's subclusters? Doesn't this rule apply on the subclusters in new sandbox?

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,202 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+					fmt.Sprintf("cannot terminate sandbox %q that has Shutdown field set to true",
+						sandboxToBeRemoved))
+				allErrs = append(allErrs, err)
+			} // else looks good
+		} // else is expected
+	}
+	return allErrs
+}
+
+// checkSboxForShutdown checks if sbox or its scluster has shutdown field set to true in spec/status
```
cchen-vertica, line 1908: Enrich the comment to explain the returned value.

api/v1/verticadb_webhook.go
LiboYu2: This is from the JIRA: " If a sandbox has `shutdown`(spec) true and any of its subclusters has the annotation "vertica.com/shutdown-driven-by-sandbox" set to true in the old vdb, we should prevent the user from updating that subcluster's shutdown field. "
Here we check the subclusters that exist in both new and old vdbs.

api/v1/verticadb_webhook.go
LiboYu2: I will make code change to cover that scenario and set up a new test case.

api/v1/verticadb_webhook.go
roypaulin: Again, if `oldSandboxMap[sandboxToBeRemoved].Shutdown` is false then there is no point in running all this code. Check `oldSandboxMap[sandboxToBeRemoved].Shutdown` at the beginning of the loop and if false skip and go to next sandbox. Something like:
```
for _, sandboxToBeRemoved := range sandboxesToBeRemoved {
       if !oldSandboxMap[sandboxToBeRemoved].Shutdown {
              continue
       }
```

LiboYu2: That is a good point. I will check oldSandboxMap[sandboxToBeRemoved].Shutdown earlier. If it is false, let it skip the rest of the current loop.

api/v1/verticadb_webhook.go
```diff
@@ -1701,6 +1728,211 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+					fmt.Sprintf("cannot change the image for sandbox %q because %q",
+						sandboxName, strings.Join(errMsgs, ",")))
+				allErrs = append(allErrs, err)
+			}
+		}
+	}
+	return allErrs
+}
+
+// checkTerminatingSandboxes ensure the sandbox to terminate is not shut down
```
cchen-vertica, line 1885: This comment doesn't match the function implementation. I don't think we need to check if all subclusters inside the sandbox are terminating. If a "shutdown" sandbox is removed, we should error out.

api/v1/verticadb_webhook.go
roypaulin: No we should check newSandbox. Here is the scenario this covers: sandbox sand1 has a subcluster sc1. you set sand1.shutdown to true, the operator will set sc1.shutdown to true and add the annotation to sc1. So now sand.shutdown = true, sc1.shutdown. true and sc1 has "vertica.com/shutdown-driven-by-sandbox". We do not want a user to come and change sc1.shutdown because that annotation is a proof that the sandbox drives the shutdown.

This is relevant only if sand1.shutdown is still true(newvdb)

api/v1/verticadb_webhook.go
```diff
@@ -148,6 +150,11 @@ func (v *VerticaDB) validateImmutableFields(old runtime.Object) field.ErrorList
 	allErrs = v.checkImmutableStsName(oldObj, allErrs)
 	allErrs = v.checkValidSubclusterTypeTransition(oldObj, allErrs)
 	allErrs = v.checkSandboxesDuringUpgrade(oldObj, allErrs)
+	allErrs = v.checkShutdownSandboxImage(old, allErrs)
+	allErrs = v.checkTerminatingSandboxes(old, allErrs)
+	allErrs = v.checkUnsandboxShutdownConditions(old, allErrs)
+	allErrs = v.checkAnnotatedSubclustersInShutdownSandbox(old, allErrs)
+	allErrs = v.checkNewSBoxOrSClusterShutdownUnset(allErrs)
```
roypaulin, line 157: Directly pass oldObj to these functions to stay consistent.

api/v1/verticadb_webhook.go
LiboYu2: I will change the code to adopt the new logic. Could you update the rule in the JIRA as well?

api/v1/verticadb_webhook.go
```diff
@@ -1828,20 +1828,18 @@ func (v *VerticaDB) checkAnnotatedSubclustersInShutdownSandbox(oldObj *VerticaDB
 	for sandboxName := range persistSandboxes {
 		newSandbox := newSandboxMap[sandboxName]
 		oldSandbox := oldSandboxMap[sandboxName]
-		if newSandbox.Shutdown {
+		if newSandbox.Shutdown { // !oldSandbox.Shutdown &&
```
roypaulin, line 1831: Why this comment? I do not think oldSandbox.shutdown value matters.

LiboYu2: This is useless. I will remove it.

api/v1/verticadb_webhook.go
```diff
@@ -1833,20 +1828,18 @@ func (v *VerticaDB) checkAnnotatedSubclustersInShutdownSandbox(old runtime.Objec
...
```
cchen-vertica, line 1831: Remove the comment in this line.

api/v1/verticadb_webhook.go
```diff
@@ -1822,8 +1818,7 @@ func (v *VerticaDB) checkUnsandboxShutdownConditions(old runtime.Object, allErrs
 
 // checkAnnotatedSubclustersInShutdownSandbox ensures: when a sandbox is to be shutdown, if a subcluster in it has annotation
```
cchen-vertica, line 1819: Update the comment here

api/v1/verticadb_webhook.go
```diff
@@ -1828,20 +1828,18 @@ func (v *VerticaDB) checkAnnotatedSubclustersInShutdownSandbox(oldObj *VerticaDB
...
-						index := newSubclusterIndexMap[subclusterName.Name]
-						p := field.NewPath("spec").Child("subclusters")
-						err := field.Invalid(p.Index(index).Child("shutdown"),
-							newSubcluster.Shutdown,
-							fmt.Sprintf("cannot change Shutdown field for subcluster %q whose shutdown is driven by sandbox %q",
-								subclusterName.Name, oldSandbox.Name))
-						allErrs = append(allErrs, err)
-					}
+				newSubcluster, oldSclusterPersist := newSubclusterMap[subclusterName.Name]
+				if oldSclusterPersist && oldSubcluster.Shutdown != newSubcluster.Shutdown && !newSubcluster.Shutdown {
```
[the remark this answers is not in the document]
LiboYu2: Do you mean this one "!newSubcluster.Shutdown"?  It is used to check if a user sets the shutdown to false while the sandbox is marked for shutdown. It is needed.

api/v1/verticadb_webhook.go
```diff
@@ -1833,20 +1828,18 @@ func (v *VerticaDB) checkAnnotatedSubclustersInShutdownSandbox(old runtime.Objec
 	for sandboxName := range persistSandboxes {
```
cchen-vertica, line 1828: Why do we need the sandbox to be persistent?

api/v1/verticadb_webhook.go
```diff
@@ -1882,33 +1874,22 @@ func (v *VerticaDB) checkShutdownSandboxImage(old runtime.Object, allErrs field.
...
-		numOfSclustersInSbox := len(oldSandboxMap[sandboxToBeRemoved].Subclusters)
-		sclustersToTerminate := numOfSclustersInSbox
-		for _, subcluster := range oldSandboxMap[sandboxToBeRemoved].Subclusters {
-			if _, ok := newSubclusterMap[subcluster.Name]; ok {
-				sclustersToTerminate--
-			}
-		}
-		// sandboxToBeRemoved and all its subclusters are not found in new spec
-		if sclustersToTerminate == numOfSclustersInSbox {
+		} else {
```
cchen-vertica, line 1885: You can remove the else here since you already used a "continue" in line 1884

api/v1/verticadb_webhook.go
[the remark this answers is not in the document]
roypaulin: We want to prevent any kind of change so "oldSubcluster.Shutdown != newSubcluster.Shutdown" is enough

api/v1/verticadb_webhook.go
LiboYu2: removed

api/v1/verticadb_webhook.go
LiboYu2: updated.

api/v1/verticadb_webhook.go
LiboYu2: Because of this condition "oldSubcluster.Shutdown != newSubcluster.Shutdown"
The rule applies when a user tries to mess with an existing subcluster. If it is a new subcluster, it will be covered by other rules.

api/v1/verticadb_webhook.go
LiboYu2: with the "else", it is easier to read and understand the logic.

api/v1/verticadb_webhook.go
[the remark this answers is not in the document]
LiboYu2: If we only have "oldSubcluster.Shutdown != newSubcluster.Shutdown", that will stop the operator from the setting the shutdown field to true from false.

api/v1/verticadb_webhook.go
cchen-vertica: That condition is for persistent subclusters, not for persistent sandboxes.

LiboYu2: If the sandbox is not persistent, it will be a new sandbox, For new sandboxes to be created, its shutdown field will not be true. We are only concerned about subclusters in persistent sandboxes.

api/v1/verticadb_webhook.go
cchen-vertica: Then line 1884 doesn't make sense. You can just use `if oldSandboxMap[sandboxToBeRemoved].Shutdown`, what's the point to use an else block.

api/v1/verticadb_webhook.go
cchen-vertica: If we have a subcluster that isn't in a sandbox of old vdb, but we put it in the sandbox of new vdb, your code cannot cover that subcluster, right?

cchen-vertica: Do we have a rule to prevent a "shutdown" subcluster to add to a sandbox in your PR?

LiboYu2: Correct myself. Rule 1 of https://jira.verticacorp.com/jira/browse/VER-97368 covers that scenario.

LiboYu2: > If we have a subcluster that isn't in a sandbox of old vdb, but we put it in the sandbox of new vdb, your code cannot cover that subcluster, right?

Right. That is the sandboxing scenario. It will be covered in https://jira.verticacorp.com/jira/browse/VER-97254

cchen-vertica (approved):

PR #975, opened 2024-11-04
Fixed istio tests

cchen-vertica:
This PR enhances the stability of the Istio tests by upgrading the istioctl version and refining the create_db process in vclusterOps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -715,7 +715,7 @@ $(OPERATOR_SDK):
 	chmod +x $(OPERATOR_SDK)
 
 ISTIOCTL = $(shell pwd)/bin/istioctl
-ISTIOCTL_VERSION = 1.17.2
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -715,7 +715,7 @@ $(OPERATOR_SDK):
 	chmod +x $(OPERATOR_SDK)
 
 ISTIOCTL = $(shell pwd)/bin/istioctl
-ISTIOCTL_VERSION = 1.17.2
+ISTIOCTL_VERSION = 1.23.3
```
HaoYang0000, line 718: What's the difference between these two version?

cchen-vertica: The new version fixed some iptables issues I have in my local environment. I believe the new version will be more stable.

roypaulin (approved):

PR #978, opened 2024-11-09
Added a new helm parameter controllers.vdbMaxBackoffDuration

cchen-vertica:
This PR introduces a new Helm parameter, controllers.vdbMaxBackoffDuration, to address the requeue-hang issue in revive_db. This parameter allows us to set a maximum requeue duration for the VDB controller, providing more control over the requeue behavior.


pkg/opcfg/config.go
```diff
@@ -375,3 +375,13 @@ func getStackTrace() zap.Option {
 	}
 	return zap.AddStacktrace(zapcore.LevelEnabler(lvl))
 }
+
+// GetVdbMaxBackoffDuration returns maximum backoff requeue duration in milliseconds for vdb controller
+func GetVdbMaxBackoffDuration() int {
+	duration := lookupIntEnvVar("VDB_MAX_BACKOFF_DURATION", envCanNotExist)
+	if duration == 0 {
```
HaoYang0000, line 382: If we set the value VDB_MAX_BACKOFF_DURATION=0, should we expect 0ms duration?

cchen-vertica: Yes, we do expect 0ms duration. I will change this one to "<0".

pkg/opcfg/config.go
```diff
@@ -378,10 +378,15 @@ func getStackTrace() zap.Option {
 
 // GetVdbMaxBackoffDuration returns maximum backoff requeue duration in milliseconds for vdb controller
 func GetVdbMaxBackoffDuration() int {
-	duration := lookupIntEnvVar("VDB_MAX_BACKOFF_DURATION", envCanNotExist)
-	if duration == 0 {
+	const envName = "VDB_MAX_BACKOFF_DURATION"
+	duration := lookupStringEnvVar(envName, envCanNotExist)
```
roypaulin, line 382: Why not use `lookupIntEnvVar`? If the issue is the default value, maybe we can change this to `lookupStringEnvVar(envName string, mustExist bool, defaultValue int)`?

cchen-vertica: `lookupIntEnvVar` returns 0 when it cannot find the env var which cannot differentiate the case with user input 0. I don't want to update `lookupIntEnvVar` too much since it's called in many other places.

pkg/opcfg/config.go
```diff
@@ -375,3 +375,18 @@ func getStackTrace() zap.Option {
...
+func GetVdbMaxBackoffDuration() int {
+	const envName = "VDB_MAX_BACKOFF_DURATION"
+	duration := lookupStringEnvVar(envName, envCanNotExist)
+	if duration == "" {
+		// Default value is 1000ms
+		return 1000
+	}
+	valInt, err := strconv.Atoi(duration)
+	if err != nil || valInt < 0 {
+		dieIfNotValid(envName)
```
roypaulin, line 389: You should return the default value in this case too. see lookupIntEnvVar. Also use a constant for the default value.

cchen-vertica: `lookupIntEnvVar` doesn't return any value when err != nil (dieIfNotValid() will exit the process).  I think if the user provided the wrong value, the operator should exit and tell them their value isn't acceptable. We shouldn't replace their wrong value with a default value.

If the default value occurred more than once in the code, I will use a constant value, but it didn't so I just use a literal constant.

roypaulin (approved):

PR #979, opened 2024-11-12
Ver 97254 shutdown webhook

LiboYu2:
Here are the rules implemented in this PR:

- you cannot sandbox a subcluster with shutdown true(spec or status).
- you cannot add a subcluster to a sandbox that has the shutdown field set to true
- you cannot add a subcluster to a sandbox that already has at least one subcluster with shutdown true
- you cannot scale (up/down) a subcluster with shutdown true(spec or status)
- You cannot remove a subcluster with shutdown true(spec or status)

### DIFF
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -1795,7 +1795,10 @@ func (v *VerticaDB) checkShutdownForScaleUpOrDown(old runtime.Object, allErrs fi
 	return allErrs
 }
 
-// checkSClusterToBeSandboxedShutdownUnset ensures a subcluster to be sandboxed has Shutdown field set to false
+// checkSClusterToBeSandboxedShutdownUnset ensures the following when a subcluster (A) is to be added to a sandbox (B)
+// 1 A's Shutdown field is set to false
+// 2 B's Shutdown field is set to false
+// 3 B does not have any subcluster in it that has Shutdown set to true (spec/status)
 func (v *VerticaDB) checkSClusterToBeSandboxedShutdownUnset(allErrs field.ErrorList) field.ErrorList {
 	statusSClusterMap := v.GenStatusSubclusterMap()
 	newSclusterMap := v.GenSubclusterMap()
@@ -1808,14 +1811,14 @@ func (v *VerticaDB) checkSClusterToBeSandboxedShutdownUnset(allErrs field.ErrorL
 	for subclusterName, sandboxName := range newSclusterSboxMap {
 		oldSandboxName, isInOldSandbox := statusScluterSboxMap[subclusterName]
 		_, foundSubcluster := newSclusterMap[subclusterName]
-		if !foundSubcluster { // this error should be captured by other piece of code
-			continue
+		if !foundSubcluster {
+			continue // avoid panic. this error should have been reported by other functions
 		}
 		_, hasStatus := statusSClusterMap[subclusterName]
-		// the subcluster to be sandboxed is either a new subcluster (not found in status) to be added
-		// or an existing subcluster (found in status) which includes two scenarios:
+		// the subcluster to be sandboxed is either a new subcluster to be added (not found in status)
+		// or an existing subcluster (found in status), the latter of which includes two scenarios:
 		//   1 the subcluster was in a sandbox (whose name is different from current sandbox name)
-		//   2 the subcluster was not in a sandbox
+		//   2 the subcluster was not in a sandbox previously
 		if !hasStatus || (hasStatus && (!isInOldSandbox || isInOldSandbox && oldSandboxName != sandboxName)) {
 			sandbox := newSandboxMap[sandboxName]
 			errMsgs := v.checkSboxForShutdown(sandbox, newSclusterMap, statusSclusterIndexMap)
--- a/api/v1/verticadb_webhook.go
+++ b/api/v1/verticadb_webhook.go
@@ -1773,30 +1773,28 @@ func (v *VerticaDB) checkNewSBoxOrSClusterShutdownUnset(allErrs field.ErrorList)
 // live in a sandbox where Shutdown is set to false in spec/status for all pieces
 func (v *VerticaDB) checkShutdownForScaleUpOrDown(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {
 	newSclusterMap := v.GenSubclusterMap()
-	newSclusterSboxMap := v.GenSubclusterSandboxMap()
-	newSandboxMap := v.GenSandboxMap()
-	statusSclusterIndexMap := v.GenStatusSClusterIndexMap()
 	oldSclusterMap := oldObj.GenSubclusterMap()
 	newSclusterIndexMap := v.GenSubclusterIndexMap()
-	for subclusterName, sandboxName := range newSclusterSboxMap {
+	statusSubclusterMap := v.GenStatusSubclusterMap()
+	sclusterSboxMap := v.GenSubclusterSandboxMap()
+	for subclusterName, newScluster := range newSclusterMap {
 		oldScluster, foundSclusterInOld := oldSclusterMap[subclusterName]
 		if !foundSclusterInOld {
 			continue // we only need to deal with persistent subclusters
 		}
-		newScluster, foundSclusterInNew := newSclusterMap[subclusterName]
-		if !foundSclusterInNew {
-			continue // avoid unexpected panic when something is wrong
+		_, inSandbox := sclusterSboxMap[subclusterName]
+		if inSandbox {
+			continue // this scenario is handled by another webhook rule
 		}
-		sandbox := newSandboxMap[sandboxName]
 		if oldScluster.Size != newScluster.Size { // scale up/down
-			errMsgs := v.checkSboxForShutdown(sandbox, newSclusterMap, statusSclusterIndexMap)
-			if len(errMsgs) != 0 {
+			subclusterStatus, hasStatus := statusSubclusterMap[subclusterName]
+			if newScluster.Shutdown || (hasStatus && subclusterStatus.Shutdown) {
 				i := newSclusterIndexMap[subclusterName]
 				p := field.NewPath("spec").Child("subclusters").Index(i).Child("size")
 				err := field.Invalid(p,
 					newScluster.Size,
-					fmt.Sprintf("cannot scale up/down subcluster %q in sandbox %q because %q",
-						subclusterName, sandboxName, strings.Join(errMsgs, ",")))
+					fmt.Sprintf("cannot scale up/down %q which is marked for shutdown or has been shut down",
+						subclusterName))
 				allErrs = append(allErrs, err)
 			}
 		}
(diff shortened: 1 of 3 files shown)


api/v1/verticadb_webhook.go
```diff
@@ -1765,6 +1769,85 @@ func (v *VerticaDB) checkNewSBoxOrSClusterShutdownUnset(allErrs field.ErrorList)
...
+// checkShutdownForScaleUpOrDown ensures a subcluster to be scaled up/down has Shutdown field set to false and
+// live in a sandbox where Shutdown is set to false in spec/status for all pieces
+func (v *VerticaDB) checkShutdownForScaleUpOrDown(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {
+	newSclusterMap := v.GenSubclusterMap()
+	newSclusterSboxMap := v.GenSubclusterSandboxMap()
+	newSandboxMap := v.GenSandboxMap()
+	statusSclusterIndexMap := v.GenStatusSClusterIndexMap()
+	oldSclusterMap := oldObj.GenSubclusterMap()
+	newSclusterIndexMap := v.GenSubclusterIndexMap()
+	for subclusterName, sandboxName := range newSclusterSboxMap {
```
cchen-vertica, line 1781: This loop will only check subclusters in the sandbox. What if the subclusters not in a sandbox?

api/v1/verticadb_webhook.go
```diff
@@ -1765,6 +1769,85 @@ func (v *VerticaDB) checkNewSBoxOrSClusterShutdownUnset(allErrs field.ErrorList)
...
+// 3 B does not have any subcluster in it that has Shutdown set to true (spec/status)
+func (v *VerticaDB) checkSClusterToBeSandboxedShutdownUnset(allErrs field.ErrorList) field.ErrorList {
+	newSclusterMap := v.GenSubclusterMap()
+	statusScluterSboxMap := v.GenSubclusterSandboxStatusMap()
+	newSclusterSboxMap := v.GenSubclusterSandboxMap()
+	newSandboxIndexMap := v.GenSandboxIndexMap()
+	newSandboxMap := v.GenSandboxMap()
+	statusSclusterIndexMap := v.GenStatusSClusterIndexMap()
+	sandboxWithError := map[string]bool{}
+	for subclusterName, sandboxName := range newSclusterSboxMap {
```
cchen-vertica, line 1819: Minor: rename subclusterName to newSC, sandboxName to newSB for bettern readability.

api/v1/verticadb_webhook.go
```diff
@@ -1765,6 +1769,85 @@ func (v *VerticaDB) checkNewSBoxOrSClusterShutdownUnset(allErrs field.ErrorList)
...
+		oldSandboxName, isInOldSandbox := statusScluterSboxMap[subclusterName]
+		_, foundSubcluster := newSclusterMap[subclusterName]
+		if !foundSubcluster {
+			continue // avoid panic. this error should have been reported by other functions
+		}
+		// the subcluster to be sandboxed is either a new subcluster to be added (not found in status)
+		// or an existing subcluster (found in status), the latter of which includes two scenarios:
+		//   1 the subcluster was in a sandbox (whose name is different from current sandbox name)
+		//   2 the subcluster was not in a sandbox previously
+		if !isInOldSandbox || isInOldSandbox && oldSandboxName != sandboxName {
```
cchen-vertica, line 1829: Use parenthesis to separate the conditions for readability.

cchen-vertica: The case `isInOldSandbox && oldSandboxName != sandboxName` is for unsandboxing. I remember this is checked in another rule. Will we get two errors for this case from two different rules?

LiboYu2: Good memory! Here is the related rule from my previous PR:
You cannot unsandbox a subcluster with `shutdown` true(spec and/or status) or part of a sandbox with `shutdown` true 

There will be two error messages:
When a subcluster is moved from one sandbox to another,
-  this related rule will raise an error for **unsandboxing** scenario.
- The rule implmented in this PR will raise another error messages for **sandboxing** scenario.

LiboYu2: Removed this check "isInOldSandbox && oldSandboxName != sandboxName" per offline discussion to avoid duplicate error messages.

api/v1/verticadb_webhook.go
cchen-vertica: Isn't this function for this rule: you cannot scale (up/down) a subcluster with shutdown true(spec or status)?

api/v1/verticadb_webhook.go
```diff
@@ -1900,6 +1979,26 @@ func (v *VerticaDB) checkShutdownForSandboxesToBeRemoved(oldObj *VerticaDB, allE
...
 
+// checkShutdownForSubclustersToBeRemoved ensures the subclusters to be removed are not shut down
+func (v *VerticaDB) checkShutdownForSubclustersToBeRemoved(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {
+	oldSubclusterMap := oldObj.GenSubclusterMap()
+	newSubclusterMap := v.GenSubclusterMap()
+	statusSubclusterMap := v.GenStatusSubclusterMap()
+	subclustersToBeRemoved := vutil.MapKeyDiff(oldSubclusterMap, newSubclusterMap)
+	for _, subclusterToBeRemoved := range subclustersToBeRemoved {
+		subclusterstatus, foundInStatus := statusSubclusterMap[subclusterToBeRemoved]
+		if oldSubclusterMap[subclusterToBeRemoved].Shutdown || foundInStatus && subclusterstatus.Shutdown {
```
roypaulin, line 1990: minor
```suggestion
		if oldSubclusterMap[subclusterToBeRemoved].Shutdown || (foundInStatus && subclusterstatus.Shutdown) {
```

api/v1/verticadb_webhook.go
```diff
@@ -1810,38 +1810,34 @@ func (v *VerticaDB) checkShutdownForScaleUpOrDown(oldObj *VerticaDB, allErrs fie
...
-		oldSandboxName, isInOldSandbox := statusScluterSboxMap[subclusterName]
-		_, foundSubcluster := newSclusterMap[subclusterName]
+	for newSC, newSB := range newSclusterSboxMap {
+		_, foundSubcluster := newSclusterMap[newSC]
 		if !foundSubcluster {
 			continue // avoid panic. this error should have been reported by other functions
 		}
 		// the subcluster to be sandboxed is either a new subcluster to be added (not found in status)
 		// or an existing subcluster (found in status), the latter of which includes two scenarios:
 		//   1 the subcluster was in a sandbox (whose name is different from current sandbox name)
```
cchen-vertica, line 1825: Remove this line and modify the comment accordingly.

api/v1/verticadb_webhook.go
```diff
@@ -1900,6 +1979,26 @@ func (v *VerticaDB) checkShutdownForSandboxesToBeRemoved(oldObj *VerticaDB, allE
...
+// checkShutdownForSubclustersToBeRemoved ensures the subclusters to be removed are not shut down
+func (v *VerticaDB) checkShutdownForSubclustersToBeRemoved(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {
+	oldSubclusterMap := oldObj.GenSubclusterMap()
+	newSubclusterMap := v.GenSubclusterMap()
+	statusSubclusterMap := v.GenStatusSubclusterMap()
+	subclustersToBeRemoved := vutil.MapKeyDiff(oldSubclusterMap, newSubclusterMap)
+	for _, subclusterToBeRemoved := range subclustersToBeRemoved {
+		subclusterstatus, foundInStatus := statusSubclusterMap[subclusterToBeRemoved]
+		if oldSubclusterMap[subclusterToBeRemoved].Shutdown || foundInStatus && subclusterstatus.Shutdown {
+			p := field.NewPath("spec").Child("subclusters")
```
roypaulin, line 1991: Set this outside of the loop.

api/v1/verticadb_webhook.go
```diff
@@ -1900,6 +1979,26 @@ func (v *VerticaDB) checkShutdownForSandboxesToBeRemoved(oldObj *VerticaDB, allE
...
+			err := field.Invalid(p,
```
roypaulin, line 1992:
```suggestion
			err := field.Invalid(p.Index(...),
```

api/v1/verticadb_webhook.go
```diff
@@ -1810,38 +1810,34 @@ func (v *VerticaDB) checkShutdownForScaleUpOrDown(oldObj *VerticaDB, allErrs fie
...
+	for newSC, newSB := range newSclusterSboxMap {
+		_, foundSubcluster := newSclusterMap[newSC]
 		if !foundSubcluster {
 			continue // avoid panic. this error should have been reported by other functions
 		}
 		// the subcluster to be sandboxed is either a new subcluster to be added (not found in status)
 		// or an existing subcluster (found in status), the latter of which includes two scenarios:
 		//   1 the subcluster was in a sandbox (whose name is different from current sandbox name)
 		//   2 the subcluster was not in a sandbox previously
-		if !isInOldSandbox || isInOldSandbox && oldSandboxName != sandboxName {
```
cchen-vertica: We still need the condition `if !isInOldSandbox` to cover the case 2: the subcluster was not in a sandbox previously

api/v1/verticadb_webhook.go
roypaulin: This will not include subclusters in the main cluster. You should iterate over `new.spec.subclusters`, for each subcluster , check if size has changed, if true, check if shutdown(spec/status) is true, if true append an error, then check if it is part of a sandbox, if true check sandbox.Shutdown, if true, append an error

roypaulin: Rectification, given that there is already a rule that prevents scaling a subcluster that is part of a sandbox, you  do not need to check the sandbox.

api/v1/verticadb_webhook.go
LiboYu2: After testing, I have to put back the condition "isInOldSandbox && oldSandboxName != sandboxName".
This is what I found:
When a subcluster is moved from sandbox A to sandbox B and A is shut down, , the previous rule kicks in.
When A is not shut down but B is shut down, the new rule applies.  
Only when both A and B are shut down, there will be two different error messages: one from unsandboxing and one from sandboxing.

api/v1/verticadb_webhook.go
```diff
@@ -1773,30 +1773,28 @@ func (v *VerticaDB) checkNewSBoxOrSClusterShutdownUnset(allErrs field.ErrorList)
...
 		oldScluster, foundSclusterInOld := oldSclusterMap[subclusterName]
 		if !foundSclusterInOld {
 			continue // we only need to deal with persistent subclusters
 		}
-		newScluster, foundSclusterInNew := newSclusterMap[subclusterName]
-		if !foundSclusterInNew {
-			continue // avoid unexpected panic when something is wrong
+		_, inSandbox := sclusterSboxMap[subclusterName]
+		if inSandbox {
+			continue // this scenario is handled by another webhook rule
```
cchen-vertica, line 1787: Which rule handles this one? If you don't check the subclusters in sandboxes, you need to remove comment in line 1773

LiboYu2: line 1631.
// checkImmutableSubclusterInSandbox ensures we do not scale and remove any subcluster that is in a sandbox
func (v *VerticaDB) checkImmutableSubclusterInSandbox(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {

comment is removed.

cchen-vertica: Minor: update this line's comment to "this scenario is handled by checkImmutableSubclusterInSandbox()" in your next PR.

api/v1/verticadb_webhook.go
roypaulin: No, the subcluster is not moved yet. just use oldObj.

api/v1/verticadb_webhook.go
```diff
@@ -106,6 +106,7 @@ func (v *VerticaDB) ValidateCreate() error {
 	verticadblog.Info("validate create", "name", v.Name, "GroupVersion", GroupVersion)
 
 	allErrs := v.validateVerticaDBSpec()
+	allErrs = v.checkNewSBoxOrSClusterShutdownUnset(allErrs)
```
roypaulin, line 109: Move this to `validateVerticaDBSpec()` and remove it from `validateImmutableFields()`.

PR #980, opened 2024-11-14
Remove proxy when subcluster removed

HaoYang0000:
When you remove the subcluster from the vdb, its pods will get deleted after the subcluster is removed. The client proxy becomes useless and must be removed too together with the configMap.

Also, when a subcluster size is scaled down to 0(e.g. stop_subcluster), we should scale down proxy size to 0 as well. And this should be done after stop_subcluster(all connections are draining down).

TODO: Update e2e test case


pkg/controllers/vdb/dbremovenode_reconciler.go
```diff
@@ -146,6 +148,14 @@ func (d *DBRemoveNodeReconciler) removeNodesInSubcluster(ctx context.Context, sc
 			return ctrl.Result{}, fmt.Errorf("failed to update subcluster status: %w", err)
 		}
 
+		if vmeta.UseVProxy(d.Vdb.Annotations) {
```
cchen-vertica, line 151: Move the condition "sc.Size == 0" here since we don't want to build proxy deployment when sc.Size != 0 and it will be more readable.

pkg/controllers/vdb/dbremovenode_reconciler.go
```diff
@@ -146,6 +148,14 @@ func (d *DBRemoveNodeReconciler) removeNodesInSubcluster(ctx context.Context, sc
...
+			// Scale donw client proxy size to 0
+			err := d.scaleDonwClientProxy(ctx, sc)
```
cchen-vertica, line 153: Fix the typo "Donw" in this file.

pkg/controllers/vdb/dbremovesubcluster_reconciler.go
```diff
@@ -199,6 +208,25 @@ func (d *DBRemoveSubclusterReconciler) updateSubclusterStatus(ctx context.Contex
...
+	cmName := names.GenVProxyConfigMapName(d.Vdb, sc)
+	vpName := names.GenVProxyName(d.Vdb, sc)
+	vpDep := builder.BuildVProxyDeployment(vpName, d.Vdb, sc)
+	cmDep := builder.BuildVProxyConfigMap(cmName, d.Vdb, sc)
+
+	if sc.Size == 0 {
+		d.Log.Info("Deleting client proxy config map", "Name", cmName)
+		err := d.VRec.GetClient().Delete(ctx, cmDep)
+		if err != nil {
+			return err
```
cchen-vertica, line 222: Add a error log here.

cchen-vertica:
We don't need to manually delete config maps or deployments in the operator. We should let subcluster sts own proxy deployment, and proxy deployment own config map. So when the subcluster is deleted, the proxy pod and its config map will be auto deleted.

pkg/controllers/vdb/dbremovenode_reconciler.go
HaoYang0000: moved in https://github.com/vertica/vertica-kubernetes/pull/980/commits/6ac1b322c3ca3e79772bf448f5d47c8a63fa304d

HaoYang0000:
> Creating, updating and deleting k8s objects is generally done in `obj_reconciler.go`. This can be simpler, in obj_reconciler, just before a subcluster's statefulset is deleted, delete its client proxy. Also in obj_reconciler, there is a place where the sts is updated, all you have to do is to check the sts size and if it is zero, set the client proxy deployment size to zero too, and update it.
> 
> Let me know if you have any confusion.

For the scale down to 0 part, I think I found the right place to update the client proxy size to 0. But for remove subcluster part, is it under method "checkForDeletedSubcluster"? 
Also, do we need to worry about the order of the reconcile call? For example, when I removed the subcluster sts in vdb, the dbremovesubcluster_reconciler and obj_reconciler will be called at same time, or in different order?

HaoYang0000:
> We don't need to manually delete config maps or deployments in the operator. We should let subcluster sts own proxy deployment, and proxy deployment own config map. So when the subcluster is deleted, the proxy pod and its config map will be auto deleted.

Based on the existing structure, I see there is proxy lives inside the subcluster already in verticadb_types.go. So if the subcluster is deleted, does that mean the proxy pod will be deleted as well? Do we need to add logic for this part?

roypaulin:
> > Creating, updating and deleting k8s objects is generally done in `obj_reconciler.go`. This can be simpler, in obj_reconciler, just before a subcluster's statefulset is deleted, delete its client proxy. Also in obj_reconciler, there is a place where the sts is updated, all you have to do is to check the sts size and if it is zero, set the client proxy deployment size to zero too, and update it.
> > Let me know if you have any confusion.
> 
> For the scale down to 0 part, I think I found the right place to update the client proxy size to 0. But for remove subcluster part, is it under method "checkForDeletedSubcluster"? Also, do we need to worry about the order of the reconcile call? For example, when I removed the subcluster sts in vdb, the dbremovesubcluster_reconciler and obj_reconciler will be called at same time, or in different order?

Both scale down and deployment should occur in obj_reconciler. The reconcilers are never run at the same time, the order is what you see in verticadb_controller. But if you look closely in obj_reconciler, you will see we never delete a sts or scale it down before remove_subcluster/remove_node.
Let's talk tomorrow, I will explain it.

HaoYang0000:
Adding Qin's e2e test commits in https://github.com/vertica/vertica-kubernetes/pull/980/commits/c5b4b9cb2eabd5cc9bda4b5213c8fd7073da9e22

pkg/controllers/vdb/k8s.go
```diff
@@ -92,11 +92,29 @@ func createSts(ctx context.Context, vrec config.ReconcilerInterface, expSts *app
...
+	}
+	return vrec.GetClient().Update(ctx, vpDep)
+}
+
+// deleteDep will delete an existing deployment. It assumes the deployment exist already.
+func deleteDep(ctx context.Context, vrec config.ReconcilerInterface, vpDep *appsv1.Deployment, vdb *vapi.VerticaDB) error {
+	err := ctrl.SetControllerReference(vdb, vpDep, vrec.GetClient().Scheme())
+	if err != nil {
+		return err
+	}
```
roypaulin, line 118:
```suggestion
```

pkg/controllers/vdb/k8s.go
```diff
@@ -92,11 +92,29 @@ func createSts(ctx context.Context, vrec config.ReconcilerInterface, expSts *app
...
 	}
 	return vrec.GetClient().Create(ctx, vpDep)
 }
+
+// updateDep will update an existing deployment. It assumes the deployment exist already.
+func updateDep(ctx context.Context, vrec config.ReconcilerInterface, vpDep *appsv1.Deployment, vdb *vapi.VerticaDB) error {
+	err := ctrl.SetControllerReference(vdb, vpDep, vrec.GetClient().Scheme())
+	if err != nil {
+		return err
+	}
```
roypaulin, line 109:
```suggestion
```

pkg/controllers/vdb/k8s.go
```diff
@@ -92,11 +92,29 @@ func createSts(ctx context.Context, vrec config.ReconcilerInterface, expSts *app
 	return vrec.GetClient().Create(ctx, expSts)
 }
 
-// createSts will create a new deployment. It assumes the deployment doesn't already exist.
+// createDep will create a new deployment. It assumes the deployment doesn't already exist.
 func createDep(ctx context.Context, vrec config.ReconcilerInterface, vpDep *appsv1.Deployment, vdb *vapi.VerticaDB) error {
 	err := ctrl.SetControllerReference(vdb, vpDep, vrec.GetClient().Scheme())
 	if err != nil {
 		return err
 	}
```
roypaulin, line 100: Now the deployment belongs to the sts so you should remove this.

api/v1/helpers.go
```diff
@@ -462,6 +462,11 @@ func (s *Subcluster) GetVProxyConfigMapName(vdb *VerticaDB) string {
 	return fmt.Sprintf("%s-%s-proxy-cm", vdb.Name, s.Name)
 }
 
+// GetVProxyConfigMapName returns the name of the client proxy config map
+func (v *VerticaDB) GetVProxyConfigMapName(scName string) string {
```
roypaulin, line 466: There is already `names.GenVProxyConfigMapName`, what do you need this for?

pkg/names/names.go
```diff
@@ -76,6 +76,12 @@ func GenVProxyConfigMapName(vdb *vapi.VerticaDB, sc *vapi.Subcluster) types.Name
 	return GenNamespacedName(vdb, sc.GetVProxyConfigMapName(vdb))
 }
 
+// GenVProxyConfigMapNameFromVDB returns the name of the client proxy configmap from
+// subcluster name string
+func GenVProxyConfigMapNameFromVDB(vdb *vapi.VerticaDB, scName string) types.NamespacedName {
```
roypaulin, line 81: Don't need this, use GenVProxyConfigMapName

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -272,6 +272,16 @@ func (o *ObjReconciler) checkForDeletedSubcluster(ctx context.Context) (ctrl.Res
 		if err != nil {
 			return ctrl.Result{}, err
 		}
+		// Delete vproxy config map
+		if vmeta.UseVProxy(o.Vdb.Annotations) {
+			scName := stss.Items[i].Labels[vmeta.SubclusterNameLabel]
```
roypaulin, line 277: you should check that sts contains the subcluster label.

HaoYang0000: Sure, added validation in https://github.com/vertica/vertica-kubernetes/pull/980/commits/ec75345ef8c24a0d98579ef71e4996f440a966ef

pkg/controllers/vdb/k8s.go
HaoYang0000: Removed in https://github.com/vertica/vertica-kubernetes/pull/980/commits/2cb807d2a895bc50d4b73093cb4bd4e3f2a49028

pkg/controllers/vdb/k8s.go
HaoYang0000: Removed in https://github.com/vertica/vertica-kubernetes/pull/980/commits/2cb807d2a895bc50d4b73093cb4bd4e3f2a49028

api/v1/helpers.go
HaoYang0000: The names.GenVProxyConfigMapName takes params: vdb *vapi.VerticaDB, sc *vapi.Subcluster
But when the time I need to generate the configMapName, I don't have the sc object available.
Only the subclusterName, which is retrieved from sts label.

pkg/names/names.go
HaoYang0000: Duplicate conversation as https://github.com/vertica/vertica-kubernetes/pull/980#discussion_r1851608704

tests/e2e-leg-11/client-proxy/45-assert.yaml
```diff
@@ -0,0 +1,67 @@
...
+spec:
+  subclusters:
+    - name: pri1
+      size: 4
+      proxy:
+        replica: 2
+    - name: sec1
+      size: 0
+      proxy:
+        replica: 0
```
cchen-vertica, line 67: You need to assert the subcluster status as well.

tests/e2e-leg-11/client-proxy/47-assert.yaml
```diff
@@ -0,0 +1,37 @@
...
+apiVersion: vertica.com/v1
+kind: VerticaDB
+metadata:
+  name: v-client-proxy
+spec:
+  subclusters:
+    - name: pri1
+      size: 4
+      proxy:
+        replica: 2
```
cchen-vertica, line 37: You need to assert the subcluster status. And you need another assert file to verify the deployment and config map don't exist.

tests/e2e-leg-11/client-proxy/45-assert.yaml
HaoYang0000: Updated in https://github.com/vertica/vertica-kubernetes/pull/980/commits/2ba5e9177241b32e2fa786d009b93a1fab9a0849

tests/e2e-leg-11/client-proxy/47-assert.yaml
HaoYang0000: Updated the subcluster status check in: https://github.com/vertica/vertica-kubernetes/pull/980/commits/a0b499152a983d8631970f4b920581aae28a87b7

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -480,6 +516,13 @@ func (o *ObjReconciler) checkVProxyDeployment(ctx context.Context, sc *vapi.Subc
 	curDep := &appsv1.Deployment{}
 	vpDep := builder.BuildVProxyDeployment(vpName, o.Vdb, sc)
 	vpErr := o.Rec.GetClient().Get(ctx, vpName, curDep)
+
+	if sc.Size == 0 {
+		*vpDep.Spec.Replicas = 0
+		o.Log.Info("Scale down client proxy", "Name", vpName, "Size", vpDep.Spec.Replicas, "Image", vpDep.Spec.Template.Spec.Containers[0].Image)
+		return o.Rec.GetClient().Update(ctx, vpDep)
+	}
```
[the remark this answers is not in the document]
HaoYang0000: Changed to use sts.replicas in https://github.com/vertica/vertica-kubernetes/pull/980/commits/975c93a7ec550203014e65bcf0c95d8fd56d14ea

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -286,6 +293,20 @@ func (o *ObjReconciler) checkForDeletedSubcluster(ctx context.Context) (ctrl.Res
...
+		return ctrl.Result{}, err
+	}
+
+	for i := range deploy.Items {
+		err = o.Rec.GetClient().Delete(ctx, &deploy.Items[i])
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
```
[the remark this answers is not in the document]
HaoYang0000: Removed in https://github.com/vertica/vertica-kubernetes/pull/980/commits/2953813501736fb47ff5c45b08399154f05262e2

tests/e2e-leg-11/client-proxy/47-assert.yaml
HaoYang0000: The verify case added in https://github.com/vertica/vertica-kubernetes/pull/980/commits/18086da1b6a6ff75bcb0858ee24885900dfb86b1

roypaulin:
You can take it from here. please add more unit tests.

tests/e2e-leg-11/client-proxy/47-assert.yaml
```diff
@@ -0,0 +1,41 @@
...
+  name: v-client-proxy-pri1
+status:
+  currentReplicas: 4
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: v-client-proxy-pri1-proxy
+status:
+  readyReplicas: 2
```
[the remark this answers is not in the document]
HaoYang0000: added assert in https://github.com/vertica/vertica-kubernetes/pull/980/commits/4d73485709b6883ab3e4cfa229cc65925b68faf5

tests/e2e-leg-11/client-proxy/55-kill-proxy-pod.yaml
```diff
@@ -25,19 +25,20 @@ data:
...
-    # new proxy pod name should be different to the original one
-    for pod_new in $PROXY_PRI1_NEW; do
-      for pod_orig in $PROXY_PRI1_ORIG; do
-        if [ "$pod_new" == "$pod_orig" ]; then
-          echo "Assertion failed: expect a new proxy pod to be created, got the original pod $PROXY_PRI1_ORIG"
-          exit 1
-        fi
-      done
-    done
+    # # new proxy pod name should be different to the original one
```
HaoYang0000, line 33: Meeting an problem when delete pod, the svc seems failed to update due to conflicts of object update:
 logger.go:42: 08:05:19 | client-proxy | 2024-11-29 07:55:04 +0000 UTC	Warning	Endpoints v-client-proxy		FailedToUpdateEndpoint	Failed to update endpoint kuttl-test-improved-fawn/v-client-proxy: Operation cannot be fulfilled on endpoints "v-client-proxy": the object has been modified; please apply your changes to the latest version and try again

roypaulin:
@cchen-vertica, having the sts own the deployment has a limitation which is proxy pods can't trigger the controller, so if a proxy pod/deployment is deleted, as they aren't owned by vdb, nothing will happen on the operator, so if there is any reconciler that depends on the proxy, it won't be run. Example: if you delete proxy pod, a new one will be created but the routing label won't be added because it won't trigger the controller.
I decided to explicitly delete the proxy dep and configmap when a subcluster is removed.

tests/e2e-leg-11/client-proxy/45-assert.yaml
```diff
@@ -0,0 +1,71 @@
...
+metadata:
+  name: v-client-proxy
+spec:
+  subclusters:
+    - name: pri1
+      size: 4
+      proxy:
+        replica: 2
+    - name: sec1
+      size: 0
```
cchen-vertica, line 65: We should assert "proxy.replica: 0" here as well.

tests/e2e-leg-11/client-proxy/55-errors.yaml
```diff
@@ -0,0 +1,17 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: v-client-proxy-pri1-proxy
```
cchen-vertica, line 17: We need to add one more step to verify the deleted proxy deployment can be recreated.

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -468,49 +483,113 @@ func (o *ObjReconciler) checkVProxyConfigMap(ctx context.Context, cmName types.N
...
-		o.PFacts.Invalidate()
-		return ctrl.Result{}, createSts(ctx, o.Rec, expSts, o.Vdb)
+	stsCreated, err := o.ensureStsExists(ctx, nm, curSts, expSts)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	var curDep *appsv1.Deployment
+	if vmeta.UseVProxy(o.Vdb.Annotations) {
+		if curDep, err = o.reconcileVProxy(ctx, sc); err != nil {
```
cchen-vertica, line 585: We should only create proxy deployment after a subcluster sts is created. Do reconcileVProxy() only in line 591 and line 648.

roypaulin: I don't need to. reconcileVProxy will create the proxy deployment whenever it is missing. If you look at the code, you will see I always check first if the proxy exists
- If I move this to line 591, if I manually delete the deployment, it won't get recreated because the sts already exists.
- In line 648, I only want to recreate the sts and that has nothing to do with the proxy deployment. Keep in mind sts no longer owns the deployment so deleting it won't delete the proxy.

tests/e2e-leg-11/client-proxy/55-errors.yaml
roypaulin: Step 56 does that.

pkg/controllers/vdb/obj_reconciler.go
cchen-vertica: I don't want the proxy to be created when a statefulset is updated. During online upgrade, we removed the old main and the proxy. However, when we rename subclusters in the new main, the old proxy will be created again due to a statefulset update. 

We probably need a more sophisticated proxy reconciler to handle the rename case.

roypaulin: Or we can reuse the annotation trick. I think what you want is that when we rename the subclusters, the stss and proxies do not get rebuilt/recreated.

roypaulin:
@LiboYu2 Can you also review this PR?

cchen-vertica (approved):

PR #977, opened 2024-11-05
Create vertica client proxy pods in subcluster

qindotguan:
This is the first step to supporting session transfer with client proxy: if the proxy field is defined, a deployment with vertical client pods will be created in each subcluster. 

For example:
```
spec:
   subclusters:
   - name: sc1
     proxy:
        image: vertica-client-proxy:latest
        replica: 2 (default: 1)
        log-level: DEBUG (default: INFO)
        port: 5433 (default: 5433, we don't support using other ports in MVP)
        tls-certs: xxxx 
```

The client proxy configurations are stored in a config map, which is mounted into the proxy pods as /config/config.yaml.

PR #981, opened 2024-11-17
update proxy config maps when add subcluster/node or remove subcluster/node

qindotguan:
#977 implemented the function to create a proxy deployment and config map when creating a subcluster. 

This PR will address the following situations:
1. remove subclusters: remove the proxy deployments and config maps for the subclusters
2. add nodes to a subcluster: update the Vertica nodes list with newly added nodes in the config map of that subcluster
3. remove nodes from a subcluster: update the Vertica nodes list to remove the nodes from the config map of that subcluster


pkg/builder/builder.go
```diff
@@ -866,7 +866,7 @@ func BuildVProxyDeployment(nm types.NamespacedName, vdb *vapi.VerticaDB, sc *vap
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        nm.Name,
 			Namespace:   nm.Namespace,
-			Labels:      MakeLabelsForVProxyObject(vdb, sc, true),
+			Labels:      MakeLabelsForVProxyObject(vdb, sc, false),
```
qindotguan, line 869: Change `forPod ` to false to add label `vertica.com/subcluster-name` to the proxy deployment. Only objects with `vertica.com/subcluster-name` will be deleted In `checkForDeletedSubcluster`.

Labels added to the object when forPod is true:
```
    vertica.com/subcluster-selector-name: v-client-proxy-sec1
    vertica.com/subcluster-svc: sec1
```
Labels added to the object when forPod is false:
```
    vertica.com/subcluster-name: sec1
    vertica.com/subcluster-type: secondary
```

roypaulin:
As we discussed, let's not include the delete client-proxy here.

cchen-vertica:
You can remove your proxy-delete code in this PR. We don't need to manually delete config map or deployments in the operator. We should let subcluster sts owns proxy deployment, and proxy deployment owns config map. So when subcluster is deleted, proxy pod and its config map will be auto-deleted. I will ask Hao to implement the new ownerships.

PR #983, opened 2024-11-18
Added client routing label to proxy pods

cchen-vertica:
This PR added client routing label to proxy pods for routing traffic from service to client proxy pods. After this change, the service will only communicate with the proxy pods rather than vertica pods.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -862,20 +862,30 @@ func makeScrutinizeInitContainers(vscr *v1beta1.VerticaScrutinize, vdb *vapi.Ver
 
 // BuildVProxyDeployment builds manifest for a subclusters VProxy deployment
 func BuildVProxyDeployment(nm types.NamespacedName, vdb *vapi.VerticaDB, sc *vapi.Subcluster) *appsv1.Deployment {
+	proxyLabels := MakeLabelsForVProxyObject(vdb, sc, true)
+	depSelectorLabels := MakeDepSelectorLabels(vdb, sc)
+	// pod should have both proxy labels and deployment selector labels
+	depPodLabels := make(map[string]string)
+	for k, v := range proxyLabels {
+		depPodLabels[k] = v
+	}
+	for k, v := range depSelectorLabels {
+		depPodLabels[k] = v
+	}
 	return &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        nm.Name,
 			Namespace:   nm.Namespace,
-			Labels:      MakeLabelsForVProxyObject(vdb, sc, true),
+			Labels:      proxyLabels,
 			Annotations: MakeAnnotationsForVProxyObject(vdb),
 		},
 		Spec: appsv1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{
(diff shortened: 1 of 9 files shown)


pkg/builder/builder.go
```diff
@@ -862,20 +862,30 @@ func makeScrutinizeInitContainers(vscr *v1beta1.VerticaScrutinize, vdb *vapi.Ver
 
 // BuildVProxyDeployment builds manifest for a subclusters VProxy deployment
 func BuildVProxyDeployment(nm types.NamespacedName, vdb *vapi.VerticaDB, sc *vapi.Subcluster) *appsv1.Deployment {
+	proxyLabels := MakeLabelsForVProxyObject(vdb, sc, true)
```
roypaulin, line 865:
```suggestion
	proxyLabels := MakeLabelsForVProxyObject(vdb, sc, false)
```
because it is not a pod. However you probably to update `MakeCommonLabels()`, maybe add an argument `isProxy`.
If you have a better idea, feel free to implement it

pkg/builder/builder.go
```diff
@@ -862,20 +862,30 @@ func makeScrutinizeInitContainers(vscr *v1beta1.VerticaScrutinize, vdb *vapi.Ver
...
+	depSelectorLabels := MakeDepSelectorLabels(vdb, sc)
+	// pod should have both proxy labels and deployment selector labels
+	depPodLabels := make(map[string]string)
+	for k, v := range proxyLabels {
+		depPodLabels[k] = v
+	}
+	for k, v := range depSelectorLabels {
+		depPodLabels[k] = v
+	}
```
roypaulin, line 874: If you do what I mentioned above, you won't need this, you can set vmeta.ProxyPodSelectorLabel in MakeCommonLabels

cchen-vertica: I don't see how MakeCommonLabels will simplify the code here. The proxy labels are for proxy deployment, here I am collecting labels for proxy pods.

cchen-vertica: I made the changes in makeLabelsForObject which is the only caller of MakeCommonLabels.

cchen-vertica:
> Looks good! Is the tls issue fixed?

It's fixed. We don't need "-m disable" to connect to vertica anymore.

Makefile
```diff
@@ -439,6 +436,18 @@ else
 	scripts/push-to-kind.sh -i ${VLOGGER_IMG}
 endif
 
+.PHONY: docker-push-vproxy
+docker-push-vproxy:  ## Push vertica logger docker image
```
qindotguan, line 440: should be `client proxy docker image`

Makefile
```diff
@@ -439,6 +436,18 @@ else
...
+ifneq ($(strip $(VPROXY_IMG)),)
+ifeq ($(shell $(KIND_CHECK)), 0)
+	docker push ${VPROXY_IMG}
```
qindotguan, line 443: should we not push the client proxy image to docker from here?

cchen-vertica: If VPROXY_IMG is specified, we need to push it to docker.

Makefile
cchen-vertica: I will modify it in my next PR.

roypaulin (approved):

PR #985, opened 2024-11-22
Client proxy spec update

HaoYang0000:
We will move the common proxy attributes from spec.subcluster[].proxy to spec.proxy.

Old format:
spec
 subclusters:
   -name: sc1
    proxy:
      image:
      replica:
      tlsSecret:
      logLevel:
      resources:
    -name: sc2
     proxy:
      image:
      replica:
      tlsSecret:
      logLevel:
      resources:

New format:
spec
 proxy:
  image:
  tlsSecret:
  logLevel:
  subclusters:
    -name: sc1
     replica:
     resources:
    -name: sc2
     replica:
     resources:
 subclusters:
   -name: sc1 (no proxy fields here)
   -name: sc2 (no proxy fields here)


tests/e2e-leg-11/client-proxy/40-add-subcluster.yaml
```diff
@@ -11,13 +11,21 @@
...
-      "value": {"name": "sec1", "size": 2, "type": "secondary"}
-      }]'
-    namespaced: true
+apiVersion: vertica.com/v1
+kind: VerticaDB
+metadata:
+  name: v-client-proxy
+spec:
+  proxy:
+    image: kustomize-vproxy-image
```
roypaulin, line 20: I don't like this. It will change the proxy image to kustomize-vproxy-image which is an invalid image. Also, make sure the client-proxy test case passes in your local machine first.

qindotguan: This should be fine. It allows user to customize their image through the env variable $VPROXY_IMG instead of using the default one.

roypaulin: It does noest work as you think, it only works at step 15 when we use `kustomize`. Here it will just update the image to kustomize-vproxy-image which is wrong

api/v1/verticadb_types.go
```diff
@@ -840,6 +840,14 @@ type Proxy struct {
 	// use a secret path reference prefix, such as gsm://. Everything after the
 	// prefix is the name of the secret in the service you are storing.
 	TLSSecret string `json:"tlsSecret,omitempty"`
+}
+
+type ProxySubclusterConfig struct {
```
roypaulin, line 845: Let's make `resources` a pointer. `resources` in the manifest even when you do not set it. Making a pointer will allow us to omit subclusters[].proxy from the manifest when proxy is disabled.

api/v1/verticadb_types.go
```diff
@@ -840,6 +840,14 @@ type Proxy struct {
...
+	// +kubebuilder:default:=1
```
roypaulin, line 846: Let's remove the default value otherwise the proxy field will always show up in the manifest even when client proxy is disabled.

tests/e2e-leg-11/client-proxy/40-add-subcluster.yaml
HaoYang0000: If we want to use image value from the "kustomize-vproxy-image", what is the proper way to do that?

HaoYang0000: And btw, how is the kustomize-vproxy-image generated? I can't find any other reference of "kustomize-vproxy-image" rather than the setup db yaml

api/v1/verticadb_types.go
HaoYang0000: Done, updated in https://github.com/vertica/vertica-kubernetes/pull/985/commits/5135a6a307167492fc151fb108a08e94a7c9b0b0

api/v1/verticadb_types.go
HaoYang0000: Updated in https://github.com/vertica/vertica-kubernetes/pull/985/commits/201c6fff001e9faf91be18d9aa69184c3d8c7c76

api/v1/verticadb_types.go
```diff
@@ -816,12 +822,6 @@ type Proxy struct {
 	// The docker image name that contains the Vertica proxy server.
 	Image string `json:"image,omitempty"`
```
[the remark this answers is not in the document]
HaoYang0000: The default image value is removed in https://github.com/vertica/vertica-kubernetes/pull/985/commits/af8b77209961e2dbbc8638f26e6abb35edfa5cc6

HaoYang0000:
> The jira needs to enable the user to change replica num or other custom fields of client proxy. Could we add a test for that case? For example, changing replica num from 2 to 1 after db is running.

I added one test case in client-proxy test step 43, increasing the replicate number from 1 to 2 after db is running.

roypaulin:
> The jira needs to enable the user to change replica num or other custom fields of client proxy. Could we add a test for that case? For example, changing replica num from 2 to 1 after db is running.

The jira yes, but this PR is only about the API change, which is needed for release. I suggest we wrap up the api change first. "changing replica num from 2 to 1" is already taken care of by obj_reconciler, I can add a test.

cchen-vertica:
> > The jira needs to enable the user to change replica num or other custom fields of client proxy. Could we add a test for that case? For example, changing replica num from 2 to 1 after db is running.
> 
> The jira yes, but this PR is only about the API change, which is needed for release. I suggest we wrap up the api change first. "changing replica num from 2 to 1" is already taken care of by obj_reconciler, I can add a test.

Hao told me the "replica num change" part had been done in his last PR and tests were already added. You don't need to do that now.

tests/e2e-leg-11/client-proxy/40-add-subcluster.yaml
qindotguan: @roypaulin you are right. It works only if we run `kustomize build` as step 15. So here we can just don't set image and use the default value.

@HaoYang0000, this is my understanding of how it was generated:
The script scripts/setup-kustomize.sh will put $VPROXY_IMG into the key data.vproxyImage in the e2e configmap. Then each time `kustomize build` will replace kustomize-vproxy-image with that value stored in the key data.vproxyImage. So if the yaml file was not generated by `kustomize build`, kustomize-vproxy-image won't be replaced.

PR #987, opened 2024-11-26
Add VerticaReplicator webhook

jhamilton-ot:
Add webhook for VerticaReplicator. It enforces the following rules:
- Only one of the following should be specified: table, schema, pattern, or pattern with exclusion
- Make sure the new field "mode" only has one of the following fields: async, sync or empty string(for backward compatibility)
- Object name, include/exclude pattern, and target namespace cannot be used in sync mode


roypaulin (approved):
Looks good to me!
I would also like someone else familiar with the vcluster implementation to approve.

PR #990, opened 2024-12-02
fix leg 6 log rorate output

HaoYang0000:


HaoYang0000:
Seems like there is a config change on the log rorate, the proper fix should be adding something in/vertica-kubernetes/tests/e2e-leg-6/verify-server-logrotate/sql-input/logrotate.sql, to setup the log level or something so we don't get the new generated lines. But I can't find anything on the public doc: https://docs.vertica.com/24.4.x/en/admin/monitoring/rotating-log-files/

cchen-vertica (approved):

PR #999, opened 2024-12-10
use annotations proxy log level

HaoYang0000:
The proxy config map should use the annotation value of vertica.com/client-proxy-log-level. This pr is for updating the logic by replacing the hardcoded value.


roypaulin:
Looks good! Minor: can you add a webhook rule to check that the log level is one of these : TRACE|1=DEBUG|2=INFO|3=WARN|4=FATAL|5=NONE?

HaoYang0000:
> Looks good! Minor: can you add a webhook rule to check that the log level is one of these : TRACE|1=DEBUG|2=INFO|3=WARN|4=FATAL|5=NONE?

Sure, I will add one in the webhook PR

PR #1002, opened 2024-12-11
Prevent proxy deployment update

roypaulin:
When creating a deployment, a replicaset is created for each replica. When the update strategy is `RollingUpdate`, each time there is a change in the deployment spec, an extra pod will be created with the update and the old one will be deleted once the new one is ready. So there is a timing where you may see twice the number of pods with respect to what you expect. 
During testing, QA found that each time we create a proxy deployment, at first we had twice the number of required pods before getting the correct number after a few seconds. That is because we updated the deployment by adding the server version annotation when it is available in the vdb. That annotation does't serve any purpose so I just removed to avoid updating the deployment.

### DIFF
--- a/pkg/builder/labels_annotations.go
+++ b/pkg/builder/labels_annotations.go
@@ -171,8 +171,7 @@ func MakeAnnotationsForStsObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster) map[s
 // MakeAnnotationsForVProxyObject builds the list of annotations that are included
 // in the proxy config object.
 func MakeAnnotationsForVProxyObject(vdb *vapi.VerticaDB) map[string]string {
-	annotations := MakeAnnotationsForObject(vdb)
-	return annotations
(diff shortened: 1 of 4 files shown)


pkg/builder/labels_annotations.go
```diff
@@ -172,9 +172,6 @@ func MakeAnnotationsForStsObject(vdb *vapi.VerticaDB, sc *vapi.Subcluster) map[s
 // in the proxy config object.
 func MakeAnnotationsForVProxyObject(vdb *vapi.VerticaDB) map[string]string {
 	annotations := MakeAnnotationsForObject(vdb)
-	if ver, ok := vdb.Annotations[vmeta.VersionAnnotation]; ok {
-		annotations[vmeta.VersionAnnotation] = ver
-	}
 	return annotations
```
cchen-vertica, line 175: Minor: `return MakeAnnotationsForObject(vdb)`

cchen-vertica (approved):

PR #1006, opened 2024-12-13
Fix sandbox-upgrade-happy-path

roypaulin:
A bug was found in sandbox-upgrade-happy-path test case. The sandbox is upgraded and then removed (unsandboxed). After unsandboxing, we restart all the pods of the former sandbox. The issue is that the subcluster that joins the main cluster is not running the same image, so we get stuck for a while in startnode(HTTPSPollNodeStateIndirectOp) until the timeout(1200s). After timeout expires, it will requeue and run `UnsandboxImageVersionReconciler` which will recreate the pods with the correct image and next the next startnode attempt will succeed.  The solution consists in calling `UnsandboxImageVersionReconciler` just before the restart reconciler.

### DIFF
--- a/.github/actions/run-e2e-leg/action.yaml
+++ b/.github/actions/run-e2e-leg/action.yaml
@@ -164,7 +164,8 @@ runs:
           fi
         fi
 
-        if [[ "${{ inputs.leg-identifier }}" == "leg-4" ]]; then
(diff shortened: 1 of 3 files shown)


roypaulin:
I have tried more than 5 times locally and it works every time. Increasing the timeout of step55 to give it more time to pass on github actions

cchen-vertica (approved):

PR #1008, opened 2024-12-17
update vcluster package version

LiboYu2:
Picked up Vivek's two unsandboxing PRs.

### DIFF
--- a/pkg/controllers/vdb/upgrade.go
+++ b/pkg/controllers/vdb/upgrade.go
@@ -134,6 +134,7 @@ func (i *UpgradeManager) isVDBImageDifferent(ctx context.Context, sandbox string
 		return false, err
 	}
 	clusterImage := ""
+	differentImgFromVdb := false
 	for inx := range stss.Items {
 		sts := stss.Items[inx]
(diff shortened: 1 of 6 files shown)


roypaulin (approved):

pkg/controllers/vdb/upgrade.go
```diff
@@ -154,11 +155,11 @@ func (i *UpgradeManager) isVDBImageDifferent(ctx context.Context, sandbox string
 		}
 
 		if cntImage != targetImage {
-			return true, nil
+			differentImgFromVdb = true
```
cchen-vertica, line 158: Why do we need this change? Here if `differentImgFromVdb` is true, we should stop the for loop and return.

roypaulin: Say you have 3 statefulsets  sts1: img2 sts2: img2 sts3: img1 and in vdb image=img1. Normally we should not trigger upgrade but given that the 1st stafulset of the loop (sts1) has a different image from vdb, we will. With this change, you wait until you have finished comparing the images across statefulsets.

cchen-vertica (approved):

PR #1012, opened 2024-12-23
add setup-prometheus-adapter

qindotguan:
To scale up/down the Vertica DB subcluster size based on prometheus metrics, we need to install the prometheus adapter. This PR provided a way to deploy the promutheus adapter in our e2e test environment.
```
make setup-prometheus-adapter
```


Makefile
```diff
@@ -378,6 +378,12 @@ else
 	exit 1
 endif
 
+.PHONY: setup-prometheus-adapter
+setup-prometheus-adapter:  ## Setup prometheus adapter for VerticaAutoscaler
+	helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+	helm repo update
+	helm install prometheus-adapter prometheus-community/prometheus-adapter --namespace monitoring --create-namespace
```
roypaulin, line 385: Make the release name and namespace a parameter. Default release name and namespace can be prometheus-adapter.

- There must be an undeploy command too
- The user must be able to set the prometheus url, prometheus port , the number of replicas.

roypaulin (commented):
Look at Hao's PR. I think his PR needs to go in before yours, because I would like you to test the dapter when prometheus is running.

Makefile
qindotguan: - add PROMETHEUS_ADAPTER_NAME and PROMETHEUS_ADAPTER_NAMESPACE
- added undeploy-prometheus-adapter
- added example values with `--set prometheus.url`, `--set prometheus.port`, and `--set replicas`. I will make them configurable later

prometheus/adapter.yaml
```diff
@@ -0,0 +1,24 @@
+# Url to access prometheus
+prometheus:
+  url: http://prometheus.default.svc
+  port: 9090
+  path: ""
+
+# Number of prometheus adapter instances
+replicas: 1
```
roypaulin, line 8:
```suggestion
```
This is not needed. The default replica is already 1

roypaulin (commented):
Update Hao's test case

LiboYu2:
@roypaulin If a vdb has sandboxes, how do we exclude them when collecting metrics?

Makefile
```diff
@@ -142,22 +142,33 @@ OLM_CATALOG_IMG ?= olm-catalog:$(TAG)
...
 MINIMAL_VERTICA_IMG ?=
 # Name of the helm release that we will install/uninstall
 HELM_RELEASE_NAME?=vdb-op
-# Prometheus variables that we wil be used for deployment 
-PROMETHEUS_HELM_NAME?=prometheus
-PROMETHEUS_INTERVAL?=5s
 DB_USER?=dbadmin
 DB_PASSWORD?=
 VDB_NAME?=verticadb-sample
 VDB_NAMESPACE?=default
```
roypaulin, line 165: Move these next to other prometheus variables.

roypaulin:
Try prometheus-sanity first on your local machine to make sure it works.

Makefile
roypaulin: Can you address this?

roypaulin (approved):
Looks good! Can you also update DEVELOPER.md like Hao did in his open PR?

PR #1019, opened 2025-01-07
Create hpa from CR

roypaulin:
A new field `CustomAutoscaler` was added. When set, there is a reconciler that will create an hpa whose values will come from that field. If that field change, the hpa is updated to reflect the changes. If the hpa is deleted, the reconciler will make sure it is recreated.


LiboYu2:
> @LiboYu2 @qindotguan @HaoYang0000 thoughts, comments, questions?

I will take a look this afternoon.

api/v1/version.go
```diff
@@ -58,6 +58,8 @@ const (
 	// Starting in v24.1.0, the default deployment method supported changes from admintools to vclusterops
 	// for official releases of vertica-k8s images
 	VcluseropsAsDefaultDeploymentMethodMinVersion = "v24.1.0"
+	// The min version that exposed prometheus metrics through the http service
+	PrometheusMetricsMinVersion = "v12.0.4"
```
cchen-vertica, line 62: So the new auto-scaler can work with both admintools and vcluster?

api/v1beta1/verticaautoscaler_types.go
```diff
@@ -84,16 +84,15 @@ type VerticaAutoscalerSpec struct {
...
 }
 
 // CustomAutoscalerSpec customizes VerticaAutoscaler
 type CustomAutoscalerSpec struct {
-
 	// +kubebuilder:Minimum:=0
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// The miminum number of pods when scaling.
-	MinReplicas int32 `json:"minReplicas,omitempty"`
+	MinReplicas *int32 `json:"minReplicas,omitempty"`
```
cchen-vertica, line 95: Why are we using different types for "MinReplicas" and "MaxReplicas"?

api/v1beta1/verticaautoscaler_types.go
```diff
@@ -244,7 +243,39 @@ func MakeVAS() *VerticaAutoscaler {
 	}
 }
 
+func MakeVASWithMetrics() *VerticaAutoscaler {
```
cchen-vertica, line 246: Add a comment for this function. Is this function only used for the test?

pkg/builder/builder.go
```diff
@@ -863,6 +864,26 @@ func makeScrutinizeInitContainers(vscr *v1beta1.VerticaScrutinize, vdb *vapi.Ver
 	return cnts
 }
 
+func BuildHorizontalPodAutoscaler(nm types.NamespacedName, vas *v1beta1.VerticaAutoscaler) *autoscalingv2.HorizontalPodAutoscaler {
```
cchen-vertica, line 867: Add a comment for this function.

pkg/controllers/vas/hpareconciler.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/vertica/vertica-kubernetes/pkg/builder"
+	"github.com/vertica/vertica-kubernetes/pkg/controllers"
+	"github.com/vertica/vertica-kubernetes/pkg/names"
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// RefreshSelectorReconciler is a reconciler to update the pod selector in the
```
cchen-vertica, line 34: Update the comment

pkg/controllers/vas/hpareconciler.go
```diff
@@ -0,0 +1,87 @@
...
+	Vas  *vapi.VerticaAutoscaler
+	Log  logr.Logger
+}
+
+func MakeHorizontalPodAutoscalerReconciler(v *VerticaAutoscalerReconciler, vas *vapi.VerticaAutoscaler,
+	log logr.Logger) controllers.ReconcileActor {
+	return &HorizontalPodAutoscalerReconciler{VRec: v, Vas: vas, Log: log.WithName("HorizontalPodAutoscalerReconciler")}
+}
+
+// Reconcile will handle updating the selector in the status portion of a VerticaAutoscaler
```
cchen-vertica, line 47: Where does the reconcile() update the status portion?

pkg/controllers/vas/hpareconciler_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/vertica/vertica-kubernetes/pkg/v1beta1_test"
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var _ = Describe("subclusterresize_reconcile", func() {
+	ctx := context.Background()
+
+	It("should requeue if VerticaDB doesn't exist", func() {
```
cchen-vertica, line 34: Update the description of this test.

tests/e2e-leg-12/create-hpa/95-errors.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: Service
+metadata:
+  labels:
+    app.kubernetes.io/managed-by: verticadb-operator
+---
+apiVersion: vertica.com/v1
+kind: VerticaDB
+---
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
```
cchen-vertica, line 27: We could add one more step: delete the CR, and then verify HPA is gone.

api/v1beta1/verticaautoscaler_types.go
roypaulin: Yes

api/v1beta1/verticaautoscaler_types.go
roypaulin: Because that is how it is in the hpa. I decided to keep the same types as in the hpa.

pkg/controllers/vas/hpareconciler.go
roypaulin: That' a copy/paste error. I will update the comment

api/v1/version.go
roypaulin: To be honest I was not sure what must be the min version. Technically you can use the new field to set non prometheus metrics like cpu/mem and it would work. But it is not going to work with prometheus metrics so I will pit 24.1.0 instead.

cchen-vertica (approved):

HaoYang0000 (approved):

PR #1021, opened 2025-01-08
Define prometheus metrics to k8s

HaoYang0000:


prometheus/adapter.yaml
```diff
@@ -0,0 +1,73 @@
...
+      resources:
+        overrides:
+          namespace:
+            resource: namespace
+          pod:
+            resource: pod
+      # name:
+      #   matches: "^vertica_sessions_running_counter$"
+      #   as: "vertica_sessions_running_counter"
+      metricsQuery: 'sum(increase(vertica_sessions_running_counter[60m])) by (namespace, pod)'
```
roypaulin, line 59: Using `increase` on gauges is generally not a good idea.

prometheus/adapter.yaml
```diff
@@ -0,0 +1,73 @@
...
+      resources:
+        overrides:
+          namespace:
+            resource: namespace
+          pod:
+            resource: pod
+      # name:
+      #   matches: "^vertica_queued_requests_total$"
+      #   as: "vertica_queued_requests_total"
+      metricsQuery: 'sum(increase(vertica_queued_requests_total[60m])) by (namespace, pod)'
```
roypaulin, line 73: why `60m`?

prometheus/adapter.yaml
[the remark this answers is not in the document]
HaoYang0000: Yeah, I ran the Vertica API to get the value and compared it with the Prometheus API result. For example, percentage value, like CPU/memory usage in an average of time, it gives for example 10174m means 10.174% per hour in average, using the avg_over_time function. I left the example in the code.

Ref: https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/docs/walkthrough.md#quantity-values

HaoYang0000: This is a draft so I am open to suggestions. I think this is a counter, and it gives an integer value back. According to the doc: "increase should only be used with counters and native histograms where the components behave like counters."
Ref: https://prometheus.io/docs/prometheus/latest/querying/functions/#increase

prometheus/adapter.yaml
HaoYang0000: Based on the spike doc: https://confluence.verticacorp.com/display/~cchen/Spike+for+Identifying+Useful+Vertica+Metrics

tests/e2e-leg-11/prometheus-sanity/30-assert.yaml
qindotguan, line 1: We don't need an assert file for step 30-verify-prometheus-adapter.yaml. The script is just a query to the endpoint so no status change, and the scripts is executed in a TestStep and no test pods were created.

HaoYang0000: Got it, I merged main but the file still here, I removed manually in: https://github.com/vertica/vertica-kubernetes/pull/1021/commits/d2b35d79331ec7db1c0944181dcc92b043aa9e08

roypaulin (approved):

LiboYu2 (changes requested):
This is an error from a local run of the test case:
    logger.go:42: 16:50:45 | prometheus-sanity/10-deploy-prometheus | Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "prometheus-kube-prometheus-operator" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "kuttl-test-talented-mongrel": current value is "prometheus"
    logger.go:42: 16:50:45 | prometheus-sanity/10-deploy-prometheus | make: *** [Makefile:667: deploy-prometheus] Error 1
    case.go:399: failed in step 10-deploy-prometheus
    case.go:401: command "cd ../../.. && make deploy-prometheus PROMETHEUS_NAMESPACE=$NAMESPACE" failed, exit status 2

roypaulin:
A thing I think would be interesting is to show an example of VerticaAutoscaler where the metrics are set, so it can be a reference, and we know how to properly use it. It can be a file in prometheus folder that contains an example for each.

LiboYu2:
> A thing I think would be interesting is to show an example of VerticaAutoscaler where the metrics are set, so it can be a reference, and we know how to properly use it. It can be a file in prometheus folder that contains an example for each.

There is one in config./sample. We can add another one for custom metric.

HaoYang0000:
> This is an error from a local run of the test case: logger.go:42: 16:50:45 | prometheus-sanity/10-deploy-prometheus | Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "prometheus-kube-prometheus-operator" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "kuttl-test-talented-mongrel": current value is "prometheus" logger.go:42: 16:50:45 | prometheus-sanity/10-deploy-prometheus | make: *** [Makefile:667: deploy-prometheus] Error 1 case.go:399: failed in step 10-deploy-prometheus case.go:401: command "cd ../../.. && make deploy-prometheus PROMETHEUS_NAMESPACE=$NAMESPACE" failed, exit status 2
> 
> We should not create extra namespace. All test related resources must be installed in kuttl namespace and when the test is finished the namespace will be deleted. It may not be possible to use a single namespace. We can create extra ones but with names that will not easily duplicate any existing ones. After the test case is ready, the extra namespaces can be deleted.

The current Prometheus integration test used the kuttl namespace, we didn't use extra namespace there.
This error happens because you installed Prometheus in your local env, and kuttl shares the same env which Prometheus has already been installed. To run the test, you will need to undeploy Prometheus in your local env. 
Sometimes the test will fail on the local env and the Prometheus resources will be leftover, you can use the following cmd as reference to clean up the resources:
kubectl delete clusterrole  prometheus-kube-prometheus-operator
kubectl delete clusterrole  prometheus-kube-prometheus-prometheus
kubectl delete clusterrolebinding prometheus-kube-prometheus-operator
kubectl delete clusterrolebinding prometheus-kube-prometheus-prometheus
kubectl delete svc prometheus-kube-prometheus-kube-proxy -n kube-system 
kubectl delete svc prometheus-kube-prometheus-kubelet -n kube-system
kubectl delete MutatingWebhookConfiguration prometheus-kube-prometheus-admission
kubectl delete ValidatingWebhookConfiguration prometheus-kube-prometheus-admission
kubectl delete clusterrole  prometheus-kube-prometheus-prometheus
kubectl delete clusterrolebinding prometheus-kube-prometheus-operator
kubectl delete svc prometheus-kube-prometheus-kubelet -n kube-system
kubectl delete ValidatingWebhookConfiguration prometheus-kube-prometheus-admission

HaoYang0000:
@roypaulin @LiboYu2 I added an example for the autoscaler with custom metrics in another PR: https://github.com/vertica/vertica-kubernetes/pull/1033

PR #1036, opened 2025-01-16
Add metric vertica_queued_requests_total

qindotguan:
This is the first step to enable the metric vertica_queued_requests_total. Some future work would include adding new test cases to test together with hpa.


roypaulin:
Can you show how you use it in the hpa?

roypaulin (approved):

PR #1042, opened 2025-01-27
Support autoscaling with KEDA

roypaulin:
KEDA (Kubernetes-based Event Driven Autoscaling) is an open-source that allows for fine-grained autoscaling (including to/from zero) for event driven Kubernetes workloads. This PR integrates KEDA's ScaledObject to the VerticaAutoscaler CR. Now users can pick which autoscaler to use (scaledObject/hpa). 
API changes were made now the customAutoscaler field has 3 subfields: type, hpa and scaledObject. When type is "HPA", hpa must be set with the metric's definition. The metric can either be cpu/mem or a prometheus metric.
A new test case was added to scale based on a scaledObject.


kuttl-test.yaml
```diff
@@ -80,3 +80,5 @@ commands:
   - script: if [ "$NEED_PROMETHEUS" = "true" ]; then make deploy-prometheus; fi
   - script: make undeploy-prometheus-adapter || true
   - script: if [ "$NEED_PROMETHEUS" = "true" ]; then make deploy-prometheus-adapter; fi
+  - script: make undeploy-keda || true
+  - script: if [ "$NEED_PROMETHEUS" = "true" ]; then make deploy-keda; fi
```
cchen-vertica, line 84: Does all tests in e2e-leg-12 need keda?

cchen-vertica:
> KEDA (Kubernetes-based Event Driven Autoscaling) is an open-source that allows for fine-grained autoscaling (including to/from zero) for event driven Kubernetes workloads. This PR integrates KEDA's ScaledObject to the VerticaAutoscaler CR. Now users can pick which autoscaler to use (scaledObject/hpa). API changes were made now the customAutoscaler field has 3 subfields: type, hpa and scaledObject. When type is "HPA", scaledObject must be set with the metric's definition. The metric can either be cpu/mem or a prometheus metric. Anew test case was added to scale based on a scaledObject.

Is "When type is "HPA", scaledObject must be set with the metric's definition" true a typo? I think they are mutual exclusive.

The type can only be "HPA“ or "scaledObject", right? HPA means no-keda, and scaledObject means keda, right?

roypaulin:
> Is "When type is "HPA", scaledObject must be set with the metric's definition" true a typo? I think they are mutual exclusive.
> 
> The type can only be "HPA“ or "scaledObject", right? HPA means no-keda, and scaledObject means keda, right?

Yes that's a typo.

kuttl-test.yaml
roypaulin: No, but we will be adding many tests and keda operator is cluster-scope so one instance will manage multiple namespaces.

roypaulin:
We need to make some server changes to support scaling entire db with keda. Prometheus queries on keda do not rely on the selector but use the labels to find the target pods. We do not have a label that can regroup all pods that belong to the same cluster. I will create a follow-up ticket.

prometheus_metrics_latest.json
```diff
@@ -43,10 +43,20 @@
             "type": "counter"
         },
         {
-            "description": "A summary of the pause duration of garbage collection cycles.",
+            "description": "A summary of the wall-time pause (stop-the-world) duration in garbage collection cycles.",
```
cchen-vertica, line 46: How the changes in this file is generated? The prometheus metrics should be the same as before, right?

cchen-vertica (approved):
Looks good. Leave some minor comments. You can fix them in a follow-up jira if you want.

cchen-vertica:
> We need to make some server changes to support scaling entire db with keda. Prometheus queries on keda do not rely on the selector but use the labels to find the target pods. We do not have a label that can regroup all pods that belong to the same cluster. I will create a follow-up ticket.

We need another follow-up jira to change the term "scale up" to "scale out", and "scale down" to "scale in" in K8s repo.

prometheus_metrics_latest.json
roypaulin: These are prometheus go built-in metrics, so the change with the library version.

PR #1045, opened 2025-01-29
Add routing label to sandbox pods

roypaulin:
When a sandbox pod is deleted, the operator does not put back the routing label. This adds the ClientRoutingLabelReconciler in the sandbox controller. It also adds an annotation in the configMap to sometimes disable routing, like in the case of online upgrade where routing to the sandbox must not be allowed until replication is done.


cchen-vertica (commented):
Looks good. Only one concern: if we enable client-proxy and restart a client-proxy pod of a sandbox, the client-proxy pod should still have the routing label. Your code seems covered this case, but we don't have an e2e test for it. You can leave a comment in your jira to let QA test this case.

cchen-vertica (approved):

PR #1051, opened 2025-02-03
Support scaling the entire database

roypaulin:
So far, we have been scaling based on a service name. Now we can consider the entire database for scaling.
Now if serviceName is empty all the pods will be selected, and when scaling is needed all the subclusters will be considered.


pkg/meta/labels.go
```diff
@@ -135,6 +135,11 @@ const (
 
 	// This indicates the name of the client proxy deployment.
 	ClientProxyNameLabel = "vertica.com/client-proxy-name"
+
+	// Indicates if a pod belongs to the main cluster. It is mainly use as a selector in
+	// VAS to filter out sandbox pods.
+	IsSandboxLabel = "vertica.com/is-sandbox"
+	IsSandboxFalse = "false"
```
HaoYang0000, line 142: Is this label only used by VAAS? If we do sandboxing for a subcluster, should we change this flag to true?

roypaulin: You are correct, we need to change this to true for sandboxes.

tests/e2e-leg-1/autoscale-db-by-subcluster/50-remove-last-scaled-subcluster.yaml
```diff
@@ -0,0 +1,17 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: kubectl scale -n $NAMESPACE verticaautoscaler/v-autoscale-db --replicas 0
```
cchen-vertica, line 17: Since ksafety is not 0, the primary subcluster will not be removed even we set "--replicas 0", right? If we don't violate ksafety check, will scale-down remove the existing subclusters? For example, we have pri1 and sec1 which are created manually by the customers and we set "--replicas 0", a. if ksafety is 0, will both subclusters be removed? b. if ksafety is not 0, will sec1 be removed?

changes/unreleased/Fixed-20250203-220428.yaml
```diff
@@ -0,0 +1,5 @@
+kind: Fixed
+body: Routing traffic to a sandbox pod after restart
```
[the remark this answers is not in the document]
roypaulin: Yes. I forgot to add a changie entry.

tests/e2e-leg-1/autoscale-db-by-subcluster/30-assert.yaml
```diff
@@ -0,0 +1,39 @@
...
+kind: StatefulSet
+metadata:
+  name: v-autoscale-db-by-subcluster-as-1
+status:
+  replicas: 2
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: v-autoscale-db-by-subcluster-as
```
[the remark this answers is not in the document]
roypaulin: Only if you set the serviceName in the template. If you don't each subcluster will have its own service whose name is the SC name.

tests/e2e-leg-1/autoscale-db-by-subcluster/50-remove-last-scaled-subcluster.yaml
roypaulin: First of all, `kubectl scale` does manual scaling, it is just for testing, to trigger the scaling logic.
That being said, if target size is lower than currentsize, that yes we will try to remove sec1 (providing -1*(target-curentSize) is <= sec1 size).
scaling to zero will not be possible because it will violates ksafety and minReplicas min value is 1.

pkg/controllers/vas/scaledown_reconciler.go
```diff
@@ -78,7 +78,7 @@ func (s *ScaledownReconciler) Reconcile(ctx context.Context, req *ctrl.Request)
 			s.Log.Info("Metric's value is lower than the scale-down threshold.", "metric", mStatus.name)
 			newMinReplicas = *s.Vas.Spec.CustomAutoscaler.MinReplicas
 		} else {
-			newMinReplicas = s.Vas.Spec.TargetSize
+			newMinReplicas = s.Vas.Status.CurrentSize
```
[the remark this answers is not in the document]
roypaulin: currentSize represents the number of existing pods/replicas, targetSize is the desired state. In this case we want minReplicas to be equal to the current number of replicas so we do not trigger scale down. currentSize is more accurate.

cchen-vertica (approved):

PR #1056, opened 2025-02-06
Jc4244441 Replace password with certificate 

LiboYu2:
In my previously merged PR, support for using env var to pass secret name is added for NMA.
Here are the major new changes added:
1 The operator will use password to create db, execute DDLs to configure TLS.
   Only when DDLs are executed properly will the operator starts to use certificate.
2 The decision about whether to use password or certificate is based on the following factors:
    -     vdb version
    -     tls annotation
    -     whether DB is initialized
 3 I realize it is hard to let functions in vadmin/ package to share the data as it is a 
    multithreading env. To make it easier to set/pass a flag, I added a new data structure. 
    VdbContext. It is a thread safe map that supports multiple databases. We can use it 
    to store bool or string values in it. This data structure will also be used when we rotate 
    certificate.

Use of "UseTLSCert" flag
This vdb-scoped global flag indicates whether we should use TLS instead of password.
It will be set up before a reconcile cycle starts, based on 
    -     vdb version
    -     tls annotation
    -     whether DB is initialized
We use it as it makes it a lot easier to pass parameter from other packages to vadmin/


roypaulin:
We need to change our strategy. PR has been open for more than one month. Now that we know it won't be part of next release, we can take our time and do it bit by bit. We don't need to pass all the tests to merge it.
Let's wrap up and so what we are going to do in the last 2 days of this sprint:
- Set back the knob default value to false (turn certs off)
- In e2e.yml we are going to add some new legs: e2e-leg*-vcluster-certs that will run with certs. For now we can just pick one test suite e2e-leg-1 (very simple). We will progressively test new legs with certs (while still keeping the corresponding test with password), only when the feature is stable, and we have successfully run all tests with certs, are we going to set the knob to true by default (or outright remove it).
- I will start reviewing your PR
- We will create more tickets to try to fix other operations (like revivedb, etc...) and will gradually enable more tests with certs. This will allow more devs to work on this in parallel (we need more tickets for next sprint)

@LiboYu2 You have been working on this for more than a month and I think it will do you good to also work on something else.
We cannot fix everything in a 3-point story. Add a comment about any issues you found in the ticket in ValueEdge.

@cchen-vertica @ningdeng let me know what you think.

pkg/vadmin/vc.go
```diff
@@ -73,6 +79,24 @@ func retrieveNMACerts(ctx context.Context, fetcher *cloud.SecretFetcher, vdb *va
...
+		secretName = meta.GetNMATLSSecretName(vdb.Annotations)
+	} else {
+		secretName = vdb.Spec.NMATLSSecret
+	}
+	if secretName == "" {
+		return "", fmt.Errorf("failed to retrieve nma secret name")
+	}
+	return secretName, nil
+}
+
```
roypaulin, line 99: Let's not deal with cert rotation here, there is a separate ticket for that.

api/v1/helpers.go
```diff
@@ -94,6 +95,18 @@ func (v *VerticaDB) FindTransientSubcluster() *Subcluster {
...
+	return MakeVersionedVDB(LegacyVersion)
+}
+
+// MakeVDB is a helper that constructs a fully formed VerticaDB struct using the sample name.
+// This is intended for test purposes.
+func MakeMTLSVDB() *VerticaDB {
+	return MakeVersionedVDB(NMATLSCertRotationMinVersion)
+}
+
+// MakeVDB is a helper that constructs a fully formed VerticaDB struct using the sample name.
```
cchen-vertica, line 107: Change this function description.

pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -51,7 +52,7 @@ const (
 	// This is a file that we run with the create_db to run custome SQL. This is
 	// passed with the --sql parameter when running create_db. This is no longer
 	// used starting with versions defined in vapi.DBSetupConfigParameters.
-	PostDBCreateSQLFile = "/home/dbadmin/post-db-create.sql"
+	PostDBCreateSQLFile = "/tmp/post-db-create.sql"
```
cchen-vertica, line 55: Why are we changing this path?

pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -170,33 +194,66 @@ func (c *CreateDBReconciler) preCmdSetup(ctx context.Context, initiatorPod types
 func (c *CreateDBReconciler) generatePostDBCreateSQL(ctx context.Context, initiatorPod types.NamespacedName) (ctrl.Result, error) {
 	// On newer server versions we moved over the SQL to config parameters. So,
```
cchen-vertica, line 195: Update the description here.

api/v1/version.go
```diff
@@ -74,6 +74,8 @@ const (
 	// the minimum versions of the NMA that we support them.
 	NMATLSSecretInGSMMinVersion               = "v24.1.0"
 	NMATLSSecretInAWSSecretsManagerMinVersion = "v24.2.0"
+	// The NMA TLS certificate can be rotated
+	NMATLSCertRotationMinVersion = "v25.2.0"
```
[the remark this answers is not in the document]
cchen-vertica: We also need to update v25.2.0 to v25.3.0

pkg/vadmin/add_node_vc.go
```diff
@@ -77,11 +77,17 @@ func (v *VClusterOps) genAddNodeOptions(s *addnode.Parms, certs *HTTPSCerts) vop
 	}
 
 	// auth options
-	opts.Key = certs.Key
-	opts.Cert = certs.Cert
-	opts.CaCert = certs.CaCert
-	opts.UserName = v.VDB.GetVerticaUser()
-	opts.Password = &v.Password
+	if v.shouldUseCertAuthentication() {
```
cchen-vertica, line 80: line 80-90 can be wrapped up in a helper function. Using "DatabaseOptions" to do that.

tests/e2e-leg-4/revivedb-1-node/15-assert.yaml
```diff
@@ -19,12 +19,11 @@ status:
   replicas: 1
   readyReplicas: 1
 ---
-apiVersion: vertica.com/v1beta1
+apiVersion: vertica.com/v1
```
cchen-vertica, line 22: You can add a new test suite for revivedb with customized tls cert. We still want to verify old revivedb behavior with v1beta1 version.

cchen-vertica (commented):
@roypaulin We can discuss about Libo's VDB context implementation tomorrow. I prefer using annotations to store the info rather than using cache.

pkg/controllers/vdb/createdb_reconciler.go
roypaulin: It should say the same for admintools at least.

roypaulin:
> @roypaulin We can discuss about Libo's VDB context implementation tomorrow. I prefer using annotations to store the info rather than using cache.

Yes, let's do that. It goes hand in hand with my design on rollback after failure. Also caching using shared state is not a best practice.

pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -170,33 +194,66 @@ func (c *CreateDBReconciler) preCmdSetup(ctx context.Context, initiatorPod types
...
+			\"namespace\":\"%s\"}';`,
+			c.Vdb.Spec.NMATLSSecret, corev1.TLSPrivateKeyKey, c.Vdb.ObjectMeta.Namespace))
+
+		sb.WriteString(fmt.Sprintf(
+			`CREATE CA CERTIFICATE https_ca_cert_0 SECRETMANAGER KubernetesSecretManager SECRETNAME '%s' CONFIGURATION '{\"data-key\":\"%s\", 
+			\"namespace\":\"%s\"}';`,
+			c.Vdb.Spec.NMATLSSecret, paths.HTTPServerCACrtName, c.Vdb.ObjectMeta.Namespace))
+
+		sb.WriteString(fmt.Sprintf(
+			`CREATE CERTIFICATE https_cert_0 SECRETMANAGER KubernetesSecretManager SECRETNAME '%s' CONFIGURATION '{\"data-key\":\"%s\",
```
roypaulin, line 244: We should make this a function to cover other secret managers.

tests/e2e-leg-4/revivedb-1-node/15-assert.yaml
LiboYu2: I fixed the issues with the original test cases. I will keep the old cases and create new ones.

roypaulin:
@LiboYu2 can you address my comments above?

cchen-vertica:
@LiboYu2 You can change the target branch to "25.3-mutual-tls"

ningdeng:
@LiboYu2 , @roypaulin , @cchen-vertica and me had a meeting for moving this PR forward, we've decided to merge this PR to master by Monday (March 24, 2025) with the feature flag turned off, such that 
1. this will not affect the e2e tests. 
2. QA can use the daily build to provide early feedback of this feature with the feature flag turned on. 

The follow up work of this PR would be that:
1. K8s team assign 1 or 2 engineers to debug the e2e failures with Libo, this can speed up the work.
2. Any tasks that were blocked by this PR can be moved forward.

pkg/meta/annotations.go
```diff
@@ -207,6 +207,8 @@ const (
 	HTTPSTLSConfGenerationAnnotationFalse = "false"
 	HTTPSTLSConfGenerationDefaultValue    = true
 
+	NMATLSSECRETAnnotation          = "vertica.com/nma-tls-secret"           // #nosec G101
+	CLIENTSERVERTLSSecretAnnotation = "vertica.com/client-server-tls-secret" // #nosec G101
```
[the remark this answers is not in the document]
LiboYu2: CLIENTSERVERTLSSecretAnnotation will be removed as it is not required right now.
NMATLSSECRETAnnotation is still required. Even though we don't support cert rotation, users can specify the secret name by themselves. If they specify a wrong/invalid secret name, we don't want to interrupt our operator. Therefore, after the db is successfully created, the secret name is saved to this annotation and all operator code will get secret name from this annatation.

pkg/vadmin/vdb_context.go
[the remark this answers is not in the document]
LiboYu2: This vdb_context.go servers two purposes.
1.  As you said, Share mutable state. However, it must be thread safe. Previously we did not have any data structure to support that. I tried to use a flag to control the TLS configuration and then I realized this issue. 
2.  Thread safe in memory cache. Currently all _vc.go in vadmin/ will read the secret from Kubernetes. It is static content and we have to make a rest call to read it every time. It is slow and waste of bandwidth and cpu.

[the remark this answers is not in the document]
roypaulin: Using share mutable state in k8s go operator is strongly discouraged. If we ever decide to use caching it would a local cache. We can worry about it later. For now, just remove it.

pkg/vadmin/vc.go
LiboYu2: copied from above:
NMATLSSECRETAnnotation is still required. Even though we don't support cert rotation, users can specify the secret name by themselves. If they specify a wrong/invalid secret name, we don't want to interrupt our operator. Therefore, after the db is successfully created and DDLs are run properly, the secret name will be saved to this annotation and all operator code will get secret name from this annatation.

pkg/meta/annotations.go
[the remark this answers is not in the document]
roypaulin: The question is: is this needed in this PR? I think it is not. The scenario you mentioned will come up during cert rotation. This PR is only for password replacement, so I do not want anything extra to be included here.

roypaulin:
@LiboYu2 let's just keep what is related to replacing password. Let's not deal with anything related with cert rotation here. There is a separate ticket for cert rotation.

pkg/vadmin/vdb_context.go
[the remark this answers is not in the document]
roypaulin: I do understand your concern, for remote secret manager (no need to cache on k8s, the secret is in the same namespace), and we can have a ticket to add one, but it will not be this way. Let's remove vdb_context.go.

pkg/meta/annotations.go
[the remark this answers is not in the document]
LiboYu2: Assume this scenario: 
A user runs "k edit vdb" and changes "NMATLSSecret" to some non-existent secret name.
If our code reads the secret name from Spec.NMATLSSecret, then the operator will stop functioning.

pkg/vadmin/vdb_context.go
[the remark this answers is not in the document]
LiboYu2: I don't agree. If you think we don't need cache and performance, that is fine.
We still need it for sharing the the TLS flag. All those annotations added to vdb are shared mutable variables. They
are not thread safe. They are slow. You have to make rest calls to read the latest value or make changes to it.
There is nothing wrong with mutable variable. The issue is whether you can use it properly or not.

pkg/vadmin/vc.go
```diff
@@ -21,34 +21,40 @@ import (
...
 		EVWriter: v.EVWriter,
 	}
-	return retrieveNMACerts(ctx, &fetcher, v.VDB)
+	return retrieveNMACerts(ctx, fetcher, v.TargetVDB)
 }
 
-// retrieveTargetNMACerts will retrieve the certs from NMATLSSecret for calling target NMA endpoints
-func (v *VClusterOps) retrieveTargetNMACerts(ctx context.Context) (*HTTPSCerts, error) {
+// retrieveNMACerts will retrieve the certs from NMATLSSecret for calling NMA endpoints
+func (v *VClusterOps) retrieveNMACerts(_ context.Context) (*HTTPSCerts, error) {
```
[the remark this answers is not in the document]
LiboYu2: The two fucntions' positions are switched in the file. I've put them back.

pkg/vadmin/add_node_vc.go
LiboYu2: DatabaseOptions is defined in vclusterops and I will have to create another PR in server repo in order to change it.
Instead, I will add it to vadmin/vc.go. All the users of this function are in vadmin/ package. Another advantage is that HTTPSCerts is also defined in vadmin/. We can always move it later if you want.

pkg/meta/annotations.go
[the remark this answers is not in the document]
roypaulin: 1. If they pass a wrong secret when creating vdb, the operator will throw an error and that is fine. We are suposed to let the user know their secret does not exist so they can fixit.
2. If vdb already exist and they are changing spec.NMATLSSecret, that is already in the realm of cert rotation and we do not deal with cert roration in this PR.

I am not saying we don't have to cache the secret name, I am saying we don't have to do it in this PR.
So please, remove it for now.

pkg/vadmin/vdb_context.go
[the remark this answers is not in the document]
roypaulin: First, they are not shared mutable variables, the k8s API server is the single source of truth, second this is how the framework is designed, reconciliation loops would have their own copy of the CR.

Again it is not a best practice in k8s go operator. Please remove it.

pkg/vadmin/vc.go
LiboYu2: I will repeat for the last time:
This has nothing to do with certificate rotation.

roypaulin:
Resolve the conflicts so the tests can run.

pkg/controllers/vdb/createdb_reconciler.go
roypaulin: We don't have to change this for older versions that use atmintools

pkg/vadmin/vc.go
roypaulin: Changing the secret litteraly means you want to do cert rotation. There's no other reason.
spec.nnaTLSSecret should always be the source of truth, the annotation will be used for rollback(update spec.nmaTLSSecret to its previous value).
But all of this is irrelevant in this PR.
We only need the DDLs, to pass the certs without the password (if usecert is enabled).

.github/workflows/build-images.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       uses: spilchen/switch-case-action@v2
       id: full_vertica_image
       with:
-        default: docker.io/opentext/vertica-k8s-private:latest-test-master
+        default: lyu2584/vertica_k8s_test_repo_public:20250307
```
roypaulin, line 83: @cchen-vertica can we already change this back?

pkg/vadmin/vdb_context.go
roypaulin, line 1: Let's remove vdb_context. Let's avoid using shared state even with mutex. Each reconciliation must stay independent.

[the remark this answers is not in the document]
LiboYu2: The vdbcontext is merely a thread safe map. 

Why do we need it?

1 The dozens of _vc.go in vadmin/ package cannot hold a reference of vdb and any other data types from other packages, as that will introduce a dependency cycle.

2 We can simply define a variable at package level and use it. As the operator will work on multiple vdb in different threads, it will be shared not in a safe way.

3 To pass any information from other packages to _vc.go files in vadmin/, we have to change function parameters for all the functions in the calling stack of them. That is definitely not a good idea.

The vdbcontext allows those _vc.go files to easily access shared information. 

If you look at the changes in file vertica_controller.go, you will see the TLS flag is set based on static information from the vdb. It is just simple caching. 

If you look at those _vc.go files, all of them will read the secret from the Kubernetes directly. Each read is a rest call. It makes no sense to make so many rest calls to read the same piece of information. It slows the operator down.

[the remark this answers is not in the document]
LiboYu2: I pushed one more commit. Now the TLS flag is immutable after being set.

roypaulin:
@LiboYu2 Resolve the conflicts so the tests can run.

pkg/controllers/vdb/tlsservercertgen_reconciler.go
```diff
@@ -127,10 +124,9 @@ func (h *TLSServerCertGenReconciler) createSecret(secretFieldName, secretName st
 	caCert security.Certificate) (*corev1.Secret, error) {
 	secret := corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace:       h.Vdb.Namespace,
-			Annotations:     builder.MakeAnnotationsForObject(h.Vdb),
-			Labels:          builder.MakeCommonLabels(h.Vdb, nil, false, false),
-			OwnerReferences: []metav1.OwnerReference{h.Vdb.GenerateOwnerReference()},
```
roypaulin, line 133: Put this back.

roypaulin:
I am merging this. We will do this incrementally. 
There is probably a server change that broke e2e-leg-3-admintools and e2e-leg-4-admintools-vdb-gen, so I set the legacy image to the 20250320 build (last successful) until the issue is fixed.

PR #1072, opened 2025-02-17
Add back nma certs mount

roypaulin:
In a previous PR, `mount-nma-certs` was removed to support getting the certs from secret. 
That caused a few issues:
- when coming from an older operator version with a vdb that had `nma-certs-mount` true, the pods are crashing
- will break scrutinize as it still supports nma certs mount
Until these issues are fixed, we will still use `mount-nma-certs` to allow a smooth transition from an older operator version. Moreover, I added another annotation `enable-nma-tls-rotation` that will temporarily served as feature flag until we are sure nma certs rotation will not break the existing code.

### DIFF
--- a/pkg/controllers/vdb/tlsservercertgen_reconciler.go
+++ b/pkg/controllers/vdb/tlsservercertgen_reconciler.go
@@ -23,6 +23,7 @@ import (
 	vapi "github.com/vertica/vertica-kubernetes/api/v1"
 	"github.com/vertica/vertica-kubernetes/pkg/builder"
 	"github.com/vertica/vertica-kubernetes/pkg/controllers"
+	vmeta "github.com/vertica/vertica-kubernetes/pkg/meta"
 	"github.com/vertica/vertica-kubernetes/pkg/names"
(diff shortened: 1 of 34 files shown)


pkg/controllers/vdb/tlsservercertgen_reconciler.go
```diff
@@ -58,8 +59,10 @@ func MakeTLSServerCertGenReconciler(vdbrecon *VerticaDBReconciler, log logr.Logg
 // Reconcile will create a TLS secret for the http server if one is missing
 func (h *TLSServerCertGenReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
 	secretFieldNameMap := map[string]string{
-		ClientServerTLSSecret: h.Vdb.Spec.ClientServerTLSSecret,
```
LiboYu2, line 61: Could you add missing clientServerTLSSecret?

roypaulin: It is at line 65, behind the feature flag.

LiboYu2 (approved):

PR #1086, opened 2025-02-24
Fix vas svc name with underscore

HaoYang0000:
VerticaAutoscaler fails to add new nodes when name/ServiceName in spec.template contains underscore (e.g.sc_vas)

### DIFF
--- a/pkg/controllers/vas/subclusterscale_reconciler.go
+++ b/pkg/controllers/vas/subclusterscale_reconciler.go
@@ -156,9 +156,10 @@ func (s *SubclusterScaleReconciler) genNextSubclusterName(scMap map[string]*vapi
 	if baseName == "" {
 		baseName = s.Vas.Name
 	}
+	preferredName := v1beta1.GenCompatibleFQDNHelper(baseName)
 	i := 0
 	for {
-		name := fmt.Sprintf("%s-%d", baseName, i)
+		name := fmt.Sprintf("%s-%d", preferredName, i)
 		_, ok := scMap[name]
 		if !ok {
 			return name
--- a/pkg/controllers/vas/subclusterscale_reconciler_test.go
+++ b/pkg/controllers/vas/subclusterscale_reconciler_test.go
@@ -226,4 +226,38 @@ var _ = Describe("subclusterscale_reconcile", func() {
 		Expect(k8sClient.Get(ctx, vdbName, fetchVdb)).Should(Succeed())
 		Expect(len(fetchVdb.Spec.Subclusters)).Should(Equal(1))
 	})
+
+	It("should replace underscore with hyphen in service name", func() {
+		vdb := vapi.MakeVDB()
+		const serviceName = "vas_1"
+		const preferedServiceName = "vas-1"
+		vdb.Spec.Subclusters[0].ServiceName = serviceName
+		test.CreateVDB(ctx, k8sClient, vdb)
+		defer test.DeleteVDB(ctx, k8sClient, vdb)
+
+		vas := v1beta1.MakeVAS()
+		vas.Spec.ServiceName = serviceName
+		vas.Spec.ScalingGranularity = v1beta1.SubclusterScalingGranularity
+		vas.Spec.Template = v1beta1.Subcluster{
+			ServiceName: serviceName,
+			Name:        serviceName,
+			Size:        8,
+		}
+		vas.Spec.TargetSize = vas.Spec.Template.Size*2 + 3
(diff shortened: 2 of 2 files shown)


cchen-vertica (approved):

roypaulin (changes requested):
Shouldn't we add a wabhook rule to check the template? I think we should let the useer know there is wrong value in the template so they can fix it

roypaulin:
> Shouldn't we add a wabhook rule to check the template? I think we should let the user know there is wrong value in the template so they can fix it

The issue is not the sc name only but also the service name. I think webhook rules would be better as template is technically a subcluster so we can reuse some of the rules in vdb webhook.

HaoYang0000:
> Shouldn't we add a wabhook rule to check the template? I think we should let the useer know there is wrong value in the template so they can fix it

I see your point, and will add a webhook rule to check the subcluster template. Following are for clarification. 
We have two things we need to handle here, following the convention we are using from vdb: 
1. The name of subcluster: the value "sc_1" is a valid value for vdb, and what we do is replace '_' by '-'.
2. The name of the service: the value "sc_1" is not valid and will be rejected by webhook.

pkg/controllers/vas/subclusterscale_reconciler.go
```diff
@@ -155,9 +155,10 @@ func (s *SubclusterScaleReconciler) genNextSubclusterName(scMap map[string]*vapi
 	if baseName == "" {
 		baseName = s.Vas.Name
 	}
+	preferredName := vapi.GenCompatibleFQDNHelper(baseName)
```
cchen-vertica, line 158: Will this name be used as subcluster name in vertica as well? We want the subcluster name have underscore, and the service name has hyphen.

cchen-vertica: We need to modify one of keda e2e test to cover this case.

roypaulin: If we add webhook rules then we can remove changes here.

HaoYang0000: Sure, I will remove this change.

HaoYang0000: Changes revertd in https://github.com/vertica/vertica-kubernetes/pull/1086/commits/81c8146e7ff9acbfff960720dd1d3e6567967ee3 
Test added in https://github.com/vertica/vertica-kubernetes/pull/1086/commits/9d18feffe5c6da4e7a0e1a42534880f10e19275b

HaoYang0000:
Hey @roypaulin, please take another look, thanks

api/v1/verticaautoscaler_webhook.go
roypaulin, line 1: You should also check v.Spec.Template.Name (if non-empty) as it will be use as the base name for the statefulset

api/v1/verticaautoscaler_webhook.go
```diff
@@ -124,6 +124,15 @@ func (v *VerticaAutoscaler) validateScalingGranularity(allErrs field.ErrorList)
 // validateSubclusterTemplate will validate the subcluster template
 func (v *VerticaAutoscaler) validateSubclusterTemplate(allErrs field.ErrorList) field.ErrorList {
 	pathPrefix := field.NewPath("spec").Child("template")
+	// Validate each template service name
+	if v.Spec.Template.ServiceName != "" && !IsValidServiceName(v.Spec.Template.ServiceName) {
```
roypaulin, line 128:
```suggestion
	if v.CanUseTemplate() && v.Spec.Template.ServiceName != "" && !IsValidServiceName(v.Spec.Template.ServiceName) {
```

api/v1/verticaautoscaler_webhook.go
HaoYang0000: But we allow subcluster name to have under score, right? In the vdb the subcluster name sc_1 is valid.

roypaulin: underscore is not the only think we check. Something like`--sc_1` is not a valid name. Look at verticadb_webhook.go

HaoYang0000: The name sc_1 is a valid name, because in vdb spec we will convert it to sc-1 during creating the sts. I added logic to validate the subcluster name in VAS as well, similar to what we do for vdb. Changes added in https://github.com/vertica/vertica-kubernetes/pull/1086/commits/c210f21494f6087cefb2f55585939b3f36b604cc

PR #1071, opened 2025-02-17
add new webhook for autoscaler

HaoYang0000:
- scaling granularity must be subcluster if service name is empty
- When spec.customAutoscaler.hpa.scaledownThreshold is set, scaledown stabilisation window must be 0
- spec.customAutoscaler.scaledObject.metrics[].type must be one of "prometheus", "cpu", "memory" or empty
- if spec.customAutoscaler.scaledObject.metrics[].type is: 
  1) "prometheus", spec.customAutoscaler.scaledObject.metrics[].metricType must be one of "Value", "AverageValue".
  2) "cpu" or "memory" spec.customAutoscaler.scaledObject.metrics[].metricType must be one of "Utilization" or "AverageValue"
- customAutoscaler.type must be one of "HPA" or "ScaledObject " or empty string

PR #1091, opened 2025-02-25
webhook rule custom autoscaler

qindotguan:
Add webhook rules for custom autoscaler as continued work of #1071:
- maxReplicas cannot be less than minReplicas
- maxReplicas must be set
- cannot set customAutoscaler after CR creation
- 2 metrics cannot have the same name
- if spec.customAutoscaler.scaledObject.metrics[].type is "prometheus", spec.customAutoscaler.scaledObject.metrics[].prometheus must not be nil, if it is "cpu" or "memory", spec.customAutoscaler.scaledObject.metrics[].resource must not be nil
- scaledownThreshold must be of the same type as the threshold used for scaleup.
- customAutoscaler.hpa must be non-nil if customAutoscaler.type == "HPA" / - customAutoscaler.scaledObject must be non-nil if customAutoscaler.type == "ScaledObject" or empty


roypaulin:
The history is not clean

qindotguan:
Tracking in #1097.

PR #1114, opened 2025-03-10
Add types in status.subclusters

qindotguan:
This PR is the Task 1 to support subcluster promotion/demotion: adds types in status.subclusters. 

status:
  subclusters:
  - name: sc1
    type: primary/secondary/sandboxprimary/sandboxsecondary (default as empty if it's not in the db yet)

Following tasks will be:
Task 2. add new types in spec.sandboxes.subclusters.type. For example:
Task 3. Remove the updates in spec.subclusters.subcluster.type

### DIFF
--- a/api/v1/verticadb_types.go
+++ b/api/v1/verticadb_types.go
@@ -1031,6 +1031,12 @@ type SubclusterStatus struct {
 	Oid string `json:"oid"`
 
 	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// +optional
+	// The type of the subcluster. It could be primary, secondary, transient, or empty if
+	// it's not in the db yet.
+	Type string `json:"type"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=status
 	// A count of the number of pods that have been added to the database for this subcluster.
 	AddedToDBCount int32 `json:"addedToDBCount"`
 
--- a/pkg/controllers/vdb/suite_test.go
+++ b/pkg/controllers/vdb/suite_test.go
@@ -126,6 +126,7 @@ func defaultPodFactOverrider(_ context.Context, _ *vapi.VerticaDB, pf *podfacts.
 	pf.SetStartupInProgress(false)
 	pf.SetUpNode(true)
 	pf.SetSubclusterOid("123456")
+	pf.SetSubclusterType("Primary")
 	pf.SetShardSubscriptions(1)
 	return nil
 }
--- a/pkg/podfacts/podfacts.go
+++ b/pkg/podfacts/podfacts.go
@@ -63,6 +63,9 @@ type PodFact struct {
 	// The oid of the subcluster the pod is part of
 	subclusterOid string
 
+	// The type of the subcluster the pod is part of
+	subclusterType string
+
 	// The sandbox a pod is part of
 	sandbox string
 
@@ -260,6 +263,7 @@ func (p *PodFacts) ConstructsDetail(subclusters []vapi.Subcluster, upNodes []uin
 			pf := PodFact{
 				name:           types.NamespacedName{Name: fmt.Sprintf("%s-%d", sc.Name, j)},
 				subclusterName: sc.Name,
+				subclusterType: sc.Type,
 				isPrimary:      sc.IsPrimary(),
 				shutdown:       sc.Shutdown,
 				upNode:         isUp,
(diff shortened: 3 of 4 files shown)


api/v1/verticadb_types.go
```diff
@@ -1030,6 +1030,12 @@ type SubclusterStatus struct {
 	// Object ID of the subcluster.
 	Oid string `json:"oid"`
 
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// +optional
+	// The type of the subcluster. It could be primary, secondary, transient, or empty if
+	// it's not in the db yet.
+	Type string `json:"type"`
```
HaoYang0000, line 1037: Can we make the string a type:
type SubclusterType string

And for the type string we can do:
PrimarySubcluster SubclusterType  = "primary"

Its a minor thing but it will make our code more consistent.

pkg/controllers/vdb/suite_test.go
```diff
@@ -126,6 +126,7 @@ func defaultPodFactOverrider(_ context.Context, _ *vapi.VerticaDB, pf *podfacts.
 	pf.SetStartupInProgress(false)
 	pf.SetUpNode(true)
 	pf.SetSubclusterOid("123456")
+	pf.SetSubclusterType("Primary")
```
HaoYang0000, line 129: Use the constant in verticadb_type.go: PrimarySubcluster

pkg/podfacts/podfacts.go
```diff
@@ -717,6 +722,11 @@ func (p *PodFact) GetSubclusterOid() string {
 	return p.subclusterOid
 }
 
+// GetSubclusterType returns the string value of subclusterOid in PodFact
```
HaoYang0000, line 725: // GetSubclusterType returns the string value of subclusterType in PodFact

api/v1/verticadb_types.go
```diff
@@ -1030,6 +1030,12 @@ type SubclusterStatus struct {
 	// Object ID of the subcluster.
 	Oid string `json:"oid"`
 
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// +optional
+	// The type of the subcluster. It could be primary, secondary, transient, or empty if
```
cchen-vertica, line 1035: transient shouldn't be in the list. We need add "sandboxprimary" and "sandboxsecondary" in the list.

pkg/podfacts/podfacts.go
```diff
@@ -63,6 +63,9 @@ type PodFact struct {
 	// The oid of the subcluster the pod is part of
 	subclusterOid string
 
+	// The type of the subcluster the pod is part of
+	subclusterType string
```
cchen-vertica, line 67: We don't need "subclusterType" here. It is a duplicate with "isPrimary".

qindotguan: Thanks! Removed subclusterType and added function GetSubclusterStatusType() to set the status type.

pkg/podfacts/podfacts.go
```diff
@@ -717,6 +717,25 @@ func (p *PodFact) GetSubclusterOid() string {
...
+				// TODO: return SandboxSecondarySubcluster if pod is in sandbox with type secondary
+				return vapi.SandboxPrimarySubcluster
+			} else {
+				return vapi.SecondarySubcluster
+			}
+		}
+	}
+
+	// return empty if it's not in a subcluster yet
+	return ""
```
roypaulin, line 736:
```suggestion
	if !p.dbExists {
		// return empty if it's not in a subcluster yet
		return ""
	}
	if p.isPrimary {
		if p.sandbox != vapi.MainCluster {
			return vapi.SandboxPrimarySubcluster
		}
		return vapi.PrimarySubcluster
	}
	// TODO: return SandboxSecondarySubcluster if pod is in sandbox with type secondary
	return vapi.SecondarySubcluster
```

api/v1beta1/verticadb_types.go
```diff
@@ -1084,6 +1084,12 @@ type SubclusterStatus struct {
 	// Object ID of the subcluster.
 	Oid string `json:"oid"`
 
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// +optional
+	// The type of the subcluster. It could be primary, secondary, transient, or empty if
```
roypaulin, line 1089: I don't think transient should be part of this list; it is not real subcluster type. In the database a transient sc would be secondary so that's the type that should appear here.
Also add other types like sandboxprimary/sandboxsecondary.

roypaulin (approved):

cchen-vertica:
Let's hold this PR for one more day. Since this PR still has follow-up tasks to do, we can merge it after we release 25.2.

PR #1115, opened 2025-03-10
4245450 rotate NMA and HTTPS TLS certificates

LiboYu2:
The change will rotate the TLS certificates for NMA and HTTPS service.
Here is the life cycle of a new secret name during a rollback:
1 user change secret name in spec
2 rotate secret
3 after rotation, the new secret name in spec is also saved in annotation "vertica.com/nma-tls-secret-in-use".
   When the operator code need to use cert, they will get secret name from that annotation. After rotation,
  the original secret name saved in that annotation is now saved in "vertica.com/nma-tls-secret-previously-used", 
  which will be used if rollback is required.
  
  Here is the bootstrap process for db creation:
1 call create db API.
2 execute DDLs to set up key/cert
3 secret name in the spec is also saved in annotation "vertica.com/nma-tls-secret-in-use".
   When the operator code need to use cert, they will get secret name from that annotation.

Here is a high level description of the reconcile progress.
1 check preconditions before rotation can start
2 check the remote host and port and verify if the cert has already been rotated. 
   If rotation is already done, nothing happens. This is to make sure the Reconcile() is idempotent.
3 rotation https cert first.
   verify the rotation is succescful.
4 rotation nma cert first.
   verify the rotation is succescful.
5 If rotation is successful, 
      copy secret name from "vertica.com/nma-tls-secret-in-use" to
                  "vertica.com/nma-tls-secret-previously-used". 
      copy secret name from spec to "vertica.com/nma-tls-secret-in-use"
      send an event to vdb.


HaoYang0000:
Is this one ready for review?

ningdeng:
What's the status of this PR? is this deprecated or is this still ongoing after 1 month?

PR #1116, opened 2025-03-10
Make final changes for the autoscaler

roypaulin:
- Adds a new field `useCachedMetric` to read metric from the cache first.
- Adds annotations to pause the autoscaler
- Remove threshold adjustment field as it is not used

### DIFF
--- a/api/v1/verticaautoscaler_types.go
+++ b/api/v1/verticaautoscaler_types.go
@@ -54,7 +54,7 @@ type VerticaAutoscalerSpec struct {
 	ServiceName string `json:"serviceName,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:fieldDependency:scalingGranularity:Subcluster"
 	// When the scaling granularity is Subcluster, this field defines a template
 	// to use for when a new subcluster needs to be created.  If size is 0, then
 	// the operator will use an existing subcluster to use as the template.  If
@@ -281,9 +281,16 @@ type PrometheusSpec struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=false
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch","urn:alm:descriptor:com.tectonic.ui:advanced"}
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	// Used for skipping certificate check e.g: using self-signed certs.
 	UnsafeSsl bool `json:"unsafeSsl,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
+	// Enables caching of metric values during polling interval. It is Used to control whether the autoscaler should use cached metrics for scaling
+	// decisions rather than querying the external metric provider (e.g., Prometheus) on each scale event. This feature is not supported for cpu and memory.
+	UseCachedMetrics bool `json:"useCachedMetrics,omitempty"`
 }
 
 type CPUMemorySpec struct {
@@ -299,12 +306,6 @@ type CPUMemorySpec struct {
 
 // MetricDefinition defines increment and metric to be used for autoscaling
 type MetricDefinition struct {
-
-	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
-	// +kubebuilder:Minimum:=0
-	// The value used to increase the threshold after a scale out or a scale in.
-	ThresholdAdjustmentValue int `json:"thresholdAdjustmentValue,omitempty"`
 
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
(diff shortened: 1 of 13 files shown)


pkg/builder/labels_annotations.go
```diff
@@ -190,6 +195,23 @@ func MakeAnnotationsForSandboxConfigMap(vdb *vapi.VerticaDB, forUpgrade bool) ma
 	return annotations
 }
 
+// MakeAnnotationsForScaledObject builds the list of annotations that are included
+// in the scaledobject.
+func MakeAnnotationsForScaledObject(vas *vapi.VerticaAutoscaler) map[string]string {
+	annotations := make(map[string]string, len(vas.Annotations))
+	for k, v := range vas.Annotations {
```
HaoYang0000, line 202: Seems like we just want to change the key in the annotations so we don't need a loop here. We can do a copy of vas.Annotations, and assign the value if key exists and remove the old key in the copied map?

api/v1beta1/verticaautoscaler_types.go
```diff
@@ -281,9 +281,16 @@ type PrometheusSpec struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	// Used for skipping certificate check e.g: using self-signed certs.
 	UnsafeSsl bool `json:"unsafeSsl,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
+	// Enables caching of metric values during polling interval. It is Used to control whether the autoscaler should use cached metrics for scaling
+	// decisions rather than querying the external metric provider (e.g., Prometheus) on each scale event. This feature is not supported for cpu and memory.
+	UseCachedMetrics bool `json:"useCachedMetrics,omitempty"`
```
HaoYang0000, line 293: Is this a build in option on k8s autoscaling? We just read it and passed it down

api/v1beta1/verticaautoscaler_types.go
```diff
@@ -281,9 +281,16 @@ type PrometheusSpec struct {
...
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch","urn:alm:descriptor:com.tectonic.ui:advanced"}
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	// Used for skipping certificate check e.g: using self-signed certs.
 	UnsafeSsl bool `json:"unsafeSsl,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
+	// Enables caching of metric values during polling interval. It is Used to control whether the autoscaler should use cached metrics for scaling
+	// decisions rather than querying the external metric provider (e.g., Prometheus) on each scale event. This feature is not supported for cpu and memory.
```
HaoYang0000, line 292: If this is not supported for cpu or memory, maybe adding a new webhook rule for it?

api/v1/verticaautoscaler_types.go
```diff
@@ -281,9 +281,16 @@ type PrometheusSpec struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=false
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch","urn:alm:descriptor:com.tectonic.ui:advanced"}
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
```
qindotguan, line 284: Why was it removed?

pkg/builder/labels_annotations.go
roypaulin: No we want to copy all the annotations and if we find either of those 2, we change them.

api/v1beta1/verticaautoscaler_types.go
roypaulin: Yes, the caching is done by keda, we just need to set this field in the scaledobject.

api/v1beta1/verticaautoscaler_types.go
roypaulin: We don't need, it is in `PrometheusSpec` so it is not possible to set it for cpu/memory.

pkg/meta/annotations.go
```diff
@@ -387,6 +387,15 @@ const (
 	// Annotation set in a sandbox configMap. Indicates that routing must be disabled
 	// on the sandbox nodes.
 	DisableRoutingAnnotation = "vertica.com/disable-routing"
+
+	// It will pause scaling immediately and use the current instance count.
+	// This only works for a scaledobject.
+	PausingAutoscalingAnnotation = "vertica.com/paused"
```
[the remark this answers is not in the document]
roypaulin: I agree with you; I just don't want to expose `scaledObject` too much as it is an external component. A compromise would be to document that it is supported only for a scaledobject. What do you think?

api/v1/verticaautoscaler_types.go
roypaulin: Because there is no "advanced" section yet for VerticaAutoscaler in Openshift. VerticaAutoscaler is not big enough to have an advanced section.

pkg/meta/annotations.go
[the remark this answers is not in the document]
qindotguan: What about "vertica.com/paused-scaling"? It sounds like vetica will pause if using "vertica.com/paused".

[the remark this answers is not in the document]
roypaulin: Sounds good!

pkg/builder/labels_annotations.go
```diff
@@ -190,6 +195,23 @@ func MakeAnnotationsForSandboxConfigMap(vdb *vapi.VerticaDB, forUpgrade bool) ma
...
+		key := k
+		if key == vmeta.PausingAutoscalingAnnotation {
+			key = kedaPausingAutoscalingAnnotation
```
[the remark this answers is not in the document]
roypaulin: Opening a PR after this to add webhook rules.

PR #1118, opened 2025-03-12
Add sandboxes subclusters types

qindotguan:
As a follow-up task after #1114, this is the task 2 to add sandbox subcluster types. For example:
```
spec:
  sandboxes:
  - name: sand1
    subclusters:
    - name: sc1
      type: primary (default)
    - name: sc3
      type: secondary
```


api/v1/verticadb_types.go
```diff
@@ -1013,11 +1022,14 @@ const (
 	PrimarySubcluster   = "primary"
 	SecondarySubcluster = "secondary"
 	TransientSubcluster = "transient"
-	// A sandbox primary subcluster is a secondary subcluster that was the first
-	// subcluster in a sandbox. These subclusters are primaries when they are
-	// sandboxed. When unsandboxed, they will go back to being just a secondary
-	// subcluster
+	// In subclusters status, a sandbox primary subcluster is a primary
+	// subcluster in a sandbox. These subclusters are primaries when they
+	// are sandboxed. When unsandboxed, they will go back to being just a
```
cchen-vertica, line 1027: Minor: ... they will revert to a secondary subcluster

pkg/podfacts/podfacts.go
```diff
@@ -717,6 +717,26 @@ func (p *PodFact) GetSubclusterOid() string {
...
 }
 
+// GetSubclusterStatusType returns the subcluster status type depends on its type in subclusters and sandboxes
+func (p *PodFact) GetSubclusterStatusType() string {
+	if !p.dbExists {
+		// return empty if it's not in a subcluster yet
+		return ""
+	}
+	if p.isPrimary {
+		if p.sandbox != vapi.MainCluster {
```
cchen-vertica, line 727: This is not accurate. From the pod fact, we know a pod will be sandboxed, but we don't know if it is already in the sandbox at this point(in status reconciler). You should update this type in unsandbox/sandbox_reconciler.

In this function, we can do if p.sandbox != vapi.MainCluster return "". Only set the type through this function when the pod is not in a sandbox. Afterwards, when we have promote_subcluster reconciler, we can remove this function and update the "type" status in that reconciler.

roypaulin:
Can you fix the history now that other PR is merged?

PR #1121, opened 2025-03-17
Add e2e test to scale out db using wildcard

qindotguan:
- Create 2 vdb in the same namespace
- Set up the autoscaler to scale the 1st vdb
- Make sure the metric's value changes in the 2nd vdb do not trigger scaling
- Make sure you only take into account the 1st vdb nodes and only scale the 1st vdb


roypaulin:
Should rebase onto main, it will fix the tests.

roypaulin (approved):

PR #1126, opened 2025-03-19
Remove v1 verticadb

HaoYang0000:
Replace e2e test VerticaDB from v1beta1 to v1.


roypaulin:
E2e tests are not the only thing to change. You have to actually remove Verticadb v1beta1 from all over the code.

roypaulin:
Look everywhere Verticadb v1beta1 is used and removed all the dependent code. By code I mean the entire repo.

PR #1135, opened 2025-03-24
[Rebased] add sandboxes subclusters type

qindotguan:
Rebased #1118 on the main branch after #1114 was merged.


pkg/controllers/vdb/status_reconciler.go
```diff
@@ -182,6 +183,24 @@ func (s *StatusReconciler) calculateClusterStatus(stat *vapi.VerticaDBStatus) {
 	}
 }
 
+// getSubclusterStatusType returns the subcluster status type depends on its type in subclusters and sandboxes
+func (s *StatusReconciler) getSubclusterStatusType(podName types.NamespacedName) string {
```
roypaulin, line 187: We want to rely on the sandboxes status to make this decision, instead of podfacts.

roypaulin (commented):
Add webhook to check that the type is either primary or secondary.
Also we are not going to remove sandboxprimary/secondary is this PR?

qindotguan:
Removing sandboxprimary/secondary will be addressed in "Task 3: Remove the updates in spec.subclusters.subcluster.type".

api/v1/verticadb_webhook.go
```diff
@@ -1258,6 +1258,14 @@ func (v *VerticaDB) validateSubclustersInSandboxes(allErrs field.ErrorList) fiel
 	for i, sandbox := range sandboxes {
 		scInSandbox := make(map[string]int)
 		for _, sc := range sandbox.Subclusters {
+			// check if sandbox type is either primary or secondary
+			if sc.Type != "" && sc.Type != PrimarySubcluster && sc.Type != SecondarySubcluster {
+				err := field.Invalid(path.Index(i),
+					sandboxes[i],
+					fmt.Sprintf("if sandbox subcluster type is not empty, it should be either %s or %s", PrimarySubcluster, SecondarySubcluster))
```
cchen-vertica, line 1265: When you have this webhook, the operator cannot update the subcluster type to "SandboxPrimary" which should fail some e2e tests. I don't know why your e2e tests can pass. We need to double check if this webhook works.

If the webhook works fine, we should move the webhook to "Task 3: Remove the updates in spec.subclusters.subcluster.type".

pkg/controllers/vdb/status_reconciler.go
roypaulin, line 187:
```
func (s *StatusReconciler) getSubclusterStatusType(podName types.NamespacedName) string {
	if !s.PFacts.DoesDBExist() {
		return ""
	}

	sandboxStatus := s.Vdb.GetSandboxStatus(s.PFacts.SandboxName)
	isSandbox := sandboxStatus != nil && sandboxStatus.Name != vapi.MainCluster

	if s.PFacts.Detail[podName].GetIsPrimary() {
		if isSandbox {
			return vapi.SandboxPrimarySubcluster
		}
		return vapi.PrimarySubcluster
	}

	if isSandbox {
		return vapi.SandboxSecondarySubcluster
	}

	return vapi.SecondarySubcluster
}

```

pkg/controllers/vdb/status_reconciler.go
```diff
@@ -213,8 +232,8 @@ func (s *StatusReconciler) calculateSubclusterStatus(ctx context.Context, sc *va
 		if pf.GetSubclusterOid() != "" {
 			curStat.Oid = pf.GetSubclusterOid()
 		}
-		if pf.GetSubclusterStatusType() != "" {
-			curStat.Type = pf.GetSubclusterStatusType()
+		if s.getSubclusterStatusType(podName) != "" {
+			curStat.Type = s.getSubclusterStatusType(podName)
```
roypaulin, line 236:
```suggestion
                scType := s.getSubclusterStatusType(podName)
		if scType != "" {
			curStat.Type = scType
```
Also, move this out of the loop. We only need to call the function once, not once for each pod in the subcluster.

api/v1/verticadb_webhook.go
qindotguan: This webhook checks the spec.sandboxes[].subclusters[].type. The webhook works as the operator updated is spec.subclusters[].type.

pkg/controllers/vdb/status_reconciler.go
qindotguan: Good catch. Thanks for the suggestion.

pkg/controllers/vdb/status_reconciler.go
qindotguan: Thanks! Also updated to pass sc instead of the podName as the parameter.

api/v1/verticadb_webhook.go
cchen-vertica: I see. We need this webhook to guard spec.sandboxes[].subcsluters[].type as well. You could add that webhook in Task 3.

PR #1137, opened 2025-03-24
Add new jobs to run legs with cert

roypaulin:
This allows us to run existing vcluster legs with and without cert.


LiboYu2 (approved):
Had a call and Roy explained to me how this works.

PR #1146, opened 2025-03-26
Fix e2e leg4

LiboYu2:
Changes included in this PR:
1 add password to vsql command line for running DDLs.
2 fix all e2e cases in leg4. 
3 add a run with cert for leg4.


pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -129,9 +129,20 @@ func (c *CreateDBReconciler) execCmd(ctx context.Context, initiatorPod types.Nam
...
+		fakePodRunner, ok2 := c.PRunner.(*cmds.FakePodRunner)
+		if ok2 {
+			passwd = fakePodRunner.VerticaSUPassword
+		}
+	}
+
 	if c.Vdb.IsCertRotationEnabled() {
 		cmd := []string{
-			"-f", PostDBCreateSQLFileVclusterOps,
+			"--password", clusterPodRunner.VerticaSUPassword, "-f", PostDBCreateSQLFileVclusterOps,
```
qindotguan, line 145: Should here use passwd`?
```suggestion
			"--password", passwd "-f", PostDBCreateSQLFileVclusterOps,
```

qindotguan (commented):
You can run `make test` to check for the lint issues.

pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -129,9 +129,20 @@ func (c *CreateDBReconciler) execCmd(ctx context.Context, initiatorPod types.Nam
 	if res, err := c.Dispatcher.CreateDB(ctx, opts...); verrors.IsReconcileAborted(res, err) {
 		return res, err
 	}
+	passwd := ""
+	clusterPodRunner, ok := c.PRunner.(*cmds.ClusterPodRunner)
```
HaoYang0000, line 133: There is a GetSuperuserPassword method in verticadb_controller, maybe add it here in the reconciler?

pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -129,9 +129,20 @@ func (c *CreateDBReconciler) execCmd(ctx context.Context, initiatorPod types.Nam
 	if res, err := c.Dispatcher.CreateDB(ctx, opts...); verrors.IsReconcileAborted(res, err) {
 		return res, err
 	}
+	passwd := ""
```
ningdeng, line 132: I assume you'll follow Qin's suggestion to use `passwd`, if so can you do `var password string`, requesting because this could cause a "empty password" vulnerability in scanners. 

Golang default string vars to be empty string, so just declare the var works.

cchen-vertica:
Change tlsMode to "verify_ca", and then rerun the test.

ningdeng:
What's the status of this? I see no response for the comments. @LiboYu2

pkg/controllers/vdb/createdb_reconciler.go
LiboYu2: That is definitely a better approach. I will use it. Thanks for sharing.

pkg/controllers/vdb/createdb_reconciler.go
LiboYu2: I used Hao's proposed method. Now the passwd is defined as follows:

	passwd, err := vk8s.GetSuperuserPassword(ctx, c.VRec.Client, c.Log, c.VRec.GetEventRecorder(), c.Vdb)
	if err != nil {
		return ctrl.Result{}, err
	}

Will this pass the security scanning?

LiboYu2: Reverted the original change. There is a change in main branch to use PRunner.ExecVSQL to execute the SQL script. That function already takes password into consideration.

ningdeng: > I used Hao's proposed method. Now the passwd is defined as follows:
> 
> ```
> passwd, err := vk8s.GetSuperuserPassword(ctx, c.VRec.Client, c.Log, c.VRec.GetEventRecorder(), c.Vdb)
> if err != nil {
> 	return ctrl.Result{}, err
> }
> ```
> 
> Will this pass the security scanning?

As long as you don't initialize it explicitly as "", you should be fine.

pkg/controllers/vdb/createdb_reconciler.go
ningdeng: @LiboYu2 -- if you don't plan to take the suggestion from a comment, can you reply saying why you don't take it instead of simply ignore the comment?

Or have you already addressed this?

LiboYu2: > @LiboYu2 -- if you don't plan to take the suggestion from a comment, can you reply saying why you don't take it instead of simply ignore the comment?
> 
> Or have you already addressed this?

Reverted the original change. There is a change in main branch to use PRunner.ExecVSQL to execute the SQL script. That function already takes password into consideration.

LiboYu2:
> What's the status of this? I see no response for the comments. @LiboYu2
It is ready for review. Cai wants me to run some tests with verify_ca. I am working on that.

LiboYu2:
> Change tlsMode to "verify_ca", and then rerun the test.

All test cases passed with "verify_ca".

tests/e2e-leg-4/revivedb-1-node/cert.yaml
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+
+# This is applied through envsubst so that the $NAMESPACE gets filled in
+
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: serving-cert
+spec:
+  commonName: dbadmin
```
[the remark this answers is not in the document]
LiboYu2: > This value shouldn't be hardcoded. It should set by "vertica-superuser-name".

Currently, we have to use 'dbadmin' as it is hardcoded in our code. After we support super admin specified by the user, we can make this change.

cchen-vertica (approved):
All looks good. You need to fix one comment, and then you can merge it.

PR #1147, opened 2025-03-26
Fix leg 12

roypaulin:
This fixes a password issue and run leg 12 with cert.


roypaulin:
If any leg 12 test fails I will fix it in this PR.

PR #1165, opened 2025-04-03
Fix upgrade with certs

roypaulin:
The base version for upgrade is 24.4.0 so if you upgrade to 25.3.0, as there is no tls config, http calls after upgrade will all fail.
This picks a 25.3.0 image if certs are enabled .


roypaulin (commented):
Looks good!

cchen-vertica (approved):

PR #1173, opened 2025-04-07
in vlogger add option to filter log

HaoYang0000:
This PR enables vlogger to include logs from different levels. 
Valid log levels are: DEBUG INFO WARNING ERROR

Parameters: 
LOG_LEVEL: Defines the minimum log severity level to print.
LOG_FILTER: Comma-separated list (e.g., INFO,ERROR) supports including specific levels in the list.

Example usage:
No env variable setup, prints all logs under INFO level: ./tail_vertica_log.sh

Minimal setup: show INFO and above(INFO,WARNING,ERROR)
LOG_LEVEL=INFO ./tail_vertica_log.sh

Show WARNING and above(WARNING,ERROR):
LOG_LEVEL=WARNING ./tail_vertica_log.sh
or LOG_FILTER=WARNING,ERROR ./tail_vertica_log.sh

Show INFO and WARNING(INFO,WARNING):
LOG_FILTER=INFO,WARNING ./tail_vertica_log.sh

If both LOG_LEVEL and LOG_FILTER are set, LOG_LEVEL will be ignored(WARNING,ERROR):
LOG_LEVEL=INFO LOG_FILTER="WARNING,ERROR" ./tail_vertica_log.sh

Fallback to CLI usage if env not set: ./tail_vertica_log.sh $LOG_LEVEL $LOG_FILTER
e.g: ./tail_vertica_log.sh WARNING "WARNING,ERROR"

Sample crd to include WARNING and ERROR log level.
  sidecars:
    - name: vlogger
      image: vertica-logger:2.0.0
      env:
        - name: LOG_LEVEL
          value: "WARNING "
Or 
  sidecars:
    - name: vlogger
      image: vertica-logger:2.0.0
      env:
        - name: LOG_FILTER
          value: "WARNING,ERROR"


roypaulin:
I have suggestion:
- Let's also support traditional logging: if user pass log level WARNING, we log WARNING, ERROR, FATAL. So add an env var LOG_LEVEL. default will be INFO. This is the most widespread use of the logging level system so most users will know it
- You can still keep LOG_LEVELS but it is better if the syntax is "WARNING,ERROR", what is not in the comma separated list is automatically excluded. If present, takes precedence.

No env var set: it internally uses LOG_LEVEL default value.
LOG_LEVEL set: use that value and do an hierarchical logging
LOG_LEVELS set: use the value and log all the levels in the list
LOG_LEVEL and LOG_LEVELS set: LOG_LEVELS takes precedence

Makefile
```diff
@@ -11,7 +11,7 @@ export VERSION
 # order to have a different release cadence.
 #
 # When changing this, be sure to update the tags in docker-vlogger/README.md
-VLOGGER_VERSION ?= 1.0.1
+VLOGGER_VERSION ?= 2.0.1
```
cchen-vertica, line 14: The version should be 2.0.0

Makefile
```diff
@@ -453,16 +453,16 @@ docker-build-operator: manifests generate fmt vet ## Build operator docker image
 		--build-arg GO_VERSION=${GO_VERSION} \
 		-f docker-operator/Dockerfile .
 
-.PHONY: docker-build-vlogger
 docker-build-vlogger:  ## Build vertica logger docker image
 	docker pull ${VLOGGER_BASE_IMG}:${VLOGGER_ALPINE_VERSION} # Ensure we have the latest alpine version
-	docker buildx build \
+	cd docker-vlogger \
```
cchen-vertica, line 458: why do we need this change?

docker-vlogger/Dockerfile
```diff
@@ -21,12 +21,10 @@ FROM ${BASE_IMG}:${ALPINE_VERSION}
...
 RUN apk add --no-cache tini
 ENTRYPOINT ["/sbin/tini", "--"]
-# $DBPATH is set by the operator and is the /<localDataPath>/<dbName>.
-# The tail can't be done until the vertica.log is created.  This is because the
-# exact location isn't known until the server pod has come up and is added to
-# the cluster. 
-# Note: we use the '-F' option with tail so that it survives log rotations.
-CMD ["sh", "-c", "FN=$DBPATH/v_*_catalog/vertica.log; until [ -f $FN ]; do sleep 5; done; tail -n 1 -F $FN"]
+
+COPY ./tail_vertica_log.sh /tmp/tail_vertica_log.sh
```
cchen-vertica, line 25: Could we put this script in /opt/vertica/bin or /usr/local/bin?

docker-vlogger/tail_vertica_log.sh
```diff
@@ -0,0 +1,98 @@
...
+# exact location isn't known until the server pod has come up and is added to
+# the cluster. 
+# Note: we use the '-F' option with tail so that it survives log rotations.
+
+# -------------------------------
+# Configuration & Valid Log Levels
+# -------------------------------
+
+# Ref: https://docs.vertica.com/23.3.x/en/connecting-to/client-libraries/client-drivers/install-config/ado-netr/ado-net-logging/
+VALID_LEVELS="TRACE DEBUG INFO WARNING ERROR FATAL"
```
cchen-vertica, line 18: The reference here is for client logs, not Vertica logs. Remove "TRACE" and "FATAL" in the VALID_LEVELS.

cchen-vertica (commented):
We still need a new field in CRD and a webhook to validate the value of it, right?

docker-vlogger/Dockerfile
roypaulin: this is vertica-logger image, it does not have /opt/vertica/

HaoYang0000:
> I have suggestion:
> 
> * Let's also support traditional logging: if user pass log level WARNING, we log WARNING, ERROR, FATAL. So add an env var LOG_LEVEL. default will be INFO. This is the most widespread use of the logging level system so most users will know it
> * You can still keep LOG_LEVELS but it is better if the syntax is "WARNING,ERROR", what is not in the comma separated list is automatically excluded. If present, takes precedence.
> 
> No env var set: it internally uses LOG_LEVEL default value. LOG_LEVEL set: use that value and do an hierarchical logging LOG_LEVELS set: use the value and log all the levels in the list LOG_LEVEL and LOG_LEVELS set: LOG_LEVELS takes precedence

We can do LOG_LEVEL, which follows traditional logging, and let user set LOG_LEVEL=INFO, or LOG_LEVEL=WARNING based on their usage.
And add LOG_FILTER(instead of LOG_LEVELS) to avoid confusion. We will let LOG_FILTER overwrite the LOG_LEVEL setup. And take LOG_FILTER=INFO,ERROR,-WARNING as input. '-' means exclude, otherwise means include.

docker-vlogger/Dockerfile
HaoYang0000: No need to do that. The script is currently stored in k8s operator repo, alone with vlogger docker file. The vlogger side car docker image was simply a cmd run, but docker CMD doesn't work well with pipeline redirection. That's why I put it in the script.

Makefile
HaoYang0000: The script is under docker-vlogger path. If we don't add the docker-vlogger there we need to add it in the docker file under docker-vlogger. I checked other make target, this is the convention we used for docker-vertica and docker-vertica-v2

roypaulin:
> > We still need a new field in CRD and a webhook to validate the value of it, right?
> 
> No field needed for the CRD. Because we use K8S sidecar crd here, and we only need to check if the env values is there with sidecar image similar to "vertica-logger:2.0.0"

I think the validation needs to happen in the script itself. You can't use the image to know if it is a vertica logger container as the image name can be anything.

roypaulin:
> > I have suggestion:
> > 
> > * Let's also support traditional logging: if user pass log level WARNING, we log WARNING, ERROR, FATAL. So add an env var LOG_LEVEL. default will be INFO. This is the most widespread use of the logging level system so most users will know it
> > * You can still keep LOG_LEVELS but it is better if the syntax is "WARNING,ERROR", what is not in the comma separated list is automatically excluded. If present, takes precedence.
> > 
> > No env var set: it internally uses LOG_LEVEL default value. LOG_LEVEL set: use that value and do an hierarchical logging LOG_LEVELS set: use the value and log all the levels in the list LOG_LEVEL and LOG_LEVELS set: LOG_LEVELS takes precedence
> 
> We can do LOG_LEVEL, which follows traditional logging, and let user set LOG_LEVEL=INFO, or LOG_LEVEL=WARNING based on their usage. And add LOG_FILTER(instead of LOG_LEVELS) to avoid confusion. We will let LOG_FILTER overwrite the LOG_LEVEL setup. And take LOG_FILTER=INFO,ERROR,-WARNING as input. '-' means exclude, otherwise means include.

I am ok with LOG_FILTER. But it should just be the list of levels you want to log. Any level not in the list will not be logged.

roypaulin:
Again, let's make it simple:
No env var set: it internally uses LOG_LEVEL default value.
LOG_LEVEL set: use that value and do an hierarchical logging
LOG_FILTER set: use the value and log all the levels in the list. Let's not add that "-WARNING"  syntax. If the user wants to log a subset of log levels they just have to list them this way: "loglevel1,loglevel2,..."
LOG_LEVEL and LOG_FILTER set: LOG_FILTER takes precedence, meaning LOG_LEVEL is completely ignored

 LOG_LEVEL and LOG_FILTER should be mutually exclusive and save different purpose with the former being for hierarchical logging (log the given log level and above) and the latter being for logging a given list of log level. LOG_FILTER takes precedence **if set**. The 2 use cases never intersect.

HaoYang0000:
> Again, let's make it simple: No env var set: it internally uses LOG_LEVEL default value. LOG_LEVEL set: use that value and do an hierarchical logging LOG_FILTER set: use the value and log all the levels in the list. Let's not add that "-WARNING" syntax. If the user wants to log a subset of log levels they just have to list them this way: "loglevel1,loglevel2,..." LOG_LEVEL and LOG_FILTER set: LOG_FILTER takes precedence, meaning LOG_LEVEL is completely ignored
> 
> LOG_LEVEL and LOG_FILTER should be mutually exclusive and save different purpose with the former being for hierarchical logging (log the given log level and above) and the latter being for logging a given list of log level. LOG_FILTER takes precedence **if set**. The 2 use cases never intersect.

The existing code already covers the use case as you mentioned here, but there are comments I want to add so we are on the same page:
For no env var set: It should have the same behavior as version 1.0.1, which prints all the logs in vertica.log. Here not only contains log info, but also some query statements and other messages included in the vertica.log.
LOG_LEVEL set: We are good on the existing setup.
LOG_FILTER set: I think we should still add the "-WARNING" syntax. As mentioned above, the vertica.log prints not only logs. If user want to simply exclude <INFO> to save costs but keeps the other messages. The - syntax will help to achieve that.

roypaulin:
This PR is only for log level, we can keep the lines without any log level.

My concern with that syntax is if users want only ERROR logs, they will have to set these:
LOG_LEVEL=ERROR 
LOG_FILTER=-DEBUG,-INFO,-NOTICE,... and that's quite confusing.
What I am saying is that it would be easier to just do this:
LOG_FILTER=ERROR.

HaoYang0000:
> This PR is only for log level, we can keep the lines without any log level.
> 
> My concern with that syntax is if users want only ERROR logs, they will have to set these: LOG_LEVEL=ERROR LOG_FILTER=-DEBUG,-INFO,-NOTICE,... and that's quite confusing. What I am saying is that it would be easier to just do this: LOG_FILTER=ERROR.

This PR contains both LOG_LEVEL and LOG_FILTER functionality: https://github.com/vertica/vertica-kubernetes/pull/1173/files#diff-08e80447e449a3fae022bf5529f48e91bc88b6e12dfe112a62eb0513bf8e60c1R34-R35

For setting up the ERROR log, the user can simply do LOG_LEVEL=ERROR, no LOG_FILTER needs to be set. If you read the example in PR description or the comment in the code, the LOG_FILTER will overwrite values in LOG_LEVEL.

HaoYang0000:
@roypaulin @cchen-vertica Please take a look.

roypaulin:
Referring to the description above:



> Override log level filter. Show ERROR and above, but keep the DEBUG level(DEBUG,ERROR)
LOG_LEVEL=ERROR LOG_FILTER="DEBUG" ./tail_vertica_log.sh
or LOG_FILTER="DEBUG,ERROR" ./tail_vertica_log.sh

These 2 should not be used together. And if they are, LOG_FILTER should take precedence (not override) i.e we just ignore LOG_LEVEL. So to achieve the example above, users should simply do the following: LOG_FILTER="DEBUG,ERROR".

HaoYang0000:
> Referring to the description above:
> 
> > Override log level filter. Show ERROR and above, but keep the DEBUG level(DEBUG,ERROR)
> > LOG_LEVEL=ERROR LOG_FILTER="DEBUG" ./tail_vertica_log.sh
> > or LOG_FILTER="DEBUG,ERROR" ./tail_vertica_log.sh
> 
> These 2 should not be used together. And if they are, LOG_FILTER should take precedence (not override) i.e we just ignore LOG_LEVEL. So to achieve the example above, users should simply do the following: LOG_FILTER="DEBUG,ERROR".

I updated the usage example in: https://github.com/vertica/vertica-kubernetes/pull/1173/commits/c5599f027a38ce84fa1d2592b51722ab76ea0137

HaoYang0000:
@roypaulin @cchen-vertica Updated as discussed:
Treat untagged lines as INFO level(but not print DEBUG lines).

docker-vlogger/tail_vertica_log.sh
```diff
@@ -0,0 +1,147 @@
...
+print_logs() {
+  tail -n 1 -F $LOG_FILE | while read -r line; do
+    upper_line=$(to_upper "$line")
+    show=1
+    # untagged lines, treated as INFO level
+    if ! echo "$upper_line" | grep -qE "<(DEBUG|INFO|WARNING|ERROR)>"; then
+      # if <INFO> in level pattern, print it
+      if echo "$LEVEL_PATTERN" | grep -q "<INFO>"; then 
+        # If Debug tag found, not print it
+        echo "$upper_line" | grep -q "<DEBUG>" && show=0
```
cchen-vertica, line 125: Don't need to grep <DEBUG> again.

cchen-vertica (approved):
Left a minor comment.

docker-vlogger/tail_vertica_log.sh
HaoYang0000: We need this, because this will exclude the <DEBUG> tag logs when we have <INFO> in the log level. Because we treat INFO to print everything, so we need this.

PR #1183, opened 2025-04-11
Add sandbox option to VerticaScrutinize, to run using sandbox hosts instead of main cluster

kcmackeen:
Added "sandbox" to VerticaScrutinize. When set, we will collect Scrutinize using the hosts of the specified sandbox; if omitted, we will collect from main cluster, as before. Also added error handling when invalid sandbox name is specified.

### DIFF
--- a/pkg/controllers/vscr/vdbverify_reconciler.go
+++ b/pkg/controllers/vscr/vdbverify_reconciler.go
@@ -72,8 +72,8 @@ func (s *VDBVerifyReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (c
 
 	// If sandbox is defined and sandbox does not exist, we want to exit early
 	sandbox := s.Vscr.Spec.Sandbox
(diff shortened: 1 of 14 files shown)


pkg/controllers/vscr/vdbverify_reconciler.go
```diff
@@ -69,6 +69,16 @@ func (s *VDBVerifyReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (c
 		return ctrl.Result{}, s.updateStateAndScrutinizeReadyCondition(ctx, metav1.ConditionFalse, events.VerticaDBNotFound,
 			fmt.Sprintf("NotReady:%s", events.VerticaDBNotFound))
 	}
+
+	// If sandbox is defined and sandbox does not exist, we want to exit early
+	sandbox := s.Vscr.Spec.Sandbox
+	if sandbox != "" && s.Vdb.GetSandbox(sandbox) == nil {
```
roypaulin, line 75: Use sandbox status instead

roypaulin (approved):

PR #1186, opened 2025-04-14
Small refactoring

roypaulin:
- some refactoring in vc.go
- Add sandbox upgrade/shutdown at the beginning of the reconcile loop. This will allow sandboxes to be stuck when the is an error in the reconcile loop, this will make sandbox controller more reactive
- some logs to make debugging easier 
- fix draining


pkg/controllers/vdb/drainnode_reconciler.go
```diff
@@ -91,12 +92,15 @@ func (s *DrainNodeReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (c
 	elapsed := time.Since(drainStartTime)
 	timeout := time.Duration(timeoutInt) * time.Second
 	// If timeout has expired, we move on to the next reconciler (DBRemoveNodeReconciler|DBRemoveSubclusterReconciler)
+	s.VRec.Log.Info("Draining in progress", "start", drainStartTime, "elapsed", elapsed, "timeout", timeout)
 	if elapsed >= timeout {
 		s.VRec.Log.Info("Draining timeout has expired")
+		s.killConnectionsToPendingDeletePods(ctx, pfs)
```
roypaulin, line 98: I added killing connections back. Because in remove_node/subcluster reconcilers, there is a very small window where if vdb has changed since the last collection, we will requeue and go through draining again. But killing sessions is just best-effort, even if it fails, we still proceed with remove_node/subcluster as they will kill the sessions anyway

pkg/controllers/vdb/drainnode_reconciler.go
```diff
@@ -163,3 +166,62 @@ func (s *DrainNodeReconciler) setDrainStartAnnotation(ctx context.Context) error
...
+}
+
+// calculateRequeueDelay calculates how long we wait between consecutive requeues
+func calculateRequeueDelay(elapsed, timeout time.Duration) time.Duration {
+	remaining := timeout - elapsed
+
+	if remaining <= 0 {
+		return 0
+	}
+	if remaining > 10*time.Second {
```
cchen-vertica, line 222: #1193 limits the requeue time to 1 second. This behavior probably cannot work.

pkg/controllers/vdb/sandboxshutdown_reconciler.go
```diff
@@ -67,8 +69,13 @@ func (s *SandboxShutdownReconciler) reconcileSandboxShutdown(ctx context.Context
 	sbName := sb.Name
 	sbStatus := s.Vdb.GetSandboxStatus(sbName)
 	if sbStatus == nil {
-		s.Log.Info("Requeue because the sandbox does not exist yet", "sandbox", sbName)
-		return ctrl.Result{Requeue: true}, nil
+		if s.Requeue {
+			s.Log.Info("Requeue because the sandbox does not exist yet", "sandbox", sbName)
+		} else {
+			s.Log.Info("sandbox does not exist in the database yet", "sandbox", sbName)
```
cchen-vertica, line 75: Why don't we need to requeue when the subcluster is not sandboxed? If a user specify the vdb like:
spec:
   sandbox:
       name: sand1
       shutdown: true
and the sand1 isn't sandboxed yet, will webhook prevent it? If not, we probably won't stop it.

cchen-vertica (approved):

PR #1187, opened 2025-04-14
test leg 3 cert

LiboYu2:


.github/workflows/e2e.yml
```diff
@@ -352,6 +352,28 @@ jobs:
           communal-storage-type: hostpath
           e2e-retry-times: ${{ inputs.e2e_retry_times || 2 }}
 
+  e2e-leg-3-vcluster-cert:
+    if: ${{ ! github.event.pull_request.head.repo.fork  && (inputs.e2e_test_suites == 'all' || inputs.e2e_test_suites == 'vcluster leg 3 cert' || inputs.e2e_test_suites == '')}}
```
cchen-vertica, line 356: Define "vcluster leg 3 cert" in input.

tests/e2e-leg-3/revive-with-different-local-paths/vdb-to-create/base/setup-vdb.yaml
```diff
@@ -19,6 +19,7 @@ spec:
   image: kustomize-vertica-image
   communal:
     includeUIDInPath: false
+  httpServerTLSSecret: custom-cert
```
cchen-vertica, line 22: I want to see if httpServerTLSSecret is not set, could we generate the correct tls certificate. Could you update the test in leg-4 since it has a different superuser name? Or you could use another superuser name in this leg and remove this line.

LiboYu2: I tested in my local environment.  If httpServerTLSSecret is not set, we can generate the correct tls certificate if when a customized superadmin username is used.  When we revive the db, we have to use the secret name that was used to create the db. That is why we have to use a predefined secret name for those revive test cases in leg 3 and leg 4.
Leg 4 has two non-revive test cases but they are both v1 and don't have httpServerTLSSecret field.

cchen-vertica: I see. The user is possible to lose the auto-generated secret. I think the default behavior should still be "username & password" authentication.

.github/workflows/e2e.yml
cchen-vertica: This is till need to be addressed.

cchen-vertica (approved):

PR #1193, opened 2025-04-14
Allow the user to set maximum backoff duration for sandbox controller

cchen-vertica:
I noticed that the sandbox controller can hang after requeuing multiple times, even though we don’t set a large requeueAfter value. To address this, I added a new parameter that allows the controller to pick up tasks more quickly.


cchen-vertica:
> You should also add that new param in Makefile and make sure it appears in bundle/manifests

I didn't put it in Makefile because I think the default value(1s) is good enough for our development. The customers won't use the Makefile.

config/manager/operator-envs
```diff
@@ -18,3 +18,4 @@ CONCURRENCY_SANDBOXCONFIGMAP=${CONCURRENCY_SANDBOXCONFIGMAP}
 CONCURRENCY_VERTICAREPLICATOR=${CONCURRENCY_VERTICAREPLICATOR}
 BROADCASTER_BURST_SIZE=${BROADCASTER_BURST_SIZE}
 VDB_MAX_BACKOFF_DURATION=${VDB_MAX_BACKOFF_DURATION}
+SANDBOX_MAX_BACKOFF_DURATION=${SANDBOX_MAX_BACKOFF_DURATION}
```
roypaulin, line 21: This is the reason why I asked to add it in the Makefile. We are getting the values from there.

cchen-vertica: I added that in Makefile. I was trying to say: the environment variable here is only used by the developers. The customers won't use this one.

roypaulin (approved):

PR #1198, opened 2025-04-15
4541256 support https tls mode

LiboYu2:
Major changes: 
1 support both https TLS mode and client server TLS mode. The current tls modes are saved in status fields and new tls modes are saved in spec fields. 
2 support TLS mode ratation by calling https cert rotation API.
3 use http/v2/health for liveness/readiness/startup probe. It is also used in podfacts' script.

Fen's change is already merged into master. Some e2e tests failed in my local environment because of issues from the image built from Fen's branch and I want to see whether the new daily build can resolve those failures.


pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -140,6 +140,13 @@ func (c *CreateDBReconciler) execCmd(ctx context.Context, initiatorPod types.Nam
 			c.Log.Error(err2, "failed to execute TLS DDLs after db creation stderr - "+stderr)
 			return ctrl.Result{}, err2
 		}
+		httpsTLSMode := vapi.MakeNMATLSMode(c.Vdb.Spec.HTTPSTLSMode)
+		clientTLSMode := vapi.MakeClientServerTLSMode(c.Vdb.Spec.ClientServerTLSMode)
+		err = vdbstatus.UpdateTLSModes(ctx, c.VRec.GetClient(), c.Vdb, []*vapi.TLSMode{httpsTLSMode, clientTLSMode})
```
cchen-vertica, line 145: Here we just update TLSMode in vdb.status, right? We should update TLSMode in database first, then we update the status.

pkg/controllers/vdb/createdb_reconciler.go
```diff
@@ -342,7 +349,6 @@ func (c *CreateDBReconciler) generateAWSTlsSQL(sb *strings.Builder) {
 	fmt.Fprintf(sb, "ADD CA CERTIFICATES server_ca_cert TLSMODE 'TRY_VERIFY';\n")
```
cchen-vertica, line 349: Here the mode is still hardcoded.

pkg/controllers/vdb/init_db.go
```diff
@@ -98,14 +98,21 @@ func (g *GenericDatabaseInitializer) runInit(ctx context.Context) (ctrl.Result,
...
+			return ctrl.Result{}, err
+		}
+		clientSec := vapi.MakeClientServerTLSSecretRef(g.Vdb.Spec.ClientServerTLSSecret)
+		if err := vdbstatus.UpdateSecretRef(ctx, g.VRec.GetClient(), g.Vdb, clientSec); err != nil {
+			return ctrl.Result{}, err
+		}
+		sRefs := []*vapi.SecretRef{
+			sec, clientSec,
+		}
+		if err := vdbstatus.UpdateSecretRefs(ctx, g.VRec.GetClient(), g.Vdb, sRefs); err != nil {
```
cchen-vertica, line 113: Why do we need to update the secret refs again?

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -671,7 +671,7 @@ func (o *ObjReconciler) handleStatefulSetUpdate(ctx context.Context, sc *vapi.Su
 	// If the NMA deployment type is changing, we cannot do a rolling update for
 	// this change. All pods need to have the same NMA deployment type. So, we
 	// drop the old sts and create a fresh one.
-	if isNMADeploymentDifferent(curSts, expSts) {
+	if isNMADeploymentDifferent(curSts, expSts) || isLivenessPortDifferent(curSts, expSts) {
```
cchen-vertica, line 674: Enrich the comment a little bit.

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -815,6 +815,11 @@ func isNMADeploymentDifferent(sts1, sts2 *appsv1.StatefulSet) bool {
 	return vk8s.HasNMAContainer(&sts1.Spec.Template.Spec) != vk8s.HasNMAContainer(&sts2.Spec.Template.Spec)
 }
 
+// isLivenessPortDifferent will return true if the two stateful sets use different http port for livness check
+func isLivenessPortDifferent(sts1, sts2 *appsv1.StatefulSet) bool {
+	return vk8s.GetLivenessProbePort(&sts1.Spec.Template.Spec) != vk8s.GetLivenessProbePort(&sts2.Spec.Template.Spec)
```
cchen-vertica, line 820: Could we make this more generic: isHealthCheckDifferent(), return when livenessProbe or startupProbe or readinessProbe is different. Not only compare the port, but we also compare path and scheme.

pkg/controllers/vdb/tlsmode_reconciler.go
```diff
@@ -0,0 +1,275 @@
...
+		VRec:       vdbrecon,
+		Vdb:        vdb,
+		Log:        log.WithName("TLSModeReconciler"),
+		Dispatcher: dispatcher,
+		PRunner:    prunner,
+		Pfacts:     pfacts,
+	}
+}
+
+// Reconcile will create a TLS secret for the http server if one is missing
```
cchen-vertica, line 68: update the comment here

pkg/controllers/vdb/tlsmode_reconciler.go
```diff
@@ -0,0 +1,275 @@
...
+		return ctrl.Result{}, nil
+	}
+	if h.Vdb.Spec.HTTPSTLSMode == h.Vdb.GetNMATLSModeInUse() &&
+		h.Vdb.Spec.ClientServerTLSMode == h.Vdb.GetClientServerTLSModeInUse() {
+		return ctrl.Result{}, nil
+	}
+	h.Log.Info(fmt.Sprintf("https: current tls mode - %s, spec tls mode - %s", h.Vdb.GetNMATLSModeInUse(), h.Vdb.Spec.HTTPSTLSMode))
+	h.Log.Info(fmt.Sprintf("client: current tls mode - %s, spec tls mode - %s", h.Vdb.GetClientServerTLSModeInUse(),
+		h.Vdb.Spec.ClientServerTLSMode))
+	h.VRec.Eventf(h.Vdb, corev1.EventTypeNormal, events.NMATLSModeUpdateStarted,
```
cchen-vertica, line 81: The event should be more generic: TLSModeUpdateStarted

pkg/controllers/vdb/tlsmode_reconciler.go
```diff
@@ -0,0 +1,275 @@
...
+
+	if currentHTTPSTLSMode != newHTTPSTLSMode {
+		h.Log.Info(fmt.Sprintf("ready to change HTTPS TLS mode from %s to %s", currentHTTPSTLSMode, newHTTPSTLSMode))
+	}
+	if currentClientTLSMode != newClientTLSMode {
+		h.Log.Info(fmt.Sprintf("ready to change client TLS mode from %s to %s", currentClientTLSMode, newClientTLSMode))
+	}
+
+	currentCert := string(currentSecret.Data[corev1.TLSCertKey])
+	keyConfig := fmt.Sprintf("{\"data-key\":%q, \"namespace\":%q}", corev1.TLSPrivateKeyKey, h.Vdb.Namespace)
```
cchen-vertica, line 139: Will AWS key have this format? AWS key shouldn't have namespace.

pkg/controllers/vdb/tlsmode_reconciler.go
```diff
@@ -0,0 +1,275 @@
...
+		h.Log.Info("tls modes retrieved from db are saved into vdb status")
+		h.VRec.Eventf(h.Vdb, corev1.EventTypeNormal, events.NMATLSModeUpdateSucceeded,
+			"Successfully updated TLS modes after reviving. https - %s, client - %s", h.Vdb.Spec.HTTPSTLSMode,
+			h.Vdb.Spec.ClientServerTLSMode)
+	}
+	return ctrl.Result{}, nil
+}
+
+func (h *TLSModeReconciler) loadTLSModeAfterRevive(ctx context.Context, tlsConfig int) (bool, ctrl.Result, error) {
+	currentTLSMode, _ := h.getCurrentTLSMode(tlsConfig)
```
cchen-vertica, line 203: Here you get tls mode from status, but where did you put the correct tls mode into the status?

tests/e2e-leg-12/scale-in-db-with-threshold-hpa/41-wait-for-prometheus.yaml
```diff
@@ -14,4 +14,4 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
-  - script: "sleep 60"
\ No newline at end of file
+  - script: "sleep 60"
```
cchen-vertica, line 17: Why do we need this sleep? Can we explicitly specify the object we are waiting for?

cchen-vertica (commented):
After we have tlsmode_reconciler, we shouldn't distinguish create_db, revive_db, update_db use cases. In that reconciler, we can get the current tlsMode from status, and then compare it with spec.tlsMode to see if we need to update it. The status should come from vertica server.

Moreover, I see we rotate certs too in tlsmode_reconciler. We could have an https_tls_reconciler to reconcile https's tlsmode and certificates, and a client_server_tls_reconciler to reconcile server's tlsmode and certificates.

cchen-vertica:
Add changie files since you added new fields.

LiboYu2:
> After we have tlsmode_reconciler, we shouldn't distinguish create_db, revive_db, update_db use cases. In that reconciler, we can get the current tlsMode from status, and then compare it with spec.tlsMode to see if we need to update it. The status should come from vertica server.
> 
> Moreover, I see we rotate certs too in tlsmode_reconciler. We could have an https_tls_reconciler to reconcile https's tlsmode and certificates, and a client_server_tls_reconciler to reconcile server's tlsmode and certificates.

I used rotate API to change TLS mode as we don't have vclusterops API for that. It would be convenient if we have a generic vclusterops API to run any DDLs.

The https and client server reconcile processes are very similar. We can do a lot of optimization when handling them together.
For instance, we can make one meta upgrade call and restart NMA once for both. We can have a single tls_reconciler to handle them at the same time.

api/v1/verticadb_types.go
```diff
@@ -369,6 +369,19 @@ type VerticaDBSpec struct {
 	// Create client proxy pods for the subcluster if defined
 	// All incoming connections to the subclusters will be routed through the proxy pods
 	Proxy *Proxy `json:"proxy,omitempty"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:hidden"
```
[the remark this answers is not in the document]
LiboYu2: Originally NMATLSMode was used and later we decided to use HTTPSTLSMode. 
v1beta1 is using HTTPServerTLSSecret.  If we change NMATLSSecret to HttpsTLSSecret, we have to convert it between two different v1 versions. Do we need to create a v1.1 for that?

api/v1/verticadb_types.go
```diff
@@ -1002,6 +1022,15 @@ type SecretRef struct {
 	Type string `json:"type"`
 }
 
+type TLSMode struct {
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// TLS mode
+	Mode string `json:"name"`
```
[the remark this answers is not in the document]
LiboYu2: That is copied from "SecretRef".  It should be "name".

api/v1/helpers.go
```diff
@@ -1505,6 +1553,27 @@ func SetSecretRef(refs *[]SecretRef, newRef SecretRef) (changed bool) {
 	return changed
 }
 
+// SetTLSMode updates the slice with a new TLSMode by Type, and returns true if any changes occurred.
+func SetTLSMode(refs *[]TLSMode, newRef TLSMode) (changed bool) {
+	if refs == nil {
```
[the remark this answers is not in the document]
LiboYu2: That is a good question! It took me quite some time to figure out the answer. 
I think this nil checking is not required. Here is how to function is called:
      vapi.SetTLSMode(&vdb.Status.TLSModes, *tlsMode)
The refs is a pointer that stores the address of vdb.Status.TLSModes, which will never be nil. 
I will remove it.

tests/e2e-leg-12/scale-in-db-with-threshold-hpa/41-wait-for-prometheus.yaml
LiboYu2: I did not write the original test case. I added a "\n" at the end of the line.

pkg/controllers/vdb/tlsmode_reconciler.go
roypaulin, line 1: After yesterday's conversation, this should not be a separate reconciler, even temporarily. We are calling the same vclusterops api as httpscertrotation reconciler. There is a lot of code duplication. 
Moreover, a user can change the secret and tls mode at the same time. Normally, we would call rotate_https_cert with the new cert and tls mode, but with this we will call it twice, once for certrotation and the 2nd time for changing tls mode; that's not right.
https_cert_rotation reconciler only needs some minor changes to support changing tls mode.
Later when that new api (we discuss about yesterday) is available, we can rename httpscertrotation_reconciler and expand it even more to support setting tls config.

@cchen-vertica you are the main reviewer so I will let you decide if Libo should enforce this. But we already know that there will be one reconciler for tls so let's not add code we know we will remove.

pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -255,12 +255,12 @@ func (r *VerticaDBReconciler) constructActors(log logr.Logger, vdb *vapi.Vertica
...
 		MakeStatusReconciler(r.Client, r.Scheme, log, vdb, pfacts),
-		// Creates or updates any k8s objects the CRD creates. This includes any
-		// statefulsets and service objects.
-		MakeObjReconciler(r, log, vdb, pfacts, ObjReconcileModeAll),
 		// Set version info in the annotations and check that the deployment is
 		// compatible with the image.
 		MakeImageVersionReconciler(r, log, vdb, prunner, pfacts, false /* enforceUpgradePath */),
+		// Creates or updates any k8s objects the CRD creates. This includes any
+		// statefulsets and service objects.
+		MakeObjReconciler(r, log, vdb, pfacts, ObjReconcileModeAll),
```
roypaulin, line 263: Why did you move this reconciler?

roypaulin (changes requested):
@LiboYu2 you should merge main into your branch to get the changes I made.

pkg/controllers/vdb/httpscertrotation_reconciler.go
```diff
@@ -68,8 +68,8 @@ func (h *HTTPSCertRotationReconciler) Reconcile(ctx context.Context, _ *ctrl.Req
 	if h.Vdb.IsStatusConditionTrue(vapi.HTTPSCertRotationFinished) && h.Vdb.IsStatusConditionTrue(vapi.TLSCertRotationInProgress) {
 		return ctrl.Result{}, nil
 	}
-	currentSecretName := h.Vdb.GetNMATLSSecretNameInUse()
-	newSecretName := h.Vdb.Spec.NMATLSSecret
+	currentSecretName := h.Vdb.GetHTTPSTLSSecretNameInUse()
+	newSecretName := h.Vdb.Spec.HTTPSTLSSecret
```
roypaulin, line 72: We cannot straight out replace nmaTLSSecret with httpsTLSSecret. At least not until it is completely removed. We need to think about backward compatibility.
If a user is coming from an older version, they will have nmaTLSSecret set. We will ignore their secret and generate a new one because httpsTLSSecret is empty. We do not want that.
We also do not want the user to be able to set both nmaTLSSecret and httpsTLSSecret.

The scope of this PR is already pretty so you can do ir in a follow-up PR. Please just make sure you track it somewhere.
We need to do this before releasing.

roypaulin (changes requested):

pkg/controllers/vdb/verticadb_controller.go
LiboYu2: ObjReconciler needs to know the image version when building the pod. That way,  HTTP probe endpoint will be used instead of HTTPS one. That is for VERIFY_CA.

pkg/controllers/vdb/tlsmode_reconciler.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	sql := fmt.Sprintf("select mode from tls_configurations where name='%s';", tlsConfigName)
+	cmd := []string{"-tAc", sql}
+	stdout, stderr, err := h.PRunner.ExecVSQL(ctx, initiatorPod.GetName(), names.ServerContainer, cmd...)
+	h.Log.Info(fmt.Sprintf("%s tls mode from db - %s", tlsConfigName, stdout))
+	if err != nil || strings.Contains(stderr, "Error") {
+		h.Log.Error(err, fmt.Sprintf("failed to retrieve %s TLS mode after reviving db, stderr - %s", tlsConfigName, stderr))
+		return false, ctrl.Result{}, err
+	}
+	currentTLSMode = h.getTLSMode(stdout)
```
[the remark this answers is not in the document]
LiboYu2: That is for revive_db.

pkg/controllers/vdb/httpscertrotation_reconciler.go
LiboYu2: As discussed, we will make "nmaTLSSecret" immutable in webhook after it is initialized. And we will set the initial value of httpsTLSSecret from nmaTLSSecret if it is not empty. For cert rotation, users should use httpsTLSSecret

pkg/controllers/vdb/tlsservercertgen_reconciler.go
```diff
@@ -64,7 +64,6 @@ func (h *TLSServerCertGenReconciler) Reconcile(ctx context.Context, _ *ctrl.Requ
 			h.Log.Error(err, "failed to initialize httpsTLSSecret from nmaTLSSecret")
 			return ctrl.Result{}, err
 		}
-		h.Vdb.Spec.HTTPSTLSSecret = h.Vdb.Spec.NMATLSSecret
```
roypaulin: This is not needed. updating vdb will update the pointer.

pkg/controllers/vdb/verticadb_controller.go
roypaulin: There is already another MakeImageVersionReconciler before obj_reconciler at line 236. No need to change the order.

api/v1/verticadb_types.go
```diff
@@ -1038,7 +1038,7 @@ type SecretRef struct {
 type TLSMode struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=status
 	// TLS mode
-	Mode string `json:"name"`
+	Mode string `json:"mode"`
```
roypaulin, line 1041: It is best practice to have the json name match the firld name

pkg/controllers/vdb/httpscertrotation_reconciler.go
```diff
@@ -197,3 +255,22 @@ func GetAWSCertsConfig(vdb *vapi.VerticaDB) (keyConfig, certConfig, caCertConfig
 	caCertConfig = fmt.Sprintf("{\"json-key\":%q, \"region\":%q}", paths.HTTPServerCACrtName, region)
 	return
 }
+
+func (h *HTTPSCertRotationReconciler) updateTLSConfig() int {
```
roypaulin, line 259: precheck to know if any https tls field changed

pkg/controllers/vdb/httpscertrotation_reconciler.go
```diff
@@ -121,65 +127,117 @@ func (h *HTTPSCertRotationReconciler) Reconcile(ctx context.Context, _ *ctrl.Req
...
-			rotatehttpscerts.WithTLSMode(h.Vdb.Spec.HTTPSTLSMode),
-			rotatehttpscerts.WithInitiator(initiatorPod.GetPodIP()),
+	h.VRec.Eventf(h.Vdb, corev1.EventTypeNormal, events.HTTPSCertRotationSucceeded,
+		"Successfully rotated https cert with secret name %s and mode %s", h.Vdb.Spec.HTTPSTLSSecret, tlsData.tlsMode)
+
+	return err
+}
+
+// buildHTTPSTLSUpdateData constructs the data needed for tls update
+func (h *HTTPSCertRotationReconciler) buildHTTPSTLSUpdateData(ctx context.Context,
```
roypaulin, line 187: collect tls info for vclusterops in a single place

api/v1/verticadb_types.go
LiboYu2: Please try to avoid refactoring the code. My setup tls config branch relies on this branch. You are creating a lot of trouble for me.

roypaulin:
Also, please add unit tests. They are very important to make sure new code doesn't introduce bugs.

@cchen-vertica please make tls-related code has good unit test coverage. It can be done in this PR or in a follow up PR. But it has to be done.

pkg/controllers/vdb/httpscertrotation_reconciler.go
ningdeng: address the comments?

.github/actions/run-e2e-leg/action.yaml
```diff
@@ -192,6 +195,7 @@ runs:
 
           # Run test with cert rotation enabled
           export USE_CERT="${{ inputs.use-cert }}"
+          export E2E_TEST_DIRS=$E2E_TEST_DIRS_EXCLUDE_VC_ONLY
```
cchen-vertica, line 199: e2e-leg-6-vc-only won't be added to E2E_TEST_DIRS so the tests in it won't be run.  Check "USE_CERT" in line 188 and 189, then remove E2E_TEST_DIRS_EXCLUDE_VC_ONLY

cchen-vertica:
@LiboYu2 Could you rebase the branch on master? There are too many conflicted commits, and you should know which needs to keep and which doesn't.

cchen-vertica (approved):

PR #1199, opened 2025-04-15
Set archive max number of restore points when saving restore point

kcmackeen:
We can store a restore point by setting "status.conditions.SaveRestorePointNeeded" to true. This will automatically create an archive, if it does not exist. Previously, this archive would have max restore points set to 0, meaning we can save an unlimited number of restore points for this archive.

Now, the maximum number of restore points can be controlled by setting "spec.restorePoints.numRestorePoints". When this maximum is exceeded, we will overwrite the oldest restore point. Default is 0, meaning that number of restore points is still unlimited.


kcmackeen:
One problem I have realized with this fix:

The archive is only created the first time you set SaveRestorePointNeeded to True. Every subsequent time, it will re-use the existing archive (understandably so). So if you did something like

1. Set numRestorePoints: 2 in VDB
2. SaveRestorePointNeeded to True
3. Update VDB and set numRestorePoints: 1

The archive will still continue to have max restore points of 2. Which is probably confusing to a user.

To implement this, we would need the vcluster equivalent ofALTER ARCHIVE:

https://docs.vertica.com/25.1.x/en/sql-reference/statements/alter-statements/alter-archive/

I looked for something like this in vclusterops but could not find it.

roypaulin:
Looks good! Let's just add a webhook rule to make sure `numRestorePoints` is >= 0.

roypaulin:
> One problem I have realized with this fix:
> 
> The archive is only created the first time you set SaveRestorePointNeeded to True. Every subsequent time, it will re-use the existing archive (understandably so). So if you did something like
> 
> 1. Set numRestorePoints: 2 in VDB
> 2. SaveRestorePointNeeded to True
> 3. Update VDB and set numRestorePoints: 1
> 
> The archive will still continue to have max restore points of 2. Which is probably confusing to a user.
> 
> To implement this, we would need the vcluster equivalent of ALTER ARCHIVE:
> 
> https://docs.vertica.com/25.1.x/en/sql-reference/statements/alter-statements/alter-archive/
> 
> I looked for something like this in vclusterops but could not find it.

You are right. Please bring this up during the standup, you might end up creating a ticket.

kcmackeen:
See https://github.com/vertica/vertica-kubernetes/pull/1209

PR #1209, opened 2025-04-17
Set max number of restore points for archive when saving restore point

kcmackeen:
We can store a restore point by setting "status.conditions.SaveRestorePointNeeded" to true. This will automatically create an archive, if it does not exist. 

Previously, this archive would have max restore points set to 0, meaning we can save an unlimited number of restore points for this archive.

Now, the maximum number of restore points can be controlled by setting "spec.restorePoints.numRestorePoints". When this maximum is exceeded, we will overwrite the oldest restore point. Default is 0, meaning that number of restore points is still unlimited.

NOTE: this only applies when creating an archive. If you have an existing archive, it will not update the max number of restore points on it. This scenario can be handled in a separate PR (as it requires vcluster API changes).

(New PR; made a mistake doing a rebase)


api/v1beta1/verticadb_conversion.go
```diff
@@ -174,9 +174,10 @@ func convertToSpec(src *VerticaDBSpec) v1.VerticaDBSpec {
 	}
 	if src.RestorePoint != nil {
 		dst.RestorePoint = &v1.RestorePointPolicy{
-			Archive: src.RestorePoint.Archive,
-			Index:   src.RestorePoint.Index,
-			ID:      src.RestorePoint.ID,
+			Archive:          src.RestorePoint.Archive,
```
roypaulin, line 177: This is convertTo v1, you also need to update convertFrom v1.

roypaulin (commented):
It looks good! Cai will approve after you address my last comment.

api/v1/verticadb_types.go
```diff
@@ -405,6 +405,11 @@ type RestorePointPolicy struct {
 	// The identifier of the restore point in the restore archive to restore from.
 	// Specify either index or id exclusively; one of these fields is mandatory, but both cannot be used concurrently.
 	ID string `json:"id,omitempty"`
+	// +kubebuilder:default:=0
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// Number of restore points to save for this archive.
```
cchen-vertica, line 411: Modify the comment to: //Maximum number of restore points to store in the archive.

api/v1beta1/verticadb_types.go
```diff
@@ -459,6 +459,11 @@ type RestorePointPolicy struct {
 	// The identifier of the restore point in the restore archive to restore from.
 	// Specify either index or id exclusively; one of these fields is mandatory, but both cannot be used concurrently.
 	ID string `json:"id,omitempty"`
+	// +kubebuilder:default:=0
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// Number of restore points to save for this archive.
```
cchen-vertica, line 465: Ditto

cchen-vertica:
Add a changie file for this PR.

roypaulin (approved):

cchen-vertica (approved):

PR #1216, opened 2025-04-18
Support aws secret version id

roypaulin:
In aws secrets manager, a secret can have multiple versions that show how it changed over time. Previously, users had to create a new secret in aws secrets manager, each time they want to rotate the old one.
Now we allow users to pass a specific version of a secret in aws secrets manager.
The new format is: awssm://<secret-arn>@<version-id> where `version-id` is the specific version of the secret you want to use.
This applies to all secrets used in the CRs

Another change is the choice of caching the the current secret names in the status instead of the annotations. For that a new field `secretRefs` has been added to the status. It contains the names of the secrets currently in use.
Ex: spec.nmaTLSSecret = "sec1", spec.passwordSecret = "sec2" The status will look like this:
```
secretRefs:
- name: sec1
   type: NMATLSType
- name: sec2
   type: PASSWORDType
```
As you can see, this will be very easy to extend and we don't have to worry about users accidently changing them.


roypaulin:
@cchen-vertica I don't think we should use annotations for caching the secret:
- They have a limited size and user can edit them 
- We are diverting from what annotations are meant for, we should not store critical data there.
- Info about the current state should be in status.
- There is more than one secret we want to rotate so caching all of them in the annotations will complicate things
- We can easily cache all the secrets (that need rotation) in a vector in status. Users cannot edit it.

I strongly believe status makes more sense here because a secret name is a persistent critical data.

I open this draft to show you what I have in mind. I will continue on Tuesday.

roypaulin:
> So we are caching previous/old secret name in status, right?

Yes

roypaulin:
@cchen-vertica ready for review.

cchen-vertica:
Looks good. You could verify the secret name in status of step 17 and step 40 in e2e-leg-13.

tests/e2e-leg-13/nma-https-cert-rotate/17-assert.yaml
```diff
@@ -24,6 +24,8 @@ kind: VerticaDB
 metadata:
   name: v-tls-certs
 status:
+  secretRefs:
+    - type: NMATLSSecretType
```
roypaulin, line 28: Can't check the name here as it is randomly generated.

cchen-vertica (approved):

PR #1226, opened 2025-04-23
Fixed Issue with Restarting Pods in Sandbox Subcluster

cchen-vertica:
In K8s realdb, when a sandbox subcluster is shut down and then restarted, it may remain down and never come back up. This happens because, during the restart reconciliation, the pods in the sandbox subcluster are not in a Running state—so the reconciler skips the restart and does not requeue the task. As a result, the sandbox restart reconciler is never triggered again.
This PR addresses that issue.


roypaulin (approved):
Looks good!

PR #1230, opened 2025-04-24
support subclusters promotion and demotion

qindotguan:
This is the last part to support subcluster promotion and demotion. It will remove the type sandboxprimary from subcluster types and stop updating the subcluster type.

- Task 3. Remove the updates in spec.subclusters.subcluster.type

Previous tasks:

- #1114: Task 1: add type in status.subclusters
- #1135: Task 2: add type in spec.sandboxes.subclusters


api/v1/verticadb_webhook.go
```diff
@@ -200,7 +200,7 @@ func (v *VerticaDB) checkValidSubclusterTypeTransition(oldObj *VerticaDB, allErr
 		} else if oldType == SecondarySubcluster && sc.Type != SecondarySubcluster {
 			_, found := scToSbMap[sc.Name]
 			// You can only transition out of a secondary subcluster if its during sandboxing.
-			if sc.Type != SandboxPrimarySubcluster || !found {
+			if sc.IsSandboxPrimary(oldObj) || !found {
 				invalidStateTransitionErr(i)
```
roypaulin, line 204: Now that there is a type under spec.sandboxes, we should no longer allow spec.subclusters.type to change.

api/v1/verticadb_webhook.go
```diff
@@ -1292,6 +1292,22 @@ func (v *VerticaDB) validateProxyLogLevel(allErrs field.ErrorList) field.ErrorLi
...
+		}
+	}
+
+	err := field.Invalid(field.NewPath("spec").Child("sandboxes"),
+		sand.Subclusters,
+		fmt.Sprintf("there must be at least one primary subcluster in the sandbox %s", sand.Name))
+	allErrs = append(allErrs, err)
+
+	return allErrs
+}
```
roypaulin, line 1309: Refactor this to a boolean func that returns true if the sandbox has at least one primary sc. 
```suggestion
func (sand *Sandbox) hasPrimarySubcluster() bool {
	for i := range sand.Subclusters {
	        sc := &sand.Subclusters[i]
		if sc.Type == PrimarySubcluster {
			return true
		}
	}

	return false
}

func (v *VerticaDB) validateSandboxPrimarySubcluster(allErrs field.ErrorList, sbIndex int) field.ErrorList {
	sb := &v.Spec.Sandboxes[i]
	if sb != nil && !sb.hasPrimarySubcluster() {
		err := field.Invalid(field.NewPath("spec").Child("sandboxes").Index(sbIndex),
			sb.Subclusters,
			fmt.Sprintf("there must be at least one primary subcluster in the sandbox %s", sb.Name))
		allErrs = append(allErrs, err)
	}
	return allErrs
}
```
then call validateSandboxPrimarySubcluster in validateSubclustersInSandboxes

api/v1/verticadb_webhook.go
```diff
@@ -1832,31 +1848,57 @@ func (v *VerticaDB) checkSandboxPrimary(allErrs field.ErrorList, oldObj *Vertica
...
+			// calculate the number of primary subcluster nodes to be removed from the old sandbox
+			offsetMap[oldSbName] += int(sc.Size)
+			// check if the old sandbox has enough primary subcluster nodes
+			if !oldObj.DoesSandboxHaveQuorum(oldSbName, offsetMap[oldSbName]) {
+				i := oldScIndexMap[oldScName]
+				err := field.Invalid(path.Index(i),
+					oldObj.Spec.Subclusters[i],
+					fmt.Sprintf("quorum not meet, cannot remove the primary subcluster %q from sandbox %q", oldScName, oldSbName))
+				allErrs = append(allErrs, err)
+			}
```
roypaulin, line 1862: Question for @qindotguan and @cchen-vertica: I think we should not use the "quorum" term here. Previously, there was always only 1 primary in a sandbox so you could not remove that primary, now we can have more than one. Shouldn't it be possible to remove a primary if there will be at least one primary left? In the server What's the policy when unsandboxing a primary subcluster when there is more than one primary in the sandbox?

roypaulin (commented):
This the first batch of comments.

api/v1/verticadb_webhook.go
cchen-vertica: We should have the same limitations as the subclusters in main cluster: we cannot remove the primary subcluster if that will cause a quorum loss.

roypaulin: Currently, you can remove a primary subcluster from main as long as you respect the ksafety constraint (at least 3 primaries if ksafety is one). There is no "quorum" check for main cluster. that's why I am confused.

api/v1/helpers.go
```diff
@@ -1193,6 +1252,21 @@ func (v *VerticaDB) GetSandboxStatusCheck(sbName string) (*SandboxStatus, error)
 	return sb, nil
 }
 
+// DoesSandboxHaveQuorum returns true if the sandbox will keep quorum
+func (v *VerticaDB) DoesSandboxHaveQuorum(sbName string, offset int) bool {
```
roypaulin, line 1256: I don't fully understand this function. Can you add more details with examples.

api/v1/verticadb_webhook.go
qindotguan: Should we update to check the "quorum" for the main cluster as well? In the server, the DB requires that more than half of the primary nodes in the database be up:
https://docs.vertica.com/25.2.x/en/glossary/primary-node/

api/v1/verticadb_webhook.go
qindotguan: Thanks. Refactored.

api/v1/helpers.go
qindotguan: Example added.

api/v1/helpers.go
```diff
@@ -1203,6 +1263,32 @@ func (v *VerticaDB) GetSandboxStatusCheck(sbName string) (*SandboxStatus, error)
...
+	scMap := v.GenSubclusterMap()
+	sb := v.GetSandbox(sbName)
+	for i := range sb.Subclusters {
+		sc := sb.Subclusters[i]
+		if sc.Type != PrimarySubcluster {
+			continue
+		}
+		totalPrimaryCount += int(scMap[sc.Name].Size)
+	}
+	return 2*(totalPrimaryCount-offset) > totalPrimaryCount
```
cchen-vertica, line 1289: We should calculate the up primary nodes. You can get the node state from status.

api/v1/verticadb_webhook.go
```diff
@@ -1333,6 +1356,9 @@ func (v *VerticaDB) validateSubclustersInSandboxes(allErrs field.ErrorList) fiel
 		for sc, index := range scInSandbox {
 			seenScWithSbIndex[sc] = index
 		}
+
+		// sandboxHasPrimarySubcluster checks if there is a primary subcluster in the sandboxe
```
cchen-vertica, line 1360: Minor: fix "sandboxe"

api/v1/verticadb_webhook.go
cchen-vertica: If we can remove a primary subcluster from main, that's good. I couldn't remove a primary subcluster from main last Friday in Kevin's real-db, that could be a bug. 

When we remove/demote the primary subcluster, we shouldn't do the quorum check since the base of the quorum decreases. Instead, we need to make sure the node number matches the ksafety(the formula is probably num-of-primary-nodes >= 2*ksafety + 1).

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
roypaulin, line 1: A major change you need to make is altering the sandbox subclusters type in the db, to match the type in sandbox.subcluster.type. There is already a reconciler for that, and it is being used in online upgrade. The first sc in a sandbox is the primary (If I remember correctly it is still the case. correct me if I am wrong) and all the others will be secondary. So after you sandbox a subcluster, you need to call that reconciler (if needed) to apply the desired type.

We can discuss about that.

roypaulin (commented):
I think there is a misunderstanding, promotion/demotion here means from primary to secondary and vice-versa.

The e2e test to add must create a sandbox with 3 scs and make sure their type in the db matches the expected type.

For simplicity I suggest we limit this to sandboxes for now.

pkg/controllers/vdb/status_reconciler.go
```diff
@@ -182,26 +182,29 @@ func (s *StatusReconciler) calculateClusterStatus(stat *vapi.VerticaDBStatus) {
 	}
 }
 
-// getSubclusterStatusType returns the subcluster status type depends on its type in subclusters and sandboxes
-func (s *StatusReconciler) getSubclusterStatusType(sc *vapi.Subcluster) string {
+// calculateSubclusterStatusType will figure out the statusw type for the given subcluster
```
cchen-vertica, line 185: Minor: change "statusw" to "status". 

Also, as we discussed, the status of subcluster should be calculated using podfacts(sandbox-name and subcluster-type from FetchNodeDetails).

api/v1/verticadb_webhook.go
qindotguan: Thanks. I removed the quorum check from the sandbox and updated it to check if there is at least one primary in the sandbox in hasPrimaryAfterRemoval(). 

For the main cluster, only check k-safety when a primary subcluster is removed or demoted in IsKSafetyAfterRemoval().

api/v1/verticadb_webhook.go
qindotguan: Updated to not allow subcluster type change if it's in a sandbox.

api/v1/helpers.go
qindotguan: DoesSandboxHaveQuorum() was removed. Now when calculating k-safety in IsKSafetyAfterRemoval(), it uses UpNodeCount in the subcluster status.

api/v1/verticadb_webhook.go
qindotguan: As discussed, changed to do the same k-safety check for both sandbox and main cluster.

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -616,7 +616,7 @@ func (r *OnlineUpgradeReconciler) promoteReplicaBSubclusters(ctx context.Context
 		return ctrl.Result{}, err
 	}
 	sbPFacts.Invalidate()
-	actor := MakeAlterSubclusterTypeReconciler(r.VRec, r.Log, r.VDB, sbPFacts, r.Dispatcher)
+	actor := MakeAlterSubclusterTypeReconciler(r.VRec, r.Log, r.VDB, sbPFacts, r.Dispatcher, true /* forUpgrade */)
```
roypaulin, line 619: If you add alter subcluster to sandboxsubcluster_reconiler, do you still need alter here?

roypaulin:
@cchen-vertica @qindotguan I have a suggestion:
sandboxsubcluster_reconciler is now overcrowded, and this might slow down sandboxing.
Why not let sandbox controller handle subcluster type change.
Here is the workflow:
- sandboxsubcluster_reconciler: create a sandbox (make sure the 1st subcluster tobe sandboxed has type primary in spec) and its configmap.
- sandbox_controller: once the configmap is created or a new subcluster is added, sandbox_controller reconcile that sandbox and enforce the subclusters type.

That way we delegate sandbox jobs to sandbox_controller and and we don't slow down sandboxsubcluster_reconciler and verticadb_controller.

Moreover, later when we want to allow user to change the type of an existing subcluster, we can let sandbox_controller handle subclusters in a sandbox and verticadb_controller handle subclusters in the main.

pkg/controllers/vdb/altersubcluster_reconciler.go
```diff
@@ -83,9 +90,26 @@ func (a *AlterSubclusterTypeReconciler) findSandboxSubclustersToAlter() ([]*vapi
...
+			if found && targetType == vapi.PrimarySubcluster && !sc.IsPrimary(a.Vdb) {
+				scs = append(scs, sc)
+			}
+		} else {
+			// for sandbox subcluster with multiple primary subclusters, vclusterops only
+			// set the first one as primary
+			// if sandbox subcluster type is primary but podfacts (from database) is_primary is false,
+			// we need to change the subcluster is_primary to true in the database
+			pn := names.GenPodName(a.Vdb, sc, int32(i))
+			pf, ok := a.PFacts.Detail[pn]
```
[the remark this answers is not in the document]
qindotguan: Thanks, updated to use `PFacts.FindFirstUpPod`.

pkg/controllers/vdb/altersubcluster_reconciler.go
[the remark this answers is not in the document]
qindotguan: I updated findSandboxSubclustersToAlter() to include sc with both promotion and demotion. Later we meed to monitor the sandbox subcluster type updates and call AlterSubclusterTypeReconciler directly.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -156,19 +154,15 @@ func (s *SandboxSubclusterReconciler) sandboxSubclusters(ctx context.Context) (c
 	}
 	s.PFacts.Invalidate()
 
-	return ctrl.Result{}, nil
+	return s.updateSandboxSubclusterType(ctx, scSbMap)
```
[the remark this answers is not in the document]
qindotguan: Good point. I moved updateSandboxSubclusterType() into executeSandboxCommand() right after the subcluster sandboxed.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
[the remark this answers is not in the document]
qindotguan: Nice use case. Before sandboxing, I changed to sort the subcluster by its type (primary, secondary) first, so that it will always sandbox the primary subcluster.

pkg/controllers/vdb/onlineupgrade_reconciler.go
qindotguan: It should not. When duplicating the main cluster, the subcluster types are recorded into the sandbox subcluster types in moveReplicaGroupBSubclusterToSandbox(). 
So onlineupgrade_reconciler should not alter it again. I will do more testing before removing this.

roypaulin:
> @cchen-vertica @qindotguan I have a suggestion: sandboxsubcluster_reconciler is now overcrowded, and this might slow down sandboxing. Why not let sandbox controller handle subcluster type change. Here is the workflow:
> 
> * sandboxsubcluster_reconciler: create a sandbox (make sure the 1st subcluster tobe sandboxed has type primary in spec) and its configmap.
> * sandbox_controller: once the configmap is created or a new subcluster is added, sandbox_controller reconcile that sandbox and enforce the subclusters type.
> 
> That way we delegate sandbox jobs to sandbox_controller and and we don't slow down sandboxsubcluster_reconciler and verticadb_controller.
> 
> Moreover, later when we want to allow user to change the type of an existing subcluster, we can let sandbox_controller handle subclusters in a sandbox and verticadb_controller handle subclusters in the main.

The more I look at this the more I think we have to do it. Think about RealDB with all those sandboxes each with potentially many subclusters. when creating multiple sandboxes, it will be very slow because you have to also change some subclusters' type. And later when we support changing the type of a subcluster already in  a sandbox, it will slow down certicadb_controller if many users are changing their subclusters' type. Thy might affect RealDB perf which is already slow.
I will try this myself, just focus on finishing this PR.

pkg/controllers/vdb/status_reconciler.go
```diff
@@ -182,26 +182,32 @@ func (s *StatusReconciler) calculateClusterStatus(stat *vapi.VerticaDBStatus) {
...
-	// if in a sandbox, set subcluster type according to the sandbox status
-	sandboxStatus := s.Vdb.GetSandboxStatus(s.PFacts.SandboxName)
-	isSandbox := sandboxStatus != nil && sandboxStatus.Name != vapi.MainCluster
-	if isSandbox {
-		if sc.IsSandboxPrimary() {
+	pf, ok := s.PFacts.FindFirstUpPod(true, sc.Name)
+	// If we don't have any up pods (could be read-only) in the subcluster, we
+	// can't determine the type of the subcluster.
+	if !ok {
+		s.Log.Info("No up pods found in the subcluster to calculate its status type",
```
cchen-vertica, line 197: Why do we need an up pod to determine the type? Are you worried podfact won't be accurate? I think we can simply trust the podfacts even when we don't have an up node. When you return "", the type won't be more accurate than the podfacts's.

pkg/controllers/vdb/sandboxsubcluster_reconciler.go
```diff
@@ -154,25 +153,32 @@ func (s *SandboxSubclusterReconciler) sandboxSubclusters(ctx context.Context) (c
...
 	// in a sandbox is the primary.
 	for i := range s.Vdb.Spec.Sandboxes {
-		vdbSb := &s.Vdb.Spec.Sandboxes[i]
+		sbs := &s.Vdb.Spec.Sandboxes[i]
 		var sbName string
 		sbScs := []string{}
+
+		// sort sandbox subclusters according to the order of sandbox subclusters type (primary, secondary)
+		// to make sure the first subcluster is always sandbox primary
+		vdbSb := sbs.DeepCopy()
```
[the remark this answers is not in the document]
qindotguan: You mean removing sorting? We should sort it in case the first subcluster in sandbox is secondary:

pkg/controllers/vdb/status_reconciler.go
qindotguan: You are right; we can trust the podfact. In podcats, the initial type is from vdb, and then it's set according to NodeDetails, which requires a running pod.

pkg/controllers/vdb/altersubcluster_reconciler.go
```diff
@@ -101,35 +111,18 @@ func (a *AlterSubclusterTypeReconciler) alterSubclusters(ctx context.Context, sc
...
-		err = a.updateSubclusterTypeInVDB(ctx, sc)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
 	}
 	return ctrl.Result{}, nil
 }
 
-// updateSubclusterTypeInVDB updates the given subcluster's type in VDB
-func (a *AlterSubclusterTypeReconciler) updateSubclusterTypeInVDB(ctx context.Context, sc *vapi.Subcluster) error {
```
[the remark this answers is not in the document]
qindotguan: It's removed because this was to update vdb subcluster type from secondary to sandboxprimary. In the next PR, we can add it back (in this reconcile or a new one) and update to handle promotion and demotion.

pkg/controllers/vdb/altersubcluster_reconciler.go
```diff
@@ -101,35 +111,18 @@ func (a *AlterSubclusterTypeReconciler) alterSubclusters(ctx context.Context, sc
 		if err != nil {
 			return ctrl.Result{}, err
 		}
-		a.PFacts.Invalidate()
```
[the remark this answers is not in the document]
qindotguan: Yes, this reconciler only updates the type in the database now. We can add this back later for promotion and demotion cases.

pkg/controllers/vdb/altersubcluster_reconciler.go
```diff
@@ -39,6 +37,7 @@ type AlterSubclusterTypeReconciler struct {
 	Vdb        *vapi.VerticaDB // Vdb is the CRD we are acting on.
 	PFacts     *podfacts.PodFacts
 	Dispatcher vadmin.Dispatcher
+	IsUpgrade  bool
```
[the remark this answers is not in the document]
qindotguan: Removed. Thanks.

PR #1237, opened 2025-04-28
Add brackets to IPv6 address to check if node up

qindotguan:
On the IPv6-enabled k8s cluster, the operator would restart the Vertica pods after the database creation. The root cause is that it could not detect the node's up status with the IPv6 URL. A bracket was added to the IPv6 address to resolve the issue.


cchen-vertica:
Add a changie file for the fix.

changes/unreleased/Fixed-20250428-160725.yaml
```diff
@@ -0,0 +1,5 @@
+kind: Fixed
+body: Added brackets for IPv6 address to check if nodes up
```
cchen-vertica, line 2: Refine it to "Resolve the issue where the cluster keeps restarting when using IPv6."

PR #1250, opened 2025-05-01
Added retry times for build process

cchen-vertica:
Build process is not stable recently so I add retry process to it. By default, the build will automatically retry up to 3 times upon failure.

### DIFF
--- a/docker-vertica-v2/packages/package-checksum-patcher.py
+++ b/docker-vertica-v2/packages/package-checksum-patcher.py
@@ -123,7 +123,11 @@ def patch_dir(dir, old_checksum):
      None
     """
     libname = 'lib' + os.path.basename(dir) + '.so'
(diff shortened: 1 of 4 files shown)


docker-vertica-v2/packages/package-checksum-patcher.py
```diff
@@ -123,7 +123,11 @@ def patch_dir(dir, old_checksum):
      None
     """
     libname = 'lib' + os.path.basename(dir) + '.so'
-    xsum_out = str(subprocess.check_output(['md5sum', dir + '/lib/' + libname]))
+    lib_path = os.path.join(dir, 'lib', libname)
+    if not os.path.isfile(lib_path):
```
cchen-vertica, line 127: This change is for fixing devops CICD problem: the images after 04/23/2025 cannot be pushed to dockerhub due to missing .so file.

roypaulin (approved):

PR #1262, opened 2025-05-06
Fix sandbox_shutdown

roypaulin:
In a previous PR, I decided to call sandboxshutdown _reconciler early in the reconcile loop, to prevent some failures to affect some sandbox ops. The issue is that it might be possible that (ex: realdb) someone is adding a new sandbox +subclusters and that will fail sandboxshutdown _reconciler. The 2nd sandboxshutdown _reconciler does not have this issue because it happens towards the end when all critical reconcilers have been run.
To fix this, I added a field IgnoreError that is true for the 1st sandboxshutdown _reconciler. When true, if a subcluster is not yet in the status, we just skip that sc knowing a later reconciler will add it to the db.

### DIFF
--- a/pkg/controllers/vdb/shutdownspec_reconciler.go
+++ b/pkg/controllers/vdb/shutdownspec_reconciler.go
@@ -59,10 +59,21 @@ func (r *ShutdownSpecReconciler) updateSubclustersShutdownState(ctx context.Cont
 func (r *ShutdownSpecReconciler) updateSubclustersShutdownStateCallback() (bool, error) {
 	needUpdate := false
 	scMap := r.Vdb.GenSubclusterMap()
+	scStatusMap := r.Vdb.GenSubclusterStatusMap()
 	for i := range r.Vdb.Spec.Sandboxes {
 		sb := &r.Vdb.Spec.Sandboxes[i]
+		sbStatus := r.Vdb.GetSandboxStatus(sb.Name)
+		if sbStatus == nil {
+			continue
+		}
 		for j := range sb.Subclusters {
-			sc := scMap[sb.Subclusters[j].Name]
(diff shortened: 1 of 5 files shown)


pkg/controllers/vdb/shutdownspec_reconciler.go
```diff
@@ -59,10 +59,21 @@ func (r *ShutdownSpecReconciler) updateSubclustersShutdownState(ctx context.Cont
...
+		if sbStatus == nil {
+			continue
+		}
 		for j := range sb.Subclusters {
-			sc := scMap[sb.Subclusters[j].Name]
+			scName := sb.Subclusters[j].Name
+			sc := scMap[scName]
+			scStatus := scStatusMap[scName]
+			if !sbStatus.IsSubclusterInSandbox(scName) || scStatus == nil {
+				break
```
cchen-vertica, line 74: Why do we "break" here? If one subcluster isn't in the sandbox, we skip updating the annotation for the other subclusters in the sandbox?

roypaulin: shutdownspec_reconciler is called twice: early in the reconcile loop and towards the end after sandboxsubcluster_reconciler. we "break" because we know a subcluster not being in the sandbox means it first have to go through sandbox_subcluster. We rely on the 2nd shutdownspec_reconciler to then update the annotation after sandboxsubcluster_reconciler.
So we can only get to this "break" in the 1st shutdownspec_reconciler.
Keep in mind this is very unlikely to happen in the first place, especially with the webhook rules we have.


Note: I removed the 2nd shutdownspec_reconciler call in a previous PR. I am adding it back.

cchen-vertica (approved):

PR #1270, opened 2025-05-13
Added leg5 cert test

cchen-vertica:


roypaulin:
Not directly related to this but now that we removed ddls from createdbreconciler, we need an even more recent base image for upgrade tests. One that has those nma changes. So 20250513. That's probably why those tests fail.

tests/e2e-leg-8-online-cert/online-upgrade-down-cluster/00-create-communal-creds.yaml
roypaulin, line 1: Don't we want to keep running this without cert too?

roypaulin (approved):

PR #1277, opened 2025-05-19
Add tls config vclusterops api

roypaulin:
Add the new set_tls_config vclusterops api to the operator.


pkg/vadmin/set_tls_config_vc.go
```diff
@@ -0,0 +1,77 @@
...
+	certs *HTTPSCerts) *vops.VSetTLSConfigOptions {
+	opts := vops.VSetTLSConfigOptionsFactory()
+
+	opts.RawHosts = append(opts.RawHosts, s.InitiatorIP)
+	opts.DBName = v.VDB.Spec.DBName
+
+	opts.IsEon = v.VDB.IsEON()
+	opts.IPv6 = net.IsIPv6(s.InitiatorIP)
+
+	opts.Key = certs.Key
```
cchen-vertica, line 65: Could we use v.setAuthentication() to configure the credentials?

roypaulin: Now that I think about it, I think we always need to pass the password as before calling this there's no TLS config so we can only rely on the password for running the ddls with vertica sql go library.

pkg/vadmin/set_tls_config_vc.go
```diff
@@ -0,0 +1,76 @@
...
+
+	opts.RawHosts = append(opts.RawHosts, s.InitiatorIP)
+	opts.DBName = v.VDB.Spec.DBName
+
+	opts.IsEon = v.VDB.IsEON()
+	opts.IPv6 = net.IsIPv6(s.InitiatorIP)
+
+	v.setAuthentication(&opts.DatabaseOptions, v.VDB.GetVerticaUser(), &v.Password, certs)
+	if v.Password != "" {
+		opts.Password = &v.Password
```
roypaulin, line 67: This is needed to run those ddls and set tls config.

PR #1278, opened 2025-05-19
Add new tool to generate tls cert

roypaulin:
This adds a new binary that can be used to generated a valid tls secret for nma/https and client-server. This can be handy when you do not want to rely on the operator to automatically generate a secret as that secret will get deleted with the vdb. This will also make testing easier as we won't have to copy an autogenerated secret and won't have to manually create a tls secret (which might be invalid). We also do not have to store secret yaml files for the e2e tests.

This tool can be improved and even exposed to users and QA.


pkg/security/generate.go
```diff
@@ -105,3 +110,25 @@ func NewCertificate(ca Certificate, commonName string, dnsNames []string) (Certi
 		tlsCrt: pem.EncodeToMemory((&pem.Block{Type: "CERTIFICATE", Bytes: keyCert})),
 	}, nil
 }
+
+func GetDNSNames(namespace string) []string {
+	return []string{
+		fmt.Sprintf("*.%s.svc", namespace),
```
cchen-vertica, line 116: We should add DNS for cluster-ip and load-balancer-ip(if exists) as well. If it takes much time, you can create a follow-up ticket for that.

tests/e2e-leg-4/revivedb-3-node/12-create-client-cert.yaml
```diff
@@ -14,4 +14,4 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
-  - command: sh -c "envsubst < client_cert.yaml | kubectl apply -n $NAMESPACE -f -"
+  - script: cd ../../.. && make create-tls-secret SECRET_NAME=client-cert SECRET_NAMESPACE=$NAMESPACE DB_USER=$VERTICA_SUPERUSER_NAME
```
cchen-vertica, line 17: Do we need to do this for other legs1(,2,5,6,7,8,9, upgrade) or only revive tests need to do this change?

roypaulin: Yes, we should do this for any tests that need custom cert as it will make sure we are running tests with valid cert.

roypaulin:
I did not cover all the tests here, I first want the tool to go in. I added the cluster ip and load balancer ips.

For future improvements we can open tickets

cmd/cert-gen/main.go
```diff
@@ -45,15 +46,19 @@ func encodeSecretData(secret *corev1.Secret) map[string]string {
...
-	fmt.Printf("Usage: %s <secret-name> <secret-namespace> <common-name>\n", os.Args[0])
+	fmt.Printf("Usage: %s [OPTIONS] <secret-name> <secret-namespace> <common-name>\n", os.Args[0])
 	flag.PrintDefaults()
 }
 
 func main() {
 	opts := certgen.Options{}
 	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	flag.Usage = usage
+	flag.StringVar(&opts.ClusterIPs, "cluster-ips", "", "Comma-separated list of cluster IPs")
```
cchen-vertica, line 57: My bad, I found ClusterIP was not useful. It should be the node's IP or host(for NodePort service). We can use only one flag: "extra-san-entries"(comma-separated list of customized Subject Alternative Name entries). Then people can put their desired IPs or hostnames in that flag. We shouldn't validate the IPs since it could be hosts or URIs.

cchen-vertica (approved):

PR #1279, opened 2025-05-20
Remove v1beta1 VDB API; remove webhook to convert v1beta1 VDB to v1

kcmackeen:
Remove v1beta1 VerticaDB API, as well as webhook to convert from v1beta1 to v1. This required several small changes throughout the code to use v1 version of functions instead of v1beta1. Unit tests are now passing using just v1 VDB.

Also, I had to make **MANY** changes to the e2e tests so that all tests only use v1 VDB. This also required changing the syntax for many setup-vdb.yaml files, for example to use:

```
  annotations:
    vertica.com/k-safety: "0"
```

Instead of:

`  kSafety: "0"
`

Some important notes from this PR:

1) We can no longer upgrade (directly or indirectly) from 1.X versions to 25.3+. This is because v1 was introduced in 2.0. As of 2.0, any VDB created in v1beta1 will be auto-converted to v1. However, any VDB created before this version and then upgraded will never follow the conversion logic, and it will remain as v1beta1. With v1beta1 removed in 25.3, this VDB will no longer be supported. It must be re-created using v1 API.
2) We cannot upgrade the verticadb CRD from <25.3 to 25.3+ using "kubectl apply". This is because the <25.3 CRD will contain webhook conversion logic, and "kubectl apply" cannot overwrite this. We must use "kubectl patch"; for example, use the below command to clear conversion before using "kubectl apply":

kubectl patch crd verticadbs.vertica.com --type merge -p '{\"spec\":{\"conversion\":null}}'
3) In v1beta1, if encryptSpreadComm was not specified, it would be saved as "disabled"; in v1, it will now be "vertica"
4) httpServerMode was deprecated and was only included in v1beta1; it has now been removed


api/v1beta1/verticaautoscaler_types.go
```diff
@@ -70,7 +71,7 @@ type VerticaAutoscalerSpec struct {
 	// here it will be used as a prefix for the new subcluster.  Otherwise, we
 	// use the name of this VerticaAutoscaler object as a prefix for all
 	// subclusters.
-	Template Subcluster `json:"template"`
+	Template v1.Subcluster `json:"template"`
```
roypaulin, line 74: Let's v1beta1.Subcluster for v1beta1 VerticaAutoscaler.Template, for backward compatibility. Next PI we plan to remove v1beta1 VerticaAutoscaler too anyway.

pkg/security/webhook_test.go
```diff
@@ -88,19 +88,6 @@ var _ = Describe("webhook", func() {
 		Expect(files[1].Name()).Should(Equal(corev1.TLSPrivateKeyKey))
 	})
 
-	It("should add annotations to the CRD", func() {
```
roypaulin, line 91: Why did you remove this? I don't see how it is related to v1beta VDB

pkg/security/webhook_test.go
```diff
@@ -112,7 +99,6 @@ var _ = Describe("webhook", func() {
 		defer deleteSecret(ctx, secretName)
 
 		Expect(PatchConversionWebhookFromSecret(ctx, &logger, restCfg, secretName, prefixName, ns)).Should(Succeed())
-		verifyCertForConversionEquals(ctx, mockCert)
```
roypaulin, line 115: Instead of removing this line, could we this test case for verticaautoscaler? At line 92 use verticaautoscaler. We no longer need this for VerticaDB as we are removing the conversion webhook.

roypaulin (commented):
You should also update config/crd/patches/webhook_in_verticadbs.yaml.

tests/e2e-server-upgrade/offline-upgrade-rename-cm/50-assert.yaml
qindotguan, line 1: The same to offline-upgrade-rename-cm/40-assert.yaml, do the change at line 24:
image

tests/e2e-server-upgrade/online-upgrade-kill-transient/setup-vdb/base/setup-vdb.yaml
```diff
@@ -11,31 +11,32 @@
...
       size: 1
-      isPrimary: true
+      type: primary
     - name: sec
       size: 1
-      isPrimary: false
+      type: secondary
   temporarySubclusterRouting:
     template:
       name: transient
```
qindotguan, line 37: Haven't tested this yet, but if it failed, try to add type as secondary in the template:
```suggestion
      name: transient
      type: secondary
```

roypaulin:
@kcmackeen Please prepare your env for Tuesday. We will manually debug the failed tests so we can merge this week. Please remind me of this during the standup if I forget.

roypaulin:
@kcmackeen I tried your branch in my local machine and that test every time. I ended up running that test in a separate job and now it passes. This is now ready to be merged when the dependencies will be resolved.

roypaulin:
Once you do that and the tests pass, we can merge this.

roypaulin:
create-enable-rotate-add still fails because "installCount" does not exist in v1 so remove the status from 15-assert.yaml.

roypaulin (commented):
Some things that I missed:
- add a changie entry with type "Remove"
- Remove v1beta1_verticadb.yaml file also from config/samples/kustomization.yaml

Merge again the main into your branch.

PR #1282, opened 2025-05-21
Remove owner reference from tls secrets

roypaulin:
When mutual tls is enabled, we do not want to delete auto-generated secrets as they could be reused for revive_db.
This will remove the vdb owner reference from the tls secrets only when the version supports mutual tls.


roypaulin:
To test it I updated revivedb-3-node. I removed the steps where the certs are created so we will use auto-generated tests for revivedb.

tests/e2e-leg-13/nma-https-cert-rotate/95-errors.yaml
```diff
@@ -11,7 +11,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
cchen-vertica, line 13: Could we add an assert step here to verify the secret is still there:
apiVersion: v1
kind: Secret
metadata:
  labels:
    app.kubernetes.io/name: vertica
    app.kubernetes.io/component: database
    vertica.com/database: ...
    app.kubernetes.io/instance: ...
  name: custom-cert

roypaulin: YES, will do it in my upcoming PR.

PR #1283, opened 2025-05-21
Applied New Option UsePasswordForSQLClientOnly in Operator

cchen-vertica:
When establishing a TLS connection for HTTPS and a non-TLS connection for Client-Server, we need to provide both the TLS certificates and the password to vclusterOps, and set UsePasswordForSQLClientOnly to true. I’ve added this logic to the vclusterOps invocation.


api/v1/helpers.go
```diff
@@ -1482,6 +1482,27 @@ func (v *VerticaDB) GetNMATLSSecretNameInUse() string {
...
 }
 
+// IsCertNeededForClientServerAuth returns true if certificate is needed for client-server authentication
+func (v *VerticaDB) IsCertNeededForClientServerAuth() bool {
+	tlsMode := strings.ToLower(v.Spec.ClientServerTLSMode)
+	return tlsMode != tlsModeDisable && tlsMode != tlsModeEnable
+}
+
+// GetNMAClientServerTLSMode returns the tlsMode for NMA client-server communication
+func (v *VerticaDB) GetNMAClientServerTLSMode() string {
```
roypaulin, line 1492: Why don't we straight out return the mode in VDB? If there is a valid reason you should add it to the description or add a comment.

pkg/builder/builder.go
```diff
@@ -85,6 +85,7 @@ const (
 	NMASecretNameEnv            = "NMA_SECRET_NAME"             // #nosec G101
 	NMAClientSecretNamespaceEnv = "NMA_CLIENT_SECRET_NAMESPACE" // #nosec G101
 	NMAClientSecretNameEnv      = "NMA_CLIENT_SECRET_NAME"      // #nosec G101
+	NMAClientSecretTLSModeEnv   = "NMA_CLIENT_TLS_MODE"         // #nosec G101
```
roypaulin, line 88: Does NMA ever read this? Are we going to add the same for https?

LiboYu2: No, we don't need to add that for https.

LiboYu2: Why do we need to pass this new environmental variable to NMA? Do they use it to double check our inputs?

api/v1/helpers.go
```diff
@@ -1482,6 +1482,27 @@ func (v *VerticaDB) GetNMATLSSecretNameInUse() string {
...
+	tlsMode := strings.ToLower(v.Spec.ClientServerTLSMode)
```
LiboYu2, line 1493: We can move this validation logic to webhook. If users enter an incorrect tls mode, they have to fix it. When we use it, we don't need to do any validation.

api/v1/helpers.go
cchen-vertica: This mode is NMA client server tls mode, not Vertica client server tls mode. I added some new constants for making it clear.

api/v1/helpers.go
cchen-vertica: This isn't a validation. This is a conversion.

PR #1292, opened 2025-05-29
Modify annotations for TLS authentication

kcmackeen:
Currently, to enable TLS authentication, we must set two different annotations:

- enable-tls-certs-rotation must be set to True
- mount-nma-certs must be set to False

This is inconvenient. We should be able to set a single annotation to enable/disable TLS. So, we have renamed:

`enable-tls-certs-rotation`

To

`enable-tls-auth`

And this will now be the only annotation required for TLS; mount-nma-certs will no longer impact TLS.

We also have a requirement that mount-nma-certs be set to False by default, for security reasons.


LiboYu2:
When "enable-tls-auth" and "mount-nma-certs" are both set to true, what will happen? @cchen-vertica Could you confirm that?
I expect "mount-nma-certs" to be overwritten to "false" in this case.  Maybe this should be done in my current setup tls config branch.

LiboYu2 (approved):

roypaulin:
I heard from Cai that we won't allow tls if mount-nma-cert is true. Are you going to add that check in a separate ticket?

kcmackeen:
> Merge the main into your branch and rerun the test.

Done!

cchen-vertica (approved):

PR #1296, opened 2025-06-03
Add client server tls reconciler

roypaulin:
This adds a new reconciler that will handle tls config, cert rotation and/or tls mode change for the client server tls config.

Due to the similarities between clientservertls_reconciler and httpstls_reconciler, I moved the shared logic to a separate file controller by a new struct "TLSConfigManager".


cchen-vertica:
> Question:
> 
> * after client server cert rotation, do we also need to run nma cert rotation as we do for https'?

Yes, we need. Both https and client-server cert rotation should follow an nma cert rotation(nma restart). So client-server cert rotation should create another condition like https.

roypaulin:
Work in progress!

roypaulin:
Another round of improvements.
still to do :
- add unit tests
- add comments around the code
- add more e2e tests
- remove gettlsconfigmodeafterrevive_reconciler
- improve nmacert rotation reconciler
- rename rotate_https_cert to rotate_tls_cert

LiboYu2:
Here is my advice:
Your work should focus on client server rotation and my PR takes care of everything else.
In my PR, I've created reusable code to handle client server tls setup separately. You should 
not change the scope of your work. Now there are many conflicts. We should have a meeting
to talk about it.

roypaulin:
Another round of improvements.
still to do :
- add more e2e tests
- add more comments around the code

PR #1297, opened 2025-06-03
Run leg-10 with cert

roypaulin:


roypaulin:
All leg-10 tests pass except vrep's.

PR #1305, opened 2025-06-09
Add validation for TLS certificate

kcmackeen:
Create validation for secret certification. We want to check the following:

1. Can certificate be decoded correctly?
2. Does Extended Key Usage have Client Auth and Server Auth?
3. Are certificate dates correct? Is it expired or not yet valid?
4. If the certificate is about to expire, alert the user
5. For CA certs only, does it have Subject Key Identifier?

Also in this PR, I made a change to rename IsCertRotationEnabled to IsTLSAuthEnabled (since this is more accurate).

### DIFF
--- a/.github/workflows/build-release-assets.yml
+++ b/.github/workflows/build-release-assets.yml
@@ -31,6 +31,12 @@ jobs:
         make vdb-gen
         ls -lhrt bin/
 
+    - name: Build cert-gen
+      run: | 
+        mkdir -p bin
+        make cert-gen
+        ls -lhrt bin/
+
     - name: Build Release yaml
       run: |
         IMG_REPO=docker.io/opentext/ make config-transformer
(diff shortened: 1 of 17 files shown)


pkg/controllers/vdb/tlsservercertgen_reconciler.go
```diff
@@ -57,6 +58,10 @@ func MakeTLSServerCertGenReconciler(vdbrecon *VerticaDBReconciler, log logr.Logg
 
 // Reconcile will create a TLS secret for the http server if one is missing
 func (h *TLSServerCertGenReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	secretStatus := h.Vdb.GetSecretStatus(httpsNMATLSSecret)
+	if secretStatus != nil && secretStatus.Name == h.Vdb.Spec.HTTPSNMATLSSecret {
```
roypaulin, line 62: The condition should also include clientServerTLSSecret.

tests/e2e-leg-9/invalid-cert-validation/00-create-creds.yaml
```diff
@@ -0,0 +1,19 @@
+# (c) Copyright [2021-2024] Open Text.
```
cchen-vertica, line 1: The new test should be moved to leg-13 which is dedicated to tls-auth.

PR #1316, opened 2025-06-17
Disable cache duration

roypaulin:
Cache duration does not work and fails all the "cert" tests. This is a temporary change to allow tests to pass until it is fixed.


roypaulin:
I ended up opening a MR in the server to fix this issue. In case this is merged, I will revert the changes when that MR [23246](https://gitlab.otxlab.net/analytics-database/ver/server/-/merge_requests/23246) is merged and available in daily builds.

PR #1318, opened 2025-06-17
Add timeout to drain during read-only online upgrade

kcmackeen:
During read-only online upgrade, we do a shutdown of all primary subclusters and all secondary subclusters. Before completing this shutdown, we re-route all client traffic and then do a drain of all existing connections. Unlike in drainnode_reconciler (which will eventually timeout and kill the sessions), this drain will continue infinitely. 

We would like to add a timeout to this drain logic. If it drains for more than a certain amount of time, we will auto-kill the sessions and proceed with the the shutdown and upgrade.

For this, we will re-use much of the logic from drainnode_reconciler. So the timeout can be configured via annotation vertica.com/active-connections-drain-seconds, defaulted to 60. However, in this case, we are draining subcluster by subcluster, so we cannot re-use the entire reconciler.

FUTURE WORK: Currently, it will drain subclusters sequentially. However, in the future, it would be nice to include multiple subclusters in a single drain (for example, all primaries and then all secondaries). This would speed up the process for customers with many subclusters and many active connections.


pkg/controllers/vdb/readonlyonlineupgrade_reconciler.go
```diff
@@ -454,17 +457,59 @@ func (o *ReadOnlyOnlineUpgradeReconciler) drainSubcluster(ctx context.Context, s
...
+}
+
+func (o *ReadOnlyOnlineUpgradeReconciler) runSubclusterDrain(ctx context.Context, scName string) (ctrl.Result, error) {
+	drainSubclusterAnnotation := "vertica.com/drain-start-time-" + scName
+	pfs := o.PFacts.FindPodsBySubcluster(scName)
 
-		o.Log.Info("starting check for active connections in subcluster", "name", scName)
-		return o.Manager.isSubclusterIdle(ctx, o.PFacts, scName)
+	// If no active connections, return
+	if res, err := o.Manager.isSubclusterIdle(ctx, o.PFacts, scName); !verrors.IsReconcileAborted(res, err) {
```
roypaulin, line 478: If err != nil, we want to let the user know so we should return the error

pkg/controllers/vdb/readonlyonlineupgrade_reconciler.go
```diff
@@ -454,17 +457,59 @@ func (o *ReadOnlyOnlineUpgradeReconciler) drainSubcluster(ctx context.Context, s
...
+	if err := o.routeClientTraffic(ctx, scName, true); err != nil {
+		return ctrl.Result{}, nil
+	}
+
+	o.Log.Info("starting check for active connections in subcluster", "name", scName)
+	return o.runSubclusterDrain(ctx, scName)
+}
+
+func (o *ReadOnlyOnlineUpgradeReconciler) runSubclusterDrain(ctx context.Context, scName string) (ctrl.Result, error) {
+	drainSubclusterAnnotation := "vertica.com/drain-start-time-" + scName
```
roypaulin, line 474: Add this to annotations.yaml, the prefix will be a constant and we will use a function to build the final annotation.

You should also check early if there is already an annotation with prefix "vertica.com/drain-start-time-" but a different subcluster. That would mean we are already draining a draining a subcluster which is not the current one so we should just return.

pkg/controllers/vdb/readonlyonlineupgrade_reconciler.go
```diff
@@ -454,17 +457,59 @@ func (o *ReadOnlyOnlineUpgradeReconciler) drainSubcluster(ctx context.Context, s
...
+	}
+	elapsed := time.Since(drainStartTime)
+	timeout := time.Duration(timeoutInt) * time.Second
+	// If timeout has expired, we move on to the next reconciler (DBRemoveNodeReconciler|DBRemoveSubclusterReconciler)
+	o.Log.Info("Draining in progress", "start", drainStartTime, "elapsed", elapsed, "timeout", timeout)
+	if elapsed >= timeout {
+		o.VRec.Eventf(o.Vdb, corev1.EventTypeWarning, events.DrainSubclusterTimeout,
+			"Timed out while draining connections for subcluster '%s'; killing connections", scName)
+		o.Log.Info("Draining timeout has expired")
+		killConnectionsToPendingDeletePods(ctx, o.VRec, o.PFacts.PRunner, pfs)
```
roypaulin, line 507: This will make us try to kill connections in every pod in the subcluster (even pods with no active connections) and that's expensive. I think we should have another function like isSubclusterIdle (with a slightly different query) that will return the list of session ids in the subcluster. If the list is empty, we know there are no active connections in the subcluster. If non-empty, we can later reuse that sesssion ids list to kill the sessions.

pkg/meta/annotations.go
```diff
@@ -875,3 +880,26 @@ func getResource(annotations map[string]string, annotationName, defValStr string
...
+
+// Generate annotation name for subcluster-specific drain start time
+// eg vertica.com/drain-start-time-sc1
+func GenSubclusterDrainStartAnnotationName(scName string) string {
+	return genAnnotationName(DrainStartSubclusterPrefixAnnotation, scName)
+}
+
+// Find any annotation names with the drain start prefix
+// e.g., "vertica.com/drain-start-time-sc1"
+func FindDrainTimeoutSubclusterAnnotations(annotations map[string]string) ([]string, bool) {
```
roypaulin, line 903: You could extend this function to also get a suffix (sc name) and return true as soon as there is an annotation with prefix "vertica.com/drain-start-time-" but suffix different from the given suffix. Then you could add a helper in api/helpers.go , owned by VDB, that will call this function.

tests/e2e-server-upgrade/online-upgrade-drain-timeout/25-create-regular-user.yaml
roypaulin, line 1: why not call this "20-wait-for-createdb

roypaulin (commented):
Looks good!

pkg/controllers/vdb/readonlyonlineupgrade_reconciler.go
```diff
@@ -448,23 +451,85 @@ func (o *ReadOnlyOnlineUpgradeReconciler) isMatchingSubclusterType(sts *appsv1.S
...
+	o.Log.Info("Draining in progress", "start", drainStartTime, "elapsed", elapsed, "timeout", timeout)
+	if elapsed >= timeout {
+		o.VRec.Eventf(o.Vdb, corev1.EventTypeWarning, events.DrainSubclusterTimeout,
+			"Timed out while draining connections for subcluster '%s'; killing connections", scName)
+
+		pf, found := o.PFacts.FindFirstUpPod(true, scName)
+
+		// If we cannot find a pod to use for killing, just skip
+		if found {
+			killSessions(ctx, o.VRec, o.PFacts.PRunner, sessionIds, pf)
```
roypaulin, line 525: If we just skip, the annotation is still there so next time we will retry again for sc. I think if no sc nodes are up, it obviously means there will be no active connections so we should remove the annotation (if set) so next time we skip that sc.
What do you think?

roypaulin (commented):
Could you a changie entry?

roypaulin (approved):

PR #1330, opened 2025-06-25
Set cache duration

roypaulin:
This adds an annotation to control how long to keep the cache.


roypaulin:
[b15c2d7](https://github.com/vertica/vertica-kubernetes/pull/1330/commits/b15c2d7d04289600a0731a826127931df5b5e36f) fixes the issue pointed by Abhishek about cert rotation with aws secret with version-id. We were passing awssecret@versionid, instead of passing aws secret and versionid separately.

PR #1331, opened 2025-06-26
Fixed operator upgrade issue

cchen-vertica:
When we upgrade operator to 25.3, we will update the statefulSet due to the change of NMA env vars and mounted volume. This PR added check for nma-mount-cert and it can keep the mounted volume and only use mounted envs after the upgrade.

Moreover, I removed operator-version annotation from pod template so that the future upgrade won't trigger rolling restart of the database.

### DIFF
--- a/pkg/controllers/vdb/httpstls_reconciler.go
+++ b/pkg/controllers/vdb/httpstls_reconciler.go
@@ -43,7 +43,7 @@ func MakeHTTPSTLSReconciler(vdbrecon *VerticaDBReconciler, log logr.Logger, vdb 
 	return &HTTPSTLSReconciler{
 		VRec:       vdbrecon,
 		Vdb:        vdb,
-		Log:        log.WithName("MakeHTTPSTLSReconciler"),
+		Log:        log.WithName("HTTPSTLSReconciler"),
 		Dispatcher: dispatcher,
 		PRunner:    prunner,
 		Pfacts:     pfacts,
--- a/pkg/controllers/vdb/obj_reconciler_test.go
+++ b/pkg/controllers/vdb/obj_reconciler_test.go
@@ -47,6 +47,7 @@ import (
 var _ = Describe("obj_reconcile", func() {
 	ctx := context.Background()
 	const trueStr = "true"
+	const falseStr = "false"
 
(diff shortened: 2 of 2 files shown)


pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -137,6 +143,50 @@ func (o *ObjReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Re
 	return ctrl.Result{}, nil
 }
 
+// recordAnnotations will record the annotations in the vdb so we can get the correct annotation
+// value after operator is updated and the default value of the annotation is changed
+func (o *ObjReconciler) recordAnnotations(ctx context.Context) error {
+	// When the annotation is already set, no need to record
```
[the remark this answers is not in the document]
cchen-vertica: When the annotation is changed by the user, we will update the mounted volume accordingly.

pkg/controllers/vdb/obj_reconciler.go
```diff
@@ -137,6 +143,50 @@ func (o *ObjReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Re
...
+		return nil
+	}
+
+	svcName := names.GenHlSvcName(o.Vdb)
+	curSvc := &corev1.Service{}
+	err := o.Rec.GetClient().Get(ctx, svcName, curSvc)
+	svcNotExist := err != nil && kerrors.IsNotFound(err)
+	useNMAMount := false
+	if svcNotExist {
+		useNMAMount = vmeta.UseNMACertsMount(o.Vdb.Annotations)
```
roypaulin, line 160: you already know that o.Vdb.Annotations does not contain MountNMACertsAnnotation, so what's the point of this?

pkg/controllers/vdb/obj_reconciler.go
roypaulin, line 1: We can add a unit test for that new function.

pkg/controllers/vdb/obj_reconciler.go
cchen-vertica: Here I assign the default value to it when it doesn't exist. Then in next reconcile loop, we will skip retrieving Headless service.

PR #1335, opened 2025-07-01
Create HTTPS/NMA TLS rollback skeleton

kcmackeen:
This lays the foundations for rollback after tls cert rotation failure. 
The failure scenarios are the following (simplified summary):
- https cert rotation fails but before any change was made in the tls config. We are still using the old cert so we just need to undo the secret name change. (no-op on nma)
- https cert rotation fails, we altered the tls config but polling the new cert health failed (it is unlikely): We need to rollback the rotation by re-running cert rotation with reverse params; to get back to the old cert. (no-op on nma)
- htts cert rotation succeeded but nma cert rotation failed: we need to roll back both nma and https. First nma, we need to restart nma containers with the old cert. Once nma is back running with old cert, we re-run cert rotation with reverse params;

I added a new reconciler that will incapsulate the rollback logic. In the cert rotation reconcilers, we will set a status condition with a specific "reason". The rollback reconciler will use that condition to decide how far it needs to go to undo what has been done.

This is only the skeleton and currently does not do any rollback.

### DIFF
--- a/api/v1/helpers.go
+++ b/api/v1/helpers.go
@@ -617,6 +617,38 @@ func (v *VerticaDB) IsCertRotationInProgress() bool {
 	return v.IsStatusConditionTrue(TLSCertRotationInProgress)
 }
 
+func (v *VerticaDB) IsTLSCertRollbackNeeded() bool {
+	return v.IsStatusConditionTrue(TLSCertRollbackNeeded)
+}
+
+func (v *VerticaDB) FindTLSCertRollbackNeededCondition() *metav1.Condition {
+	return v.FindStatusCondition(TLSCertRollbackNeeded)
+}
+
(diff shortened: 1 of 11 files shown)


pkg/controllers/vdb/httpstlsupdate_reconciler.go
```diff
@@ -68,6 +68,10 @@ func (h *HTTPSTLSUpdateReconciler) Reconcile(ctx context.Context, req *ctrl.Requ
 		return ctrl.Result{}, nil
 	}
 
+	if h.Vdb.IsTLSCertRollbackNeeded() {
```
roypaulin, line 71: Add this condition to line  67 to fix gocyclo error

PR #1337, opened 2025-07-02
Add more webhook rules for tls

roypaulin:
1. "enable-tls-auth" and "mount-nma-certs" cannot be both set to true
2. when revive_db, if tls auth is enabled, https and client-server secrets must be present on vdb
3. cannot change both httpsNMATLS and clientServerTLS at the same time
4. on an existing vdb, cannot enable tls and change the https secret at the same time.


api/v1/verticadb_webhook.go
```diff
@@ -1450,6 +1452,42 @@ func (v *VerticaDB) validateSubclustersInSandboxes(allErrs field.ErrorList) fiel
...
+	return allErrs
+}
+
+// hasTLSSecretsWhenRevive checks if the TLS secrets are set when initPolicy is set to revive
+func (v *VerticaDB) hasTLSSecretsWhenRevive(allErrs field.ErrorList) field.ErrorList {
+	if v.Spec.InitPolicy != CommunalInitPolicyRevive {
+		return allErrs
+	}
+
+	if vmeta.UseTLSAuth(v.Annotations) && !vmeta.ShouldBypassSecretCheckOnRevive(v.Annotations) {
```
cchen-vertica, line 1472: Why do we need this bypass check?

api/v1/verticadb_webhook.go
```diff
@@ -2300,17 +2338,44 @@ func (v *VerticaDB) checkImmutableClientProxy(oldObj *VerticaDB, allErrs field.E
...
+	if !v.IsSetForTLS() {
+		return allErrs
+	}
+
+	httpsTLSConfigChanged := oldObj.GetHTTPSNMATLSSecret() != v.GetHTTPSNMATLSSecret() ||
+		oldObj.GetHTTPSNMATLSMode() != v.GetHTTPSNMATLSMode()
+	clientTLSConfigChanged := oldObj.GetClientServerTLSSecret() != v.GetClientServerTLSSecret() ||
+		oldObj.GetClientServerTLSMode() != v.GetClientServerTLSMode()
+	if v.IsTLSConfigUpdateInProgress() {
+		if oldObj.GetHTTPSNMATLSSecret() != v.GetHTTPSNMATLSSecret() ||
```
cchen-vertica, line 2355: Just use httpsTLSConfigChanged

api/v1/verticadb_webhook.go
```diff
@@ -2300,17 +2338,44 @@ func (v *VerticaDB) checkImmutableClientProxy(oldObj *VerticaDB, allErrs field.E
...
+			oldObj.GetHTTPSNMATLSMode() != v.GetHTTPSNMATLSMode() {
+			err := field.Invalid(field.NewPath("spec").Child("httpsNMATLS"),
+				v.Spec.HTTPSNMATLS,
+				"httpsNMATLS cannot be changed when tls config update is in progress")
+			allErrs = append(allErrs, err)
+		}
+		if oldObj.GetClientServerTLSSecret() != v.GetClientServerTLSSecret() ||
```
cchen-vertica, line 2362: Use clientTLSConfigChanged

api/v1/verticadb_webhook.go
```diff
@@ -2300,17 +2338,44 @@ func (v *VerticaDB) checkImmutableClientProxy(oldObj *VerticaDB, allErrs field.E
...
+			oldObj.GetClientServerTLSMode() != v.GetClientServerTLSMode() {
+			err := field.Invalid(field.NewPath("spec").Child("clientServerTLS"),
+				v.Spec.ClientServerTLS,
+				"clientServerTLS cannot be changed when tls config update is in progress")
+			allErrs = append(allErrs, err)
+		}
+	} else if httpsTLSConfigChanged && clientTLSConfigChanged {
```
cchen-vertica, line 2369: Do we really have this limitation?

api/v1/verticadb_webhook.go
```diff
@@ -2337,63 +2402,77 @@ func (v *VerticaDB) hasValidTLSMode(tlsModeToValidate, fieldName string, allErrs
...
-	// - when the annotation is true, don't allow the user to change it to false.
-	// - when that annotation is false, don't allow the user to change HTTPSNMATLSSecret,
-	//   clientServerTLSSecret and tls modes.
+	// Rule 2: TLS Auth transition restrictions
 	if vmeta.UseTLSAuth(oldObj.Annotations) {
 		if !vmeta.UseTLSAuth(v.Annotations) {
-			err := field.Forbidden(field.NewPath("spec").Child("httpsNMATLS"),
-				"cannot disable mutual TLS after it's enabled")
-			allErrs = append(allErrs, err)
+			prefix := field.NewPath("metadata").Child("annotations")
```
cchen-vertica, line 2419: Remove this check since forbid-disable-tls rule already exists.

api/v1/verticadb_webhook.go
```diff
@@ -2300,17 +2338,44 @@ func (v *VerticaDB) checkImmutableClientProxy(oldObj *VerticaDB, allErrs field.E
...
-	// If cert rotation is in progress, httpsNMATLS can not be changed
-	if v.IsSetForTLS() && v.IsTLSConfigUpdateInProgress() &&
-		(oldObj.GetHTTPSNMATLSSecret() != v.GetHTTPSNMATLSSecret() ||
-			oldObj.GetHTTPSNMATLSMode() != v.GetHTTPSNMATLSMode()) {
+// checkImmutableTLSConfig validates the TLS config fields in vdb
+// It checks if the TLS config is being changed while the TLS config update is in progress.
+// It also checks if user is trying to change both httpsNMATLS and clientServerTLS at the same time.
+func (v *VerticaDB) checkImmutableTLSConfig(oldObj *VerticaDB, allErrs field.ErrorList) field.ErrorList {
+	// If the vdb is not set for TLS, we don't need to check anything.
+	if !v.IsSetForTLS() {
```
cchen-vertica, line 2346: We should use !vmeta.UseTLSAuth(v.Annotations) since we have a delay in getting the vertica version.

api/v1/verticadb_webhook.go
roypaulin: Yes. After https cert rotation, if you do client server cert rotation it will fail because nma is still using the old cert.

api/v1/verticadb_webhook.go
roypaulin: I am not adding a new check I just refactored the one that was there.

api/v1/verticadb_webhook.go
cchen-vertica: I see. I will create a follow-up ticket for this.

PR #1351, opened 2025-07-17
Support rollback after failed HTTPS cert rotation when DB has been updated

kcmackeen:
For HTTPS cert rotation only, there are two cases. One is rolling back before the DB has been updated, which is covered by PR https://github.com/vertica/vertica-kubernetes/pull/1345. The other is a failure during HTTPS polling, which takes place at the end of the cert rotation process, after the DB has been updated. This rollback is supported by this PR.

This rollback is more complex. We need to run a second HTTPS cert rotation going from the failed secret (which is now in the spec) to the last good cert (which is still saved in the status). This takes place during the rollback reconciler; when rollback is in progress, we consider the spec secret to be the "In Use" secret and the status secret to be "requested" secret, which is the opposite of how cert rotation typically works.

Another consideration is NMA. In the previous PR, we made a change so that the NMA config map was not updated if cert rotation failed. However, we found that HTTPS cert rotation does not work well if HTTPS is using one cert (the failed one) and NMA is using another (the previous one). So, in the case of a failure AFTER the DB is updated, we need to still update the NMA config map and rotate the NMA to use the failed secret. Then, after HTTPS has successfully been rotated to the previous secret, we can update NMA again, to use the previous secret as well.

One exception to this is when there is a failure when only HTTPS mode has been changed. Currently, if you change HTTPS mode only, this does not require an NMA restart. So, if we have a failure when only HTTPS mode was changed, we don't need the two NMA restarts; we can just keep the NMA using the previous good secret.


kcmackeen:
Closing this PR. New PR (https://github.com/vertica/vertica-kubernetes/pull/1386) has been opened, which is a clean and simplified version of this one.

PR #1355, opened 2025-07-24
Removed auto-generated https and client secrets

cchen-vertica:
This PR did below things:
1. when enable-tls-auth is "false", operator shouldn't auto-generate https and client-server secrets.

2. when enable-tls-auth is "false", operator should use nmaTLSSecret to set NMA container and leave httpsNMATLS.secret empty.

3. when enable-tls-auth changes to "true"

    a. operator should auto-generate https and client-server secrets if they don't exist

    b. operator should restart nma container if httpsNMATLS.secret != nmaTLSSecret, then set tls config for https/client-server


roypaulin:
Will review it today.

LiboYu2:
> This PR did below things:
> 
> 1. when enable-tls-auth is "false", operator shouldn't auto-generate https and client-server secrets.
> 2. when enable-tls-auth is "false", operator should use nmaTLSSecret to set NMA container and leave httpsNMATLS.secret empty.
> 3. when enable-tls-auth changes to "true"
>    a. operator should auto-generate https and client-server secrets if they don't exist
>    b. operator should restart nma container if httpsNMATLS.secret != nmaTLSSecret, then set tls config for https/client-server

When we rotate certs, do we need to rotate nmaTLSSecret as well?

pkg/builder/builder.go
```diff
@@ -2198,12 +2198,20 @@ func GetTarballName(cmd []string) string {
...
 	secretMap := map[string]string{
 		NMASecretNamespaceEnv:       vdb.ObjectMeta.Namespace,
-		NMASecretNameEnv:            vdb.GetHTTPSNMATLSSecret(),
-		NMAClientSecretNamespaceEnv: vdb.ObjectMeta.Namespace,
-		NMAClientSecretNameEnv:      vdb.GetClientServerTLSSecret(),
-		NMAClientSecretTLSModeEnv:   vdb.GetNMAClientServerTLSMode(),
+		NMASecretNameEnv:            vdb.GetNMATLSSecret(),
+		NMAClientSecretNamespaceEnv: clientSecretNamespace,
+		NMAClientSecretNameEnv:      clientSecretName,
+		NMAClientSecretTLSModeEnv:   clientSecretTLSMode,
```
kcmackeen, line 2214: This is going to clash with the rollback PR I just merged. We want the config map to get the status value if there is a TLS update in progress and/or if there was a cert rotation failure; otherwise, get the spec value. It looks like this change will always get from the spec (which is what we were doing previously).

See the helper function GetValueForTLSConfigMap; we would need to do something similar here.

pkg/controllers/vdb/tlsservercertgen_reconciler.go
```diff
@@ -61,27 +63,47 @@ func MakeTLSServerCertGenReconciler(vdbrecon *VerticaDBReconciler, log logr.Logg
 }
 
 // Reconcile will create a TLS secret for the http server if one is missing
+//
+//nolint:gocyclo
 func (h *TLSServerCertGenReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
cchen-vertica: I will do that.

pkg/builder/builder.go
cchen-vertica: I will resolve the conflicts with your new code.

PR #1356, opened 2025-07-24
Fix shutdown subcluster issue

roypaulin:
We found that restarting a subcluster after shutdown did not work because obj_reconciler will overwrite the expected size with the current size which in this case is zero. This fixes that issue.


roypaulin:
TBH It is a surprise the shutdown e2e test has not failed.

roypaulin:
@qindotguan I think this should fix that issue we were debugging.

PR #1358, opened 2025-07-28
Create service monitor when prometheus is enabled

roypaulin:
This will create a service monitor for each vdb, when prometheus is enabled. This will allow prometheus to automatically scrape the metrics from the db, and grafana will display them through the dashboards.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/onsi/ginkgo/v2 v2.22.1
 	github.com/onsi/gomega v1.36.1
 	github.com/pkg/errors v0.9.1
+	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.74.0
 	github.com/prometheus/client_golang v1.20.5
 	github.com/vertica/vcluster v1.0.0
 	github.com/vertica/vertica-sql-go v1.1.1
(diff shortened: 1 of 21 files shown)


roypaulin:
One major limitation is that prometheus cannot read secrets from remote sources (aws/gcp).

cchen-vertica:
> One major limitation is that prometheus cannot read secrets from remote sources (aws/gcp).

Okay, we can create a follow-up spike for removing this limitaiton.

helm-charts/verticadb-operator/tests/manager-role_test.yaml
```diff
@@ -31,7 +31,6 @@ tests:
               - scaledobjects
             verbs:
               - create
-              - delete
```
cchen-vertica, line 34: Why don't we need the delete permission?

cchen-vertica (commented):
With this implementation, we will create multiple service-monitor objects when we have multiple vdbs, right? As my understanding, we should create only one service-monitor object along with Prometheus deployment and then add its label to vdb services in builder.go.

helm-charts/verticadb-operator/tests/manager-role_test.yaml
roypaulin: Because the operator is not calling the Delete api. The scaledObject will be deleted by Kubernetes API Server's built-in garbage collector. We do not want to give the operator permissions it does not need.

PR #1360, opened 2025-07-29
Check for cert not matching key and for cert not matching common name

kcmackeen:
Added two new cases to the function that will validate TLS certificates for httpsNMA and clientServer secrets:

1. Validate if private key and certificate are matching; if not, produce event and requeue.
2. Validate if subject.commonName of cert matches the database user;  if not, produce event and requeue. By default, the dbuser will be assumed to be the annotation value vertica.com/superuser-name (default is dbadmin). However, for clientServer, this can be overridden using the new spec field below.

For httpsNMA, we will only ever validate against superuser-name. However, for clientServer, we have added a new field to the spec section clientServerTLS: commonName. When specified, clientServer secret's commonName will be validated against this value instead of superuser-name.


tests/e2e-leg-13/invalid-cert-validation/30-change-clientserver-cert-to-mismatched-cert copy.yaml
roypaulin, line 1: Fix this file name.

pkg/security/verify.go
roypaulin, line 1: Add unit tests for the new functions.

api/v1/helpers.go
```diff
@@ -1753,6 +1753,16 @@ func (v *VerticaDB) IsCertNeededForClientServerAuth() bool {
 	return tlsMode != tlsModeDisable && tlsMode != tlsModeEnable
 }
 
+// GetExpectedCertCommonName returns the expected common name for the TLS certificate.
+// For httpsNMATLS, this is the DB admin. For clientServerTLS, it will also default to
+// DB admin, but it can be overridden using clientServerTLS.commonName.
+func (v *VerticaDB) GetExpectedCertCommonName(configName string) string {
+	if configName == ClientServerTLSConfigName && v.Spec.ClientServerTLS != nil && v.Spec.ClientServerTLS.CommonName != "" {
```
LiboYu2, line 1760: If I understand correctly the comments,  clientService and httpsNMATLS can have their own common names, which can be different. Then we should have two such functions for each of them. Did I miss anything here?

PR #1362, opened 2025-07-31
Support scraping prometheus using tls cert

roypaulin:
Now prometheus can scrape metrics from the db using tls cert. When tls is disabled, it will use the user/password, when tls is enabled, it will the cert in httpsNMATLS.secret.

### DIFF
--- a/helm-charts/verticadb-operator/README.md
+++ b/helm-charts/verticadb-operator/README.md
@@ -35,10 +35,21 @@ This helm chart will install the operator and an admission controller webhook.  
 | serviceAccountAnnotations | A map of annotations that will be added to the serviceaccount created. | |
 | serviceAccountNameOverride | Controls the name given to the serviceaccount that is created. | |
 | tolerations | Any [tolerations and taints](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) used to influence where a pod is scheduled. This parameter is provided as a list. | Not set |
-| webhook.caBundle | A PEM encoded CA bundle that will be used to validate the webhook's server certificate.  This option is deprecated in favour of providing the CA bundle in the webhook.tlsSecret with the ca.crt key. This option will be removed in a future release.| |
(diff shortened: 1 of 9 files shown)


pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -96,7 +97,7 @@ func (r *VerticaDBReconciler) SetupWithManager(mgr ctrl.Manager, options control
 		Owns(&appsv1.Deployment{})
 
 	discoveryClient := discovery.NewDiscoveryClientForConfigOrDie(mgr.GetConfig())
-	if isServiceMonitorObjectInstalled(discoveryClient) {
+	if opcfg.IsPrometheusEnabled() && isServiceMonitorObjectInstalled(discoveryClient) {
```
qindotguan, line 100: Is this related to removing the service monitor? Should we add a test for this?

helm-charts/verticadb-operator/README.md
```diff
@@ -35,10 +35,21 @@ This helm chart will install the operator and an admission controller webhook.
 | serviceAccountAnnotations | A map of annotations that will be added to the serviceaccount created. | |
 | serviceAccountNameOverride | Controls the name given to the serviceaccount that is created. | |
 | tolerations | Any [tolerations and taints](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) used to influence where a pod is scheduled. This parameter is provided as a list. | Not set |
-| webhook.caBundle | A PEM encoded CA bundle that will be used to validate the webhook's server certificate.  This option is deprecated in favour of providing the CA bundle in the webhook.tlsSecret with the ca.crt key. This option will be removed in a future release.| |
```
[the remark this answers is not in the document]
roypaulin: Yes, been removed since https://github.com/vertica/vertica-kubernetes/issues/519

pkg/controllers/vdb/verticadb_controller.go
roypaulin: This is related to an issue Abhishek found. Without this change, if prometheus is disabled and the servicemonitor CRD is installed in the cluster, the operator will try to own the servicemonitor, but because it does not have permissions (we add servicemonitor rule in the clusterrole only when prometheus is enabled) it will crash.

We do not need to add a test. Abhishek will do the testing.

PR #1364, opened 2025-08-01
allow users to change tls version and cipher suites

LiboYu2:
Two new fields are added vdb crd. One for tls version one for cipher suites.
set_config_parameter_vc.go is used to make changes to the two parameters.
A new polling API is added (yet to finish) to poll https server after making config change.

The whitelist in NMA code must be updated to include the new paramters.
The branch name in server gitlab repo:
tls_ver_cipher


LiboYu2:
> Please undraft the PR if you think it is ready for review. Also could you add some context to the description. It will help reviewers understand why we are making these changes

The working code is ready for review. I still need to add an e2e test case. After that, I will change its draft status. I will update the description to make it easier to understand the logic.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+}
+
+func (t *DBTLSConfigReconciler) shouldSkipReconcile() bool {
+	return t.Vdb.IsStatusConditionTrue(vapi.TLSConfigUpdateInProgress) ||
+		t.Vdb.IsTLSCertRollbackNeeded() || !t.Vdb.IsStatusConditionTrue(vapi.DBInitialized) ||
+		t.Vdb.IsStatusConditionTrue(vapi.UpgradeInProgress)
+}
+
+// Reconcile will create a TLS secret for the http server if one is missing
```
kcmackeen, line 56: This is inaccurate. I assume it is just accidentally copied from another reconciler

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	t.VRec.Eventf(t.Vdb, corev1.EventTypeNormal, events.HTTPSTLSUpdateSucceeded,
+		"Successfully updated tls version to %d", newTLSVersion)
+	return nil
+}
+
+func (t *DBTLSConfigReconciler) getHostGroups() (upHosts,
+	mainClusterHosts []string) {
+	for _, detail := range t.Pfacts.Detail {
+		upHosts = append(upHosts, detail.GetPodIP())
```
cchen-vertica, line 148: You need to check if the pod is up before you add the pod to upHosts.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,311 @@
...
+
+func (t *DBTLSConfigReconciler) updateCipherSuites(ctx context.Context, initiatorPod *podfacts.PodFact) error {
+	newCipherSuites := t.Vdb.Spec.DBTLSConfig.CipherSuites
+	t.VRec.Eventf(t.Vdb, corev1.EventTypeNormal, events.HTTPSTLSUpdateStarted,
+		"Started to update tls cipher suites to %s", t.placeholderForAll(newCipherSuites))
+	err := t.setCipherSuites(ctx, initiatorPod, t.Vdb.Spec.DBTLSConfig.TLSVersion, t.Vdb.Spec.DBTLSConfig.CipherSuites)
+	if err != nil {
+		t.Log.Info("failed to update cipher suites", "TLSVersion", t.Vdb.Spec.DBTLSConfig.TLSVersion, "cipherSuites",
+			newCipherSuites)
+		t.VRec.Eventf(t.Vdb, corev1.EventTypeNormal, events.HTTPSTLSUpdateFailed,
```
cchen-vertica, line 164: Use a new event rather than HTTPSTLSUpdateFailed .

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,311 @@
...
+			"failed to update tls cipher suites to %s", t.placeholderForAll(newCipherSuites))
+		return err
+	}
+	hosts, mainClusterHosts := t.getHostGroups()
+	err = t.pollHTTPS(ctx, hosts, mainClusterHosts)
```
cchen-vertica, line 169: Move this out.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,311 @@
...
+	opts := []pollhttps.Option{
+		pollhttps.WithInitiators(upHosts),
+		pollhttps.WithMainClusterHosts(strings.Join(mainClusterHosts, ",")),
+	}
+	return t.Dispatcher.PollHTTPS(ctx, opts...)
+}
+
+func (t *DBTLSConfigReconciler) compareTLSConfig(dbTLSConfigInSpec, dbTLSConfigInStatus *vapi.DBTLSConfig) (updateVersion,
+	updateCipherSuites bool) {
+	if *dbTLSConfigInSpec == *dbTLSConfigInStatus {
```
cchen-vertica, line 196: This is not accurate. You should rely on your t.equal() function. More importantly, you should check if they are nil before you dereference them.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,311 @@
...
+		setconfigparameter.WithUserName(t.Vdb.GetVerticaUser()),
+		setconfigparameter.WithInitiatorIP(initiatorPod.GetPodIP()),
+		setconfigparameter.WithConfigParameter(paramName),
+		setconfigparameter.WithValue(value),
+	}
+	return t.Dispatcher.SetConfigurationParameter(ctx, opts...)
+}
+
+func (t *DBTLSConfigReconciler) equal(cipherSuitesInUse, cipherSuitesInSpec string) bool {
+	listOfCipherSuitesInUse := strings.Split(cipherSuitesInUse, ",")
```
cchen-vertica, line 257: Before you split the string, you should replace the separator based on tls version.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,311 @@
...
+	t.Vdb.Status.DBTLSConfig = &vapi.DBTLSConfig{
+		TLSVersion:   tlsVersion,
+		CipherSuites: cipherSuites,
+	}
+	return t.saveTLSConfigInStatus(ctx, cipherSuites, tlsVersion)
+}
+
+func (t *DBTLSConfigReconciler) saveCipherSuitesInStatus(ctx context.Context, cipherSuites string) error {
+	refreshStatusInPlace := func(vdb *vapi.VerticaDB) error {
+		vdb.Status.DBTLSConfig.CipherSuites = cipherSuites
```
cchen-vertica, line 293: Check if DBTLSConfig is nil.

api/v1/verticadb_types.go
```diff
@@ -936,6 +943,20 @@ type Affinity struct {
 	PodAntiAffinity *corev1.PodAntiAffinity `json:"podAntiAffinity,omitempty" protobuf:"bytes,3,opt,name=podAntiAffinity"`
 }
 
+type DBTLSConfig struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=2
```
cchen-vertica, line 948: Remove the default value of it since it will be empty when deployment method is admintools.

pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -241,6 +241,7 @@ func (r *VerticaDBReconciler) constructActors(log logr.Logger, vdb *vapi.Vertica
 		MakeAutoCertRotateReconciler(r, log, vdb, true /* init */),
 		// Always generate cert first if nothing is provided
 		MakeTLSServerCertGenReconciler(r, log, vdb),
+		MakeDBTLSConfigReconciler(r, log, vdb, prunner, dispatcher, pfacts),
```
cchen-vertica, line 244: Your reconciler is executed too earlier. If you don't have a up pod, your cluster restart will keep failing.

This should be put after MakeDeploymentMethodReconciler() so once db is created, we can update its tls config params in one loop.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
LiboYu2: Good catch. Now they both use colon. Here colon will be used.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
LiboYu2: At the start of Reconcile(), line 80 checks if DBTLSConfig is nil in status. If it is, line 81 will initialize it and populate its fields using values from the db. When we use DBTLSConfig  from status, we don't need to do nil checking.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+	}
+	t.Log.Info("tls config update required, ", "updateVersion", strconv.FormatBool(updateTLSVersion), "updateCipherSuites",
+		strconv.FormatBool(updateCipherSuites))
+	if updateTLSVersion {
+		err := t.updateTLSVersion(ctx, initiatorPod)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		_, updateCipherSuites = t.compareTLSConfig(t.Vdb.Spec.DBTLSConfig, t.Vdb.Status.DBTLSConfig)
```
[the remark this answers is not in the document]
LiboYu2: When updateTLSVersion() is called, it will update TLS version first. Then it will load cipher suites from db and save them into the status.DBTLSConfig. 
That is required because:
TLS 1.2 and 1.3 use different configuration parameters to store their cipher suites. Those variables are sticky. 
When TLS is changed from 1.2 to 1.3, "tlsciphersuites" is used and "enabledciphersuites" is no longer used. But "enabledciphersuites" still stores the cipher suites previously set for TLS 1.2. 
When TLS is changed from 1.3 to 1.2, "enabledciphersuites" is used. We have to read it and save the 1.2 ciphersuites into status.DBTLSConfig. That is why we need another compareTLSConfig on line 98.

api/v1/verticadb_webhook.go
```diff
@@ -866,6 +881,49 @@ func (v *VerticaDB) hasValidTLSModes(allErrs field.ErrorList) field.ErrorList {
 	return allErrs
 }
 
+// hasValidTLSModes checks whether the TLS version and cipher suites are valid
+func (v *VerticaDB) hasValidTLSVersionAndCipherSuites(allErrs field.ErrorList) field.ErrorList {
+	if !vmeta.UseTLSAuth(v.Annotations) {
+		return allErrs
+	}
+	if v.Spec.DBTLSConfig.TLSVersion != 2 && v.Spec.DBTLSConfig.TLSVersion != 3 {
```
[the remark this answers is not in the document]
LiboYu2: Removed the default value. Added nil checking.

api/v1/verticadb_webhook.go
```diff
@@ -1153,6 +1169,57 @@ func (v *VerticaDB) hasValidTLSModes(allErrs field.ErrorList) field.ErrorList {
 	return allErrs
 }
 
+// hasValidTLSModes checks whether the TLS version and cipher suites are valid
+func (v *VerticaDB) hasValidDBTLSConfig(allErrs field.ErrorList) field.ErrorList {
+	if !v.UseVClusterOpsDeployment() && v.Spec.DBTLSConfig != nil {
```
kcmackeen, line 1174: I think you also want to check if TLS is enabled:

```suggestion
	if !v.UseVClusterOpsDeployment() && v.Spec.DBTLSConfig != nil {
```
```suggestion
	if !v.UseVClusterOpsDeployment() && !vmeta.UseTLSAuth(v.Annotations) && v.Spec.DBTLSConfig != nil {
```

LiboYu2: The reconciler will start to reconcile tls version and cipher suites when tls is enabled and configured. This rule controls when users can add dbTlsConfig block to their yaml file. It is reasonable to assume they will add it after tls is enabled.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,369 @@
...
+	if !ok {
+		t.Log.Info("No pod found to update tls config. restart next.")
+		restartReconciler := MakeRestartReconciler(t.VRec, t.Log, t.Vdb, t.PRunner, t.Pfacts, true, t.Dispatcher)
+		res, err2 := restartReconciler.Reconcile(ctx, request)
+		return "", res, err2
+	}
+	return initiatorPodIP, ctrl.Result{}, err
+}
+
+func (t *DBTLSConfigReconciler) updateTLSVersion(ctx context.Context, initiatorPodIP string) error {
```
cchen-vertica, line 121: Rename it to "updateTLSVersionAndReadCipherSuites"

pkg/vadmin/opts/pollhttps/opts.go
```diff
@@ -0,0 +1,63 @@
...
+ limitations under the License.
+*/
+
+package pollhttps
+
+import "github.com/vertica/vertica-kubernetes/pkg/tls"
+
+type Parms struct {
+	InitiatorIPs       []string
+	MainClusterHosts   string
```
cchen-vertica, line 22: Rename this to MainClusterInitiator

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+	var updateTLSVersion, updateCipherSuites bool
+	if t.Vdb.Spec.DBTLSConfig != nil && t.Vdb.Status.DBTLSConfig != nil {
+		updateTLSVersion, updateCipherSuites, _ = t.compareTLSConfig(t.Vdb.Spec.DBTLSConfig, t.Vdb.Status.DBTLSConfig)
+		if !updateTLSVersion && !updateCipherSuites {
+			return ctrl.Result{}, nil
+		}
+	}
+	initiatorPodIP, res, err := t.getInitiatorPodIP(ctx, request)
+	if initiatorPodIP == "" {
```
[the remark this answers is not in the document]
LiboYu2: I like that idea.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+		return res, err
+	}
+	if t.Vdb.Spec.DBTLSConfig == nil || t.Vdb.Status.DBTLSConfig == nil {
+		err = t.initializeDBTLSConfigInSpecAndStatus(ctx, initiatorPodIP)
```
cchen-vertica, line 76: We should only initialize db tls config when `t.Vdb.Status.DBTLSConfig == nil`. When `t.Vdb.Status.DBTLSConfig != nil`, there is no point to read cipher and version from database.

You should remove "t.Vdb.Spec.DBTLSConfig == nil" from the if condition.

We should have two if conditions:
1. if t.Vdb.Status.DBTLSConfig == nil
     initialize db tls config in status
2. if t.Vdb.Spec.DBTLSConfig == nil
     initialize db tls config in spec

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+		if err != nil {
+			t.Log.Error(err, "failed to initialize db tls config in spec and status. skip current loop")
+			return ctrl.Result{}, err
+		}
+		updateTLSVersion, updateCipherSuites = t.compareTLSConfig(t.Vdb.Spec.DBTLSConfig, t.Vdb.Status.DBTLSConfig)
```
cchen-vertica, line 81: this check should be moved out from the if block.

pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -290,6 +290,8 @@ func (r *VerticaDBReconciler) constructActors(log logr.Logger, vdb *vapi.Vertica
 		MakeSandboxShutdownReconciler(r, log, vdb, true),
 		// Update deployment method and enable HTTPS TLS
 		MakeDeploymentMethodReconciler(r, log, vdb, prunner, pfacts, dispatcher),
+		// reconcile tls version and cipher suite
+		MakeDBTLSConfigReconciler(r, log, vdb, prunner, dispatcher, pfacts),
```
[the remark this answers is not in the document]
cchen-vertica: This still needs to be addressed.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
LiboYu2: Had an offline discussion. Here is the updated logic.

There are 3 different scenarios. For 1 and 3 when Vdb.Spec.DBTLSConfig is nil, the reconciler will do nothing.
case 1
Vdb.Spec.DBTLSConfig == nil
Vdb.Status.DBTLSConfig == nil

case 2
Vdb.Spec.DBTLSConfig != nil
Vdb.Status.DBTLSConfig == nil

case 3
Vdb.Spec.DBTLSConfig == nil
Vdb.Status.DBTLSConfig != nil

pkg/controllers/vdb/dbtlsconfig_reconciler.go
LiboYu2: resolved by offline discussion.

pkg/controllers/vdb/dbtlsconfig_reconciler.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+// Reconcile will compare TLS version and cipher suites in Spec with those in Status.
+// If they are different, it will call vcluster API to update them.
+func (t *DBTLSConfigReconciler) Reconcile(ctx context.Context, request *ctrl.Request) (ctrl.Result, error) {
+	if t.shouldSkipReconcile() || t.Vdb.Spec.DBTLSConfig == nil {
+		return ctrl.Result{}, nil
+	}
+	var updateTLSVersion, updateCipherSuites bool
+	if t.Vdb.Spec.DBTLSConfig != nil && t.Vdb.Status.DBTLSConfig != nil {
```
cchen-vertica, line 64: No need to check `t.Vdb.Spec.DBTLSConfig != nil` again.

pkg/podfacts/podfacts.go
```diff
@@ -1528,6 +1528,20 @@ func GetHostAndPodNameList(podList []*PodFact) ([]string, []types.NamespacedName
 	return hostList, podNames
 }
 
+// get all running pods and main cluster running pod
+func GetHostGroups(pfacts *PodFacts) (upHosts []string,
```
cchen-vertica, line 1532: This should be a receiver function: func (p *PodFacts)GetHostGroups(...)

PR #1366, opened 2025-08-06
add reconciler to monitor dbadmin password secret updates

qindotguan:
This PR is to automate Vertica DB admin password change in Vertica on the k8s cluster as soon as the password in the k8s secret is updated.

A new field `status.passwordSecret` was added to store the current password secret.


ningdeng:
@qindotguan can you add proper description?
@LiboYu2 can you review this by today?

api/v1/helpers.go
```diff
@@ -69,6 +69,8 @@ const (
 	DefaultServiceClientPort = 5433
 )
 
+var EmptyPassword = ""
```
roypaulin, line 72: Make it a const

api/v1/verticadb_types.go
```diff
@@ -1029,6 +1029,15 @@ type VerticaDBStatus struct {
...
 	// The list of current TLS configurations for different components
 	TLSConfigs []TLSConfigStatus `json:"tlsConfigs,omitempty"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	// +optional
+	// The secret name that contains the current password for the
+	// database's superuser. The secret must have a key named password. To
+	// store this secret outside of Kubernetes, you can use a secret path
+	// reference prefix, such as gsm://. Everything after the prefix is the name
+	// of the secret in the service you are storing.
```
roypaulin, line 1039:
```suggestion
	// The secret name that contains the current password for the
	// database's superuser.
```
This is the status, we don't need all the details from the spec

pkg/builder/builder.go
```diff
@@ -2327,6 +2331,15 @@ func GetTarballName(cmd []string) string {
 	return ""
 }
 
+// GetPasswordSecret returns the password secret
+func GetPasswordSecret(vdb *vapi.VerticaDB) string {
```
roypaulin, line 2335: It should be a "VerticaDB" method in helpers.go

pkg/builder/builder.go
```diff
@@ -2327,6 +2331,15 @@ func GetTarballName(cmd []string) string {
...
+	if vdb.Spec.PasswordSecret == vdb.Status.PasswordSecret {
+		return vdb.Spec.PasswordSecret
+	}
+	// status holds the current password
+	return vdb.Status.PasswordSecret
```
roypaulin, line 2340: It should be, if status is non-empty, return status. Otherwise return spec.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	sb := strings.Builder{}
+	sb.WriteString(fmt.Sprintf(`ALTER USER %s IDENTIFIED BY '%s';`, dbUser, *newPasswd))
+
+	cmd := []string{"-tAc", sb.String()}
+	stdout, stderr, err := a.PRunner.ExecVSQL(ctx, pf.GetName(), names.ServerContainer, cmd...)
+	if err != nil {
+		a.VRec.Log.Error(err, "failed to retrieve active sessions", "stderr", stderr)
```
roypaulin, line 128: Wrong error msg.

pkg/vk8s/helpers.go
```diff
@@ -28,19 +28,31 @@ import (
...
+	return &passwd, nil
 }
 
 // GetSuperuserPassword returns the superuser password if it has been provided
 func GetSuperuserPassword(ctx context.Context, cl client.Client, log logr.Logger,
-	e events.EVWriter, vdb *vapi.VerticaDB) (string, error) {
-	if vdb.Spec.PasswordSecret == "" {
-		return "", nil
+	e events.EVWriter, vdb *vapi.VerticaDB) (*string, error) {
+	return GetSuperuserPasswordForUpdate(ctx, cl, log, e, vdb, false)
```
roypaulin, line 44: It would have been easier to make the secret a parameter. The function should get a password secret name and get the secret. Any other check must be done by the caller.

pkg/controllers/vrpq/query_reconciler.go
```diff
@@ -204,7 +204,8 @@ func (q *QueryReconciler) makeDispatcher(log logr.Logger, vdb *vapi.VerticaDB,
 	_ *string) (vadmin.Dispatcher, error) {
 	if vmeta.UseVClusterOps(vdb.Annotations) {
 		// The password isn't needed since our API is going to strictly communicate with the NMA
-		return vadmin.MakeVClusterOps(log, vdb, q.VRec.GetClient(), "", q.VRec, vadmin.SetupVClusterOps, q.VRec.CacheManager), nil
+		return vadmin.MakeVClusterOps(log, vdb, q.VRec.GetClient(), &vapi.EmptyPassword,
```
roypaulin, line 207: Now I understand why you used a golbal variable instead of a const. But still you can just assing the empty string const to a local var and pass its pointer. It is better to avoid global vars

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -182,7 +182,7 @@ func (r *ReplicationReconciler) determineUsernameAndPassword(ctx context.Context
 		return err
 	}
 
-	if r.TargetInfo.Vdb.IsSetForTLS() && r.TargetInfo.Password == "" && r.Vrep.Spec.TLSConfig == "" {
+	if r.TargetInfo.Vdb.IsSetForTLS() && r.TargetInfo.Password == &vapi.EmptyPassword && r.Vrep.Spec.TLSConfig == "" {
```
roypaulin, line 185: Please do not compare pointers. Compare the values.

This is a general comment.

pkg/controllers/vdb/passwordsecret_reconciler.go
roypaulin, line 1: As we discussed, you need to update the pointer to password after the password changes in the db.

pkg/builder/builder.go
qindotguan: This should always be vdb.Status.PasswordSecret, or adding passwordSecret after the DB created would fail with authentication error.

api/v1/helpers.go
qindotguan: I just removed it and used the variable in separate files.

pkg/controllers/vdb/passwordsecret_reconciler.go
qindotguan: updated.

pkg/builder/builder.go
```diff
@@ -935,7 +935,11 @@ func BuildServiceMonitor(nm types.NamespacedName, vdb *vapi.VerticaDB,
 
 // BuildBasicAuthSecret constructs the secret that will be used by prometheus to authenticate
 // to the vertica db
-func BuildBasicAuthSecret(vdb *vapi.VerticaDB, name, username, password string) *corev1.Secret {
+func BuildBasicAuthSecret(vdb *vapi.VerticaDB, name, username string, password *string) *corev1.Secret {
+	passwd := ""
+	if password != nil {
+		passwd = *password
+	}
```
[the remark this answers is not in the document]
qindotguan: Removed.

api/v1/helpers.go
```diff
@@ -1392,6 +1392,16 @@ func (v *VerticaDB) GetVerticaUser() string {
 	return vmeta.GetSuperuserName(v.Annotations)
 }
 
+// GetPasswordSecret returns the password secret
+func (v *VerticaDB) GetPasswordSecret() string {
+	// status holds the current password
+	if v.Status.PasswordSecret != "" {
```
roypaulin, line 1398: Issue with this: if I first create a db with empty password (password secret empty), then I set the passwordSecret, I will be getting the secret from spec instead of status and that will be a problem until password change is done in the db.
I think if v.IsDBInitialized() is false, we should get the secret from spec and then when db is initialized, we get it from spec.

We also need to carefully think about backward compatibility. If the user had a passwordsecret in an older version and they upgrade, status will be empty, and we will get the secret from status which will be wrong. 
Let's discuss about how to handle this case.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (a *PasswordSecretReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	if err := a.PFacts.Collect(ctx, a.Vdb); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// We put the current using password secret in the status
+	// No actions needed if status content is the same to spec
+	if a.statusMatchesSpec() {
+		return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
roypaulin: I meant if db is not initialized (createdb/revivedb not done yet) we should just skip.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -0,0 +1,148 @@
...
+		Vdb:        vdb,
+		PFacts:     pfacts,
+		PRunner:    prunner,
+		Dispatcher: dispatcher,
+	}
+}
+
+func (a *PasswordSecretReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	// No-op if no up nodes found
+	if a.Vdb.Status.UpNodeCount == 0 {
```
roypaulin, line 63: Use a.Vdb.IsDBInitialized()

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -0,0 +1,154 @@
...
+	if !found {
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	sb := strings.Builder{}
+	sb.WriteString(fmt.Sprintf(`ALTER USER %s IDENTIFIED BY '%s';`, a.Vdb.GetVerticaUser(), *newPasswd))
+	cmd := []string{"-tAc", sb.String()}
+	stdout, stderr, err := a.PRunner.ExecVSQL(ctx, pf.GetName(), names.ServerContainer, cmd...)
+	if err != nil {
+		a.VRec.Log.Error(err, "failed to update the password", "stderr", stderr)
```
cchen-vertica, line 124: Since we have an event "SuperuserPasswordSecretUpdated", here we should log a failure event which is easy to know if password secret update is done.

pkg/vk8s/helpers.go
```diff
@@ -28,40 +28,35 @@ import (
 )
 
 // getPasswordFromSecret retrieves the password from the secret using the provided key
-func getPasswordFromSecret(secret map[string][]byte, key string) (string, error) {
+func getPasswordFromSecret(secret map[string][]byte, key string) (*string, error) {
+	var passwd string
 	pwd, ok := secret[key]
+	passwd = string(pwd)
```
cchen-vertica, line 34: minor: move this line to line 37.

tests/e2e-leg-3/password-secret/25-verify-superuser-password-update.yaml
```diff
@@ -0,0 +1,54 @@
...
+  name: script-verify-password-secret-update
+data:
+  entrypoint.sh: |-
+    #!/bin/bash
+    set -o errexit
+    set -o xtrace
+
+    POD_NAME=v-password-secret-sc1-0
+
+    result=$(kubectl exec $POD_NAME -i -c server -- bash -c "vsql -w topsecret -tAc 'select password from passwords'")
```
cchen-vertica, line 26: Should this be 'select count(*) from passwords'?

If the sql query is not correct, why the test can still succeed? Is that because the password contains "1"? We need to verify if error result can really fail the test.

cchen-vertica:
One question: this password change doesn't work for sandbox, right? If yes, we need to mention this in the documentation.

api/v1/helpers.go
qindotguan: As discussed, changed status.passwordSecret to be a string pointer. We only get password from status if it's not nil. This will cover the db creation, db revive, and operator upgrade cases.

api/v1/helpers.go
```diff
@@ -1411,6 +1411,16 @@ func (v *VerticaDB) GetVerticaUser() string {
 	return vmeta.GetSuperuserName(v.Annotations)
 }
 
+// GetPasswordSecret returns the password secret
+func (v *VerticaDB) GetPasswordSecret() string {
+	// status holds the current password
+	if v.IsDBInitialized() && v.Status.PasswordSecret != nil {
```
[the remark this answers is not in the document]
qindotguan: Removed. Check nil is enough. Thanks.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -0,0 +1,154 @@
...
+	newPasswd, err := vk8s.GetCustomSuperuserPassword(ctx, a.VRec.Client, a.Log, a.VRec, a.Vdb,
+		a.Vdb.Spec.PasswordSecret, names.SuperuserPasswordKey)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// No-op if password is the same
+	if *a.PFacts.VerticaSUPassword == *newPasswd {
+		// update status only if db previously using empty password
+		if a.Vdb.Status.PasswordSecret == nil {
```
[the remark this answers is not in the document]
qindotguan: Removed.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+		a.Log.Info("WARNING: password in secret is the same as current password", "current password secret",
+			a.Vdb.Status.PasswordSecret, "new password secret", a.Vdb.Spec.PasswordSecret)
+		return ctrl.Result{}, nil
+	}
+
+	if pErr := a.PFacts.Collect(ctx, a.Vdb); pErr != nil {
+		return ctrl.Result{}, pErr
+	}
+	pf, found := a.PFacts.FindFirstUpPod(false, a.Vdb.GetFirstPrimarySubcluster().Name)
```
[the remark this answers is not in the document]
qindotguan: Updated to go through all the subclusters and find the first upnode.

pkg/podfacts/podfacts.go
```diff
@@ -1572,17 +1577,28 @@ func checkIfNodeUpCmd(podIP string, isHTTP bool) string {
...
 }
 
-// FindFirstPrimaryUpPodIP returns the ip of first pod that
+// FindFirstPrimaryUpPod returns the first pod that
 // has a primary up Vertica node, and a boolean that indicates
 // if we found such a pod
-func (p *PodFacts) FindFirstPrimaryUpPodIP() (string, bool) {
+func (p *PodFacts) FindFirstPrimaryUpPod() (*PodFact, bool) {
 	initiator, ok := p.FindFirstPodSorted(func(v *PodFact) bool {
 		return v.sandbox == vapi.MainCluster && v.isPrimary && v.upNode
```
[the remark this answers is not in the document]
qindotguan: Thanks. I'll update it in the next PR.

PR #1368, opened 2025-08-06
Fix auto-cert-rotation and e2e issues

cchen-vertica:
This PR fixed two auto-cert-rotation issues:
1. After db is created, we cannot enable tls with an auto-rotate secret list
2. When auto-rotate interval is too small, we will get crashloopbackoff error when restarting NMA. As I result, I set the minimal auto-rotate interval to 10 mins.

One e2e issues:
Now all Vertica dockerhub repos has been changed to private so I need to push the images to kind cluster before the test starts.


api/v1/verticadb_webhook.go
```diff
@@ -2879,11 +2879,11 @@ func (v *VerticaDB) validateOneTLSAutoRotateConfig(configName string) field.Erro
...
 	}
 
-	// Rule 2: Interval must be >= 1
-	if interval <= 0 {
+	// Rule 2: Interval must be >= 10
+	if interval < 10 {
 		allErrs = append(allErrs,
 			field.Invalid(fldPath.Child("interval"), interval,
-				"must be greater than 0"),
+				"must be greater or equal than 10"),
```
roypaulin, line 2886: Might want to specify the unit.

ningdeng:
@LiboYu2  can you review this by today?

api/v1/verticadb_webhook.go
LiboYu2: And grammar. "must be greater than or equal to".

cchen-vertica: Will update it soon.

LiboYu2:
> This PR fixed two auto-cert-rotation issues:
> 
> 1. After db is created, we cannot enable tls with an auto-rotate secret list
> 2. When auto-rotate interval is too small, we will get crashloopbackoff error when restarting NMA. As I result, I set the minimal auto-rotate interval to 10 mins.
> 
> One e2e issues: Now all Vertica dockerhub repos has been changed to private so I need to push the images to kind cluster before the test starts.

Do we know the root cause for the crashloopbackoff? Maybe frequent restart of http service mess up liveness probe?

pkg/controllers/vdb/tls_reconciler.go
```diff
@@ -60,8 +60,8 @@ func (h *TLSReconciler) Reconcile(ctx context.Context, request *ctrl.Request) (c
 	// when we first set tls config and nma tls secret is different than https tls secret,
 	// we need to restart nma
```
LiboYu2, line 61: Why do we need to restart NMA here? Is the  nma secret name not built into environmental variables?

cchen-vertica:
> > This PR fixed two auto-cert-rotation issues:
> > 
> > 1. After db is created, we cannot enable tls with an auto-rotate secret list
> > 2. When auto-rotate interval is too small, we will get crashloopbackoff error when restarting NMA. As I result, I set the minimal auto-rotate interval to 10 mins.
> > 
> > One e2e issues: Now all Vertica dockerhub repos has been changed to private so I need to push the images to kind cluster before the test starts.
> 
> Do we know the root cause for the crashloopbackoff? Maybe frequent restart of http service mess up liveness probe?

When the container is restarted, liveness probe failure threshold will be reset so that's not the reason. It's a K8s mechanism. When it see a container failed frequently, it will mark it as crashloopbackoff.

pkg/controllers/vdb/tls_reconciler.go
cchen-vertica: Because when NMA secret is different than https/client-server tls secret, vcluster cert rotation would fail due to cert mismatch.

.github/workflows/e2e.yml
```diff
@@ -164,7 +164,7 @@ jobs:
           dockerhub-token: ${{ secrets.DOCKERHUB_TOKEN }}
           vlogger-image: ${{ needs.build.outputs.vlogger-image }}
           operator-image: ${{ needs.build.outputs.operator-image }}
-          vertica-image: "opentext/vertica-k8s:23.4.0-0-minimal"
+          vertica-image: "docker.io/opentext/vertica-k8s:23.4.0-0-minimal"
```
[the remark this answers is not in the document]
cchen-vertica: Without the prefix "docker.io", we won't log into dockerhub so we won't be able to push images to kind cluster.

PR #1380, opened 2025-08-13
Support changing both TLS secrets (httpsNMA and clientServer) at the same time

kcmackeen:
Previously, we had a webhook to block a user from changing both httpsNMATLS.secret and clientServerTLS.secret at the same time. This PR will now allow this. 

This required 3 changes:

1. Run NMA configmap update and NMA cert rotation between https and clientServer cert rotations, in addition to running them after clientServer
2. Update code that writes values to NMA configmap, such that it will only write https-related values after https rotate has finished and clientServer values after clientServer rotate has finished
3. Update the logic for TLSConfigUpdateInProgress. Previously, it was set during each rotation and unset after the NMA rotate; now, it will be set when the first rotation is triggered (whether this is https or clientServer), and then only be unset after BOTH are finished.

I also slightly changed the logic on the "trigger-tls-update-failure" testing annotation. We can now set the following values:

- client_server
- https_before_tls_update
- https_after_tls_update

### DIFF
--- a/pkg/controllers/vdb/nmacertrotation_reconciler.go
+++ b/pkg/controllers/vdb/nmacertrotation_reconciler.go
@@ -71,7 +71,7 @@ func (h *NMACertRotationReconciler) Reconcile(ctx context.Context, _ *ctrl.Reque
 	}
 
 	// nma secret
-	newSecretName := h.Vdb.GetHTTPSNMATLSSecret()
+	newSecretName := h.Vdb.GetHTTPSNMATLSSecretForConfigMap()
 
 	newSecret, res, err := readSecret(h.Vdb, h.VRec, h.VRec.GetClient(), h.Log, ctx, newSecretName)
 	if verrors.IsReconcileAborted(res, err) {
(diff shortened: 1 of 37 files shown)


cchen-vertica (approved):

pkg/controllers/vdb/nmacertrotation_reconciler.go
```diff
@@ -71,7 +71,7 @@ func (h *NMACertRotationReconciler) Reconcile(ctx context.Context, _ *ctrl.Reque
 	}
 
 	// nma secret
-	newSecretName := h.Vdb.GetHTTPSNMATLSSecret()
+	newSecretName := h.Vdb.GetHTTPSNMATLSSecretForConfigMap()
```
LiboYu2, line 74: This function's return value replies on a few status conditions, i.e., 	HTTPSTLSConfigUpdateFinished and ClientServerTLSConfigUpdateFinished. Are those status conditions reset after a rotation? If not, will there be any impact?

kcmackeen:
> > Previously, we had a webhook to block a user from changing both httpsNMATLS.secret and clientServerTLS.secret at the same time. This PR will now allow this.
> > A quick question. When both are changed, which rotation will start first? httpsNMATLS?

The flow is:

1. HTTPS rotate
2. NMA config map update and NMA rotate (HTTPS related fields only)
3. ClientServer rotate
4. NMA config map update and NMA rotate (ClientServer related fields only)
5. Then rollback if needed

PR #1386, opened 2025-08-18
Support rollback after HTTPS cert rotation when TLS has been updated

kcmackeen:
This is a clean branch and a simplification of this PR: https://github.com/vertica/vertica-kubernetes/pull/1351

Currently, we are able to rollback after a failed Client-Server cert rotation. For HTTPS, if the failure occurs **BEFORE** the TLS update takes place in the DB, this is also supported. This PR will support HTTPS rollback **AFTER** the DB has been updated; the most common cause of such a rollback is a failure in the HTTPS polling step of cert rotation.

In this rollback, we need to re-run the HTTPS cert rotation. However, rather than rotating from status (previous values) to spec (new one), we are rotating from spec (the failed secret) to status (the last good secret). However, while the HTTPS cert has been rotated to the failed one, the NMA has not yet. So, during the rollback rotation, we need to point NMA to the status value (the last one).


LiboYu2:
> This is a clean branch and a simplification of this PR: #1351
> 
> Currently, we are able to rollback after a failed Client-Server cert rotation. For HTTPS, if the failure occurs **BEFORE** the TLS update takes place in the DB, this is also supported. This PR will support HTTPS rollback **AFTER** the DB has been updated; the most common cause of such a rollback is a failure in the HTTPS polling step of cert rotation.

A quick question. How do we confirm that the DB has been updated?

tests/e2e-leg-13/cert-rotate-rollback-after-update/40-assert.yaml
```diff
@@ -0,0 +1,36 @@
...
+---
+apiVersion: v1
+kind: Event
+reason: NMATLSCertRotationSucceeded
+source:
+  component: verticadb-operator
+involvedObject:
+  apiVersion: vertica.com/v1
+  kind: VerticaDB
+  name: v-cert-rotate-rollback-after-update
```
cchen-vertica, line 36: Could we verify https-cert-rotation-succeed event here?

pkg/controllers/vdb/httpstlsupdate_reconciler.go
```diff
@@ -123,7 +125,7 @@ func (h *HTTPSTLSUpdateReconciler) Reconcile(ctx context.Context, req *ctrl.Requ
 		upHostToSandbox[p.GetPodIP()] = p.GetSandbox()
 	}
 	err = h.Manager.updateTLSConfig(ctx, initiator, upHostToSandbox)
-	if err != nil {
+	if err != nil || h.Vdb.IsTLSCertRollbackNeeded() {
```
cchen-vertica, line 128: Just to confirm: when rollback(bad-cert to previous-cert) failed in line 127, we return here and keep retrying in next reconciler loops, correct?

cchen-vertica (commented):
When we

pkg/controllers/vdb/tls_config.go
```diff
@@ -360,10 +366,18 @@ func (t *TLSConfigManager) getAWSCertsConfig(versionID, cacheDuration string) (k
 
 func (t *TLSConfigManager) setTLSUpdatedata() {
 	if t.TLSConfig == tlsConfigHTTPS {
-		t.CurrentSecret = t.Vdb.GetHTTPSNMATLSSecretInUse()
-		t.NewSecret = t.Vdb.GetHTTPSNMATLSSecret()
-		t.CurrentTLSMode = t.Vdb.GetHTTPSTLSModeInUse()
-		t.NewTLSMode = t.Vdb.GetHTTPSNMATLSMode()
+		// If rollback, go from status to spec; otherwise, go from spec to status
+		if t.Vdb.IsTLSCertRollbackInProgress() {
+			t.CurrentSecret = t.Vdb.GetHTTPSNMATLSSecret()
```
LiboYu2, line 371: After the cert rollback is complete, will t.CurrentSecret and t.NewSecret be updated?

pkg/controllers/vdb/tls_config.go
```diff
@@ -360,10 +366,18 @@ func (t *TLSConfigManager) getAWSCertsConfig(versionID, cacheDuration string) (k
 
 func (t *TLSConfigManager) setTLSUpdatedata() {
 	if t.TLSConfig == tlsConfigHTTPS {
-		t.CurrentSecret = t.Vdb.GetHTTPSNMATLSSecretInUse()
-		t.NewSecret = t.Vdb.GetHTTPSNMATLSSecret()
-		t.CurrentTLSMode = t.Vdb.GetHTTPSTLSModeInUse()
-		t.NewTLSMode = t.Vdb.GetHTTPSNMATLSMode()
+		// If rollback, go from status to spec; otherwise, go from spec to status
```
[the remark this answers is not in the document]
kcmackeen: t.Vdb.GetHTTPSNMATLSSecret() = "get spec secret"
t.Vdb.GetHTTPSNMATLSSecretInUse() = "get status secret"

On a normal rotation, the status is "current" and the spec is "new". On a rollback, the spec is "the bad failed secret that the DB is currently using" and the status is "the last good secret we want to rotate to".

So, normal rotate is spec -> status and rollback is status-> spec. That's what the comment says and that's what the code is supposed to do (as far as I can see).

Let me know if that clears it up and if there is anything I can do to make this more clear. One thought I had was rename the functions, eg:

t.Vdb.GetHTTPSNMATLSSecret() -> t.Vdb.GetHTTPSNMATLSSecretFromSpec()
t.Vdb.GetHTTPSNMATLSSecretInUse() -> t.Vdb.GetHTTPSNMATLSSecretFromStatus()

To make it more clear what we are actually doing. But that is probably outside the scope of this.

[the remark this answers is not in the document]
cchen-vertica: The code makes sense. Just the comment. When rollback, your current secret is spec secret(line 371) and your new secret is status secret(line 372). So the comment should be:
 
```suggestion
		// If rollback, go from spec to status; otherwise, go from status to spec
```

pkg/controllers/vdb/tls_config.go
```diff
@@ -101,6 +101,12 @@ func (t *TLSConfigManager) setPollingCertMetadata(ctx context.Context) (ctrl.Res
 	// httpsNMATLSSecret
 	currentSecretName := t.Vdb.GetHTTPSNMATLSSecretInUse()
 	newSecretName := t.Vdb.GetHTTPSNMATLSSecret()
+
+	// If rollback, use status as current secret and spec as new secret
```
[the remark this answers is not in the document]
cchen-vertica: Your comment says "use spec as new secret". However, the code is using status as new secret.

pkg/controllers/vdb/httpstlsupdate_reconciler.go
cchen-vertica: What I mean is rollback failure retry.

PR #1392, opened 2025-08-22
Add extra grafana parameters

roypaulin:


helm-charts/verticadb-operator/values.yaml
```diff
@@ -237,6 +237,33 @@ grafana:
     ##
     pspEnabled: false
 
+  ## Enable persistence using Persistent Volume Claims
```
LiboYu2, line 240: Could you improve the comments and let people know which component will use "persistence".

PR #1395, opened 2025-08-28
Add retry logic when HTTPS cert rotation fails during HTTPS polling

kcmackeen:
At the end of the HTTPS cert rotation, after the DB has been updated, we run HTTPS polling to verify that the HTTPS service is back up and running with the new cert. If this fails, we will be forced to do a rollback. However, HTTPS polling could potentially time out simply because the HTTPS is slow to come up or due to some other transient issue. Rather than rolling back in these situations, we want to give the user the option to retry the HTTPS polling.

This adds a new annotation::
```
vertica.com/cert-rotation-max-retries:
```

This will allow the user is set how many retries to attempt, before giving up and running rollback. This retry is only triggered if there is an error specific to HTTPS polling; all other errors will rollback immediately. The current retry is tracked in the status field ```httpsPollingCurrentRetry```.

To test this, I also updated the following testing annotation:

```
vertica.com/trigger-tls-update-failure:
```

To support a comma-separated list. This will allow us to trigger a sequence of failures, in order to fully test retry.


LiboYu2:
> At the end of the HTTPS cert rotation, after the DB has been updated, we run HTTPS polling to verify that the HTTPS service is back up and running with the new cert. If this fails, we will be forced to do a rollback. However, HTTPS polling could potentially time out simply because the HTTPS is slow to come up or due to some other transient issue. Rather than rolling back in these situations, we want to give the user the option to retry the HTTPS polling.
> 
> This adds a new annotation::
> 
> ```
> vertica.com/cert-rotation-number-of-retries:
> ```

That annotation name is kind of misleading. Could we add polling to it?
The default polling will time out in 5 minutes. That is long enough for most of users. We need at most 2 retries.

pkg/controllers/vdb/tls_config.go
```diff
@@ -449,5 +475,71 @@ func (t *TLSConfigManager) unsetForceTLSUpdateFailure(ctx context.Context) error
...
+	if len(upPods) == 0 {
+		return fmt.Errorf("no up pods available for HTTPS polling retry")
+	}
+
+	var upPodIPs []string
+	for _, pod := range upPods {
+		upPodIPs = append(upPodIPs, pod.GetPodIP())
+	}
+	opts := []pollhttps.Option{
+		pollhttps.WithMainClusterHosts(strings.Join(upPodIPs, ",")),
```
cchen-vertica, line 510: We should use only one up host from main cluster. Also, you need to set WithSyncCatalogRequired to true. You can check Libo's PR to see how he set the parameters.

pkg/controllers/vdb/tls_config.go
```diff
@@ -449,5 +475,71 @@ func (t *TLSConfigManager) unsetForceTLSUpdateFailure(ctx context.Context) error
...
+		pollhttps.WithSyncCatalogRequired(false),
+		pollhttps.WithNewHTTPSCerts(&tls.HTTPSCerts{
+			Key:    t.Key,
+			Cert:   t.Cert,
+			CaCert: t.CACert,
+		}),
+	}
+
+	lastErr := originalError
+	for attempt := 1; attempt <= maxRetries; attempt++ {
```
cchen-vertica, line 521: This line should be moved to line 499 so that each retry can have required up pods.

cchen-vertica (commented):
You can remove Libo's changes so your PR will be more clear.

pkg/controllers/vdb/tls_config.go
```diff
@@ -449,5 +475,71 @@ func (t *TLSConfigManager) unsetForceTLSUpdateFailure(ctx context.Context) error
...
+// is returned so rollback can be triggered by the caller.
+func (t *TLSConfigManager) retryHTTPSPollingIfNeeded(ctx context.Context, originalError error) error {
+	maxRetries := t.Vdb.GetCertRotateRetries()
+	if maxRetries <= 0 {
+		return fmt.Errorf("no retries configured for HTTPS polling")
+	}
+
+	upPods := t.PFacts.FindUpPods("")
+	if len(upPods) == 0 {
+		return fmt.Errorf("no up pods available for HTTPS polling retry")
```
cchen-vertica, line 502: Add a restart reconciler here to restart the cluster. If restart failed, return error; if restart succeed, collect up pods again. You could refer to getInitiatorPodIP() of Libo's branch.

In the beginning of each retry(around line 500). we need to invalid podfacts and collect new podfacts again.

pkg/controllers/vdb/tls_config.go
```diff
@@ -449,5 +475,99 @@ func (t *TLSConfigManager) unsetForceTLSUpdateFailure(ctx context.Context) error
...
+
+	// Choose one primary host for mainClusterHosts
+	mainHost, ok := t.PFacts.FindFirstPrimaryUpPodIP()
+	if !ok {
+		return nil, fmt.Errorf("no primary pod found for HTTPS polling")
+	}
+
+	opts := []pollhttps.Option{
+		pollhttps.WithMainClusterHosts(mainHost),
+		pollhttps.WithInitiators([]string{mainHost}),
```
cchen-vertica, line 563: We should get all IPs from upPods, and then assign them to initiators. You could reuse podFacts.GetHostGroups() to retrieve upHosts and mainHost.

cchen-vertica (commented):
Now we use the requeue way to do the retry, and all restartReconciler calls would fail. Then we cannot reach retry process in next reconcile loop. 

We either go back to the old version(without requeu and httpsPollingCurrentRetry) or we need to skip restartReconciler when httpsPollingCurrentRetry is set. We can discuss this issue if you want.

PR #1397, opened 2025-08-29
Rollback after invalid cert error

kcmackeen:
When you change the clientServer or httpsNMA secret, we will run several checks to validate the secret's certificate. Previous, this failure would not trigger a rollback; it would just produce errors until the user changes to a valid secret. This is different than other cert rotation failures, where we would rollback.

Now, when cert validation fails, we will trigger a rollback to the last good secret. This will only apply when changing a secret; during create db, it will still produce an error and expect the user to fix it (since there is no last good secret).


kcmackeen:
> If the secret cannot be found, it can also be treated as an invalid certificate. In this case, we can trigger a rollback as well.

This will be addressed under a future PR

cchen-vertica (approved):

PR #1400, opened 2025-09-03
Add annotation to disable "https" or "clientserver" TLS auth

kcmackeen:
Add a new annotation:

```
vertica.com/disable-tls-auth-for-config
```

Which accepts the values "https" or "clientserver". This allows us to turn on TLS auth (using ```vertica.com/enable-tls-auth```) but disable it for one of the two TLS configs.

### DIFF
--- a/api/v1/helpers.go
+++ b/api/v1/helpers.go
@@ -771,7 +771,7 @@ func (v *VerticaDB) GetTLSConfigSpecByName(tlsConfig string) *TLSConfigSpec {
 // IsAutoCertRotationEnabled checks if automatic cert rotation is enabled for
 // for a certain tlsconfig (clientServer or httpsNMA)
 func (v *VerticaDB) IsAutoCertRotationEnabled(tlsConfig string) bool {
-	if !vmeta.UseTLSAuth(v.Annotations) {
+	if !vmeta.UseTLSAuth(v.Annotations) || v.IsTLSAuthDisabledForTLSConfig(tlsConfig) {
 		return false
 	}
 	config := v.GetTLSConfigSpecByName(tlsConfig)
@@ -1516,7 +1516,7 @@ func (v *VerticaDB) IsHTTPSTLSConfGenerationEnabled() (bool, error) {
 // IsHTTPSConfigEnabled returns true if tls is enabled and https tls config
 // exists in the db. It means the db ops can start using tls
 func (v *VerticaDB) IsHTTPSConfigEnabled() bool {
-	return v.IsSetForTLS() &&
+	return v.IsSetForTLS() && !v.IsHTTPSTLSAuthDisabled() &&
 		v.GetHTTPSNMATLSSecretInUse() != ""
 }
 
@@ -1528,7 +1528,7 @@ func (v *VerticaDB) IsHTTPSConfigEnabledWithCreate() bool {
 		return v.IsHTTPSConfigEnabled()
 	}
 
-	return v.IsSetForTLS()
+	return v.IsSetForTLS() && !v.IsHTTPSTLSAuthDisabled()
 }
 
 // IsClientServerConfigEnabled returns true if tls is enabled and client-server tls config
@@ -1544,6 +1544,29 @@ func (v *VerticaDB) IsClientServerConfigEnabled() bool {
 func (v *VerticaDB) IsSetForTLS() bool {
 	return v.IsValidVersionForTLS() &&
 		vmeta.UseTLSAuth(v.Annotations)
+}
+
+// IsTLSAuthDisabledForTLSConfig returns true if tls auth is explicitly disabled for the given TLS config.
+// It does not check if tls auth is enabled/disabled for the db.
+func (v *VerticaDB) IsTLSAuthDisabledForTLSConfig(tlsConfigName string) bool {
+	disabledConfig := strings.ToLower(strings.TrimSpace(vmeta.DisableTLSAuthForConfig(v.Annotations)))
+
+	switch tlsConfigName {
+	case ClientServerTLSConfigName:
+		return disabledConfig == vmeta.DisableTLSAuthForConfigServer
+	case HTTPSNMATLSConfigName:
+		return disabledConfig == vmeta.DisableTLSAuthForConfigHTTPS
+	default:
+		return false
+	}
+}
+
+func (v *VerticaDB) IsClientServerTLSAuthDisabled() bool {
+	return v.IsTLSAuthDisabledForTLSConfig(ClientServerTLSConfigName)
+}
+
+func (v *VerticaDB) IsHTTPSTLSAuthDisabled() bool {
+	return v.IsTLSAuthDisabledForTLSConfig(HTTPSNMATLSConfigName)
 }
 
 // IsValidVersionForTLS returns true if the server version
@@ -1993,7 +2016,7 @@ func (v *VerticaDB) NoClientServerRotationNeeded() bool {
 // to include in the NMA configmap. It prioritizes the currently in-use
 // secret if an update is still in progress or a rollback is needed.
 func (v *VerticaDB) GetHTTPSNMATLSSecretForConfigMap() string {
-	if !vmeta.UseTLSAuth(v.Annotations) {
+	if !vmeta.UseTLSAuth(v.Annotations) || v.IsHTTPSTLSAuthDisabled() {
 		return v.GetNMATLSSecret()
 	}
 	return v.GetValueForTLSConfigMap(v.GetHTTPSNMATLSSecret(), v.GetHTTPSNMATLSSecretInUse(), HTTPSNMATLSConfigName)
@@ -2019,6 +2042,17 @@ func (v *VerticaDB) IsCertNeededForClientServerAuth() bool {
 	return tlsMode != tlsModeDisable && tlsMode != tlsModeEnable
 }
 
+// ShouldGenCertForTLSConfig returns true if a specific TLS config is enabled and has changed.
+func (v *VerticaDB) ShouldGenCertForTLSConfig(tlsConfigName string) bool {
+	secret := v.GetHTTPSNMATLSSecret()
+	if tlsConfigName == ClientServerTLSConfigName {
+		secret = v.GetClientServerTLSSecret()
+	}
+	return vmeta.UseTLSAuth(v.Annotations) &&
+		!v.IsTLSAuthDisabledForTLSConfig(tlsConfigName) &&
+		(secret == "" || secret != v.GetSecretInUse(tlsConfigName))
+}
+
 // GetExpectedCertCommonName returns the expected common name for the TLS certificate.
 // For httpsNMATLS, this is the DB admin. For clientServerTLS, it will also default to
 // DB admin, but it can be overridden using clientServerTLS.commonName.
@@ -2031,6 +2065,9 @@ func (v *VerticaDB) GetExpectedCertCommonName(configName string) string {
 
 // GetNMAClientServerTLSMode returns the tlsMode for NMA client-server communication
 func (v *VerticaDB) GetNMAClientServerTLSMode() string {
+	if v.IsClientServerTLSAuthDisabled() {
+		return tlsModeDisable
+	}
 	tlsMode := v.GetClientServerTLSModeForConfigMap()
 	switch tlsMode {
 	case tlsModeDisable:
@@ -2186,7 +2223,7 @@ func (v *VerticaDB) GetHTTPSNMATLSSecret() string {
 
 // GetNMATLSSecret returns the NMATLS secret based on enable-tls annotation
 func (v *VerticaDB) GetNMATLSSecret() string {
-	if !vmeta.UseTLSAuth(v.Annotations) {
+	if !vmeta.UseTLSAuth(v.Annotations) || v.IsHTTPSTLSAuthDisabled() {
 		return v.Spec.NMATLSSecret
 	}
 	return v.GetHTTPSNMATLSSecret()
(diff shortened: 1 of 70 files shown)


LiboYu2:
Why do we only support disabling one of them? What if the customers want to disable both of them?

kcmackeen:
> > Add a new annotation:
> > ```
> > vertica.com/disable-tls-auth-for-config
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Which accepts the values "https" or "clientserver". This allows us to turn on TLS auth (using `vertica.com/enable-tls-auth`) but disable it for one of the two TLS configs.
> 
> I looked at your change. The change only stops the operator from using TLS. The TLS configuration for the vertica DB is still here. I thought your change would revert all the changes that are applyed when enable-tls-auth is turned on. This is more about stopping using TLS in operator.

Great point; will be addressed in a future PR

api/v1/helpers.go
```diff
@@ -2211,12 +2248,22 @@ func (v *VerticaDB) GetClientServerTLSSecret() string {
...
-// In these cases, we skip TLS Update
-func (v *VerticaDB) ShouldSkipTLSUpdateReconcile() bool {
-	return !v.IsSetForTLS() ||
+// ShouldSkipTLSUpdateReconcile will check if TLS Update can be skipped for this config.
+// Possible reasons are:
+// 1) If TLS not enabled
+// 2) DB not initialized
+// 3) Rotate has failed (and rollback is not in progress)
+// 4) TLS auth is disabled for this config
+func (v *VerticaDB) ShouldSkipTLSUpdateReconcile(tlsConfigName string, isRollback bool) bool {
```
roypaulin, line 2257: You could simplify this by having a helper function and then have one version for each tls config.

pkg/controllers/vdb/tlsconfig_reconciler.go
```diff
@@ -155,3 +164,10 @@ func (h *TLSConfigReconciler) checkIfTLSAuthenticationCreatedInDB(ctx context.Co
 	res := strings.Trim(lines[0], " ")
 	return res == "True", nil
 }
+
+func (h *TLSConfigReconciler) shouldSkipTLSConfigReconcile() bool {
```
roypaulin, line 168: A detailed comment would help.

api/v1/helpers.go
```diff
@@ -1546,6 +1546,29 @@ func (v *VerticaDB) IsSetForTLS() bool {
 		vmeta.UseTLSAuth(v.Annotations)
 }
 
+// IsTLSAuthDisabledForTLSConfig returns true if tls auth is explicitly disabled for the given TLS config.
+// It does not check if tls auth is enabled/disabled for the db.
+func (v *VerticaDB) IsTLSAuthDisabledForTLSConfig(tlsConfigName string) bool {
+	disabledConfig := strings.ToLower(strings.TrimSpace(vmeta.DisableTLSAuthForConfig(v.Annotations)))
```
roypaulin, line 1552: Isn't it better to move this to annotations.yaml?

roypaulin (commented):
Wouldn't it be better to either have 2 annotations (enable-https-tls, enable-client-server tls) or on top of having "enable-tls-auth", add disable-https/disable-client-server?

api/v1/helpers.go
kcmackeen: I assume you mean move the strings.ToLower and strings.TrimSpace to annotations.go? That makes sense.

kcmackeen:
> Wouldn't it be better to either have 2 annotations (enable-https-tls, enable-client-server tls) or on top of having "enable-tls-auth", add disable-https/disable-client-server?

As discussed, I changed the single string annotation to be two bool annotations (disable-https-tls-auth and disable-client-server-tls-auth). I updated the description to reflect the new annotations. And I added webhooks to prevent setting both, as well as some other invalid cases. I think it is ready for re-review.

pkg/meta/annotations.go
```diff
@@ -83,10 +83,11 @@ const (
...
 
-	// Disable TLS auth for one TLS config. Options are "https" or "clientserver".
-	DisableTLSAuthForConfigAnnotation = "vertica.com/disable-tls-auth-for-config"
-	DisableTLSAuthForConfigHTTPS      = "https"
-	DisableTLSAuthForConfigServer     = "clientserver"
+	// Disable TLS auth for one TLS config. enable-tls-auth above will enable for entire VDB.
+	// When that is set, these flags can be used to disable for HTTPS or Client-Server
+	// (while leaving the other enabled). They are mutually exclusive.
+	DisableTLSAuthForHTTPSAnnotation        = "vertica.com/disable-https-tls-auth"
+	DisableTLSAuthForClientServerAnnotation = "vertica.com/disable-client-server-tls-auth"
```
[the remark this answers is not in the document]
kcmackeen: I have updated the code and the summary based on our discussion. Please re-review and make sure I didn't miss anything.

pkg/controllers/vdb/httpstlsupdate_reconciler.go
```diff
@@ -86,8 +86,7 @@ func (h *HTTPSTLSUpdateReconciler) Reconcile(ctx context.Context, req *ctrl.Requ
 		return rec.Reconcile(ctx, req)
 	}
 
-	if !h.Vdb.IsHTTPSConfigEnabled() ||
-		h.Vdb.IsStatusConditionTrue(vapi.HTTPSTLSConfigUpdateFinished) {
```
cchen-vertica, line 90: Why do we remove the condition check here?

pkg/controllers/sandbox/sandbox_controller.go
```diff
@@ -132,7 +132,7 @@ func (r *SandboxConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	prunner := cmds.MakeClusterPodRunner(log, r.Cfg, vdb.GetVerticaUser(), passwd, vmeta.UseTLSAuth(vdb.Annotations))
+	prunner := cmds.MakeClusterPodRunner(log, r.Cfg, vdb.GetVerticaUser(), passwd, vdb.IsHTTPSNMATLSAuthEnabled())
```
cchen-vertica, line 135: For pod runner, we should use IsClientServerTLSAuthEnabled()

pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -181,7 +181,7 @@ func (r *VerticaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	prunner := cmds.MakeClusterPodRunner(log, r.Cfg, vdb.GetVerticaUser(), passwd, vmeta.UseTLSAuth(vdb.Annotations))
+	prunner := cmds.MakeClusterPodRunner(log, r.Cfg, vdb.GetVerticaUser(), passwd, vdb.IsAnyTLSAuthEnabled())
```
cchen-vertica, line 184: For pod runner, we should use IsClientServerTLSAuthEnabled()

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -287,7 +286,7 @@ func (r *ReplicationReconciler) makePodFacts(ctx context.Context, vdb *vapi.Vert
 	if err != nil {
 		return nil, err
 	}
-	prunner := cmds.MakeClusterPodRunner(r.Log, r.VRec.Cfg, username, password, vmeta.UseTLSAuth(vdb.Annotations))
+	prunner := cmds.MakeClusterPodRunner(r.Log, r.VRec.Cfg, username, password, vdb.IsAnyTLSAuthEnabled())
```
cchen-vertica, line 289: For pod runner, we should use IsClientServerTLSAuthEnabled()

pkg/controllers/vrep/replication_status_reconciler.go
```diff
@@ -223,7 +223,7 @@ func (r *ReplicationStatusReconciler) makePodFacts(ctx context.Context, vdb *vap
 	if err != nil {
 		return nil, err
 	}
-	prunner := cmds.MakeClusterPodRunner(r.Log, r.VRec.Cfg, username, password, vmeta.UseTLSAuth(vdb.Annotations))
+	prunner := cmds.MakeClusterPodRunner(r.Log, r.VRec.Cfg, username, password, vdb.IsAnyTLSAuthEnabled())
```
cchen-vertica, line 226: For pod runner, we should use IsClientServerTLSAuthEnabled()

pkg/controllers/vrep/replication_reconciler.go
```diff
@@ -182,7 +181,7 @@ func (r *ReplicationReconciler) determineUsernameAndPassword(ctx context.Context
 		return err
 	}
 
-	if r.TargetInfo.Vdb.IsSetForTLS() && *r.TargetInfo.Password == "" && r.Vrep.Spec.TLSConfig == "" {
+	if r.TargetInfo.Vdb.IsAnyTLSAuthEnabledWithMinVersion() && *r.TargetInfo.Password == "" && r.Vrep.Spec.TLSConfig == "" {
```
cchen-vertica, line 184: For replication, only clientserver TLS is used so we can check if clientserver TLS is enabled.

pkg/controllers/vdb/tlsservercertgen_reconciler.go
```diff
@@ -85,16 +83,12 @@ func (h *TLSServerCertGenReconciler) reconcileSecrets(ctx context.Context) error
...
-		}
-		if vmeta.UseTLSAuth(h.Vdb.Annotations) && secretFieldName == nmaTLSSecret {
-			h.Log.Info("TLS auth is enabled. Skipping NMA secret validation and generation")
+		if h.ShouldSkipThisConfig(secretFieldName) {
 			continue
 		}
+		h.Log.Info("Reconciling TLS secret", "TLS secret", secretFieldName, "secretName", secretName)
 		// when nma secret is not empty, we can assign it to https and client TLS
-		if h.Vdb.Spec.NMATLSSecret != "" && (secretFieldName != nmaTLSSecret && secretName == "") {
+		if h.Vdb.Spec.NMATLSSecret != "" && !h.Vdb.IsAnyTLSAuthEnabled() && (secretFieldName != nmaTLSSecret && secretName == "") {
```
cchen-vertica, line 91: Why do we need `!h.Vdb.IsAnyTLSAuthEnabled()` here? When TLS is disabled, we shouldn't copy NMA secret to the other two.

pkg/controllers/vdb/clientservertlsupdate_reconciler.go
```diff
@@ -96,17 +95,19 @@ func (h *ClientServerTLSUpdateReconciler) Reconcile(ctx context.Context, req *ct
 		return ctrl.Result{}, err2
 	}
 
-	res, err := h.Manager.setPollingCertMetadata(ctx)
-	if verrors.IsReconcileAborted(res, err) {
-		return res, err
+	if h.Vdb.IsHTTPSNMATLSAuthEnabled() {
```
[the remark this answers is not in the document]
kcmackeen: The next line is:

`res, err1 := h.Manager.setPollingCertMetadata(ctx)`

Which sets the certs needed for polling HTTPS. If we have clientServerTLS enabled and httpsNMATLS disabled, we do not need these polling certs. This is something that was not possible before, which is why we didn't need the check.

pkg/controllers/vdb/httpstlsupdate_reconciler.go
kcmackeen: These checks were moved into ```ShouldSkipHTTPSTLSUpdateReconcile```

pkg/controllers/vdb/tlsservercertgen_reconciler.go
kcmackeen: The original purpose of this code was that, if you enable TLS and you already have a secret in nmaTLSSecret, we will copy it to httpsNMATLS.secret and clientServerTLS.secret, instead generating new ones.

The problem is the case where httpsNMATLS is disabled and clientServerTLS is enabled. Since httpsNMATLS is disabled, we will first generate an NMA secret. It will then copy the NMA secret into the clientServerTLS.secret. This seemed illogical to me; I feel like it should generate a separate secret for clientServer.

But, you are right; ```!IsAnyTLSAuthEnabled``` is wrong here. I feel like we should change the logic so that we only apply NMA secret to httpsNMATLS.secret; for clientServerTLS, I feel like we should always generate a new one. It is a change in functionality, but it makes more sense to me, given that the two TLS configs are separate now.

api/v1/verticadb_types.go
```diff
@@ -992,6 +992,14 @@ type TLSConfigSpec struct {
...
 	// Allow auto-rotation of a list of secrets, using a certain interval
 	AutoRotate *TLSAutoRotate `json:"autoRotate,omitempty"`
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch","urn:alm:descriptor:com.tectonic.ui:advanced"}
+	// +kubebuilder:validation:Optional
+	// Whether to enable TLS auth for this config. This replaces the
+	// "vertica.com/enable-tls-auth" annotation, which applied at the DB level.
+	// If 'enabled' is not specified for any  TLS configs, and the annotation is set,
+	// the operator will continue to honor the annotation (for backwards compatibility).
+	// Otherwise, 'enabled' takes precedence.
+	Enabled *bool `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
kcmackeen: I see 4 cases here:
1) TLS Config is defined and "enabled" is specified; easy case -> use enabled value
2) TLS Config is defined but "enabled" is not specified; if user set a secret/mode, they obviously want TLS auth -> TRUE
3) TLS Config is not defined and annotation is false or not set; user clearly does not want TLS -> FALSE
4) TLS Config is not defined and annotation is true; see below

This is a rare but possible case for customers. Typically, if customer sets TLS annotation but no TLS config, they want operator to auto-generate their secrets for them. So this case can occur in that window between when user sets TLS and when we finish generating secret and set in spec.

So, for this case, TLS config will be enabled if annotation is true; otherwise, it is false. In the future, we will remove the annotation and this case will be false.

PR #1403, opened 2025-09-04
Fixed restart issue

cchen-vertica:
This PR fixed the issue: when a subcluster is shut down, the other primary subclusters cannot be restarted in that cluster.


pkg/controllers/vdb/restart_reconciler.go
```diff
@@ -136,12 +136,11 @@ func (r *RestartReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctr
 	// ScheduleOnly.
 	if r.PFacts.GetUpNodeAndNotReadOnlyCount() == 0 &&
```
LiboYu2, line 137: Could you elaborate the condition r.PFacts.GetUpNodeAndNotReadOnlyCount() == 0? 
All the nodes are down?

cchen-vertica: It means no writable nodes are up, either all nodes are down or all up nodes are readonly nodes.

PR #1405, opened 2025-09-05
Trigger rollback when set with non-existent TLS secret

kcmackeen:
Currently, operator will trigger failure rollback when set with an invalid cert (eg expired) or when the cert rotation process fails. However, if set with a secret that does not exist, operator will not rollback to last good secret; it will remain on the failing secret until the user manually updates it.

This PR will trigger rollback in the case of a non-existent secret. The one "gotcha" with this is that obj_reconciler will requeue if HTTPS/NMA secret and Client-Server secret do not exist. But we can point them to the status secrets when rollback is pending (which is what we do for the NMA config map as well).


pkg/controllers/vdb/tlsservercertgen_reconciler.go
```diff
@@ -144,7 +144,8 @@ func (h *TLSServerCertGenReconciler) reconcileOneSecret(secretFieldName, secretN
 				if tlsStatus != nil {
 					// we do not recreate the secret as there is already
 					// a secret of this type in the status.
-					return nil
+					return h.InvalidCertRollback(ctx, "Validation of TLS Certificate %q failed; secret %q does not exist",
```
cchen-vertica, line 147: Could we directly revert secret value in the spec here using the one in status, and update Spec.AutoRotateFailedSecret to the failed one? What else does the TLSCertRollbackNeeded condition do?

cchen-vertica (approved):

PR #1410, opened 2025-09-11
Change password in sandbox and other places when passwordSecret is updated

kcmackeen:
Previously, if passwordSecret in spec was updated and VDB contained sandboxes, this was prevented via webhook. Now, after updating the main cluster, we will loop through each sandbox and update their passwords as well. However, if any sandbox is shutdown and passwordSecret is changed, we will still block this via webhook (since we cannot update them).

Also, we want any reconciler that needs to get DB password to share password logic and password cache. This will add a PasswordManager, which provides a shared password cache across all reconcilers that need it.

### DIFF
--- a/pkg/controllers/vdb/verticadb_controller.go
+++ b/pkg/controllers/vdb/verticadb_controller.go
@@ -177,6 +177,7 @@ func (r *VerticaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	}
 
+	r.InitCacheForVdb(vdb)
 	passwd, err := r.GetSuperuserPassword(ctx, log, vdb)
 	if err != nil {
 		return ctrl.Result{}, err
@@ -189,7 +190,6 @@ func (r *VerticaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	dispatcher := r.makeDispatcher(log, vdb, prunner, passwd)
 	var res ctrl.Result
 
-	r.InitCertCacheForVdb(vdb)
 	// Iterate over each actor
 	actors := r.constructActors(log, vdb, prunner, &pfacts, dispatcher)
 	for _, act := range actors {
@@ -518,7 +518,7 @@ func (r *VerticaDBReconciler) GetConfig() *rest.Config {
 	return r.Cfg
 }
 
-func (r *VerticaDBReconciler) InitCertCacheForVdb(vdb *vapi.VerticaDB) {
+func (r *VerticaDBReconciler) InitCacheForVdb(vdb *vapi.VerticaDB) {
 	fetcher := &cloud.SecretFetcher{
 		Client:   r.Client,
 		Log:      r.Log,


pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -86,26 +90,48 @@ func (a *PasswordSecretReconciler) statusMatchesSpec() bool {
...
+		a.Vdb.Spec.PasswordSecret, names.SuperuserPasswordKey, a.PasswordManager, true)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
+	res, err := a.updatePasswordSecretInSandbox(ctx, a.PFacts, newPasswd, "")
+	if verrors.IsReconcileAborted(res, err) {
+		return res, err
+	}
+	for _, sb := range a.Vdb.Spec.Sandboxes {
```
roypaulin, line 107: Could we let sandbox controller handle sandbox password change? When password secret changes, we first change the password in the main cluster and then for each sandbox, we wake up the sandbox controller which will change the password in each sandbox. That way, sandboxes password change will be done concurrently, and one failure will not affect other sandboxes.
The same logic as sandbox upgrade.

pkg/security/password.go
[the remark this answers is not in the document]
LiboYu2: @roypaulin  Could you elaborate on your previous comments?  

This is my thought. Our current cache is only for storing certs. It is not hard to make it a generic cache so we can use it in different scenarios.

pkg/cache/vdb_cache.go
```diff
@@ -136,56 +197,42 @@ func (c *CacheManangerStruct) DestroyCertCacheForVdb(namespace, name string) {
 func makeVdbCertCache(namespace string, ttl int, fetcher *cloud.SecretFetcher, enabled bool) *VdbCacheStruct {
```
LiboYu2, line 197: We can remove Cert from this function name

pkg/cache/vdb_cache.go
```diff
@@ -64,14 +67,72 @@ type CertCache interface {
 	CleanCacheForVdb([]string)
 }
 
+type ItemCache[T any] struct {
+	lock         sync.Mutex
+	items        map[string]T
+	creationTime map[string]time.Time
```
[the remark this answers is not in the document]
LiboYu2: > If they would have the same key, can we merge them into a single map?

You mean merging cert and password in the same map?
It is better to have a separate map for each type of cache.  But given that the password map has only a single item, maybe we can make its map a generic one and share it with other use cases.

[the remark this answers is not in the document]
kcmackeen: I think what you are saying is have a certCache (for TLS) and then a genericCache (containing just password for now, but we can add more items in the future). That makes a lot of sense.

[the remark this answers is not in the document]
roypaulin: my comment is specifically about lines 72-73. items and creationTime have the same key so they can be merged into a single map of struct

kcmackeen:
> I think you should fix the commit history. I see some commits that should already be in main.

Rebased again; I think it is good now

api/v1/helpers.go
```diff
@@ -1455,8 +1469,8 @@ func (v *VerticaDB) IsFetchNodeDetailsLogDisabled() bool {
 	return vmeta.IsFetchNodeDetailsLogDisabled(v.Annotations)
 }
 
-func (v *VerticaDB) GetTLSCacheDuration() uint64 {
-	duration := vmeta.GetTLSCacheDuration(v.Annotations)
+func (v *VerticaDB) GetCacheDuration() uint64 {
```
roypaulin, line 1472: At some point we need to make the decision whether we want to have a single cache duration for all elements in the cache.

api/v1/verticadb_webhook.go
```diff
@@ -518,20 +518,50 @@ func (v *VerticaDB) checkValidSubclusterTypeTransition(oldObj *VerticaDB, allErr
...
+	// 1) Check if there are any shutdown sandbox
+	shutdownSandbox := ""
+	for _, sandbox := range v.Spec.Sandboxes {
+		if sandbox.Shutdown {
+			shutdownSandbox = sandbox.Name
+			break
+		}
+	}
+
+	if shutdownSandbox != "" && oldObj.Spec.PasswordSecret != v.Spec.PasswordSecret {
```
roypaulin, line 540: What if only one subcluster in the sandbox is "shutdown"? Is password change going to work? Is there going to be any issues when that subcluster is restarted?

api/v1/verticadb_webhook.go
```diff
@@ -518,20 +518,50 @@ func (v *VerticaDB) checkValidSubclusterTypeTransition(oldObj *VerticaDB, allErr
...
+	if v.IsSandboxOpNeeded() && oldObj.Spec.PasswordSecret != v.Spec.PasswordSecret {
 		err := field.Invalid(field.NewPath("spec").Child("passwordSecret"),
 			v.Spec.PasswordSecret,
-			fmt.Sprintf("Cannot change passwordSecret from %q to %q as the vdb has sandbox",
-				oldObj.Spec.PasswordSecret, v.Spec.PasswordSecret))
+			"Cannot change passwordSecret while sandboxing/unsandboxing is in progress")
 		allErrs = append(allErrs, err)
 	}
+
+	// 3) Check if sandboxing/unsandboxing is being attempted while passwordSecret is being changed
```
roypaulin, line 556: This only covers password change on the main cluster. But if someone tries to sandbox a subcluster and password change is in progress in that sandbox, I think we should wait until password change is done.

api/v1/verticadb_webhook.go
```diff
@@ -518,20 +518,50 @@ func (v *VerticaDB) checkValidSubclusterTypeTransition(oldObj *VerticaDB, allErr
...
+	if v.IsSandboxOpNeeded() && ((v.Status.PasswordSecret == nil && v.Spec.PasswordSecret != "") ||
```
roypaulin, line 557: Minor: move the 2nd operand to a function

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -63,89 +72,221 @@ func (a *PasswordSecretReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
...
+		return a.Vdb.Spec.PasswordSecret == *a.Vdb.Status.PasswordSecret
+	}
+
+	return a.Vdb.Spec.PasswordSecret == a.Vdb.GetPasswordSecretForSandbox(sbName)
+}
+
+// getSandboxesNeedingUpdating returns sandboxes where secret in status does not match secret in spec
+func (a *PasswordSecretReconciler) getSandboxesNeedingUpdating() []vapi.Sandbox {
+	sandboxes := []vapi.Sandbox{}
+	for _, sb := range a.Vdb.Spec.Sandboxes {
```
roypaulin, line 157: It does not change much but I think we should use the status as it represents 100% sandboxes in the db.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -63,89 +72,221 @@ func (a *PasswordSecretReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
...
+	for _, sb := range sandboxes {
+		triggerUUID := uuid.NewString()
+		sbMan := MakeSandboxConfigMapManager(a.Rec, a.Vdb, sb.Name, triggerUUID)
+		triggered, err := sbMan.triggerSandboxController(ctx, PasswordChange)
+		if triggered {
+			a.Log.Info("Sandbox ConfigMap updated. The sandbox controller will drive the password change",
+				"trigger-uuid", triggerUUID, "Sandbox", sb.Name)
+		}
+		if err != nil {
+			a.Log.Error(err, "Failed to trigger sandbox password change", "sandbox", sb.Name)
```
roypaulin, line 190: > We might never retry for that sandbox for which it failed. There is a few options:
> 
> * return the error now
> * continue to next sandbox while keeping the list of failed sandboxes and return the error at the end.
> 
> I suggest we add a secret field in sandbox status. At the beginning, of the reconciler, if password update has already been done in main cluster, you can reiterate over the sandboxes and for those with that annotation unset and sandbox_status != new_password_secret you set the annotation.

> I added status field sandboxes[].passwordSecret, and I updated all logic to run reconcile when spec and status are mismatched. For this particular case, if we are not able to trigger sandbox change, it will finish the rest of the sandboxes then try again on next iteration.

I still think the issue I mentioned above can happen. If you do not return an error, you will have to wait until something else triggers the next iteration. If that does not happen, there won't be passworrd change in that sandbox.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -63,89 +72,221 @@ func (a *PasswordSecretReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
...
+	// If we are reconciling a sandbox and it matches spec, no-op
+	// Also, if we are reconciling a sandbox and main cluster is not updated yet, no-op
+	if sbName != vapi.MainCluster {
+		if a.statusMatchesSpec(sbName) || !a.statusMatchesSpec(vapi.MainCluster) {
+			return ctrl.Result{}, nil
+		}
+	}
+
+	// If everything is up-to-date, no-op
+	outdatedSandboxes := a.getSandboxesNeedingUpdating()
```
roypaulin, line 91: Do we need to calculate this when this reconciler is called by the sandbox controller?

api/v1/helpers.go
kcmackeen: Yes, I think this is a good future task

api/v1/verticadb_webhook.go
kcmackeen: Good question. I added an e2e test for when subcluster (but not sandbox) is down when password is updated, and it worked. So I guess the password change is at the sandbox level, not the subcluster level.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -63,89 +72,200 @@ func (a *PasswordSecretReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
...
-		return ctrl.Result{}, a.updatePasswordSecretStatus(ctx)
+	sbName := a.PFacts.GetSandboxName()
+
+	// Ensure status initialized for main cluster or sandbox
+	if err := a.ensureStatusInitialized(ctx, sbName); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// If we are reconciling a sandbox and it matches spec, no-op
+	// Also, if we are reconciling a sandbox and main cluster is not updated yet, no-op
```
cchen-vertica, line 83: Why does the sandbox need to wait for the main cluster update? Are we waiting for vdb-controller to update config map? 

If this is for a sandbox and trigger-id is set, we shouldn't wait for main cluster. Even if waiting is required, we should requeue instead of doing a no-op.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -63,89 +72,200 @@ func (a *PasswordSecretReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
...
+	// For main cluster, ensure status and trigger follow-ups
+	if sbName == vapi.MainCluster {
+		if err := a.updatePasswordSecretStatus(ctx); err != nil {
+			return ctrl.Result{}, err
+		}
+		// Always trigger any out-of-date sandboxes
+		return a.triggerOutOfDateSandboxes(ctx, outdatedSandboxes)
+	}
+
+	// Unset sandbox annotation
```
cchen-vertica, line 114: unset sandbox annotation should be the last step. Otherwise, if updatePasswordSecretStatusForSandbox() failed, we won't retry it.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -63,89 +72,200 @@ func (a *PasswordSecretReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
 		return ctrl.Result{}, nil
 	}
 
-	// set status to the current using secret when db is initialized
-	if a.Vdb.Status.PasswordSecret == nil {
-		return ctrl.Result{}, a.updatePasswordSecretStatus(ctx)
+	sbName := a.PFacts.GetSandboxName()
+
+	// Ensure status initialized for main cluster or sandbox
+	if err := a.ensureStatusInitialized(ctx, sbName); err != nil {
```
cchen-vertica, line 78: Will this update status.passwordSecret using spec.passwordSecret? If so, line 107 and 124 will do it, and we don't need it here.

pkg/controllers/vdb/passwordsecret_reconciler.go
```diff
@@ -63,89 +72,200 @@ func (a *PasswordSecretReconciler) Reconcile(ctx context.Context, _ *ctrl.Reques
...
+		triggerUUID := uuid.NewString()
+		sbMan := MakeSandboxConfigMapManager(a.Rec, a.Vdb, sb.Name, triggerUUID)
+		triggered, err := sbMan.triggerSandboxController(ctx, PasswordChange)
+		if triggered {
+			a.Log.Info("Sandbox ConfigMap updated. The sandbox controller will drive the password change",
+				"trigger-uuid", triggerUUID, "Sandbox", sb.Name)
+		}
+		if err != nil {
+			a.Log.Error(err, "Failed to trigger sandbox password change", "sandbox", sb.Name)
+			return ctrl.Result{}, nil
```
cchen-vertica, line 169: Why do we return nil when there is an error?

pkg/controllers/vdb/passwordsecret_reconciler.go
kcmackeen: In one of Roy's comments above, he suggested that sandbox should never be updated when main cluster has not been updated.

pkg/controllers/sandbox/sandbox_controller.go
```diff
@@ -128,7 +128,7 @@ func (r *SandboxConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	sandboxName := configMap.Data[v1.SandboxNameKey]
 	log = log.WithValues("verticadb", vdb.Name, "sandbox", sandboxName)
 
-	passwd, err := vk8s.GetSuperuserPassword(ctx, r.Client, log, r, vdb)
+	passwd, err := vk8s.GetSuperuserPassword(ctx, r.Client, log, r, vdb, r.CacheManager, sandboxName)
```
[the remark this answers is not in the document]
kcmackeen: sandbox control should already have init'ed cacheManager from main. But, yes, it makes sense to move that init up, just in case.

pkg/controllers/vdb/onlineupgrade_reconciler.go
```diff
@@ -470,7 +470,13 @@ func (r *OnlineUpgradeReconciler) queryOriginalConfigParamDisableNonReplicatable
 		r.Log.Info("No Up nodes found. Requeue reconciliation.")
 		return ctrl.Result{Requeue: true}, nil
 	}
-	vc := catalog.MakeVCluster(r.VDB, pf.VerticaSUPassword, initiator.GetPodIP(), r.Log, r.VRec.Client, r.VRec.EVRec, r.VRec.CacheManager)
+	pw := ""
```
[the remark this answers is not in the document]
kcmackeen: Hmm I don't remember why I made this change. I will revert it and see what happens.

pkg/controllers/vdb/passwordsecret_reconciler.go
kcmackeen: As discussed, I changed it so that the trigger is done first, and it does not need to wait for the main cluster. Since we are using the podfacts, which is limited to just the main cluster or sandbox, I did not need to do anything with trigger id, which made the logic a lot more simple.

pkg/controllers/vdb/verticadb_controller.go
```diff
@@ -525,7 +525,7 @@ func (r *VerticaDBReconciler) InitCertCacheForVdb(vdb *vapi.VerticaDB) {
 		Obj:      vdb,
 		EVWriter: r.EVRec,
 	}
-	r.CacheManager.InitCertCacheForVdb(vdb, fetcher)
+	r.CacheManager.InitCacheForVdb(vdb, fetcher)
```
cchen-vertica, line 528: We should iniitialize the cache earlier before we use it. You can do the same thing you did for sandbox controller.

cchen-vertica (commented):
Looks good. Only one comment needs to be addressed.

cchen-vertica (approved):

PR #1413, opened 2025-09-12
Namespace scope deployment of grafana/prometheus

roypaulin:
This is about deploying grafana/prometheus when there are namespace scope operator deployments. We don't want each operator instance to have its own grafana/prometheus (waste of resources and prometheus operator clusterrole can't be replaced by a role: shouldn't have a cluster resource in a namespace scope deployment).

Usually for namespace -scope deployment, we first deploy a cluster-scope operator with a webhook and no controllers. We will deploy grafana/prometheus in that one. Now for the namescope deployments, prometheus will be disabled (so we don't end up with multiple prometheus instances), but the user will need to pass the name of the cluster scope helm release so prometheus can find find the dbs in that namespace.
Ex:
1. helm install vdb-op -n verticadb-operator ... with webhook, controllers disabled, prometheus/grafana enabled, clusterscope
2. helm install vdb-op2 -n vertica... webhook disabled, prometheus/grafana disabled, namespace scope, cluster scope release name: vdb-op

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1395,6 +1395,7 @@ jobs:
           operator-image: ${{ needs.build.outputs.operator-image }}
           vertica-image: ${{ needs.build.outputs.minimal-vertica-image }}
           vertica-deployment-method: vclusterops
+          controllers-scope: namespace
           communal-storage-type: s3
           vertica-superuser-name: dbadmin
 


cchen-vertica:
Can the user do below:
helm install vdb-op1 -n vertica... without webhook, controller enabled, namespace scope, prometheus/grafana enabled 
helm install vdb-op2 -n vertica... webhook disabled, prometheus/grafana disabled, namespace scope, cluster scope release name: vdb-op1

I believe that is the MBIR use case.

Makefile
```diff
@@ -165,6 +165,8 @@ VDB_NAMESPACE?=default
 MINIMAL_VERTICA_IMG ?=
 # Name of the helm release that we will install/uninstall
 HELM_RELEASE_NAME?=vdb-op
+# Name of the cluster scope release that will be used for prometheus in the namespace scope release
+CLUSTER_SCOPE_RELEASE_NAME?=
```
qindotguan, line 169: If there is only one cluster-wide operator, is there a way to fetch this name automatically?

cchen-vertica (approved):

PR #1415, opened 2025-09-12
Save config object names to status

roypaulin:
This follows the work done in #1370. There was a bug. The list of configmaps/secrets used to inject env vars were saved as a comma separate list in vdb labels. That is wrong because first comma is not a valid char for a label, and second labels are not used to keep this kind of metadata.
Now I am following the right approach by saving the list in the status.

### DIFF
--- a/pkg/controllers/vdb/observedconfigobjects_reconciler.go
+++ b/pkg/controllers/vdb/observedconfigobjects_reconciler.go
@@ -47,7 +47,7 @@ func (l *ObservedConfigObjsReconciler) Reconcile(ctx context.Context, _ *ctrl.Re
 	secrets := l.getReferencedSecret()
 	configmaps := l.getReferencedConfigmap()
 	if len(secrets) > 0 || len(configmaps) > 0 {
-		if equalSets(secrets, l.Vdb.Status.ObservedSecrets) && equalSets(configmaps, l.Vdb.Status.ObservedConfigMaps) {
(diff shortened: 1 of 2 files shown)


pkg/controllers/vdb/observedconfigobjects_reconciler.go
```diff
@@ -17,64 +17,52 @@ package vdb
...
-	if len(configmaps) > 0 {
-		labels[vmeta.ConfigMapSelectorLabel] = strings.Join(configmaps, ",")
-	}
-	if len(labels) > 0 {
-		return ctrl.Result{}, l.applyLabels(ctx, labels)
+func (l *ObservedConfigObjsReconciler) Reconcile(ctx context.Context, _ *ctrl.Request) (ctrl.Result, error) {
+	secrets := l.getReferencedSecret()
+	configmaps := l.getReferencedConfigmap()
+	if len(secrets) > 0 || len(configmaps) > 0 {
+		if equalSets(secrets, l.Vdb.Status.ObservedSecrets) && equalSets(configmaps, l.Vdb.Status.ObservedConfigMaps) {
```
cchen-vertica, line 50: If we save secret and configMaps with the order in status, why do we need to use equalSets() to compare them?

roypaulin: You are right we don't need "equalSets()". We already save an ordered list of secrets/configmaps. A normal comparison is enough.

cchen-vertica (approved):

PR #1418, opened 2025-09-20
Fixed sandbox restart issue

cchen-vertica:
This PR addresses the following issue:  
Before sandboxing a subcluster, if the existing sandbox nodes restart with new IPs, the subcluster fails after sandboxing. This is a vcluster bug, so the fix is applied in vclusterOps. 

The solution ensures that config files are synchronized from the sandbox primary nodes to the new subcluster, allowing node restarts to succeed.

### DIFF
--- a/pkg/controllers/vdb/deploymentmethod_reconciler.go
+++ b/pkg/controllers/vdb/deploymentmethod_reconciler.go
@@ -96,6 +96,11 @@ func (d *DeploymentMethodReconciler) reconcileHTTPSTLS(ctx context.Context) (ctr
 		if verrors.IsReconcileAborted(res, err) {
 			return res, err
 		}
+		// after restart, we need to recollect pfacts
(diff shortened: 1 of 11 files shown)


pkg/controllers/vdb/deploymentmethod_reconciler.go
```diff
@@ -96,6 +96,11 @@ func (d *DeploymentMethodReconciler) reconcileHTTPSTLS(ctx context.Context) (ctr
 		if verrors.IsReconcileAborted(res, err) {
 			return res, err
 		}
+		// after restart, we need to recollect pfacts
+		err = d.PFacts.Collect(ctx, d.Vdb)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
roypaulin, line 103: After it recollects you need to try again to get the first up pod, to use at line 106.

PR #1424, opened 2025-09-23
Fix ND failures during cert validation tests

kcmackeen:
We were having intermittent failures during invalid-cert-validation and invalid-cert-rollback tests. There seem to be two issues here:

1. Occasionally, step 10 (where we generate the secrets) will fail to generate one of the secrets. This seems to be a concurrency issue; if we are attempting to generate multiple secrets using envsubstr, they can stomp on each other.
2. Rarely, when setting an invalid secret, we would trigger rollback but then, later, we would attempt to rotate to this failed secret. This is a timing issue, and it results from the clientservertlsupdate/httpsnmatlsupdate reconciler still having stale VDB

To fix issue 1, I separated invalid-cert-rollback into two tests: one for auto-rotate and one for non-auto-rotate; by splitting the test, we only need one "invalid" secret per test, and we don't run into any concurrency issues.

To fix issue 2, I added a requeue when cert validation fails. This will ensure that any future reconcilers (particularly, clientservertlsupdate and httpsnmatlsupdate) will have the up-to-date VDB, including the rolled-back secret.


kcmackeen:
> The invalid-cert test failed again. It could be the other issues.

It required restructuring one of the test. I also found a separate ND issue within the operator itself. See the updated summary for more details.

PR #1457, opened 2025-11-06
Always cache last license validation time

roypaulin:
This always cache the last validation check time, so we do not have to validate the license for each reconciliation loop.


cchen-vertica:
Looks good. Let's see if tests can pass.

